PR #29, opened 2024-02-19
17.0 technical onboarding berm

berm-odoo:
Hi it's my first PR for the test. I just write the manifest file.

### DIFF
--- a/estate/__init__.py
+++ b/estate/__init__.py
@@ -1,1 +1,2 @@
+from . import models
 from odoo import api
--- a/estate/__init__.py
+++ b/estate/__init__.py
@@ -1,2 +1,1 @@
 from . import models
-from odoo import api
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -8,7 +8,7 @@
     'license': "LGPL-3",
     # data files always loaded at installation
     'data': [
-        # 'views/mymodule_view.xml',
+        'security/ir.model.access.csv',
     ],
     # data files containing optionally loaded demonstration data
     'demo': [
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -9,6 +9,7 @@
     # data files always loaded at installation
     'data': [
         'security/ir.model.access.csv',
+        'views/estate_views.xml'
     ],
     # data files containing optionally loaded demonstration data
     'demo': [
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -9,6 +9,9 @@
     'data': [
         'security/ir.model.access.csv',
         'views/estate_property_views.xml',
+        'views/estate_property_offer_views.xml',
+        'views/estate_property_type_views.xml',
+        'views/estate_property_tag_views.xml',
         'views/estate_menus.xml'
     ],
     # data files containing optionally loaded demonstration data
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -6,10 +6,10 @@
     'category': 'Tutorials',
     'description': "Estate management",
     'license': "LGPL-3",
-    # data files always loaded at installation
     'data': [
         'security/ir.model.access.csv',
-        'views/estate_views.xml'
+        'views/estate_property_views.xml',
+        'views/estate_menus.xml'
     ],
     # data files containing optionally loaded demonstration data
     'demo': [
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -1,10 +1,11 @@
 {
-    'name': "berm-estate",
-    'version': '1.0',
+    'name': "Estate",
+    'version': '1.01',
     'depends': ['base'],
     'author': "berm-odoo",
-    'category': 'Sales',
-    'description': "Description text",
+    'category': 'Tutorials',
+    'description': "Estate management",
+    'license': "LGPL-3",
     # data files always loaded at installation
     'data': [
         # 'views/mymodule_view.xml',
--- a/estate/models/__init__.py
+++ b/estate/models/__init__.py
@@ -1,0 +1,3 @@
+#! env/bin/python
+
+from . import estate_property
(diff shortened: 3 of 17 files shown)


estate/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import models
+from odoo import api
```
clbr-odoo, line 2:
```suggestion
```
useless import ?

estate/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+{
+    'name': "Estate",
+    'version': '1.01',
```
clbr-odoo, line 3: We start at 1.0, which is the default value in manifest so you can even omit it here :slightly_smiling_face:

estate/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+{
+    'name': "Estate",
```
clbr-odoo, line 2: Be consistent with ' or ", doesn't matter which but keep the same across the file.

estate/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+#! env/bin/python
+
```
clbr-odoo, line 2: Useless diff

estate/models/estate_property.py
```diff
@@ -0,0 +1,40 @@
+from odoo import fields, models
+
+
+class Estate(models.Model):
+    _name = "estate.property"
+    _description = "Properties of estate entities."
+
+    name = fields.Char(required=True)
```
clbr-odoo, line 8:
```suggestion
    name = fields.Char(string="Name", required=True)
```
It's a good practice to add an explicit `string=` to fields that are meant to be in the view, else you let Odoo decide what he will use, which is a bit random.

estate/models/estate_property.py
```diff
@@ -0,0 +1,40 @@
...
+    description = fields.Text()
+    active = fields.Boolean(default=True)
+    postcode = fields.Char()
+    date_availability = fields.Date(
+            copy=False,
+            default=fields.Date.add(fields.Date.today(), fields.Date.today(), months=3))
```
clbr-odoo, line 14:
```suggestion
            default=fields.Date.add(fields.Date.today(), months=3))
```
I'm surprised it even works, not sure why this second parameter is there for :rofl:

estate/models/estate_property.py
```diff
@@ -0,0 +1,40 @@
...
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+            string='Orientation',
+            selection=[
+                ('North', 'north'),
+                ('South', 'south'),
+                ('East', 'east'),
+                ('West', 'east')],
+            help="Cardinal orientation of the garden.")
```
clbr-odoo, line 30:
```suggestion
            help="Cardinal orientation of the garden.",
        )
```
Optional comment regarding indent: when you have a fields with a lot of parameters, I try to add a comma at the end of the last and put the parenthesis on the next line, it makes the diff cleaner when another dev will add a parameter later :)

estate/models/estate_property.py
```diff
@@ -0,0 +1,40 @@
...
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+            string='Orientation',
+            selection=[
+                ('North', 'north'),
+                ('South', 'south'),
+                ('East', 'east'),
+                ('West', 'east')],
```
clbr-odoo, line 29:
```suggestion
                ('north', 'North'),
                ('south', 'South'),
                ('east', 'East'),
                ('west', 'West')],
```
The list of tuple we give to selection is something like `('unique_technical_identifier', 'Human Readable Name')`

estate/models/estate_property.py
```diff
@@ -0,0 +1,40 @@
...
+    state = fields.Selection(
+            string='State',
+            required=True,
+            copy=False,
+            default='New',
+            selection=[('New', 'new'),
+                       ('Offer Received', 'offer received'),
+                       ('Offer Accepted', 'Offer Accepted'),
+                       ('Sold', 'sold'),
+                       ('Canceled', 'Canceled')])
```
clbr-odoo, line 40:
```suggestion
            selection=[
                ('new', 'New'),
                ('offer_received', 'Offer Received'),
                ('offer_accepted', 'Offer Accepted'),
                ('sold', 'Sold'),
                ('canceled', 'Canceled')
            ],
        )
```

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,0
```
clbr-odoo, line 2: Does the tutorial ask to not give unlink (which is equivalent to delete) permissions ? (Referring to the last 0.)

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<odoo>
+
+  <menuitem id="estate_menu" action="estate_property_menu">
```
clbr-odoo, line 4:
```suggestion
  <menuitem id="estate_menu" name="Real Estate">
```
I think this action is useless, the "root" menu (a menu without parent) will always open the related App.
See comment below regarding the two actions.

estate/views/estate_property_views.xml
clbr-odoo, line 1: Your two actions (`estate_property_menu` and `estate_property_action`) are doing exactly the same thing: opening a view with records of type "estate.property" in tree (= list) view or form view, only their "title" differ.

clbr-odoo (changes requested):
Good job !
Few things to fix, and few good practice advices :+1: 
Keep up :partying_face:

PR #31, opened 2024-02-19
17.0 technical onboarding igbe

igbe-odoo:
First pull request


estate/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
hupo-odoo, line 1: There is a missing empty line at the bottom of the file. You should configure your IDE to add it automatically. This comment also applies to all files of the PR.

estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_model
```
hupo-odoo, line 1:
```suggestion
from . import estate
```
No need to add '_model' as we already know it's a model. You can change the file name as well

estate/models/estate_model.py
```diff
@@ -0,0 +1,33 @@
+from odoo import fields, models
+
```
hupo-odoo, line 2: You should have 2 empty lines before a class definition according to 'PEP 8: E302 expected 2 blank lines, found 1'

estate/models/estate_model.py
```diff
@@ -0,0 +1,33 @@
...
+        ('W', 'West')
+    ])
+    active = fields.Boolean(default=False)
+    state = fields.Selection(selection=[
+        ('N', 'New'),
+        ('OR', 'Offer Received'),
+        ('OA', 'Offer Accepted'),
+        ('S', 'Sold'),
+        ('C', 'Canceled')
+    ], required=True, copy=False, default='N')
```
hupo-odoo, line 32: I would give clearer name for the selection choices. As this will be used in several part of the code, we always want to give it a short clean name, just a simple letter is not clear enough.

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_model,access_estate_model,model_estate_property,base.group_user,1,1,1,1
```
hupo-odoo, line 2: The name of the model is always separated with dots instead of underscore
```suggestion
access_estate_model,access.estate.model,model_estate_property,base.group_user,1,1,1,1
```

hupo-odoo (commented):
Good job! I've made a few suggestions but it's a very good start. I would also suggest you to make your commit title and message follow the following guidelines: https://www.odoo.com/documentation/17.0/contributing/development/git_guidelines.html

hupo-odoo (commented):
You have done a great job! :+1: I've noticed a few things:
- In JS, the code inside {} must be spaced, like this for example : { PieChart }
- Some of your files do not finish with an empty line. You should configure your IDE to do that automatically

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,19 @@
...
+    property_ids = fields.One2many("estate.property", "property_type_id", string="Property")
+    sequence = fields.Integer(string='Sequence')
+    offer_ids = fields.One2many("estate.property.offer", "property_type_id")
+    offer_count = fields.Integer(compute="_compute_offer_count")
+
+    @api.depends("offer_ids")
+    def _compute_offer_count(self):
+        for type in self:
+            type.offer_count = len(type.offer_ids)
+        return True
```
hupo-odoo, line 19:
```suggestion
    def _compute_offer_count(self):
        for type in self:
            type.offer_count = len(type.offer_ids)
```
a compute method should not return anything

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
...
+            "line_ids": [
+                Command.create({
+                    "name": "Taxe",
+                    "quantity": "1",
+                    "price_unit": str(float(self.selling_price) * 0.06)
+                }),
+                Command.create({
+                    "name": "Administrative fees",
+                    "quantity": "1",
+                    "price_unit": "100"
```
hupo-odoo, line 21:
```suggestion
                    "quantity": 1,
                    "price_unit": 100,
```
no need for string here, as the values are int or float

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
...
+
+class EstateProperty(models.Model):
+    _inherit = "estate.property"
+
+    def estate_sold(self):
+        self.env["account.move"].create({
+            "name": "Test",
+            "move_type": "out_invoice",
+            "partner_id": self.env["estate.property"].buyer_id,
+            "line_ids": [
```
hupo-odoo, line 12:
```suggestion
            "invoice_line_ids": [
```
You are asked to use invoice_line_ids, not line_ids. Also, you forgot the line with the main amount, so far you have only the tax and admin fee

PR #43, opened 2024-03-18
[TRAINING] first commit in order to create a PR

Clement-Cardot:


estate/models/estate_property.py
```diff
@@ -0,0 +1,42 @@
...
+    state = fields.Selection([
+        ('new', 'New'),
+        ('offer_received', 'Offer Received'),
+        ('offer_accepted', 'Offer Accepted'),
+        ('sold', 'Sold'),
+        ('canceled', 'Canceled')
+    ], default='new', required=True, copy=False)
+
+    postcode = fields.Char()
+    date_availability = fields.Date(default=_compute_three_months_from_now(), copy=False)
```
kfr-odoo, line 27: `compute_xxx`methods have a specific meaning in Odoo. They should only be used for computed fields. You have two options regarding the default property:

1. Use a lambda expression
2. Rename the `_compute_three_months_from_now` => `_get_default_date_availability`

estate/models/estate_property.py
```diff
@@ -0,0 +1,42 @@
...
+    return fields.Date.today() + relativedelta(months=3)
+
+class EstateProperty(models.Model):
+
+    _name = "estate.property"
+    _description = "Real Estate Property"
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    active = fields.Boolean(default=True)
```
kfr-odoo, line 17: This is already the default value of Boolean fields, it is redudant.

```suggestion
    active = fields.Boolean()
```

estate/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
```
kfr-odoo, line 1: This header is not needed anymore. It should not be added to newly created python file in the Odoo codebase.

kfr-odoo (commented):
Your XML files are missing an end of file character. You can setup your IDE to do it automatically

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,89 @@
...
+        <field name="arch" type="xml">
+            <search string="Estate property">
+                <field name="name" string="Title"/>
+                <field name="postcode" string="Postcode"/>
+                <field name="expected_price" string="Expected Price"/>
+                <field name="bedrooms" string="Bedrooms"/>
+                <field name="living_area" string="Living Area (sqm)"/>
+                <field name="facades" string="Facades"/>
+
+                <filter string="Available" name="available" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"/>
```
kfr-odoo, line 80: https://github.com/odoo/tutorials/pull/44#discussion_r1530231420

estate/models/estate_property.py
```diff
@@ -0,0 +1,69 @@
...
+    @api.depends('offer_ids.price')
+    def _compute_best_price(self):
+        for record in self:
+            record.best_price = max(record.mapped('offer_ids.price'), default=0)
+
+    @api.onchange('garden')
+    def _onchange_garden(self):
+        if not self.garden:
+            self.garden_area = 0
+            self.garden_orientation = None
```
kfr-odoo, line 66: The "null" value used in Odoo is `False`. It does not change much but we tend to avoid setting odoo fields to `None` as the ORM will always use `False` by default.

estate/models/estate_property.py
```diff
@@ -0,0 +1,69 @@
...
+        else:
+            self.garden_area = 10
```
kfr-odoo, line 68: I don't really like magic numbers, it is always better to use some sort of constant variable with a meaningful nameeven if you only use it once

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,35 @@
...
+        ('refused', 'Refused'),
+    ], copy=False)
+
+    partner_id = fields.Many2one('res.partner', required=True)
+    property_id = fields.Many2one('estate.property', required=True)
+
+    validity = fields.Integer(string="Offer Validity (days)", default=7)
+    date_deadline = fields.Date(string="Deadline", compute="_compute_deadline_date", inverse="_inverse_deadline_date")
+
+    @api.depends('validity')
```
kfr-odoo, line 22: I know it is confusing because `create_date`is never supposed to change but you still have to tell the ORM that this method depends on it because it is used in the computation logic

estate/models/estate_property.py
```diff
@@ -0,0 +1,85 @@
...
+            self.garden_orientation = 'north'
+
+    def action_cancel(self):
+        for record in self:
+            if record.state == 'sold':
+                raise exceptions.UserError("Sold properties cannot be canceled!")
+            else:
+                record.state = 'canceled'
+        return True
```
kfr-odoo, line 77: 1. Your method was called using a button that only exists on the form view, it would not make any sense if the recordset was containing more than one record.
2. Make sure to leave the responsibility of CRUD actions to the ORM as often as possible to benefit from the implicit optimizations.
=> Note that `record.state = 'canceled` was completely fine tho
```suggestion
    def action_cancel(self):
        self.ensure_one()
        if self.state == 'sold':
            raise exceptions.UserError("Sold properties cannot be canceled!")
        
        return self.write({"state": "canceled"})
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,49 @@
...
+        for record in self:
+            if record.create_date:
+                record.validity = (record.date_deadline - fields.Date.to_date(record.create_date)).days
+            else:
+                record.validity = (record.date_deadline - fields.Date.today()).days
+
+    def action_accept(self):
+        for record in self:
+            if record.property_id.offer_ids.filtered(lambda r: r.status == 'accepted'):
+                raise exceptions.UserError("Another offer is already accepted on this property.")
```
kfr-odoo, line 40: FYI, an alternative that achieves the same result without the filter. I don't think there is a noticeable performance gap between both options
```suggestion
            accepted_offers = self.search([("property_id", "=", record.property_id), ("status", "=", "accepted")])
            if accepted_offers:
                raise exceptions.UserError("Another offer is already accepted on this property.")
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,49 @@
...
+            record.status = 'accepted'
+            record.property_id.selling_price = record.price
+            record.property_id.buyer_id = record.partner_id
```
kfr-odoo, line 43: Don't you want to also update the state of the `estate.property`record ? Also, it is debatable but I don't like that we change multiple values on a foreign model. IMO, it'd be better to have a method on `estate.property` that we could call from here and would manage the state of the record itself.

estate/models/estate_property.py
Clement-Cardot: Ok thanks,
@kfr-odoo is it better to define it inside the method or as a class constant var ?

estate/models/estate_property.py
Clement-Cardot: I understand 2. but for 1. would it be better to keep the for loop if after we want to add a feature that allow user to select some offers in order to cancel them all at once ?
Or even call this method in the accept method to cancel any other offers ?

What do you think ?

estate/models/estate_property.py
kfr-odoo: I would personnaly recommend a constant outside of the class

estate/models/estate_property.py
kfr-odoo: Avoid over-anticipation

estate/models/estate_property.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if self.state == 'canceled':
+            raise exceptions.UserError("Canceled properties cannot be sold!")
+
+        return self.write({'state': 'sold'})
+
+    def accept_offer(self, price, partner_id):
+        self.ensure_one()
+        self.selling_price = price
+        self.buyer_id = partner_id
```
kfr-odoo, line 100: Usually, we prefer to pass recordsets (odoo models) as parameter if multiple values are used.

```suggestion
    def accept_offer(self, offer):
        self.ensure_one()
        self.selling_price = offer.price
        self.buyer_id = offer.partner_id
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,113 @@
...
+    def _check_selling_price(self):
+        for record in self:
+            if not float_is_zero(record.selling_price, precision_digits=2) and float_compare(record.selling_price, record.expected_price * 0.9, precision_digits=2) == -1:
+                raise exceptions.ValidationError("The selling price cannot be lower than 90% of the expected price.")
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_only_new_or_sold(self):
+        for record in self:
+            if record.state not in ['new', 'sold']:
+                raise exceptions.UserError("You cannot delete a property that is not new or sold.")
```
kfr-odoo, line 112: Usererrors should be translatable using `_`

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,68 @@
...
+        self.property_id.accept_offer(self.price, self.partner_id)
+        return self.write({'status': 'accepted'})
+
+    def action_refuse(self):
+        self.ensure_one()
+        return self.write({'status': 'refused'})
+
+    @api.model_create_multi
+    def create(self, values):
+        best_price = self.env['estate.property'].browse(values[0]['property_id']).best_price
```
kfr-odoo, line 59: It feels off, what guarantees you that all offers have the same `property_id` ?

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,15 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
```
kfr-odoo, line 1: For information, this header is only required for modules inside Odoo/addons or enterprise/. Depending on the team you work for, you may not use it.

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="estate_property_offer_action" model="ir.actions.act_window">
```
kfr-odoo, line 3: Actions are always at the end of the file due to the sequential nature of XML parsing. To prevent reference of "not yet" created views

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,148 @@
...
+        <field name="name">estate.property.search</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <search string="Estate property">
+                <field name="name"/>
+                <field name="property_type_id"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area" filter_domain="[('living_area', '>=10_actions.html', self)]"/>
```
kfr-odoo, line 136: is it a typo or some black magic ?

kfr-odoo (commented):
Well done, it is really good.

estate/models/estate_property_offer.py
Clement-Cardot: The only way to create new offer for now is from a property page, therefore all the created offer should be link to the same property. But I agree, it is not a pretty way to handle it :/ And in the future it could be an issue...

Clement-Cardot: I tried something : https://github.com/odoo/tutorials/pull/43/commits/37417393ed8e2a81f124e1e56a2b8f91aff69cd5

awesome_owl/static/src/utils.js
```diff
@@ -0,0 +1,12 @@
+/** @odoo-module **/
+
+import { useRef, onMounted } from "@odoo/owl";
+
+function useAutoFocus(element_ref){
```
kfr-odoo, line 5: use camelcase for Javascript code
```suggestion
function useAutoFocus(elementRef){
```

awesome_owl/static/src/utils.js
```diff
@@ -0,0 +1,12 @@
+/** @odoo-module **/
+
+import { useRef, onMounted } from "@odoo/owl";
```
kfr-odoo, line 3: imports in alphabetical order

awesome_owl/static/src/todo/todo_list/todo_list.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+
+    <t t-name="awesome_owl.todo_list">
```
kfr-odoo, line 4: Pascal case here
```suggestion
    <t t-name="awesome_owl.TodoList">
```

awesome_owl/static/src/todo/todo_list/todo_list.js
```diff
@@ -0,0 +1,40 @@
...
+import { TodoItem } from "../todo_item/todo_item";
+import { useAutoFocus } from "../../utils";
+
+export class TodoList extends Component {
+    static template = "awesome_owl.todo_list";
+    static components = { TodoItem };
+
+    setup() {
+        this.todos = useState([]);
+        this.count = useState({ value: 0 });
```
kfr-odoo, line 13: It should be clear that you manipulate reactive values. However, I don't understand why `count` has to be reactive. How does it affect UI rendering ?
```suggestion
        this.state = useState({
            todos: [],
            count: 0
        })
```

awesome_owl/static/src/playground.xml
```diff
@@ -5,6 +5,27 @@
...
+        <Counter onChange.bind="incrementSum"/>
+        <Counter onChange.bind="incrementSum"/>
+
+        <p>Sum is : <t t-esc="state.value"/></p>
+
+        <Card title="'Card 1'">
+            <t t-out="this.html"/>
+        </Card>
+        <Card title="'Card 2'">
+            <p>If you read this : Hello :)</p>
```
kfr-odoo, line 17: 👋

awesome_owl/static/src/counter/counter.js
```diff
@@ -0,0 +1,22 @@
...
+
+    setup() {
+        this.state = useState({ value: 0 });
+    }
+
+    increment() {
+        this.state.value++;
+        if (this.props.onChange) {
+            this.props.onChange();
+        }
```
kfr-odoo, line 20:
```suggestion
       this.props.onChange?.()
```

PR #51, opened 2024-03-19
segn - Technical Training

seifgneedy:


estate/models/estate_property.py
```diff
@@ -20,3 +20,12 @@ class EstateProperty(models.Model):
...
         string='Orientation',
         selection=[('North', 'North'), ('South', 'South'), ('East', 'East'), ('West', 'West')])
+    state = fields.Selection(
+        default='New',
+        selection=[
+            ('New', 'New'),
+            ('Offer Received', 'Offer Received'),
+            ('Offer Accepted', 'Offer Accepted'),
+            ('Sold', 'Sold'),
+            ('Canceled', 'Canceled')])
```
vchu-odoo, line 30: It is better to use python like style names for selection as a convention, so that it would be convenient to not to think about exact user's text if it is with uppercase or not/space or underscore and etc.
```suggestion
        default='new',
        selection=[
            ('new', 'New'),
            ('offer_received', 'Offer Received'),
            ('offer_accepted', 'Offer Accepted'),
            ('sold', 'Sold'),
            ('canceled', 'Canceled')])
```

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<odoo>
+    <menuitem id="estate_menu_root" name="Real Estate">
+        <menuitem id="estate_first_level_menu" name="Advertisements">
```
vchu-odoo, line 4: Could be useful to have meaningful ids.
```suggestion
        <menuitem id="estate_property_menu_advertisements" name="Advertisements">
```

estate/__manifest__.py
```diff
@@ -5,6 +5,8 @@
     'application': True,
     'data': [
         'security/ir.model.access.csv',
+        'views/estate_property_views.xml',
+        'views/estate_menus.xml',
```
vchu-odoo, line 9: Usually we order files alphabetically.
```suggestion
        'views/estate_menus.xml',
        'views/estate_property_views.xml',
```

vchu-odoo (commented):
Good job!

Only small nitpickings.

estate/__manifest__.py
seifgneedy: I have changed the order of the files as you suggested but I'm facing the following error while activating the app in Runbot:
`ValueError: External ID not found in the system: estate.estate_property_action`

That error is because we have dependency between the files as `estate_property_action` needs to be defined first before being used in `estate_menus.xml` and The data files are sequentially loaded following their order in the `__manifest__.py` file.
That's why we need to have files in that order so that the action is defined in `estate_property_views.xml` before being used in `estate_menus.xml`.

vchu-odoo: Yes, I missed that, my bad.

estate/views/estate_property_views.xml
```diff
@@ -5,4 +5,76 @@
         <field name="res_model">estate.property</field>
         <field name="view_mode">tree,form</field>
     </record>
+
+    <record id="estate_prperty_view_tree" model="ir.ui.view">
+        <field name="name">estate.property.tree</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <tree string="Properties">
+                <field name="name" string="Title"/>
```
vchu-odoo, line 14: To not duplicate `string="Title"` everywhere you can add it directly in the field `name = fields.Char(required=True, string="Title")` in `estate_property.py`

estate/views/estate_property_views.xml
```diff
@@ -5,4 +5,76 @@
...
+        <field name="arch" type="xml">
+            <search string="Search Properties">
+                <field name="name" string="Title"/>
+                <field name="postcode" string="Postcode"/>
+                <field name="expected_price" string="Expected Price"/>
+                <field name="bedrooms" string="Bedrooms"/>
+                <field name="living_area" string="Living Area (sqm)"/> 
+                <field name="facades" string="Facades"/>
+
+                <filter name="available" string="Available" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"/>
```
vchu-odoo, line 75: Just an alternative way
`domain="[('state', 'in', ('new', 'offer_received'))]"`

vchu-odoo (commented):
Great job!

as always only small suggestions.

estate/models/estate_property.py
```diff
@@ -1,4 +1,4 @@
-from odoo import fields, models
+from odoo import fields, models, api
```
vchu-odoo, line 1: Usually we try to order it alphabetically
```suggestion
from odoo import api, fields, models
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real estate Property Field"
+
+    name = fields.Char(required=True, string='Title')
+    description = fields.Text(string='Description')
+    postcode = fields.Char(string='Postcode')
+    date_availability = fields.Date(copy=False, default=fields.Date.add(fields.Date.today(), days=90), string='Available From')
```
vchu-odoo, line 11: Not all months consist of 30 days, so to be more precise you can use months option.
```suggestion
    date_availability = fields.Date(copy=False, default=fields.Date.add(fields.Date.today(), months=3), string='Available From')
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,96 @@
...
+        <field name="arch" type="xml">
+            <search string="Search Properties">
+                <field name="name"/>
+                <field name="property_type_id"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/> 
+                <field name="facades"/>
+
```
vchu-odoo, line 90:
```suggestion
```

vchu-odoo (commented):
Good pace!

Just I noticed when tested that in the offer form view when I change deadline date the validity is not recalculated.

seifgneedy:
> Good pace!
> 
> Just I noticed when tested that in the offer form view when I change deadline date the validity is not recalculated.

Thank you, @vchu-odoo !
As the validity is recalculated by an inverse function, it's called when saving the record.
`Note that the inverse method is called when saving the record, while the compute method is called at each change of its dependencies.`

vchu-odoo:
> > Good pace!
> > Just I noticed when tested that in the offer form view when I change deadline date the validity is not recalculated.
> 
> Thank you, @vchu-odoo ! As the validity is recalculated by an inverse function, it's called when saving the record. `Note that the inverse method is called when saving the record, while the compute method is called at each change of its dependencies.`

Indeed, you are right, all good!

vchu-odoo (commented):
Congrats on completing the python tutorial! :+1: 

You can check the ci/style to polish the code :)

PR #53, opened 2024-03-20
Excercise Ch3-Ch7

stta-odoo:
1. setup manifest 2. setup access right 3. create estate model 4. create estate view


stta-odoo:
Good job!

PR #55, opened 2024-03-30
[FIX] awesome_tshirt: use {'clickable': '1'} option

kobros-tech:
This outdated clickable attribute should be assigned in options attribute object which is used for statusbar widget

### DIFF
--- a/awesome_tshirt/views/views.xml
+++ b/awesome_tshirt/views/views.xml
@@ -8,7 +8,7 @@
       <field name="arch" type="xml">
         <form>
           <header>
-            <field name="state" widget="statusbar" options="{'clickable': 1}"/>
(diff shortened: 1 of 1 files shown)


awesome_tshirt/views/views.xml
```diff
@@ -8,7 +8,7 @@
       <field name="arch" type="xml">
         <form>
           <header>
-            <field name="state" widget="statusbar" clickable="1"/>
+            <field name="state" widget="statusbar" options="{'clickable': 1}"/>
```
AntoineVDV, line 11:
```suggestion
            <field name="state" widget="statusbar" options="{'clickable': '1'}"/>
```

AntoineVDV (approved):
Thanks for the fix! :)

PR #78, opened 2024-05-27
17.0 luth JS tutorials

lucasth-dev:


awesome_owl/static/src/utils.js
```diff
@@ -0,0 +1,14 @@
...
+
+import { useRef, onMounted } from "@odoo/owl"
+
+export const useAutofocus = (refName) => {
+    const elementRef = useRef(refName)
+    onMounted(() => {
+        if(elementRef.el !== null) {
+            elementRef.el.focus()
+        }
+    })
```
davidmonnom, line 11: You can try to create a hook for this :)

awesome_owl/static/src/todolist/todoitem.js
```diff
@@ -0,0 +1,23 @@
...
+        todo: {
+            id: { type: Number },
+            description: { type: String },
+            isCompleted: { type: Boolean },
+        },
+        removeTodo: { type: Function },
+    };
+
+    toggleState() {
+        this.props.todo.isCompleted = !this.props.todo.isCompleted
```
davidmonnom, line 17: It's not really a good way to mutate a props. It's better to pass a function as a props and call it

PR #89, opened 2024-06-26
Chapter 11: Constraints

StartingFrom101:
Added: 
- A property expected price must be strictly positive
- A property selling price must be positive
- An offer price must be strictly positive
- A property tag name and property type name must be unique


StartingFrom101:
Wrong base, apologies

PR #107, opened 2024-08-02
[ADD] estate: create estate.property model and initial views

amya-odoo:
Training adding the estate module


adsh-odoo (commented):
Hello @amya-odoo

I have added some comments and a couple of suggestions.
Can you give it a try?

Thanks!!

PR #109, opened 2024-08-02
 [ADD] estate:  Created a new module called Estate Property

pach-odoo:
Training adding the estate module


estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,59 @@
...
+                            <div class="o_stat_info">
+                                <span class="o_stat_value">
+                                    <field name="offer_count"/>
+                                </span>
+                                <span class="o_stat_text">
+                                    Offers
+                                </span>
+                            </div>
+                        </button>
+                    </div>
```
adsh-odoo, line 40:
```suggestion
<div name="button_box">
                        <button class="oe_stat_button" type="action" name="%(estate.estate_property_offer_actions)d"
                            icon="fa-ticket">
                            <field string="Offers" name="offer_count" widget="statinfo"/>
                        </button>
                    </div>
```
we can do something like this to well format the stat button

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @api.depends("property_id.create_date", "validity")
+    def _compute_date_deadline(self):
+        for record in self:
+            if record.create_date:
+                record.date_deadline = record.property_id.create_date + timedelta(
+                    days=record.validity
+                )
+            else:
+                record.date_deadline = date.today() + timedelta(days=record.validity)
```
adsh-odoo, line 37:
```suggestion
       record.date_deadline = record.property_id.create_date + relativedelta(days=record.validity)
```
can you give it a try?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def _inverse_date_deadline(self):
+        for record in self:
+            if record.date_deadline and record.property_id.create_date:
+                delta = record.date_deadline - fields.Date.to_date(
+                    record.property_id.create_date
+                )
+                record.validity = delta.days
+            else:
+                record.validity = 7
```
adsh-odoo, line 47:
```suggestion
record.validity = (record.date_deadline - record.property_id.create_date.date()).days
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+            else:
+                self.write({"state": "canceled"})
+        return True
+
+    def action_sold_property(self):
+        for record in self:
+            if record.state == "canceled":
+                raise UserError(_("Canceled property can not be sold."))
+            else:
+                self.write({"state": "sold"})
```
adsh-odoo, line 124:
```suggestion
                record.state = "sold"
```
Try to avoid direct usage of write method when we are updating a single field. check in other files also
And we have to use record here instead of self as we are looping through self using record.

adsh-odoo (commented):
Hello @pach-odoo 

Overall Looking good.
I have added some comments to polish the code.
Can you please have a look?

Thanks!!

PR #111, opened 2024-08-02
[ADD] estate:  Created Real Estate Application

akpu-odoo:
Real Estate Module for training


estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+            ('Accepted', 'Accepted'),
+            ('Refused', 'Refused'),
+        ],
+        string="Status"
+    )
+    partner_id = fields.Many2one('res.partner', required=True)
+    property_id = fields.Many2one('estate.property', required=True)
+    validity = fields.Integer(default=7)
+    deadline_date = fields.Date(compute="_compute_validity", inverse="_inverse_validity",store=True)
+    created_date = fields.Date(default=lambda self: datetime.today())
```
adsh-odoo, line 22:
```suggestion
    create_date = fields.Date(default=lambda self: datetime.today())
```
I think we can override the existing create_date field here instead a new field.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+    accepted = fields.Boolean(default=False)
+
+    @api.depends("validity", "created_date")
+    def _compute_validity(self):
+        for record in self:
+            if record.created_date:
```
adsh-odoo, line 28: As we are providing the default value in the field so I think there is no need for this extra check

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+                record.deadline_date = record.created_date + timedelta(days=record.validity)
+            else:
+                record.deadline_date = False
+                
+    @api.depends("created_date", "deadline_date")
```
adsh-odoo, line 33: No need for this i think we can keep the generic behavior of inverse method that reflects the changes after saving the records in our usecase.

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+    @api.depends("offer_ids.price")
+    def _compute_price(self):
+        for record in self:
+            if record.offer_ids:
+                record.best_offer =  max(record.offer_ids.mapped('price'))
+            else:
+                record.best_offer = 0
+    
+    @api.onchange("garden")
+    def _onchange_(self):
```
adsh-odoo, line 67: Give the method name as per the use cases and try avoiding the generic keywords

adsh-odoo (commented):
Hello @akpu-odoo 

Added few remarks primarily related to guidelines for cleaning up the code.
Beside this the checkstyle of runbot is also failing due to linting issue Can you please resolve that also.

Thanks

PR #113, opened 2024-08-12
[ADD] estate: created estate module

irah-odoo:
create a estate module for training


adsh-odoo (commented):
Hello @irah-odoo 

I have left some remarks and a couple of suggestions.
Can you please have a look?

Thanks!!

estate/models/estate_property.py
```diff
@@ -0,0 +1,82 @@
+from odoo import models, fields, api
+from dateutil.relativedelta import relativedelta
+from datetime import date
+from odoo.exceptions import UserError, ValidationError
+
+
+class estate_property(models.Model):
+    _name = "estate_property"
```
adsh-odoo, line 8:
```suggestion
    _name = "estate.property"
```
we use . while giving model name as per convention make sure to check in other files also.

PR #115, opened 2024-08-14
[ADD] estate: added estate module

prgo-odoo:
After this commit
====
-Created Estate Module
-Implemented basic fields
-Implemented access rights
-Implemented basic views like tree and form view
-Implemented relational fields to the module
-Implemented compute fields and on change 
-Implemented buttons for action
-Implemented python and SQL constraints
-Implemented widgets,inline views and make the views more appealing
-Implemented python inheritance, model inheritance and view inheritance


shag-odoo:
@prgo-odoo Could you please update the PR description
Instead of added replace it with implemented and Can I know where you have implemented architecture overview?

PR #118, opened 2024-08-14
[ADD] estate: added estate module

rusi-odoo:
In this commit
====
- Implemented  models 
- Implemented views , menu and action for all the corresponding models 
- Implemented relational fields and compute fields 
- Implemented sql constraints and python constrains
- Implemented different type of button like object and action
- inherited res_user model and its view 
- Implemented method overriding


shag-odoo:
@rusi-odoo Could you please update PR title , no capitals letters in the title instead of tag
and update the PR description from overridded  to override

Thanks :innocent:

PR #166, opened 2024-10-21
Test commit & push MEPL

Mathilde411:


Mathilde411:
> Hey @Mathilde411, no worries, I'll push it again on your branch (but be careful when working with someone in the future, as you might delete his/her work). Go on with your tutorial for now and when you'll want to push, try to fetch this first (and make it work). 👍 😄

Yeah, I wasn't really thinking that someone else could have pushed on my branch right now ^^

proose:
> Yeah, I wasn't really thinking that someone else could have pushed on my branch right now ^^
I'll be more careful next time :)

People should warn you beforehand, I was just testing you a bit haha
I would suggest you use a --force-with-lease instead of a --force with your push command, to avoid this kind of issue (more info at https://stackoverflow.com/questions/52823692/git-push-force-with-lease-vs-force) :)

estate_account/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
+    'name': 'Real Estate Invoicing Link',
+    'version': '1.0',
+    'category': 'Tutorials',
+    'depends': [
+        'base',
```
proose, line 6: This is not needed as the inheritance path of account already include `base`

proose (commented):
A bit of nitpicks, but great job in general !
Very well done !

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,128 @@
...
+    invoice_id = fields.Many2one("account.move", string='Invoices', related='invoice_line_id.move_id',
+                                 readonly=True,
+                                 copy=False)
+
+    # invoice_status = fields.Selection([
+    #    ('upselling', 'Upselling Opportunity'),
+    #    ('invoiced', 'Fully Invoiced'),
+    #    ('to invoice', 'To Invoice'),
+    #    ('no', 'Nothing to Invoice')
+    # ], string='Invoice Status', compute='_get_invoice_status', store=True, readonly=True)
```
proose, line 20: Don't forget to remove your old comments

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,128 @@
...
+        """Prepare the values to create the new invoice line for a sales order line.
+
+        :param optional_values: any parameter that should be added to the returned invoice line
+        :rtype: dict
+        """
+        self.ensure_one()
+
+        res = {
+            'display_type': 'product',
+            'sequence': sequence,
```
proose, line 54: The sequence is already automatically handled, no use adding it here

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,128 @@
...
+            'quantity': 1,
+            'price_unit': self.selling_price,
+            'property_ids': [fields.Command.link(self.id)],
+        }
+
+        return res
+
+    # trying to understand sales invoice creation line by line
+    def _create_invoices(self, grouped=False, date=None):
+        if not self.env['account.move'].has_access('create'):
```
proose, line 65: The ORM handles access errors by itself, you can keep it simpler :D

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,128 @@
...
+
+                ref_invoice_vals.update({
+                    'ref': ', '.join(refs)[:2000],
+                    'invoice_origin': ', '.join(origins)
+                })
+                new_invoice_vals_list.append(ref_invoice_vals)
+            invoice_vals_list = new_invoice_vals_list
+
+            return self.env['account.move'].sudo().with_context(default_move_type='out_invoice').create(
+                invoice_vals_list)
```
proose, line 117: This whole method is way too complicated for our case. You can practically use the same example as shown in the tutorial to create a move with the correct invoice_line_ids ;)

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,128 @@
...
+    def action_view_invoice(self):
+        action = self.env['ir.actions.actions']._for_xml_id('account.action_move_out_invoice_type')
+        form_view = [(self.env.ref('account.view_move_form').id, 'form')]
+        if 'views' in action:
+            action['views'] = form_view + [(state, view) for state, view in action['views'] if view != 'form']
+        else:
+            action['views'] = form_view
+        action['res_id'] = self.invoice_id.id
+
+        return action
```
proose, line 128: This is also too complicated with what is needed. In python, you can return a dict to communicate an action (unless the method is called from a simple other python method). Your case should not need more than this (you'll probably want to adapt it to make it work):
```suggestion
        return {
            'name': 'Invoice',
            'view_mode': 'form',
            'res_model': 'account.move',
            'type': 'ir.actions.act_window',
            'res_id': self.invoice_id.id,
        }
```

proose (commented):
A bit more nitpicks :D 

More hints: usually we try to keep the file names and class names close to the model name: model 'estate.property' -> file 'estate_property.py' and class 'EstatePropery'. This allow the devs to more easily search the file we need and it appears more clear when looking into files with multiple models inside.

It's a good thing to go look into what's existing in Odoo - you'll learn a lot. But it is usefull to look at different implementations of what you're looking for: some modules are heavy in processing while others might be way more simple and quick... It depends on what's needed :man_shrugging: .

Very good work so far, you're quick to learn !

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,71 @@
...
+        </field>
+    </record>
+
+    <record id="estate_property_type_view_form" model="ir.ui.view">
+        <field name="name">estate.property.type.form</field>
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <form string="Property Types">
+                <sheet>
+                    <div class="oe_button_box" name="button_box">
```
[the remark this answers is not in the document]
Mathilde411: It is here to let modules add box buttons without stepping on each other. 
Ex: If 2 modules add buttons they already have the parent div, and do not need to step on each other to add it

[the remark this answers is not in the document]
Mathilde411: Isn't it like that to add smart buttons ? I saw it in the sales module.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,152 @@
...
+                    <button name="action_canceled" type="object" string="Cancel"
+                            invisible="state in ['sold', 'canceled']"/>
+                    <button name="action_reset" type="object" string="Reset"
+                            invisible="state not in ['sold', 'canceled']"/>
+                    <field name="state" widget="statusbar" statusbar_visible="new,offer_received,offer_accepted,sold"/>
+                </header>
+                <sheet>
+                    <div class="oe_button_box" name="button_box">
+
+                    </div>
```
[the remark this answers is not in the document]
Mathilde411: It is here to let modules add box buttons without stepping on each other.
Ex: If 2 modules add buttons they already have the parent div, and do not need to step on each other to add it

estate/models/property_offer.py
```diff
@@ -0,0 +1,74 @@
...
+                record.property_id.set_accepted_offer(None)
+            record.state = 'refused'
+        return True
+
+    def action_reset(self, propagate=True):
+        for record in self:
+            if record.state == 'accepted' and propagate:
+                record.property_id.set_accepted_offer(None)
+            record.state = 'received'
+        return True
```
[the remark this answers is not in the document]
proose: While it can be true in "classic" frameworks, in Odoo, a lot of modules depend on each other, and might want to adapt the behavior of the inherited model, or add features. If we add everything in the inherited class, it would become monstrously huge and hard to work in.

In that sense, we can apply the same thinking with files within a same module: as the "offer acceptance" comes from this file, it can handle the property variables by itself. We don't usually use setters and getters unless there are strong conditions to respect.

At the end of the day, it is just a (heavy) preference and your way of doing might be accepted, I was particularly nitpicky because your implem was very well done :smile:

estate/views/estate_property_type_views.xml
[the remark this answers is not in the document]
proose: Have you tried without ? I think this should work with just the inside node (`<button class="oe_stat_button"...`), if not, you can ignore my comment

estate/views/estate_property_views.xml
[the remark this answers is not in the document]
proose: That is good thinking if you know it will be used for inheritance ! But usually you never know if that will be the case, and we avoid adding extra nodes like this unless we know this part will particularly be inherited multiple times. Or if it can't be done otherwise, but here you can use `expr="//div[@class='oe_title']"` in your inheriting view to get it.

In this case I insists because I don't think an extra `<div class="oe_button_box" name="button_box">` is needed around the oe-stat button, but if it is, you can eventually leave it like this

estate_account/models/estate_property.py
proose: That would be nice indeed. You can then use a small method that can add a invoice line to an existing move, or put that simple code directly into the method where you create the move (assuming you change a bit the flow to allow creating a move after selecting multiple properties for example), but this code is still doing way too complicated (and useless in our case) things :P

PR #167, opened 2024-10-21
first commit

dabo-odoo:


Megaaaaaa (changes requested):
Hello :wave: here's a first review for you :smile:

Some teams won't be as nitpicking as i was but in all cases thoses guidelines should get most people to approve your code :+1:

PR #169, opened 2024-10-21
feat: estate

fres-sudo:
frca

### DIFF
--- a/estate/models/estate_property.py
+++ b/estate/models/estate_property.py
@@ -1,0 +1,9 @@
+from odoo import models, fields
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate Property"
+
+    name = fields.Char("Estate Name", required=True, translate=True)
+    price = fields.Integer("Estate Price", default=0)
--- a/estate/__init__.py
+++ b/estate/__init__.py
@@ -1,0 +1,3 @@
+from . import models
+
+
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+{
+    'name': 'Estate',
+    'category': 'Tutorials/Estate',
+    'summary': 'Real estate application',
+    'description': "",
+    'depends': [
+        'base_setup',
+    ],
+    'installable': True,
+    'application': True,
+    'auto_install': False
+}


estate/models/estate_property.py
```diff
@@ -0,0 +1,111 @@
...
+            if float_compare(record.expected_price * 0.9, record.selling_price, 3) == 1 and record.selling_price != 0:
+                raise ValidationError(f"The selling price must be at least the 90% of the expected price.")
+
+
+    def action_cancel(self):
+        if self.state != 'cancelled':
+            self.state = 'cancelled'
+
+    def action_sold(self):
+        if self.state == 'cencelled':
```
proose, line 84: Careful with keywords' spelling, or the code will just not work :)

estate/models/estate_property.py
```diff
@@ -0,0 +1,111 @@
...
+        for record in self:
+            record.total_area = record.living_area + (record.garden_area or 0)
+
+    @api.depends("offer_ids.price")
+    def _compute_best_offer(self):
+        for record in self:
+            if record.offer_ids:
+                record.best_offer = max(offer.price for offer in record.offer_ids)
+            else:
+                record.best_offer = 0  # Set to 0 if no offers are present
```
proose, line 100: It's good to want to add comments for better readability for the one that might look or review your code, but they should be reserved for when the use case is hard to comprehend. You can leave these ones, but your code is well explicit enough so you don't need to bother next time :smile: (but please write some is the code gets more complicated)
Just to complete what I write here: comments should explain *why* the code is needed, in complicated cases, not *what* it does, the code should be self-explanatory for the last part

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,79 @@
...
+                record.validity = (record.date_deadline - record.create_date.date()).days
+            else: # Fallback
+                record.validity = 0
+
+
+    @api.onchange('validity')
+    def _onchange_validity(self):
+        for record in self:
+            if record.create_date:
+                record.date_deadline = record.create_date + relativedelta(days=record.validity)
```
proose, line 79: You already have a compute that will do the same as this method, so you can remove this

estate/models/estate_property.py
```diff
@@ -0,0 +1,111 @@
...
+    total_area = fields.Float('Total Area (sqm)', compute='_compute_total_area', readonly=True)
+    best_offer = fields.Float('Best Offer', compute='_compute_best_offer', readonly=True)
+    type_id = fields.Many2one('estate.property.type')
+    _order = 'id desc'
+
+    _sql_constraints = [
+        ('expected_price', 'CHECK(expected_price > 0)',
+         'The expected price should be strictly grater than 0.'),
+        ('selling_price', 'CHECK(selling_price > 0)',
+         'The selling price should be strictly grater than 0.'),
```
proose, line 51: Don't forget that the selling price might be 0 if no offer has been accepted yet

estate/views/res_users_views.xml
proose, line 1: Be careful that the list you show on the user allows to create a new line (so a new property), but it crashes. You can test that by going to the Settings app -> Users & Companies and selecting the user you are on (Mitchell Admin if you logged in with admin credentials) -> Real Estate properties

proose (commented):
Nicely done so far ! Beware of spelling mistakes, I only picked the most important ones but there are many more. Try to correct everything I spotted here and finish by making runbot green !
Good work ! :muscle:

proose:
@fres-sudo Nicely done ! I see every comment I made were well corrected ! Last thing to do is to try to push your changes in one whole commit. Beware to only put your work in one commit, and not the work done by others before you. Tips: create a temporary branch before playing with your commits (`git checkout 18.0-tutorial-frca` -> `git branch temp-branch-before-squash`), just to be sure not to lose everything :D

PR #172, opened 2024-10-21
[ADD] estate: create real estate module

barracudapps:


estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,119 @@
...
+
+                <filter string="Status" name="status" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"/>
+                <filter string="High Value" name="high_value" domain="[('expected_price', '&gt;', 1000000)]"/>
+                <filter string="Low Value" name="low_value" domain="[('expected_price', '&lt;', 1000000)]"/>
+                <separator/>
+                <filter string="Include archived" name="archived" domain="['|', ('active', '=', True), ('active', '=', False)]"/>
+                <filter string="Archived" name="inactive" domain="[('active', '=', False)]"/>
+
+                <group expand="0" string="Group By">
+                    <filter string="Active" name="active" domain="['|', ('active', '=', True), ('active', '=', False)]" context="{'group_by': 'active'}"/>
```
Megaaaaaa, line 24: Again same as above :smile:

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,119 @@
...
+                    <filter string="Postcode" name="postcode" domain="[]" context="{'group_by': 'postcode'}"/>
```
Megaaaaaa, line 25: No need for the domain here either :+1:

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,119 @@
...
+            <search string="Search Properties">
+                <field name="name"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="facades"/>
+                <field name="tags"/>
+
+                <filter string="Status" name="status" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"/>
```
Megaaaaaa, line 16: This filter name is strange, It doesn't really tell what it filters and i doubt you can guess it from what will result of the filtering. Maybe calling it "Pending" or something like that would make it clearer ?

```suggestion
                <filter string="Pending" name="pending" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,119 @@
...
+        </field>
+    </record>
+
+    <record id="estate_property_view_form" model="ir.ui.view">
+        <field name="name">estate.property.view.form</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form string="Property">
+                <sheet>
+                    <widget name="web_ribbon" title="Archived" bg_color="bg-danger" invisible="(active)"/>
```
Megaaaaaa, line 37: No parenthesis inside invisible attribute :wink: There are a bunch of other similar issues in this file, please address those too :+1:

estate/models/estate_property.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+    description = fields.Text(
+        string="Description",
+        required=False,
+        translate=True,
+        copy=False,
+    )
+    image = fields.Image(
+        "Image",
+        copy=False,
```
Megaaaaaa, line 22: I find it strange that you use so much the `copy=False`, so I'll just explain it quickly and then you can decide whether that's what you want or not.

`copy=False` makes it so that when you duplicate a record of the model, the fields with `copy=False`'s values will not be set in the new duplicated record. Usually we use `copy=False` on fields that limitate the possibilities on a record so that the new record can be modified as much as desired before validation. For example, we don't copy the state of an invoice when we duplicate it beause if it is validated, then you can't edit the customer, the date, the products, etc on the new record. But if you put too many `copy=False` then it just acts as the `New` button because it will create an empty record. So you want to use as little `copy=False` as possible to have the closest copy of the duplicated record as possible while also maintaining manoeuvrability.

In this case, I don't think `image` should have `copy=False` for example but maybe you meant to do it so I'll let you decide which fields really needs to be `copy=False` and we'll probably talk about it again later anyway :smile:

Megaaaaaa (changes requested):
Hello :wave:

Here are a few comments for your first review. Mostly styling and guidelines, nothing major, have fun :smile:

PR #178, opened 2024-10-25
18.0 tutorial drod

vandroogenbd:


estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,85 @@
...
+                record.property_id.selling_price = 0.0
+                record.property_id.buyer = None
+                record.property_id.status = "offer_received"
+            record.status = "refused"
+        return True
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for record in vals_list:
+            _property = self.env["estate_property"].browse(record["property_id"])
```
proose, line 78: Using an underscore before a variable is usually to indicate a private variable, or a value that won't be used (if we use a method that returns a tuple of values for example). Best not to use it here

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,85 @@
...
+            if record["price"] < _property["best_offer"]:
```
proose, line 79:
```suggestion
            if record["price"] < _property.best_offer:
```
`self.env['model.name'].browse(SOMEID)` will return a recordset, so you can (and should preferably) get its fields like attributes

estate/static/description/icon.webp
proose, line 1: Why and where is this file needed ?

estate/views/estate_menu_views.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0"?>
+<odoo>
+
+    <!-- TOP MENU -->
+    <menuitem 
+        id="estate_menu_root"
+        name="Estate"
+    />
+
+    <!-- HEADER: Advertisements -->
```
proose, line 10: You don't need these comments. If you don't find this way clear enough, you can nest your menuitems:
```
<menuitem id="coucou">
    <menuitem id="salut"/>
<menuitem/>
```

"salut" sera un sous-menuitem de "coucou"

estate/models/estate_property_infos.py
```diff
@@ -0,0 +1,52 @@
...
+class Estate_Property_Tag(models.Model):
+    _name = "estate_property_tag"
+    _description = "Estate property Tags"
+    _order = "name"
+
+    name = fields.Char(required=True, string="Type")
+
+    color = fields.Integer(string="Colour")
+
+    property_estate_ids = fields.Many2many(
```
proose, line 46: This is unnecessary, we just want the tags on the properties, but we don't really care to find properties from a tag (a search by tags on the properties list view will suffice if we want to find properties with a special tag

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,173 @@
...
+                        <field name="tag_ids" widget="many2many_tags"
+                            options="{'color_field': 'color'}" />
+                    </div>
+
+                    <separator />
+
+                    <group>
+                        <group>
+                            <field name="type_id" attributes="{'can_write': false}"
+                                options="{'no_create': true, 'no_open': true}" />
```
proose, line 60: Why those options and attribute ?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,173 @@
...
+            </list>
+        </field>
+    </record>
+
+    <!-- Kanban view -->
+    <record id="estate_property_view_kanban" model="ir.ui.view">
+        <field name="name">estate_property_kanban</field>
+        <field name="model">estate_property</field>
+        <field name="arch" type="xml">
+            <kanban class="o_kanban_mobile" default_group_by="type_id">
```
proose, line 133: Don't forget your kanban view shouldn't allow drag and drop

estate_account/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# A custom real estate module, created for learning purposes
+
+{
+    'name': '[TUTO] Estate_Account',
+    'summary': 'Module to invoice sold real estate properties',
+    'depends': [
+        'base_setup', 'estate', 'account'
```
proose, line 7: 'base_setup' here is not needed, as it is already a parent of 'account' which this module depends on ;)

estate_account/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    ],
+    'license': "LGPL-3",
+    'data': [
```
proose, line 10: No need to add empty list in the manifest

proose (commented):
Very very nice work ! You thought about (almost) *every* corner cases, well done !
You can quickly work on the different comments I sent you and squash your commits in one (only yours this time :smile: ).
Well done !

PR #190, opened 2024-11-19
18.0 training amah

amah-odoo:


ushyme (commented):
Good work ! Just a few remarks we already talked about.

PR #221, opened 2025-01-16
[ADD] estate: add real estate module

nipl-odoo:
Create real estate module for training purposes.


estate/models/estate_property.py
```diff
@@ -0,0 +1,144 @@
+from odoo import models, fields, api
+from odoo.exceptions import UserError
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
```
adsh-odoo, line 4:
```suggestion
from datetime import datetime
from dateutil.relativedelta import relativedelta

from odoo import api, fields, models
from odoo.exceptions import UserError
```
Generally we first make external imports and then we import from odoo in alphabetical order.
Need to adapt in all files

adsh-odoo (commented):
Hello @nipl-odoo 
Some comments/suggestions
I can also see some extra blank lines remove them also.
Thanks!!

PR #233, opened 2025-01-22
[ADD] module_name: Display 'Book Price' on Sales Order and Invoice Lines

yame-odoo:
Added a computed field book_price to Sales Order Lines and Account Move Lines to
 display the original product price ('Book Price') based on the selected
 product

Updated Sales Order Lines list view to include the new 'Book Price' field.

Updated Invoice Lines list view (customer invoices only) to include the new 'Book Price' field.

Task-4496458


product_warranty/views/product_warranty_views.xml
```diff
@@ -0,0 +1,28 @@
...
+        <field name="model">product.warranty.configuration</field>
+        <field name="arch" type="xml">
+            <list string="Channel" editable="top">
+                <field name="name" string="name" />
+                <field name="percentage" string="percentage" />
+                <field name="years" string="years" />
+                <field name="product_template_id" string="product name"/>
+
+
+
```
adsh-odoo, line 23:
```suggestion
```

product_warranty/views/sale_menus.xml
```diff
@@ -0,0 +1,12 @@
...
+
+
+    <menuitem id="menu_warranty_config"
+        name="warranty configuration"
+        action="product_warranty_action"
+        parent="sale.menu_sale_config"
+        sequence="1" />
+
+
+</odoo>
```
adsh-odoo, line 12:
```suggestion
<odoo>
    <menuitem id="menu_warranty_config"
        name="warranty configuration"
        action="product_warranty_action"
        parent="sale.menu_sale_config"
        sequence="1" />
</odoo>
```

product_warranty/wizard/add_warranty.py
```diff
@@ -0,0 +1,53 @@
...
+        vals = super().create(vals_list)
+        sale_order = self.env["sale.order"].search(
+            [("id", "=", self.env.context.get("active_id"))]
+        )
+        for record in vals:
+            for warranty in record.list_warranty_ids:
+                if warranty.warranty_id:
+                    warranty_price = (
+                        warranty.warranty_id.percentage / 100
+                    ) * warranty.product_id.list_price
```
adsh-odoo, line 41:
```suggestion
                        warranty.warranty_id.percentage * warranty.product_id.list_price * 0.01
```
We can do something like this directly while creating sale_order_line instead of calculating here

product_warranty/wizard/add_warranty.py
```diff
@@ -0,0 +1,53 @@
...
+                    "order_line_id": product_id[2],
+                },
+            )
+        warranty_ids = self.env["product.list.warranty"].create(ids)
+        defaults["list_warranty_ids"] = warranty_ids
+        return defaults
+        
+
+    @api.model_create_multi
+    def create(self, vals_list):
```
adsh-odoo, line 31: I don't think we need to override the create method here.
Instead what we can do is we can make a custom method that will do this logic and will be called from the wizard's footer button.

adsh-odoo: something like this from wizard side 
```xml 
                <footer>
                    <button string="Add" name="add_warranties_to_order_line" type="object"/>
                    <button string="Cancel" special="cancel"/>
                </footer>```
                
  And this method add_warranties_to_order_line will perform the logic.

File: product_warranty/wizard/list_warranty.py [hunk 5]
```diff
@@ -0,0 +1,37 @@
...
+    def _onchange_enddate(self):
+        for record in self:
+            if record.warranty_id:
+                record.enddate = str(
+                    (
+                        datetime.today() + relativedelta(years=record.warranty_id.years)
+                    ).date()
+                )
+            else:
+                record.enddate = ""
```
adsh-odoo, line 35:
```suggestion
for record in self:
            record.end_date = datetime.now() + timedelta(days=record.warranty_id.year * 365)

```

adsh-odoo (commented):
Hello @yame-odoo 
Some comments/suggestions.
Need to double-check the lintings.
Thanks

PR #278, opened 2025-02-05
18.0 training rusp

rusp-odoo:


estate/__manifest__.py
```diff
@@ -0,0 +1,11 @@
...
+    'name': 'Estate',
+    'version': '1.0',
+    'depends': ['base'],
+    'author': 'Odoo - Rushil Patel',
+    'category': 'Category',
+    'description': 'Technical training chapter - 2',
+    'license': 'LGPL-3',
+    'installable': True,
+    'application': True
+}
```
shag-odoo, line 11: @rusp-odoo new line missing at the end of file
Refer codebase for file structure.

shag-odoo (changes requested):

shag-odoo:
@rusp-odoo PR title should be according to git guidelines
Reference : https://www.odoo.com/documentation/18.0/contributing/development/git_guidelines.html

estate/__manifest__.py
```diff
@@ -0,0 +1,10 @@
+{
+    'name': 'Estate',
+    'version': '1.0',
+    'depends': ['base'],
+    'author': 'Odoo - Rushil Patel',
+    'description': 'Technical training chapter - 2',
+    'license': 'LGPL-3',
+    'application': True
+}
+
```
shag-odoo, line 10:
```suggestion
```

estate/models/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import estate_property
+
```
shag-odoo, line 2:
```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
...
+    facades = fields.Integer('Facades')
+    garage = fields.Boolean('Garage')
+    garden = fields.Boolean('Garden')
+    garden_area = fields.Integer('Garden Area')
+    garden_orientation = fields.Selection(
+        string='Type',
+        selection = [('north','North'),('south','South'),('west','West'),('east','East')]
+    )
+
+
```
shag-odoo, line 25:
```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
...
+    selling_price = fields.Float('Selling Price')
+    bedrooms = fields.Integer('Number of bedrooms')
+    living_area = fields.Integer('Living Area')
+    facades = fields.Integer('Facades')
+    garage = fields.Boolean('Garage')
+    garden = fields.Boolean('Garden')
+    garden_area = fields.Integer('Garden Area')
+    garden_orientation = fields.Selection(
+        string='Type',
+        selection = [('north','North'),('south','South'),('west','West'),('east','East')]
```
shag-odoo, line 22:
```suggestion
        selection = [
        ('north','North'),
        ('south','South'),
        ('west','West'),
        ('east','East')
      ])
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
...
+    )
```
shag-odoo, line 23:
```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
+from odoo import fields, models
+
+class EstateProperty(models.Model):
+    _name = 'estate.property'
+    _description = 'Real Estate Property'
+    _order = 'sequence'
```
shag-odoo, line 6: could you explain why you have used sequence although you have not defined the `sequence` field

shag-odoo (changes requested):

estate/models/estate_property.py
rusp-odoo: The `_order` field defines how records are sorted in search results, 

Initially, I assumed that using `sequence` in `_order` would display the records in the order they were added.

After referring to the codebase, I realized that `sequence` needs to be defined or can use any other field of the table.

shag-odoo (changes requested):

estate/models/estate_property.py
```diff
@@ -0,0 +1,38 @@
...
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate Property"
+
+    name = fields.Char("Property Name", required=True)
+    active = fields.Boolean(default=True)
+    description = fields.Text("Description")
+    postcode = fields.Char("Postal Code")
+    date_availability = fields.Date("Availability Date", copy=False, default= lambda self: fields.Datetime.now() + relativedelta(months=3))
```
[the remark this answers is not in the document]
rusp-odoo: 1. `copy=False` attribute is used to prevent the field's value from being copied, setting `copy=False` ensures that the field is not duplicated but rather gets a default or empty value in the new record.
 2. The lambda function in this context is used to set a dynamic default value for `date_availability`, it ensure that default value is calculated dynamically when the record is created(not when the database is generated).

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,81 @@
...
+        <field name="arch" type="xml">
+                <search string="Property filter">
+                    <field name="name" />
+                    <field name="postcode" />
+                    <field name="expected_price" />
+                    <field name="bedrooms" />
+                    <field name="living_area" />
+                    <field name="facades" />
+                    <filter string="Available" name="state" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]" />
+                    <filter string="Postcode" name="group_by_postcode" context="{'group_by': 'postcode'}" />
```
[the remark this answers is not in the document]
rusp-odoo: It is used to organize and display the data, in this case it groups property records by postcode. 
In the search menu when user clicks the Postcode filter, all the records will be grouped bt postcode field.

estate/models/estate_property.py
```diff
@@ -63,3 +64,25 @@ class EstateProperty(models.Model):
         ],
         required=True, copy=False, default="new"
     )
+    total_area = fields.Float("Total Area(sqm)", compute="_compute_total_area")
+    best_price = fields.Float("Best Price", compute="_compute_best_price")
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
```
shag-odoo, line 74: Is loop really required?

estate/models/estate_property.py
```diff
@@ -0,0 +1,103 @@
...
+    property_type_id = fields.Many2one(
+        'estate.property.type',
+        string="Property Type"
+    )
+    offer_ids = fields.One2many(
+        "estate.property.offer",
+        "property_id",
+        string="Offers"
+    )
+    salesman = fields.Many2one(
```
shag-odoo, line 45:
```suggestion
    user_id = fields.Many2one(
```
more relevant to comodel name

estate/models/estate_property.py
```diff
@@ -0,0 +1,103 @@
...
+        "res.users",
+        string="Salesman",
+        default=lambda self: self.env.user
+    )
```
shag-odoo, line 49:
```suggestion
    )
```
do we not require partner_id field in this model?

estate/models/estate_property.py
```diff
@@ -0,0 +1,103 @@
...
+        selection = [
+            ('new', 'New'),
+            ('offer_received', 'Offer Received'),
+            ('offer_accepted', 'Offer Accepted'),
+            ('sold', 'Sold'),
+            ('cancelled', 'Cancelled')
+        ],
+        required=True, copy=False, default="new"
+    )
+    total_area = fields.Float("Total Area(sqm)", compute="_compute_total_area")
```
shag-odoo, line 68:
```suggestion
    total_area = fields.Float("Total Area(sqm)", compute="_compute_total_area")
```
what are the characteristics of compute field?

estate/models/estate_property.py
```diff
@@ -0,0 +1,103 @@
...
+        for record in self:
+            if record.state == 'cancelled':
+                raise UserError("Cancelled properties cannot be sold")
+            else:
+                record.state = 'sold'
+
+    def cancelled_property(self):
+        for record in self:
+            if record.state == 'sold':
+                raise UserError("Sold properties cannot be cancelled")
```
shag-odoo, line 101:
```suggestion
                raise UserError("Sold properties cannot be cancelled")
```
What is the difference between `UserError` and `ValidationError`

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,48 @@
...
+        selection=[
+            ('accepted', 'Accepted'),
+            ('rejected', 'Rejected')
+        ],
+        copy=False
+    )
+    partner_id = fields.Many2one("res.partner", required=True)
+    property_id = fields.Many2one("estate.property", required=True)
+    validity = fields.Integer("Validity", default=7)
+    date_deadline = fields.Date("Deadline", compute="_compute_deadline_date", inverse="_inverse_deadline_date")
```
shag-odoo, line 20:
```suggestion
    date_deadline = fields.Date("Deadline", compute="_compute_deadline_date", inverse="_inverse_deadline_date")
```
what is the significance of `inverse`?

estate/views/estate_property_views.xml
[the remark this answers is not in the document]
shag-odoo: resolve it

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,8 @@
+from odoo import fields, models
+
+
+class EstatePropertyTag(models.Model):
+    _name = "estate.property.tag"
+    _description = "Property Tags"
+
+    name = fields.Char("Tage Name", required=True)
```
shag-odoo, line 8:
```suggestion
    name = fields.Char("Tag Name", required=True)
```

shag-odoo (changes requested):

estate/models/estate_property.py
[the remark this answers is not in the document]
shag-odoo: resolve it

estate/models/estate_property.py
rusp-odoo: Not necessarily. We use loop to iterate over records, but odoo will automatically iterate over all records behind the scene. So we can avoid using loop here to iterate over the records.

estate/models/estate_property.py
rusp-odoo: If partner_id meant by `buyer` then we already have a field `buyer`, so a `partner_id` field would be unnecessary.
But if we want to track the `owner` of the property then it makes sense.

estate/models/estate_property.py
rusp-odoo: A compute field is special type of field whose value is dynamically computed using a method. Following are the key characteristics:
* Recomputed on every access
* It is not stored by Default (can be stored `store=true`)
* supports `inverse` and `search` methods

estate/models/estate_property.py
rusp-odoo: * `UserError` : It is used to restrict user actions (restricts workflow)
`eg.` In our module a sold property cannot be cancelled or vice-versa.
 
* `ValidationError` : It is used to validate database fields / data integrity (can prevent records to save)
`eg.` In our module the selling price of the property should be atleast 90% of the expected price.

estate/models/estate_property_offer.py
rusp-odoo: The `inverse` parameter along with `compute` is used to allow the computed field to be editable and allow users to manually store the value(by default the compute fields are `readonly`).

`eg` in our model (`estate_property_offer`) the `date_deadline` is computed based on `validity`, but with the inverse method user can edit the `date_deadline` manually and it will update `validity` accordingly (user has to save the record first to see the updated value.)

PR #283, opened 2025-02-05
[ADD] estate: added init and manifest files

smjo-odoo:
- added models directory and made estate_property model with init and manifest files

### DIFF
--- a/estate/__init__.py
+++ b/estate/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -1,0 +1,29 @@
+{
+    'name': "estate app",
+
+    'summary': """
+        estate module"
+    """,
+
+    'description': """
+        estate module"
+    """,
+
+    'author': "Odoo",
+    'website': "https://www.odoo.com/",
+
+    # Categories can be used to filter modules in modules listing
+    # Check https://github.com/odoo/odoo/blob/15.0/odoo/addons/base/data/ir_module_category_data.xml
+    # for the full list
+    'category': 'Tutorials/estate',
+    'version': '0.1',
+
+    # any module necessary for this one to work correctly
+    'depends': ['base', 'web'],
+    'application': True,
+    'installable': True,
+    'data': [
+      #  'views/templates.xml',
+    ],
+    'license': 'AGPL-3'
+}
--- a/estate/models/estate_property.py
+++ b/estate/models/estate_property.py
@@ -1,0 +1,20 @@
+from odoo import fields, models
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate App"
+    name = fields.Char(required=True)
+    description = fields.Text() 
+    postcode = fields.Char()
+    date_availability = fields.Date()
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float()
+    bedrooms = fields.Integer()
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        selection = [('north', 'North'), ('south', 'South'),('east', 'East'),('west', 'West'),]
+    )
--- a/estate/models/__init__.py
+++ b/estate/models/__init__.py
@@ -1,0 +1,1 @@
+from . import estate_property


estate/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
sjai-odoo, line 1:
```suggestion
from . import models

```
Add a new line at the end of each file

smjo-odoo: noted , I will add new empty lines at end of code.

sjai-odoo (changes requested):

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+{
+    'name': "estate app",
+    'summary': "estate module",
+    'description': "estate module for buying and selling properties. Users can buy/sell properties via estate agents",
```
sjai-odoo, line 4:
```suggestion
    'description': "Estate module for buying and selling properties. Users can buy/sell properties via estate agents",
```

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    'author': "Odoo",
+    'category': 'Tutorials/estate',
+    'version': '0.1',
```
sjai-odoo, line 7:
```suggestion
    'version': '1.0',
```
In version `x.y`, x refers to main version and y is for update we generally start versioning of module from 1.0

estate/models/estate_property.py
```diff
@@ -0,0 +1,22 @@
...
+    description = fields.Text() 
+    postcode = fields.Char()
+    date_availability = fields.Date()
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float()
+    bedrooms = fields.Integer()
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
```
sjai-odoo, line 17:
```suggestion
    is_garage = fields.Boolean()
    is_garden = fields.Boolean()
```
Try to define boolean fields in this way which will be easier to use and understand in code as well.

estate/models/estate_property.py
```diff
@@ -0,0 +1,22 @@
...
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        selection = [('north', 'North'), ('south', 'South'),('east', 'East'),('west', 'West'),]
```
sjai-odoo, line 20:
```suggestion
        selection = [('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West'),]
```

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<odoo>
+
+<menuitem id="root_menu" name="Real Estate">
+    <menuitem id="first_menu" name="SubMenu1">
+        <menuitem id="action_menu" name="SubSubMenu1_1" action="estate_properties_view"/>
+    </menuitem>
+</menuitem>
```
sjai-odoo, line 8:
```suggestion
  <menuitem id="root_menu" name="Real Estate">
      <menuitem id="first_menu" name="SubMenu1">
          <menuitem id="action_menu" name="SubSubMenu1_1" action="estate_properties_view"/>
      </menuitem>
  </menuitem>
```

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    'depends': ['base'],
+    'data':[
+        'security/ir.model.access.csv',
+        'views/estate_property_views.xml',
+        'views/estate_menus.xml',
+        ],
```
sjai-odoo, line 13:
```suggestion
    ],
```

sjai-odoo (changes requested):

estate/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
...
+    )
+    active=fields.Boolean(default=True)
+    state=fields.Selection(
+        selection=[('new','New'),('offer_received','Offer Received'),('offer_accepted','Offer Accepted'),('sold','Sold'),('cancelled','Cancelled')]
+        ,default='new'
+    )
+    property_type_id =fields.Many2one("estate.property.type",string="Property Type")
+    buyer_id=fields.Many2one("res.partner",string="Buyer", readonly=True)
+    seller_id=fields.Many2one("res.users",string="Salesman")
+    tag_id=fields.Many2many("estate.property.tag",string="Tags")
```
sjai-odoo, line 31:
```suggestion
    tag_ids = fields.Many2many("estate.property.tag",string="Tags")
```
- when we define any x2many field it should be defined by name_ids which will specify that it can be connected to more than one recordset of the comodel.
- always practice to give single space on both side of the field definition and inside it if you give 2-3 parameters like
(string="xyz", required=True) give a space after space to make it clear.

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,21 @@
...
+    <record id="estate_property_tag_view_list" model="ir.ui.view">
+        <field name="name">estate.property.tag.list</field>
+        <field name="model">estate.property.tag</field>
+        <field name="arch" type="xml">
+            <list string="Property Tags">
+                <field name="name"/>
+            </list>
+        </field>
+    </record>
+
```
sjai-odoo, line 20:
```suggestion
```
What is the need of this view
We just have defined a single field in the list view which will give the same view as it have by default.
We generally add the explicit view when we need a specific view different than the existing default view.

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,22 @@
...
+    <!-- List View  -->
+    <!-- <record id="property_type_list_view" model="ir.ui.view">
+        <field name="field_name">estate.property.type.list</field>
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <list string="Property Types"> 
+                <field name="name"></field>
+            </list>
+        </field>
+    </record>  -->
```
sjai-odoo, line 18:
```suggestion
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,102 @@
...
+        <field name="arch" type="xml">
+            <form string="Property Form">
+                <sheet>
+                    <group>
+                        <group>
+                            <div>
+                                <field name="name"/>
+                            </div>
+                        </group>
+                        <newline/>
```
sjai-odoo, line 40:
```suggestion
                          <div>
                              <field name="name"/>
                          </div>
```
If we give a div inside a group that will make no sense of giving group tag as if we don't need a label just remove the unnecessary tags

sjai-odoo (changes requested):

estate/models/estate_property.py
```diff
@@ -0,0 +1,102 @@
+from odoo import fields, models, api # type: ignore
+from odoo.exceptions import ValidationError, UserError # type: ignore
+from dateutil.relativedelta import relativedelta
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate App"
+    _order = "id desc"
```
sjai-odoo, line 8:
```suggestion
```
By default the ordering happens by decreasing id.

estate/models/estate_property.py
```diff
@@ -0,0 +1,102 @@
...
+            if property.measurement_unit == 'sqft':
+                property.total_area *= 10.764
+
+    @api.depends("offer_ids.offer_price")
+    def _compute_best_price(self):
+        for prop in self:
+            prop.best_price=max(prop.offer_ids.mapped("offer_price")) if prop.offer_ids else 0.0
+            
+#---------------------------------------------------- Action Methods----------------------------------------#
+    def property_sold_action(self):
```
sjai-odoo, line 69:
```suggestion
    def action_property_sold(self):  
```
put action at the beginning of the action methods to identify easily

estate/models/estate_property.py
```diff
@@ -0,0 +1,102 @@
...
+                property.total_area *= 10.764
+
+    @api.depends("offer_ids.offer_price")
+    def _compute_best_price(self):
+        for prop in self:
+            prop.best_price=max(prop.offer_ids.mapped("offer_price")) if prop.offer_ids else 0.0
+            
+#---------------------------------------------------- Action Methods----------------------------------------#
+    def property_sold_action(self):  
+        if self.state in ('sold'):
```
sjai-odoo, line 70:
```suggestion
        if self.state == 'sold':
```
If there is only one value to compare you can directly use `==` rather tha `in`

estate/models/estate_property.py
```diff
@@ -0,0 +1,102 @@
...
+            raise UserError('Property is Already sold')
+        self.state = "sold"
+        return True
+
+    def property_cancel_action(self):
+        if self.state in ('cancelled'):
+            raise UserError('Property is not for sale anymore')
```
sjai-odoo, line 77:
```suggestion
            raise UserError('Property is already cancelled')
```
improper string based on checked condition

estate/models/estate_property.py
```diff
@@ -0,0 +1,102 @@
...
+            self.garden_area = 10
+            self.garden_orientation = 'north'
+        else :
+            self.garden_area = 0
+            self.garden_orientation = False
+
+    @api.constrains('selling_price','expected_price')
+    def _check_selling_price(self):
+        for property in self:
+            if fields.Float.compare (property.selling_price, property.expected_price * 90/100, precision_rounding=0.01) < 0 and not fields.Float.is_zero(property.selling_price, precision_rounding=0.01):
```
sjai-odoo, line 94:
```suggestion
            if prop.selling_price and prop.selling_price < prop.expected_price * 0.90:
```
Make the logic easy to understand by avoiding extra checks like here it's not really necessary to precise the amounts using precision_rounding.

estate/models/estate_property.py
```diff
@@ -0,0 +1,102 @@
...
+                raise ValidationError ("Selling Price cannot be lower than '90%'of the expected price.")
+
+       #-------------------------------------CRUD methods------------------------------------------#
+    @api.ondelete(at_uninstall=False)
+    def _unlink_except_sold_or_cancelled(self):
+        for record in self:
+            if record.state not in {'new', 'cancelled'} :
```
sjai-odoo, line 101:
```suggestion
            if record.state not in ['new', 'cancelled'] :
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,102 @@
...
+
+    @api.constrains('selling_price','expected_price')
+    def _check_selling_price(self):
+        for property in self:
+            if fields.Float.compare (property.selling_price, property.expected_price * 90/100, precision_rounding=0.01) < 0 and not fields.Float.is_zero(property.selling_price, precision_rounding=0.01):
+                raise ValidationError ("Selling Price cannot be lower than '90%'of the expected price.")
+
+       #-------------------------------------CRUD methods------------------------------------------#
+    @api.ondelete(at_uninstall=False)
+    def _unlink_except_sold_or_cancelled(self):
```
sjai-odoo, line 99:
```suggestion
    @api.ondelete(at_uninstall=False)
    def _unlink_except_sold_or_cancelled(self):
```
Can you please explain why you used this approach and what this will do?
Why can't we simply use unlink method?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+    _order = "offer_price desc"
+    _sql_constraints = [
+        ('check_offer_price','CHECK(offer_price > 0)','Offer Price must be a positive amount')
+    ]
+
+    #-------------------------------------------Basic Fields-------------------------------------------#
+    offer_price = fields.Float(string="Price", required=True)
+    state = fields.Selection(string="State", selection=[('offer_received','Offer Received'), ('offer_accepted','Offer Accepted'), ('offer_rejected','Offer Rejected')],
+                             copy=False
+                            )
```
sjai-odoo, line 18:
```suggestion
    state = fields.Selection(string="State", selection=[('offer_received','Offer Received'), ('offer_accepted','Offer Accepted'), ('offer_rejected','Offer Rejected')], copy=False)
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+            date = offer.create_date.date() if offer.create_date else fields.Date.today()
+            offer.deadline_date = date + relativedelta(days=offer.validity)
+
+    def _inverse_deadline_date(self):
+        for offer in self:
+            date = offer.create_date.date() if offer.create_date else fields.Date.today()
+            offer.validity = (offer.deadline_date - date).days
+
+    # --------------------------------------------Action Methods------------------------------------------------
+    def offer_accept_action(self):
```
sjai-odoo, line 42:
```suggestion
    def action_offer_accept(self):
```
action should be in the starting of the name and change the name accordingly in xml as well

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+        if self.state == "offer_accepted":
+            raise UserError('offer is already accepted')
+        self.state = "offer_accepted"
+        self.property_id.selling_price = self.offer_price
+        self.property_id.buyer_id = self.partner_id
+        self.property_id.state = "offer_accepted"
```
sjai-odoo, line 48: Find a way to merge all three in a single line using `write` function

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+        return True
+        
+    def offer_reject_action(self):
+        self.state = "offer_rejected"
+        self.property_id.selling_price = 0
+        self.property_id.buyer_id = None
+        return True
+
+    #-------------------------------------CRUD methods------------------------------------------#
+    def create(self, vals_list):
```
sjai-odoo, line 58:
```suggestion
    @api.model_create_multi
    def create(self, vals_list):
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+        """Override the create() method to only allow offers bigger than existing
+        offers and update the property state when a new valid offer is created"""
+        if not isinstance(vals_list, list):
+            vals_list = [vals_list]
```
sjai-odoo, line 62:
```suggestion
```
This check is not needed it will always be in list

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,27 @@
...
+        <field name="view_mode">list,form</field>
+        <field name="domain">[('property_type_id', '=', active_id)]</field>
+    </record>
+
+    <!-- List View -->
+    <record id="estate_property_offer_view_list" model="ir.ui.view">
+        <field name="name">estate.property.offer.list</field>
+        <field name="model">estate.property.offer</field>
+        <field name="arch" type="xml">
+            <list editable="bottom" decoration-danger="state in ('offer_rejected')" decoration-success="state in ('offer_accepted')">
```
sjai-odoo, line 15:
```suggestion
            <list editable="bottom" decoration-danger="state == 'offer_rejected')" decoration-success="state == 'offer_accepted'">
```
Generally we use in operator when we have to check within multiple values, to check with single value we can use `==`

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,153 @@
+<?xml version="1.0"?>
+<odoo>
+    <!-- Action Window -->
+    <record id="estate_property_action" model="ir.actions.act_window">
+        <field name="name">Properties</field>
+        <field name="res_model">estate.property</field>
+        <field name="view_mode">list,form,kanban</field>
+        <field name="context"> {'search_default_' : True}</field>
```
sjai-odoo, line 8:
```suggestion
        <field name="context"> {'search_default' : True}</field>
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,102 @@
...
+    offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers")
+    
+    # --------------------------------------------Compute Methods----------------------------------------#
+    @api.depends("living_area", "garden_area", "measurement_unit")
+    def _compute_total_area(self):
+        for property in self:
+            property.total_area = property.living_area + property.garden_area
+            if property.measurement_unit == 'sqft':
+                property.total_area *= 10.764
+
```
sjai-odoo, line 62:
```suggestion
    @api.depends("living_area", "garden_area")
    def _compute_total_area(self):
        for property in self:
            property.total_area = property.living_area + property.garden_area
```
Good thinking and creativity.
But if i have a selection i will directly select that yeah i'm writing it in sqft or in sqm and enter the value in that measure

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
+from odoo import models, api, fields
+class EstateProperty(models.Model):
```
sjai-odoo, line 2:
```suggestion
from odoo import models, api, fields


class EstateProperty(models.Model):
```

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
...
+            self.env["account.move"].create({
+                'partner_id': property.buyer_id.id,
+                'move_type': 'out_invoice',
+                "journal_id": journal.id,
+                'invoice_date': fields.Date.today(),
+                "invoice_line_ids": [
+                    (0,0,{
+                            "name": property.name,
+                            "quantity": 1.0,
+                            "price_unit": property.selling_price * 6.0 / 100.0,
```
sjai-odoo, line 20:
```suggestion
                            "price_unit": property.selling_price,
```
why the property line just have 6% of the amount at which it was sold?

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
+from odoo import models, api, fields
+class EstateProperty(models.Model):
+    _inherit = "estate.property"
+
+    #------------------------------------------Action Methods----------------------------------------------
+    def property_sold_action(self):
+        """set property state as sold and Create invoice for sold property"""
+        res = super().property_sold_action()
+        journal = self.env["account.journal"].search([("type", "=", "sale")], limit=1)
```
sjai-odoo, line 9:
```suggestion
        journal = self.env["account.journal"].search([("type", "=", "sale")], limit=1)
```
why do you search for a sale journal?
What do you understand by journal?

sjai-odoo (changes requested):

estate/models/estate_property.py
smjo-odoo: sir, at_uninstall=False will ensure that upon uninstalation also , this restriction is applied. We are overriding unlink method to add a restriction based in state of the property. We can use the unlink method directly , but this restriction will not be put on deletion when a module is uninstalled, so we need the ondelete decorator to put up a restriction

estate/views/estate_property_offer_views.xml
smjo-odoo: yes sir , I will be careful and use == at single value comparisons

estate_account/models/estate_property.py
smjo-odoo: sir, we need a journal to be known by odoo to store account transactions and since we are generating invoice for a sales order, we look for a sales journal.

estate_account/models/estate_property.py
smjo-odoo: sir , as per the excercise  "Add the third step of invoice creation.
Add two invoice lines during the creation of the account.move. Each property sold will be invoiced following these conditions:
6% of the selling price
an additional 100.00 from administrative fees
"
I am unsure about whether  it is supposed to be 6% extra on the selling price or is it supposed to be 6% on the selling price

estate/models/estate_property.py
smjo-odoo: yes sir , I understood your point and I'll implement the above suggested changes

estate/models/estate_property.py
smjo-odoo: yes sir , I will make the changes following the guidelines and make the variable names according to odoo's coding conventions

estate/models/estate_property.py
sjai-odoo: Complete this point

estate/models/estate_property.py
```diff
@@ -0,0 +1,102 @@
...
+    def _compute_best_price(self):
+        for prop in self:
+            prop.best_price=max(prop.offer_ids.mapped("offer_price")) if prop.offer_ids else 0.0
+            
+    #----------------------------------------------- Action Methods----------------------------------------#
+    def action_property_sold(self):
+        if self.state == 'sold':
+            raise UserError('Property is Already sold')
+        self.state = "sold"
+        return True
```
sjai-odoo, line 73:
```suggestion
```
No need to return anything

estate/models/estate_property.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def action_property_cancel(self):
+        if self.state == 'cancelled':
+            raise UserError('Property is already cancelled')
+        self.state = "cancelled"
+        return True
```
sjai-odoo, line 79:
```suggestion
```

estate/models/estate_property_offer.py
sjai-odoo: what about this?

estate/views/estate_property_report.xml
sjai-odoo, line 1: remove if unnecessary

sjai-odoo (changes requested):

PR #285, opened 2025-02-05
[ADD] estate: added estate_property model

mkap-odoo:
In this commit:
 - Added __init__.py file in models directory.
 - Build a table structure with its relevant fields in estate_property.py file.


estate/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    App to list and sell/rent property
+    """,
+    # data files always loaded at installation
+    'data': [
+        'views/mymodule_view.xml',
+    ],
+    # data files containing optionally loaded demonstration data
+    'demo': [
+        'demo/demo_data.xml',
+    ],
```
sjai-odoo: Hi @mkap-odoo, 
Can you please explain me why have you added these data files?
What does these lines do?

estate/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+}
```
sjai-odoo, line 29: Add a new line at the end of the file.

sjai-odoo (changes requested):

estate/__manifest__.py
mkap-odoo: Honestly, I have copied the manifest file from CRM module but forgot to remove/comment unused lines.
The demo is used to load the demo data which will show only if the developer mode is on to test our applications.
And in data, we make UI for the module created for the users to interact.

estate/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    'summary': """
+        estate module"
+    """,
+
+    'description': """
+        estate module"
+    """,
+
+    'author': "Odoo",
+    'website': "https://www.odoo.com/",
```
sjai-odoo, line 13:
```suggestion
```
Generally `website` key specifies the URL related to the module, which can be the official website of the module to helps users quickly find more information about the module, for now you can remove this url.

estate/__manifest__.py
```diff
@@ -0,0 +1,29 @@
+{
+    'name': "estate app",
+
```
sjai-odoo, line 3:
```suggestion
```
Remove extra lines from whole file

estate/models/estate_property.py
```diff
@@ -0,0 +1,39 @@
...
+
+facades
+
+garage
+
+garden
+
+garden_area
+
+garden_orientation
```
sjai-odoo, line 29: If you want to add something temporarily for your reference put it in comment.

sjai-odoo (changes requested):

estate/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    'website': "https://www.odoo.com/",
+
+    # Categories can be used to filter modules in modules listing
+    # Check https://github.com/odoo/odoo/blob/15.0/odoo/addons/base/data/ir_module_category_data.xml
+    # for the full list
+    'category': 'Tutorials/estate',
+    'version': '0.1',
+
+    # any module necessary for this one to work correctly
+    'depends': ['base', 'web'],
```
sjai-odoo, line 22:
```suggestion
```
No need of any of these dependencies
`base` is a default dependency, we don't need to add it in depends

estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
...
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate App"
+
+    name = fields.Char(required=True)
+    description = fields.Text() 
+    postcode = fields.Char()
+    date_availability = fields.Date(default=lambda self: date.today() + timedelta(days=90))
+    expected_price = fields.Float()
+    selling_price = fields.Float(readonly=True , copy=False)
```
sjai-odoo, line 13:
```suggestion
    selling_price = fields.Float(copy=False)
```
If i make it readonly how will i be able to assign value to it as there is no compute method.

sjai-odoo (changes requested):

estate/models/estate_property.py
```diff
@@ -0,0 +1,57 @@
...
+    status = fields.Selection(
+        selection=[
+            ("new", "New"),
+            ("offer_reject", "Offer Rejected"),
+            ("offer_accept", "Offer Accepted"),
+            ("sold", "Sold"),
+            ("cancelled", "Cancelled"),
+        ],
+        default="new",
+        required=True,
```
sjai-odoo, line 40:
```suggestion
    status = fields.Selection(
        selection=[
            ("new", "New"),
            ("offer_reject", "Offer Rejected"),
            ("offer_accept", "Offer Accepted"),
            ("sold", "Sold"),
            ("cancelled", "Cancelled"),
        ],
        default="new",
        required=True,
        copy=False
```
make sure when duplicate record this field's value should not be copied

estate/models/estate_property.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+    active = fields.Boolean(default=True)
+    property_type_id = fields.Many2one("estate.property.type", ondelete="cascade")
```
sjai-odoo, line 43: Can you explain what `ondelete="cascade"` will do?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,110 @@
...
+                <sheet>
+                    <group>
+                        <group>
+                            <group>
+                                <h1>
+                                    <field name="name"/>
+                                </h1>
+                                <field name="tag_ids" nolabel="1" widget="many2many_tags"/>
+                            </group>
+                            <newLine/>
```
sjai-odoo, line 43:
```suggestion
                              <h1>
                                  <field name="name"/>
                              </h1>
                              <field name="tag_ids" widget="many2many_tags"/>
```
If we don't want to give field label than why do we need to add group and than again removing label by nolabel=1, That makes no sense

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,34 @@
...
+        <field name="arch" type="xml">
+            <form>
+                <sheet>
+                    <group>
+                        <field name="name"/>
+                    </group>
+                </sheet>
+            </form>
+        </field>
+    </record>
```
sjai-odoo, line 33:
```suggestion
```
We don't need to define list or form view explicitly, it's already have a view but if you want the view to be customized than only define them explicitly which will override the existing one.
But here you only have a single field than what is the need to define it seperately.

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,39 @@
...
+                <sheet>
+                    <group>
+                        <field name="price">Price</field>
+                        <field name="partner_id">Partner</field>
+                        <field name="status">Status</field>
+                    </group>
+                </sheet>
+            </form>
+        </field>
+    </record>
```
sjai-odoo, line 38:
```suggestion
```
Same issue try to understand when we should add explicit views you will get this view by default first understand why we are doing this and does this change really impact?

sjai-odoo (changes requested):

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,110 @@
...
+        <field name="arch" type="xml">
+            <search string="Search for Properties">
+                <field name="name" string="Title"/>
+                <field name="postcode" string="Postcode" />
+                <field name="expected_price" string="Expected Price" />
+                <field name="bedrooms" string="Bedrooms" />
+                <field name="living_area" string="Living Area (sqm)" />
+                <field name="facades" string="Facades" />
+                <field name="property_type_id" string="Property Type"/>
+                <filter name="status" string="Available" domain="['|',('status', '=', 'new'), ('status' , '=' , 'offer_accept')]"/>
```
sjai-odoo, line 104:
```suggestion
                <filter name="status" string="Available" domain="['|',('status', '=', 'new'), ('status' , '=' , 'offer_accept')]"/>
```
Is there any easy way where i don't have to use or (|) operator and just implement this within one single doman?

estate/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+{
+    'name': "estate app",
+    'summary': """
+        estate app
+    """,
```
sjai-odoo, line 5:
```suggestion
```
No need of this.

sjai-odoo (changes requested):

estate/models/estate_property.py
mkap-odoo: ondelete= "cascade" is used when any related record is deleted, its dependent records will be automatically deleted.
If the property type is deleted, then its related properties who have respective property type is also deleted.
Other than these, there are also 2 methods i.e restrict and set null.

estate/views/estate_property_views.xml
mkap-odoo: we can use this: <filter name="status" string="Available" domain="[('status', 'in', ['new', 'offer_accept'])]"/>

estate/models/estate_property.py
```diff
@@ -0,0 +1,146 @@
+from odoo import fields, models, api
+from datetime import date, timedelta
+from odoo.exceptions import UserError, ValidationError
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate App"
+    _order = "id desc"
```
sjai-odoo, line 9:
```suggestion
```
Default record-sets are shown with decreasing id only so no need to add separately

estate/models/estate_property.py
```diff
@@ -0,0 +1,146 @@
...
+        default="new",
+        required=True,
+        copy=False,
+    )
+    active = fields.Boolean(default=True)
+    property_type_id = fields.Many2one("estate.property.type")
+    salesman_id = fields.Many2one(
+        "res.users",
+        string="Salesman",
+        ondelete="restrict",
```
sjai-odoo, line 50: what is the meaning of ondelete="restrict"

estate/models/estate_property.py
```diff
@@ -0,0 +1,146 @@
...
+            if (
+                record.status == "new"
+                or record.status == "offer_accept"
+                or record.status == "offer_receive"
+            ):
+                record.status = "sold"
+            elif record.status == "cancelled":
+                raise UserError("Cancelled property can't be sold")
+            else:
+                raise UserError("Property already sold")
```
sjai-odoo, line 112:
```suggestion
            if record.status == "cancelled":
                raise UserError("Cancelled property can't be sold")
            elif recod.status == "sold":
                raise UserError("Property already sold")
           record.status = "sold"
```

sjai-odoo: check the requirements first than if all of them paas set directly the value.

estate/models/estate_property.py
```diff
@@ -0,0 +1,146 @@
...
+
+        return True
```
sjai-odoo, line 114:
```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,146 @@
...
+                or record.status == "offer_accept"
+                or record.status == "offer_receive"
+            ):
+                record.status = "cancelled"
+            elif record.status == "sold":
+                raise UserError("Sold property can't be cancelled")
+            else:
+                raise UserError("Property already cancelled")
+
+        return True
```
sjai-odoo, line 130: Optimize this function as well as per suggestions given in action_estate_property_sold.

estate/models/estate_property.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @api.constrains("expected_price", "selling_price")
+    def _check_exp_sel_price(self):
+        for record in self:
+            if (
+                record.selling_price < record.expected_price * 0.9
+            ) and record.selling_price > 0:
```
sjai-odoo, line 137:
```suggestion
            if record.selling_price > 0 and record.selling_price < record.expected_price * 0.9:
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,146 @@
...
+                raise ValidationError(
+                    f"The selling price must be atleast 90% of expected price"
+                )
```
sjai-odoo, line 140:
```suggestion
                raise ValidationError("The selling price must be atleast 90% of expected price")
```
What is the need of fstring here.

estate/models/estate_property.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_if_state_new_cancel(self):
```
sjai-odoo, line 143: Can you please explain what these lines do?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,93 @@
...
+    def _compute_date_deadline(self):
+        for record in self:
+            if record.create_date:
+                record.date_deadline = record.create_date.date() + relativedelta(
+                    days=record.validity
+                )
+            else:
+                record.date_deadline = datetime.today() + relativedelta(
+                    days=record.validity
+                )
```
sjai-odoo, line 49:
```suggestion
            record.date_deadline = record.create_date.date() or datetime.today() + relativedelta(days=record.validity)
```

sjai-odoo: if record.create_date.date() is there than take that value else datetime.today() will be taken into account

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def _inverse_date_deadline(self):
+        for record in self:
+            if record.date_deadline:
+                delta = record.date_deadline - datetime.today().date()
```
sjai-odoo, line 54:
```suggestion
                delta = record.date_deadline - datetime.today().date()
```
datetime.today() why always today and not create date

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,93 @@
...
+    def estate_property_offer_action_accept(self):
+        for record in self:
+            if record.status == "accepted":
+                raise UserError("Offer is already accepted")
+            else:
+                # accept the selected offer
+                record.status = "accepted"
+                record.property_id.status = "offer_accept"
+                record.property_id.buyer_id = record.partner_id
+                record.property_id.selling_price = record.price
```
sjai-odoo, line 66: Find a way to optimize this, hint is use `write` function.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,93 @@
...
+
+                # reject the other offers for same property
+                other_offers = record.property_id.offer_ids.filtered(
+                    lambda o: o.id != record.id
+                )
```
sjai-odoo, line 71:
```suggestion
                other_offers = record.property_id.offer_ids.filtered(lambda o: o.id != record.id)
```
Avoid extra lines

sjai-odoo (changes requested):

estate/models/estate_property.py
mkap-odoo: It means that when we try to delete a salesman which is linked to a property, it will restrict or will not be able to delete that salesman.

estate/models/estate_property_offer.py
mkap-odoo: Firstly I was thinking like when we are creating the offer and if we want to compute the validity from the deadline, the deadline is subtracted from the day when the record is created i.e we are creating the offer record at the current time but after that i realized my mistake that what if I create the offer record but not given the validity and deadline and we have to calculate on the basis of create date priority as per the exercise. 

We can both the methods like these:
`@api.depends("validity")
    def _compute_date_deadline(self):
        for record in self:
            create_date = (
                record.create_date.date() if record.create_date else fields.Date.today()
            )
            record.date_deadline = create_date + timedelta(days=record.validity)

    def _inverse_date_deadline(self):
        for record in self:
            if record.date_deadline:
                create_date = (
                    record.create_date.date() if record.create_date else fields.Date.today()
                )
                record.validity = (record.date_deadline - create_date).days`

estate/models/estate_property_offer.py
mkap-odoo: `record.write(
                    {
                        "status": "refused",
                        "property_id.buyer_id": False,
                        "property_id.selling_price": False,
                    }
                )`

mkap-odoo: I tried the below code in the estate_property_offer_action_refuse method but it is not working:
`record.write(
                     {
                         "status": "refused",
                         "property_id.buyer_id": False,
                         "property_id.selling_price": False,
                     }
                 )`

estate/models/estate_property.py
```diff
@@ -0,0 +1,144 @@
...
+        else:
+            self.garden_area = 0
+            self.garden_orientation = False
+
+    @api.constrains("expected_price", "selling_price")
+    def _check_exp_sel_price(self):
+        for record in self:
+            if (
+                record.selling_price < record.expected_price * 0.9
+            ) and record.selling_price > 0:
```
sjai-odoo, line 115:
```suggestion
            if record.selling_price and record.selling_price < record.expected_price * 0.9:
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,144 @@
...
+                )
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_if_state_new_cancel(self):
+        for record in self:
+            if record.status not in ["new", "cancelled"]:
+                raise UserError("Can only delete properties in new or cancelled stage")
+
+    def action_estate_property_sold(self):
+        print("Sold button clicked")
```
sjai-odoo, line 127:
```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,144 @@
...
+        for record in self:
+            if record.status == "cancelled":
+                raise UserError("Cancelled property can't be sold")
+            elif record.status == "sold":
+                raise UserError("Property already sold")
+            else:
+                record.status = "sold"
```
sjai-odoo, line 134:
```suggestion
            record.status = "sold"
```
No need of else as code reached here only if above 2 condition failed otherwise user error would be raised

estate/models/estate_property.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def action_estate_property_cancel(self):
+        print("Cancel Button Clicked")
```
sjai-odoo, line 137:
```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,144 @@
...
+        for record in self:
+            if record.status == "sold":
+                raise UserError("Sold property can't be cancelled")
+            elif record.status == "cancelled":
+                raise UserError("Property already cancelled")
+            else:
+                record.status = "cancelled"
```
sjai-odoo, line 144:
```suggestion
            record.status = "cancelled"
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,92 @@
...
+            "Offer Price for the property should be strictly positive",
+        )
+    ]
+
+    @api.depends("validity")
+    def _compute_date_deadline(self):
+        for record in self:
+            create_date = (
+                record.create_date.date() if record.create_date else fields.Date.today()
+            )
```
sjai-odoo, line 44:
```suggestion
            create_date = record.create_date.date() if record.create_date else fields.Date.today()
```
Unnecessary brackets

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,92 @@
...
+            record.date_deadline = create_date + timedelta(days=record.validity)
+
+    def _inverse_date_deadline(self):
+        for record in self:
+            if record.date_deadline:
+                create_date = (
+                    record.create_date.date() if record.create_date else fields.Date.today()
+                )
```
sjai-odoo, line 52:
```suggestion
                create_date = record.create_date.date() if record.create_date else fields.Date.today()
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,92 @@
...
+    @api.model
+    def create(self, vals):
+        offer = super().create(vals)
+        if offer.property_id.offer_ids:
+            if offer.price >= offer.property_id.best_offer:
+                offer.property_id.status = "offer_receive"
+            else:
+                raise ValidationError("The offer price should be more than best offer")
+
+        return offer
```
sjai-odoo, line 64:
```suggestion
        property = vals.get(.property_id)
        if property.offer_ids:
            if vals.get(price) < property.best_offer:
                raise ValidationError("The offer price should be more than best offer")
            property.status = "offer_receive"
        return super().create(vals)
```

sjai-odoo: We check for errors first than go to create from super, both works fine but this is better to check before to save all other creation operations, check from the values in vals. & check for error condition if that's not no need of else directly set the value as it passed from above condition ensures that it passes our check.

estate/models/estate_property_offer.py
sjai-odoo: we want to update property's field so we have to write `record.property_id.write(......` the fields and value we can't access inner fields as abc.field_name inside the string as you have done go to last point and write all the needed fields there

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def estate_property_offer_action_refuse(self):
+        for record in self:
+            if record.status == "refused":
+                raise UserError("Offer is already refused")
+            else:
+                # refuse the selected offer
+                record.status = "refused"
+                record.property_id.buyer_id = False
+                record.property_id.selling_price = False
```
sjai-odoo, line 91: merge them using `write` function

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,26 @@
+from odoo import fields, models, Command
+
+
+class EstateProperty(models.Model):
+    _inherit = "estate.property"
+
+    def action_estate_property_sold(self):
+        self.check_access('write')
```
sjai-odoo, line 8:
```suggestion
```
No need to check like this, that will be cecked when record go to write and user doesn't have that right it will give access error.

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,26 @@
...
+        print("Sold button clicked from account application")
```
sjai-odoo, line 9:
```suggestion
```

sjai-odoo (changes requested):

PR #311, opened 2025-02-05
[ADD] estate: adding a new real estate module

jmra-odoo:
Added a new module real estate for tutorial purpose


rare-odoo (changes requested):
Hello @jmra-odoo 👋,

Great work so far!

This is the first set of the review.
Please make sure to add a new line at the end of every file.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,142 @@
...
+                <sheet>
+                    <group>
+                        <h1>
+                            <field name="name" />
+                        </h1>
+                    </group>
+
+                    <group>
+                        <h5>
+                            <field name="property_tag_ids" widget="many2many_tags" />
```
rare-odoo, line 51: What is the purpose of adding widget="many2many_tags" here?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,142 @@
...
+
+                <!-- Available with NEW or OFFER RECEIVED state filter -->
+                <filter string="New or Offer Received" name="state"
+                    domain="[('date_availability', '>=', context_today()), '|', ('state', '=', 'new'), ('state', '=', 'offer_received')]" />
+
+                <!-- Group By postcode -->
+                <group expand="1" string="Group By">
+                    <filter string="Postcode" name="postcode"
+                        context="{'group_by': 'postcode', 'residual_visible': True}" />
+                </group>
```
rare-odoo, line 136: Why is this filter placed inside the group tag?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,142 @@
...
+                <field name="bedrooms" />
+                <field name="living_area" />
+                <field name="facades" />
+                <field name="garage" />
+                <field name="garden" />
+                <field name="garden_area" />
+                <field name="garden_orientation" />
+                <field name="active" />
+                <field name="state" />
+                <field name="property_tag_ids" />
```
rare-odoo, line 31: Is it necessary to show every field in the list view?
If not, how can you avoid displaying all fields?

rare-odoo (changes requested):

estate/data/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
+access_estate_property_type,access_estate_property_type,model_estate_property_type,base.group_user,1,1,1,1
+access_estate_property_tag,access_estate_property_tag,model_estate_property_tag,base.group_user,1,1,1,1
+access_estate_property_offer,access_estate_property_offer,model_estate_property_offer,base.group_user,1,1,1,1
```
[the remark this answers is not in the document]
jmra-odoo: 1 denotes the permission allowed for specific action such as create, read, write or unlink the particular model's data
in above code all the permission/access is allowed

estate/models/estate_property.py
```diff
@@ -0,0 +1,53 @@
...
+from dateutil.relativedelta import relativedelta
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate Property"
+
+    def calculate_three_months_later():
+        today = datetime.now()
+
+        return today + relativedelta(months=3)
```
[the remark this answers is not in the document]
jmra-odoo: Okay, I will take care of it from now. (Extra Line & method name)
Method is used to calculate the availability date of the property, Which should be 3 months later from the creating by default. relativedelta() method is used to do so which is imported from dateutil.relativedelta

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,32 @@
+<odoo>
+    <!-- Property Tag Action Menu -->
+    <record id="estate_property_tag_action" model="ir.actions.act_window">
+        <field name="name">Property Tag</field>
+        <field name="res_model">estate.property.tag</field>
+        <field name="view_mode">list,form</field>
```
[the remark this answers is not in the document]
jmra-odoo: 'view_mode' is used to describe that in how many ways you can see the model data. In above case its list & form

estate/views/estate_property_views.xml
jmra-odoo: It provides the each tag to look like the chips in a single line, if we do not use it then tags will be shown as table format same as order_line or sales_line.

estate/views/estate_property_views.xml
jmra-odoo: It is a group by filter option for postcode, to make group by filters we need to wrap the filters in group tag and it makes to filter placed in center column of group by category in search view

estate/views/estate_property_views.xml
jmra-odoo: No, it's not necessary, we can remove any fields as per our requirement, we can remove line 28 if we don't want the garden_orientation field in our list view.

estate/views/estate_property_tag_views.xml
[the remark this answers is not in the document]
rare-odoo: Could you kindly confirm if these are the only available attributes? 
Also, is it necessary to include the view_mode?

estate/views/estate_property_views.xml
rare-odoo: Could you please clarify if it is necessary to use `residual_visible` in the context?
Additionally, could you explain its purpose?

estate/views/estate_property_tag_views.xml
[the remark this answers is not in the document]
jmra-odoo: No, there are more attributes such as views, domain, limit, res_id, target etc.
No, view_mode by default set to list, form so not including view_mode would not affect anything, but i think adding that line could be better to understand others easily while revisiting and to modify it after.

estate/views/estate_property_views.xml
jmra-odoo: No, it's not required as i have tried removing it and it works same as old, though i have searched for it on internet and chatgpt as well i got the aswers like... for remaining data that can't be grouped in any of others, something to balance the amount and last most believable that odoo made it True by default(or deprecated) after v-15.0 so it does'nt matter now to add it or not(because i searched for it in odoo docs of 16, 17 & 18 it did't found there).

estate/models/estate_property.py
```diff
@@ -0,0 +1,109 @@
...
+            if record.property_offer_ids:
+                record.best_offer = max(record.property_offer_ids.mapped('price'), default=0)
+            else:
+                record.best_offer = 0
+
+    @api.onchange('garden')
+    def _onchange_garden(self):
+        if self.garden:    
+            self.garden_area = 10
+            self.garden_orientation = 'north'
```
rare-odoo, line 85: Why set a static value here?
Why can't this be set in the `default` or through a `compute` method?

estate/models/estate_property.py
```diff
@@ -0,0 +1,109 @@
...
+        if self.state == 'cancelled':
+            raise UserError('A cancelled property cannot be sold.')
+        else:
+            self.state = 'sold'
+
+    def action_set_cancelled_status(self):
+        if self.state == 'sold':
+            raise UserError('A sold property cannot be cancelled.')
+        else:
+            self.state = 'cancelled'
```
rare-odoo, line 101: Can this be managed using a single method?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,61 @@
...
+                    ('property_id', '=', offer.property_id.id),
+                    ('id', '!=', offer.id)  # Exclude the current offer
+                ])
+
+                other_offers.write({'status': 'refused'})
+
+                offer.property_id.selling_price = offer.price
+                offer.property_id.buyer = offer.partner_id
+
+                offer.status = 'accepted'
```
rare-odoo, line 54: Can't we access and write the `offer_ids` directly?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def action_set_offer_status_refused(self):
+        for record in self:
+            if record.status == 'accepted' or record.status == 'refused':
```
rare-odoo, line 58: Can’t this be combined into one, rather than using 'or'?

rare-odoo (changes requested):
Hello @jmra-odoo 👋

Looks good so far.
However, it seems you're still lagging behind,
and also several files are missing extra line at the end.

Anyways, This is the second set of reviews.
Thank you!

estate/models/estate_property.py
jmra-odoo: default values would be visible even before we check the garden checkbox so it cannot be used but we can definitely use the compute method instead of onchange, i have used it just for the task completion and the learning purpose of it.

estate/models/estate_property.py
jmra-odoo: I wasn't confirm but i tried it with same name and providing context with button id to both buttons and it can be done using single method and checking for button id and handling it accordingly.

estate/models/estate_property_offer.py
jmra-odoo: Yes, I should have really thought about it, this makes it more easier in just one line:
offer.property_id.property_offer_ids.status = 'refused' 
and then changing the status of current record as 'accepted'.

estate/models/estate_property_offer.py
jmra-odoo: Yes using in list concept: record.status in ['accepted', 'refused']
if there are many options to check then this can be used.

estate/models/estate_property.py
```diff
@@ -0,0 +1,120 @@
+from re import T
```
rare-odoo, line 1: Is this required?

estate/models/estate_property.py
```diff
@@ -0,0 +1,120 @@
...
+    garden = fields.Boolean(string="Garden")
+    garden_area = fields.Integer(string="Garden Area")
+    
+    garden_orientation = fields.Selection(
+        string="Garden Orientation", 
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]);
+    
+    active = fields.Boolean(string="Is Active", default=True)
+    
+    state = fields.Selection(
```
rare-odoo, line 41: Why did you make the state field `required`, even though a `default` value is being used?

estate/models/estate_property.py
```diff
@@ -0,0 +1,120 @@
...
+                self.state = 'sold'
+        elif self.env.context.get('button_id') == 'cancelled_button':
+            if self.state == 'sold':
+                raise UserError('A sold property cannot be cancelled.')
+            else:
+                self.state = 'cancelled'
+
+    def action_create_invoice(self):
+        # this method is defined to be inherited by estate_account module and create the invoice
+        return True
```
rare-odoo, line 120: Why did you add this method here?
It should be placed in the estate account module.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,67 @@
...
+        for record in self:
+            if record.date_deadline:
+                record.validity = (record.date_deadline - fields.Date.today()).days
+            else:
+                record.validity = 7
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if self.env['estate.property'].browse(vals['property_id']).best_offer:
```
rare-odoo, line 40:
```suggestion
        for vals in vals_list:
            property_id = self.env['estate.property'].browse(vals['property_id'])
            best offer = property_id.best_offer
            expected_price = property_id.expected_price
            if best_offer:
```

Since you're using browse multiple times, it's better to store the result in a variable. 
This will help prevent multiple tiime same SQL queries.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,67 @@
...
+                if vals.get('price') < self.env['estate.property'].browse(vals['property_id']).expected_price or vals.get('price') < self.env['estate.property'].browse(vals['property_id']).best_offer:
+                    max_val = max(self.env['estate.property'].browse(vals['property_id']).best_offer, self.env['estate.property'].browse(vals['property_id']).expected_price)
+                    raise UserError(f"The offer must be higher than best offer {max_val}")
+
+            self.env['estate.property'].browse(vals['property_id']).state = 'offer_received'
+        return super(EstatePropertyOffer, self).create(vals_list)
+
+    # Action Button Methods
+    def action_set_offer_status_accepted(self):
+        for offer in self:
```
rare-odoo, line 50: Why are you using a for loop?
Are you expecting multiple records?
Is it possible to handle multiple records at once?

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,21 @@
...
+
+class EstatePropertyType(models.Model):
+    _name = "estate.property.type"
+    _description = "Estate Property Type"
+    _order = "sequence, name"
+
+    name = fields.Char(string="Name", required=True)
+    property_ids = fields.One2many('estate.property', 'property_type_id', string="Properties")
+    sequence = fields.Integer(string="Sequence", default=1)
+    offer_ids = fields.One2many('estate.property.offer', 'property_type_id', string="Offers")
```
rare-odoo, line 11: What is the purpose of adding the offer ids in the property type model?

estate/models/res_users.py
```diff
@@ -0,0 +1,8 @@
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    property_ids = fields.One2many('estate.property', 'salesman', domain=[('date_availability', '>=', fields.Date.today())])
+
```
rare-odoo, line 8: Remove extra line

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,208 @@
...
+                    <button name="action_set_sold_status" type="object" string="Sold" 
+                            invisible="state == 'sold' or state == 'cancelled' or state == 'offer_accepted'  or state == 'invoiced'" 
+                            context="{'button_id': 'sold_button'}" 
+                            class="btn-secondary"
+                    />
+                    <button name="action_set_cancelled_status" type="object" string="Cancel" 
+                            invisible="state == 'sold' or state == 'cancelled' or state == 'invoiced'"  
+                            context="{'button_id': 'cancelled_button'}" 
+                    />
+                    <button name="action_create_invoice" type="object" string="Invoice"
```
rare-odoo, line 65: The `invoice` button should not be added here!
It should be added in the estate account module instead.

rare-odoo (changes requested):
Hello @jmra-odoo 👋

Great work so far 👏
This is the third set of review.

estate/models/estate_property.py
```diff
@@ -0,0 +1,120 @@
...
+    def _check_selling_price(self):
+        for record in self:
+            if not float_is_zero(record.selling_price, precision_digits=2):
+                if float_compare(record.selling_price, record.expected_price * 0.9, precision_digits=2) < 0:
+                    raise ValidationError("The selling price must be greater than the 90% of expected price.")
+
+    @api.ondelete(at_uninstall=False)
+    def _delete_property(self):
+        if self.state != 'new' and self.state != 'cancelled':
+            raise UserError("Only new and cancelled propeties can be deleted.")
```
rare-odoo, line 89: Once again, please follow the coding guidelines.
The method name should begin with `unlink_`

Also, could you clarify the purpose of adding the `at_uninstall` parameter?

rare-odoo (changes requested):

estate/models/estate_property.py
jmra-odoo: No, that was just retained by mistake was something related to regular expressions.

estate/models/estate_property.py
jmra-odoo: if we don't provide required then there is possibility of getting null value while creating property programmatically with state = None. In worst case otherwise it can be removed if there is no such creation of record programmatically.

estate/models/estate_property.py
jmra-odoo: Okay, I will do it.

estate/models/estate_property_offer.py
jmra-odoo: Yes, I had already done it in previous push i think.

estate/models/estate_property_offer.py
jmra-odoo: There shouldn't be the for loop as there is no chance to accept the offer for multiple record even after list view having the  multi_edit="1". It is not possible i think it should be removed.

estate/models/estate_property_type.py
jmra-odoo: it helps to count the no of offers that property type have.

estate/models/res_users.py
jmra-odoo: Okay, I will do it.

estate/views/estate_property_views.xml
jmra-odoo: Okay, I will do it.

estate/models/estate_property.py
jmra-odoo: Sure will go through with codebase and coding guidelines once.
at_uninstall parameter works while uninstalling the module, all data should be deleted so when reinstalling the module it doesn't make any conflicts or data inconsistency when False and set to True if there is any data that should not be deleted.

PR #316, opened 2025-02-05
[ADD] real_estate: added real estate moduel

nmak-odoo:
Created the real estate module for technical training


nmak-odoo:
Moved to https://github.com/odoo/tutorials/pull/327

PR #324, opened 2025-02-06
[ADD] estate: new estate module

apri-odoo:
- introduced new estate module


sjai-odoo (changes requested):

estate/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    'description': """ 
+        This is a base version of the Estate module for real estate management. 
+    """,
+    'license': 'LGPL-3',
+    'data': [
+        'security/ir.model.access.csv',
+    ],
+    'installable': True,
+    'application': True,
+    'auto_install': False,
```
sjai-odoo, line 14:
```suggestion
    'application': True,
```
`installable` is True by default no need to add it explicitely
`auto_install` is only useful when you have other modules as dependencies, it makes no sense here.

estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
...
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date()
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float()
+    bedrooms = fields.Integer()
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
```
sjai-odoo, line 17:
```suggestion
    is_garage = fields.Boolean()
    is_garden = fields.Boolean()
```
Try to define boolean fields in this way which make it more readable and easy to code.

sjai-odoo (changes requested):

PR #332, opened 2025-02-06
[ADD] estate:  Create an estate app

maap-odoo:
- Created an estate application for Odoo
- Added access rights in `ir.model.access.csv` for `estate.property`
- Created XML data files for actions and menus in the estate module
- Defined three-level menu structure for property management
- Implemented default values and field attributes for `estate.property`
- Added `active` and `state` fields with predefined behaviors

### DIFF
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -1,0 +1,11 @@
+{
+    "name": "estate",
+    "application":True,
+    "license": "LGPL-3",
+    "instalable": True,
+    "data" : [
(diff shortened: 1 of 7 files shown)


estate/__manifest__.py
sami-odoo, line 1: Take the help of code base to check the order in which we specify the keys of manifest

maap-odoo: @sami-odoo The changes are done. Thank you for informing.

PR #349, opened 2025-02-13
[ADD] estate: Added estate training module

smra-odoo:
- created a new module, estate.
- added init, manifest files.
- added estate_property model and access rights for the model.

### DIFF
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -1,0 +1,15 @@
+{
+    'name': "estate",
+    'version': '18.0',
+    'depends': ['base'],
+    'author': "Smit",
+    'category': 'Real Estate',
+    'description': """
+        Buy and sell properties.
+    """,
+    'installable': True,
+    'application': True,
+    'data': [
+        'security/ir.model.access.csv',
+    ]
+}
--- a/estate/security/ir.model.access.csv
+++ b/estate/security/ir.model.access.csv
@@ -1,0 +1,3 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+
+access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
--- a/estate/__init__.py
+++ b/estate/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/estate/models/__init__.py
+++ b/estate/models/__init__.py
@@ -1,0 +1,1 @@
+from . import estate_property
--- a/estate/models/estate.property.py
+++ b/estate/models/estate.property.py
@@ -1,0 +1,25 @@
+from odoo import fields,models
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "See properties"
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date()
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float()
+    bedrooms = fields.Integer()
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        string='Type',
+        selection=[('north','North'),
+                ('east','East'),
+                ('west','West'),
+                ('south','South')]
+    )


estate/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    'author': "Smit",
+    'category': 'Real Estate',
+    'description': """
+        Buy and sell properties.
+    """,
+    'installable': True,
+    'application': True,
+    'data': [
+        'security/ir.model.access.csv',
+    ]
```
sosh-odoo, line 14: Please add Licence version

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+
```
sosh-odoo, line 2: there is no need to keep the blank line in this file

estate/__manifest__.py
smra-odoo: 👍

estate/models/estate_property.py
```diff
@@ -0,0 +1,67 @@
...
+
+    @api.depends("offer_ids")
+    def _compute_best_price(self):
+        for record in self:
+            if record.offer_ids:
+                record.best_price = max(record.offer_ids.mapped("price"))
+            else:
+                record.best_price = 0
+
+    @api.onchange("garden")
```
sosh-odoo, line 60: In Odoo we prefer not to use 'onchange' generally, what's the reason behind that?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,32 @@
...
+    status = fields.Selection(
+        selection=[('accepted','Accepted'),
+                    ('refused','Refused')],
+        copy=False,
+    )
+    partner_id = fields.Many2one("res.partner", string="Partner", required=True)
+    property_id = fields.Many2one("estate.property", string="Properties", required=True, ondelete="cascade")
+
+    validity = fields.Integer(default="7")
+    date_deadline = fields.Date(compute="_compute_date_deadline",inverse="_inverse_date_deadline",store=True)
```
sosh-odoo, line 19: whats the purpose of using inverse field here? Also, why are we using store=True, what does that do?

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,34 @@
...
+                <field name="price"/>
+                <field name="partner_id"/>
+                <field name="date_deadline" string="Date Deadline" />
+                <field name="validity" string="Validity (days)" />
+                <field name="status" optional="hidden"/>
+            </list>
+        </field>
+    </record>
+
+    <record id="estate_property_offer_view_form" model="ir.ui.view">
```
sosh-odoo, line 17: what is the difference between ir.actions.act_window and ir.ui.view?

estate/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+{
+    'name': "estate",
+    'version': '1.0',
```
sosh-odoo, line 3: give proper version for the module, since its a module for Odoo v18 we will give it as 18.0, and then updating it with additional features

estate/models/estate_property.py
smra-odoo: The main reason for not using 'onchange' method generally because it does not get triggered outside of form view, so when we create any record programmatically it does not automatically gets triggered and creates inconsistency.

estate/__manifest__.py
smra-odoo: 👍

estate/views/estate_property_offer_views.xml
smra-odoo: The main difference between them is, it.actions.act_window defines an action that opens a specific model in a particular view (form, tree, kanban, etc.), it also links a menu item to view and it.ui.view defines the structure/layout of a view for a model.

sosh-odoo: can you give a little bit clarity on this?

estate/models/estate_property_tags.py
```diff
@@ -0,0 +1,14 @@
...
+
+class EstatePropertyTag(models.Model):
+    _name = "estate.property.tags"
+    _description = "Property Tags"
+    _order = "name desc"
+
+    name = fields.Char(required=True)
+    color = fields.Integer(string="Color")
+
+    _sql_constraints = [
```
sosh-odoo, line 11: what can be the substitute for this constraint? Why that substitute is not used here?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,64 @@
...
+    property_id = fields.Many2one("estate.property", string="Properties", required=True, ondelete="cascade")
+    property_type_id = fields.Many2one("estate.property.type", string="Property Types", store=True)
+
+    _sql_constraint = [
+        ('check_pos_offer_price','CHECK(price > 0)','Offered price should be greater than 0')
+    ]
+
+    @api.depends("validity","create_date")
+    def _compute_date_deadline(self):
+        for record in self:
```
sosh-odoo, line 29: why 'for record in self', why are we not directly updating self here?

estate/models/estate_property_offer.py
smra-odoo: The inverse function is used to allow bidirectional updates between a computed field (date_deadline) and its dependent field (validity) and without an inverse function, date_deadline would be read-only. The main purpose to use inverse function is to let users to manually set date_deadline, and then automatically update validity accordingly.
The store=True attribute makes sure that the computed field's value is stored in the database,so we can filter and search records by date_deadline.

estate/models/estate_property_offer.py
smra-odoo: Here, self can be a recordset containing multiple records, not just a single record so if we tried to update self.date_deadline directly, it would attempt to assign a single value to all records, which is incorrect. Using 'for record in self' we are updating each record individually with its correct date_deadline value as each record may have different create_date and validity.

estate/models/estate_property_tags.py
smra-odoo: There are two types of constraints in odoo, python constraints and sql constraints. for this constraint the substitute would be
`  @api.constrains('name')
    def _check_unique_name(self):
        for record in self:
            existing_tag = self.search([('name', '=', record.name), ('id', '!=', record.id)])
            if existing_tag:
                raise ValidationError("Property Tags should be unique!")`
this substitute is not used here as sql constraints prevent duplicates at the database level, which is faster and more efficient. sql constraints guarantee uniqueness at all times whereas python constraints can be bypassed in few cases. Using sql constraints avoids race conditions as it has no concurrency issues whereas python constraints has concurrency issues. Python constraints require a database query (search()) every time a record is saved, which is slower than an indexed sql constraint. So to get uniqueness every time the substitute is not used here.

estate/views/estate_property_offer_views.xml
smra-odoo: - ir.actions.act_window is an action definition, it defines an action that odoo executes when a menu item or button is clicked. It specifies which model to open and in which view type (list, form, kanban, etc.).
-  ir.ui.view is an view definition, it defines the structure and layout of a view. It specifies how fields are displayed in list, form, kanban, or other views. It only determines the UI structure.
- Flow - A menu item or button triggers an action (ir.actions.act_window). The action opens a model in a specified view type (list, form, etc.). The view layout is defined by ir.ui.view.

sosh-odoo: Nice

warranty/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    'version': '18.0',
+    'depends': ['sales_management','stock'],
+    'author': "Smit",
+    'category': 'Sales/Warranty',
+    'license' : 'LGPL-3',
+    'description': """
+        Extend Warranty.
+    """,
+    'data': [
+        'views/estate_actions.xml',
```
sosh-odoo, line 12: Why we have this file estate_actions.xml in this?
Its a good practice to push code when a functionality or feature is complete or we can say is workable enough and won't affect other features/functionalities.

PR #350, opened 2025-02-13
[ADD] estate training : Created a new training module

jhra-odoo:
Module Setup: Created the Real Estate module, made it installable, and added access control.
Core Features: Added the estate_property model with key fields and default inactive entries.
UI & Interactivity: Implemented menus, actions, an app icon, and improved property forms.
Navigation & Search: Added list views, filters, grouping, and property types for better categorization.
Updates: Changed license to LGPL-3 and updated the manifest.


estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,31 @@
...
+    status = fields.Selection(
+        string="Status",
+        selection=[
+            ('accepted', 'Accepted'),
+            ('refused', 'Refused')],
+        copy=False,
+    )
+    partner_id = fields.Many2one('res.partner', string="Partner", required=True)
+    property_id = fields.Many2one('estate.property', string="Property", required=True)
+    date_deadline = fields.Date(copy=False, default=fields.Date.today(), compute='_compute_date_deadline', inverse='_inverse_date_deadline', store=True)
```
sosh-odoo, line 17: whats the purpose of using inverse field here? Also, why are we using store=True, what does that do?

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="estate_property_tag_action" model="ir.actions.act_window">
```
sosh-odoo, line 3: what is the difference between ir.actions.act_window  and ir.ui.view?

estate/models/estate_property_offer.py
jhra-odoo: The inverse method (_inverse_date_deadline) allows date_deadline to be updated manually by users. When modified, it recalculates and updates the **validity** field accordingly, ensuring consistency between deadline and validity period. As, by default a calculated field will be **read-only**, so this write back action can't be performed.

Setting store=True saves date_deadline in the database, making it searchable, filterable, and sortable in Odoo views. Without it, the field is recalculated on the fly, impacting performance and preventing domain filtering.

estate/views/estate_property_tag_views.xml
jhra-odoo: ir.actions.act_window defines how a view opens, specifying the model, view type (form, tree, etc.), and behavior. ir.ui.view defines the UI layout, determining how fields and elements are displayed in forms, lists, or kanbans.

So, simply put: ir.actions.act_window opens a view, and ir.ui.view defines its appearance.

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,13 @@
...
+
+class EstatePropertyTag(models.Model):
+    _name = "estate.property.tag"
+    _description = "Property Tags"
+    _order = "name"
+
+    name = fields.Char(required=True)
+    color = fields.Integer(default=1)
+
+    _sql_constraints = [
```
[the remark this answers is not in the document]
jhra-odoo: A Python constraint (@api.constrains) could replace the SQL constraint by checking duplicates with search_count(). However, the SQL constraint is used because it's more efficient, prevents duplicates at the database level, and avoids race conditions in multi-user environments.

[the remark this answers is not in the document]
sosh-odoo: can you please let me know what is this search_count(), when and why it is used?

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,6 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,1
+estate.access_estate_property_type,access_estate_property_type,estate.model_estate_property_type,base.group_user,1,1,1,1
+estate.access_estate_property_tag,access_estate_property_tag,estate.model_estate_property_tag,base.group_user,1,1,1,1
```
sosh-odoo, line 4: in these you have mentioned 'base.group_user' what is that?

estate_account/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    'summary': """
+        Real Estate Account Management
+    """,
+    'description': """
+        Real Estate Account Management App
+    """,
+    'author': "Rahul",
+    'website': "https://www.odoo.com/",
+    'category': 'Tutorials/RealEstateAccount',
+    'version': '0.1',
```
sosh-odoo, line 13: version is 0.1? Are you serious! XD
By the way, What this version actually means?

estate/models/estate_property_tag.py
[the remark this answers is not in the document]
jhra-odoo: search_count() is an Odoo ORM method that returns the number of records matching a given domain. It is used when we only need the count of records instead of fetching full record data.

Usage Example: count = self.env['estate.property'].search_count([('state', '=', 'sold')])

estate/security/ir.model.access.csv
jhra-odoo: base.group_user is the default user group of Odoo, applying to all internal users. And that 'base' prefix is because it's defined in the base module.

basically, each row in the ir.model.access.csv file defines access control for a model and assigns it to a group.
so, by setting group_id:id to base.group_user, we are giving all internal users permissions to the specified models.

estate_account/__manifest__.py
jhra-odoo: Well, at this stage, the accounting module can only generate invoices for properties marked as sold. Without any proper accounting and transaction logic implemented, it doesn’t really deserve an integer version just yet. Basically, consider this an alpha release. 🗿

As for what that version is, it's a string that represents the version number of an Odoo module. It helps track updates, manage dependencies, and indicate compatibility with Odoo versions. And a version of 0.x indicates an experimental or alpha-stage software.

PR #352, opened 2025-02-17
[ADD] estate: add the manifest of the new estate app

AlessandroLupo:
This concludes the chapter 2 of the "Server Framework 101" tutorial


estate/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
```
rugo-odoo, line 1: We don't need those anymore.

estate/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+
+{
+    "name": "Real Estate",
+    "version": "0.1",
+    "category": "Administration",
+    "sequence": 100,
+    "license": "LGPL-3",
+    "author": "alup",
```
rugo-odoo, line 9:
```suggestion
    "author": "Odoo S.A.",
```
We work for Odoo S.A. :smile:

estate/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    "depends": [
+        "base_setup",
+    ],
+    "data": [
+        "security/ir.model.access.csv",
+        "views/estate_property_views.xml",
+        "views/estate_menus.xml",
+    ],
+    "demo": [],
+    "css": [],
```
rugo-odoo, line 22: No need if they are empty.

estate/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    "category": "Administration",
+    "sequence": 100,
+    "license": "LGPL-3",
+    "author": "alup",
+    "summary": "Track real estate",
+    "description": "",
+    "website": "https://www.odoo.com/page/estate",
+    "depends": [
+        "base_setup",
+    ],
```
rugo-odoo, line 15: ``
```suggestion
    "depends": [
        "base",
    ],
```
Right ?

estate/models/estate_property.py
```diff
@@ -0,0 +1,51 @@
...
+from odoo import fields, models
+
+
+class Property(models.Model):
+    _name = "estate.property"
+    _description = "Real estate property"
+
+    name = fields.Char("Property Name", required=True)
+    description = fields.Text("Property description")
+    active = fields.Boolean("Active", default=True)
```
rugo-odoo, line 12: By default, the label will take the name of the field and capitalize it. So in this case it will be, by default, `Active`. So we don't need to manually add it.

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- Top menu item -->
```
rugo-odoo, line 3: Comment don't really add any value. We can remove them. (when you clean up)

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,104 @@
...
+    </record>
+
+    <record id="estate_property_form" model="ir.ui.view">
+        <field name="name">estate.property.form</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form string="Property">
+                <sheet>
+                    <div class="oe_title">
+                        <h1 class="mb32">
```
rugo-odoo, line 26: We use a class but we don't have any css. Also, we use bootstrap. There is a lot of helpers than you can use instead of doing custom css (that probably already exists).

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,104 @@
...
+                <field name="living_area" string="Living Area (sqm)" />
+                <field name="facades" string="Facades" />
+
+                <filter string="Available" name="available"
+                    domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]" />
+                <filter string="Archived" name="not_active" domain="[('active', '=', False)]" />
+
+                <group expand="1" string="Group By">
+                    <filter string="Postcode" name="postcode"
+                        context="{'group_by':'postcode', 'residual_visible':True}" />
```
rugo-odoo, line 86: What is `residual_visible` ?

rugo-odoo (changes requested):
A few things.

estate/views/estate_property_views.xml
AlessandroLupo: I saw it [here](https://github.com/odoo/odoo/blob/3217cfa993efdb9da994a6786b9a72a854badcff/addons/delivery/views/delivery_carrier_views.xml#L32) and I thought that it defined whether the entries without group are showed or not. But apparentrly I was wrong, they are always listed under "none" anyway. :smile:

estate/models/estate_property.py
```diff
@@ -0,0 +1,152 @@
+from odoo import api, fields, models, tools
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools import float_compare, float_is_zero
+
+
+class Property(models.Model):
+    _name = "estate.property"
+    _description = "Real estate property"
+
+    rounding = 0.01
```
rugo-odoo, line 10: We don't usually do that but I guess it's fine here.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,215 @@
...
+                <header>
+                    <button name="action_set_sold" class="oe_highlight" type="object" string="Sold" />
+                    <button name="action_cancel" type="object" string="Cancel" />
+                </header>
+                <sheet>
+                    <group>
+                        <div class="oe_title">
+                            <h1 class="o_row">
+                                <field name="name"
+                                    placeholder="Property Name, e.g. House in Brussels" />
```
rugo-odoo, line 42:
```suggestion
                                <field name="name" placeholder="Property Name, e.g. House in Brussels" />
```
We can inline (easier to read as it's not too long).

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,215 @@
...
+                <field name="property_type_id" string="Property Type" />
+                <field name="postcode" string="Postcode" />
+                <field name="expected_price" string="Expected Price" />
+                <field name="bedrooms" string="Bedrooms" />
+                <field name="living_area" string="Living Area (sqm)" />
+                <field name="garden_area" string="Garden Area (sqm)" />
+                <field name="total_area" string="Total Area (sqm)" />
+                <field name="facades" string="Facades" />
+                <filter string="Available" name="available"
+                    domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]" />
```
rugo-odoo, line 110:
```suggestion
                    domain="[('state', 'in', ('new', 'offer_received'))]" />
```

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
```
rugo-odoo, line 2: We can do them as a list. It's easier to read.

```suggestion
<menuitem>
    <menuitem/>
    <menuitem/>
</menuitem>
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,152 @@
...
+    def action_set_sold(self):
+        flag = False
+        for record in self:
+            if record.state != "cancelled":
+                record.state = "sold"
+            else:
+                flag = True
+        if flag:
+            raise UserError("Cancelled properties can't be sold.")
+        return True
```
rugo-odoo, line 124:
```suggestion
    if self.filtered(lambda record: record.state == 'cancelled'):
        raise UserError("Cancelled properties can't be sold.")
        
    self.state = "sold"
```

1. If a `UserError` is raised, the whole transactions will rollback.
2. We don't need to write record by record, we can write in batch by using self directly.

(This should do the same.)

rugo-odoo (changes requested):

estate/models/estate_property.py
AlessandroLupo: I see that the precision is usually defined in the currency or in the UoM, but in the case none of the two options is available (right?). Is there a default setting that I can use?

estate/models/estate_property.py
AlessandroLupo: This is way better, thanks

estate/models/estate_property.py
rugo-odoo: Yep, you are right ! In this case there is not really a default, that's why it's fine here :smiley:

PR #353, opened 2025-02-17
[IMP] tutorials: first commit

roto-odoo:
modules: tutorials/awesome_clicker
files: __init__.py

first commit following the instructions of the tutorials


estate/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
```
clbr-odoo, line 1: This is not necessary in Python files for a long time already.
There are some that are still there in the code for historical reason, but please don't add them on new files :slightly_smiling_face: 
AFAIK, the licensing is also not really necessary to add on each file since it's defined in the manifest for the whole module.

estate/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+from dateutil.relativedelta import relativedelta
+from odoo import fields, models
+
```
clbr-odoo, line 5: Two empty lines between imports and class.

estate/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
...
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate"
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(default=fields.Date.today()+relativedelta(months=3), copy=False)
```
clbr-odoo, line 13: Just a small note:
`fields.Date.today()` will actually give you the server time, if you want to have the actual today of the user (which is the case most of the time for business logic) you'll have to use `context_today`.

estate/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
...
+    garden_orientation = fields.Selection(
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
+    )
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
+        selection=[('new', 'New'), ('offer_received', 'Offer Received'), ('offer_accepted', 'Offer Accepted'),
+                   ('sold', 'Sold'), ('cancelled', 'Cancelled')],
+        copy=False,
+        default='new',
+        required=True
```
clbr-odoo, line 31:
```suggestion
    state = fields.Selection(
        selection=[
            ('new', 'New'), 
            ('offer_received', 'Offer Received'), 
            ('offer_accepted', 'Offer Accepted'),
            ('sold', 'Sold'), ('cancelled', 'Cancelled'),
        ],
        copy=False,
        default='new',
        required=True,
```
Don't hesitate to use the space if it makes it more readable. Also it's a good practice to add a `,` even on the last parameter in such case: if someone needs to add something later on, he won't have to modify the last line to add it, and that will keep the git history very clean.

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,7 @@
+<odoo>
+    <menuitem id="estate_menu_root" name="Real Estate">
+        <menuitem id="estate_first_level_menu" name="Advertisements">
```
clbr-odoo, line 3: We also have convention for xml ids, although they are quite recent and therefore not that often enforced in the code for the moment.
I suggest you read this section again and update them accordingly :slightly_smiling_face: 
https://www.odoo.com/documentation/master/contributing/development/coding_guidelines.html#xml-ids-and-naming

clbr-odoo (approved):
Good job so far, keep up :+1:

awesome_clicker/__init__.py
```diff
@@ -1 +1,5 @@
 # -*- coding: utf-8 -*-
+
+#TEST COMMIT
+
+print("test")
```
rugo-odoo, line 5: We can remove this now :smile:

estate/models/estate_property.py
```diff
@@ -0,0 +1,52 @@
...
+        copy=False,
+        default='new',
+        required=True,
+    )
+    property_type_id = fields.Many2one(
+        'estate.property.type', string='Property Type',
+    )
+    partner_id = fields.Many2one(
+        'res.partner', string='Buyer',
+    )
```
rugo-odoo, line 41: Inline all of those. Let's stay consistent with the indentation we did previously.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,107 @@
...
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <search string="Search Properties">
+                <field name="name" string="Title"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area" string="Living Area (sqm)"/>
+                <field name="facades"/>
+                <filter name="available" string="Available" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"/>
```
rugo-odoo, line 84:
```suggestion
                <filter name="available" string="Available" domain="['('state', 'in', ('new', 'offer_received'))]"/>
```

rugo-odoo (changes requested):

PR #354, opened 2025-02-17
Training blse

blse-odoo:


estate/models/estate_property.py
```diff
@@ -0,0 +1,34 @@
+from odoo import models, fields
+
+class Property(models.Model):
```
hupo-odoo, line 3:
```suggestion


class Property(models.Model):
```
in python you should always put 2 lines above a class definition

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,99 @@
...
+                            </group>
+                            <group>
+                                <field name="garden"/>
+                            </group>
+                            <group>
+                                <field name="garden_area"/>
+                            </group>
+                            <group>
+                                <field name="garden_orientation"/>
+                            </group>
```
hupo-odoo, line 67: I don't think you need a group for each field, you can put them all in one group.

hupo-odoo (commented):
Good job! Everything seems to go well, I just made 2 small comments :smile:

estate/views/estate_property_views.xml
blse-odoo: Yes, thank you

PR #358, opened 2025-02-17
18.0 technical training lmer

LuneMercier:


estate/models/estate_property.py
```diff
@@ -0,0 +1,36 @@
...
+import datetime
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Property of the estate"
+
+    name = fields.Char('Name', required=True)
+    description = fields.Text('Description')
+    postcode = fields.Char('Postcode')
+    date_availability = fields.Date('Availability Date', copy=False, default=lambda _: fields.Date.today()+datetime.timedelta(days=90))
```
rugo-odoo, line 12:
```suggestion
    date_availability = fields.Date('Availability Date', copy=False, default=fields.Date.add(fields.Date.today(), months=3))
```
We can also do it like this.

rugo-odoo (changes requested):

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,155 @@
...
+        <field name="arch" type="xml">
+            <search string="Estate Property">
+                <field name="name"/>
+                <field name="property_type_id"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="facades"/>
+                <filter string="Available" name="available" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"/>
```
rugo-odoo, line 97:
```suggestion
                <filter string="Available" name="available" domain="[('state', 'in', ('new', 'offer_received'))]"/>
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,36 @@
...
+    def action_set_accepted(self):
+        self.write({'state': 'accepted'})
+
+    def action_set_refused(self):
+        self.write({'state': 'refused'})
+
+    @api.depends('validity')
+    def _compute_deadline(self):
+        for record in self:
+            record.date_deadline = (record.create_date if hasattr(record, 'create_date') else fields.Date.today()) + datetime.timedelta(days=record.validity)
```
rugo-odoo, line 32: There is 99.99% chances that we don't need to use `hasttr` here. Can we do it any other way ?

rugo-odoo (changes requested):

estate/models/estate_property.py
```diff
@@ -0,0 +1,36 @@
...
+        string='Garden Orientation',
+        selection=[('north', 'North'), ('west', 'West'), ('south', 'South'), ('east', 'East')]
+    )
+    state = fields.Selection(
+        selection=[
+            ('new', 'New'),
+            ('offer_received', 'Offer Received'),
+            ('offer_accepted', 'Offer Accepted'),
+            ('sold', 'Sold'),
+            ('cancelled', 'Cancelled')
```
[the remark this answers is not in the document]
LuneMercier: ok I see, thanks

estate/__manifest__.py
```diff
@@ -0,0 +1,9 @@
+{
```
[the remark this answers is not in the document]
LuneMercier: I put 'base_setup', should there be anything else ?

estate/models/estate_property_offer.py
LuneMercier: I have'nt found any documentation on how records properties are defined, I'll check in the source code

estate/models/estate_property.py
```diff
@@ -0,0 +1,104 @@
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare
+
+import datetime
+
```
hupo-odoo, line 6: class should be preceded by 2 empty line in python :smile:

estate/models/estate_property.py
```diff
@@ -0,0 +1,104 @@
...
+    _sql_constraints = [
+        ('selling_price', 'CHECK(selling_price > 0)', 'Prices must be strictly positive.'),
+        ('expected_price', 'CHECK(expected_price > 0)', 'Prices must be strictly positive.'),
+    ]
+
+    _order = "id desc"
+
+    def action_set_cancelled(self):
+        if self.state == 'sold':
+            raise UserError('Sold properties can not be cancelled')
```
hupo-odoo, line 61: you need to take into consideration that self might be a recordset, so therefore multiple values are possible and I think this will trigger a traceback. In your case, either you do an ensure_one to ensure your function can only be called on 1 record, or you handle the multi case.

estate/models/estate_property.py
```diff
@@ -0,0 +1,104 @@
...
+        else:
+            self.write({'state': 'cancelled'})
```
hupo-odoo, line 63: we usually use the write function when there are multiple fields to set, but as you only have one, the cleanest solution is to directly do `self.state = 'cancelled'`

estate_account/__manifest__.py
hupo-odoo, line 1: should this module be auto-install ? Haven't looked at the tutorial but looks like it's just a bridge module, which usually are auto installed

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,29 @@
+from odoo import Command, models
+
+class EstateProperty(models.Model):
+    _inherit = "estate.property"
+
+    def action_set_sold(self):
+        move = self.env['account.move'].sudo().with_context(default_move_type='out_invoice').create(
```
hupo-odoo, line 7:
```suggestion
        move = self.env['account.move'].create(
```
you can specify the move_type directly in the fields below. For the sudo, you should be very careful when using it, and I don't think you need it here unless I'm mistaken

hupo-odoo (commented):
Good job! JS is coming soon :smile:

PR #361, opened 2025-02-17
18.0 training kege

KevGerard:


estate/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "estate",
+    "version": "1.0",
+    "depends": ["base"],
+    "author": "Kevin Gerard",
+    "category": "Category",
+    "description": """
+        This is a module to manage real estate properties
+    """,
+    "data": ["data/ir.model.access.csv"],
```
clbr-odoo, line 10:
```suggestion
    "data": ["security/ir.model.access.csv"],
```
By convention, security files must be under "security" folder.

estate/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "application": True,
+    # data files always loaded at installation
+    # 'data': [
+    #     'views/mymodule_view.xml',
+    # ],
+    # # data files containing optionally loaded demonstration data
+    # 'demo': [
+    #     'demo/demo_data.xml',
+    # ],
```
clbr-odoo, line 19: Don't forget to cleanup useless comments.

estate/models/estate_property.py
```diff
@@ -0,0 +1,22 @@
+from odoo import fields, models
+
+
+class EstatePropertyModel(models.Model):
+    _name = "estate_property"
```
clbr-odoo, line 5:
```suggestion
    _name = "estate.property"
```
By convention, models `_name` are separated with dots.

estate/models/estate_property.py
```diff
@@ -0,0 +1,22 @@
...
+    _description = "Defines the model of a real estate property"
+
+    name = fields.Char(required=True)
```
clbr-odoo, line 8: Your fields looks good for now, just a small explanation on the labels of the fields:
Odoo tries to automatically compute them if there is none provided.
So `date_availability` will become `Date Availability` for example. If you want to provide one, you can do something like `date_availability = fields.Date(string="Date of availability")`

estate/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "estate",
+    "version": "1.0",
+    "depends": ["base"],
+    "author": "Kevin Gerard",
```
rugo-odoo, line 5:
```suggestion
    "author": "Odoo S.A.",
```
We work for Odoo S.A. :smile:

estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_property, estate_property_type, estate_property_tag, estate_property_offer
```
rugo-odoo, line 1:
```suggestion
from . import estate_property
from . import estate_property_type
from . import estate_property_tag
from . import estate_property_offer
```
I'm not sure why but we do it like this :smile:

estate/models/estate_property.py
```diff
@@ -0,0 +1,50 @@
...
+    _name = "estate.property"
+    _description = "Defines the model of a real estate property"
+
+    name = fields.Char(required=True, string="Title")
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        string="Date of availability",
+        copy=False,
+        default=lambda _: fields.date.today() + datetime.timedelta(days=90),
```
rugo-odoo, line 16:
```suggestion
        default=fields.Date.add(fields.Date.today(), months=3),
```
I think we can do this. It's cleaner.

estate/models/estate_property.py
```diff
@@ -0,0 +1,50 @@
...
+    bedrooms = fields.Integer(string="Number of bedrooms", default=2)
+    living_area = fields.Integer(string="Living area")
+    facades = fields.Integer(string="Number of facades")
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer(string="Garden area")
+    garden_orientation = fields.Selection(
+        [("north", "North"), ("south", "South"), ("east", "East"), ("west", "West")],
+        string="Garden orientation",
+    )
```
rugo-odoo, line 29: Indentation consistency. We have a different indentation for state (which is also a `Selection` field).

Also, the indentation in state is better because we can add values without breaking the git history.

rugo-odoo (changes requested):

estate/models/estate_property.py
KevGerard: Honestly this was done automatically by ruff depending on the line limit.
What is the usual line limit set in the CI? Maybe I can configure it so we achieve the same result :)
Otherwise, I agree with the comment ;)

estate/models/estate_property.py
```diff
@@ -0,0 +1,121 @@
...
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = None
+            self.garden_orientation = None
+
+    def action_sell_property(self):
+        if any(estate_property.state == "cancelled" for estate_property in self):
+            raise UserError("Cannot sell a cancelled property")
+        for estate_property in self:
+            estate_property.state = "sold"
```
KevGerard, line 84:
```suggestion
       self.state = "sold"
```

KevGerard (commented):
bien jouer moi

PR #362, opened 2025-02-17
[ADD] estate: added the real estate module

hel-mefe:
In this commit, I initialized the real estate module by setting up the essential configuration files, namely __manifest__.py and __init__.py.


estate/views/real_estate_views.xml
```diff
@@ -0,0 +1,14 @@
+<odoo>
+    <!-- Define the action to open your module -->
```
rugo-odoo, line 2: No need to add those comments. They add no value. We are just describing what we are doing which we can see by looking at the code.

estate/__manifest__.py
```diff
@@ -0,0 +1,11 @@
+{
+    'name': 'RealEstate',
+    'version': '1.0.0',
```
rugo-odoo, line 3:
```suggestion
    'version': '1.0',
```
We use versions with 2 numbers.

rugo-odoo (changes requested):
A few details.

estate/models/constants.py
rugo-odoo, line 1: Why did you do it like this ? Where did the idea came from ?

rugo-odoo (changes requested):

PR #364, opened 2025-02-17
18.0 - Technical Onboarding - AHSL

AhmedGamal2212:


estate/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+{
+    'name': 'Real Estate',
+    'version': '18.0',
```
rugo-odoo, line 3:
```suggestion
    'version': '1.0',
```
This is the version of the package, not the version of Odoo.

estate/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+{
```
rugo-odoo, line 1: We are missing a licence :grimacing:

rugo-odoo (changes requested):

estate/__manifest__.py
AhmedGamal2212: Added one, thanks! Please, let me know if there is some reference to read about licenses :)

rugo-odoo (changes requested):

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,44 @@
...
+        for record in self:
+            if record.create_date:
+                record.date_deadline = fields.Date.add(record.create_date, days = record.validity)
+
+    def _inverse_date_deadline(self):
+        for record in self:
+            if record.create_date and record.date_deadline:
+                create_date = fields.Date.from_string(record.create_date)
+                deadline_date = fields.Date.from_string(record.date_deadline)
+                record.validity = (deadline_date - create_date).days
```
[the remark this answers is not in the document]
AhmedGamal2212: The create_date field isn't always set to today's date (fields.Date.today()), so I used it to calculate the actual difference. Additionally, on of them is a date, while the other value is a datetime, causing a type mismatch error when subtracting them.

PR #365, opened 2025-02-20
[ADD] product_kit_sale: add a feature to sell kit products

iemu-odoo:


sale_product_kit/models/product_template.py
```diff
@@ -0,0 +1,45 @@
...
+    def _check_not_sub_product(self):
+        for template in self:
+            if template.is_kit:
+                kit_products = self.env['product.template'].search([
+                    '&',
+                    ('is_kit', '=', True),
+                    ('sub_product_ids', 'in', template.product_variant_ids.ids),
+                ])
+                if kit_products:
+                    raise ValidationError(_("This product cannot become a kit since it is a sub-product of another kit."))
```
ampa-odoo, line 33: Using a `search()` inside `for` will be very heavy I guess. `@api.constrains` are triggered before creation of each record, even if we try to create records externally using CSV import, XMLrpc etc. What do you say? :thinking: 
```suggestion
    @api.constrains('is_kit')
    def _check_not_sub_product(self):
    """Ensure kit is not a sub-product of another kit."""
    kit_records = self.filtered('is_kit')
    if kit_records:
        existing_kits = self.search([('sub_product_ids', 'in', kit_records.mapped('product_variant_ids').ids)])
        if existing_kits:
            raise ValidationError(_("A product cannot be a kit if it is already a sub-product of another kit."))

```

sale_product_kit/models/product_template.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @api.constrains('is_kit', 'sub_product_ids')
+    def _check_sub_product_ids_not_empty(self):
+        for template in self:
+            if template.is_kit and not template.sub_product_ids:
+                raise ValidationError(_("A kit product must contain at least 1 sub product."))
```
ampa-odoo, line 39: Similarly for this, instead of iterating for each product, we can use `.filtered()`.
```suggestion
    def _check_sub_product_ids_not_empty(self):
        if self.filtered(lambda p: p.is_kit and not p.sub_product_ids.exists()):
            raise ValidationError(_("A kit product must contain at least one sub-product."))
```

sale_product_kit/models/product_template.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @api.onchange('is_kit')
+    def _onchange_is_kit(self):
+        for template in self:
+            if not template.is_kit and template.sub_product_ids:
+                template.sub_product_ids = False
```
ampa-odoo, line 45: image

Can you explain a scenario supporting your code? :thinking:

sale_product_kit/models/sale_order_line.py
```diff
@@ -0,0 +1,40 @@
...
+    @api.ondelete(at_uninstall=False)
+    def _unlink_except_sub_product(self):
+        if any(self.mapped('parent_line_id')):
+            raise UserError(_("You cannot delete an order line which was added for a kit product order line."))
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_all_sub_product_lines(self):
+        for line in self:
+            if line.product_id.is_kit:
+                linked_lines = self.env['sale.order.line'].filtered(lambda line: line.parent_line_id.id == self.parent_line_id.id)
```
ampa-odoo, line 29: Firstly, `ondelete="cascade"` is given on parent_line_id field. So do we need this?
Secondly, if we will create a relation using one2many field as well then we don't need to filter this way.

Thirdly,
Aren't we in the same model? :thinking:  for self.env['sale.order.line'].

sale_product_kit/models/sale_order_line.py
```diff
@@ -0,0 +1,40 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError, ValidationError
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    is_kit = fields.Boolean(related='product_id.is_kit')
+
+    parent_line_id = fields.Many2one(comodel_name='sale.order.line', name="Linked kit product line", ondelete="cascade", domain=[('order_id', '=', 'order_id')])
```
ampa-odoo, line 12: Don't we need a one2many field creating a vice-versa relation over here?

sale_product_kit/views/sale_order_views.xml
```diff
@@ -0,0 +1,26 @@
...
+        <field name="name">sale.order.form.inherit.sale.product.kit</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='order_details']" position="inside">
+                <field name="print_in_report"/>
+            </xpath>
+            <xpath expr="//field[@name='order_line']/list" position="attributes">
+                <attribute name="decoration-warning" add="(parent_line_id)" separator="or"/>
+
```
ampa-odoo, line 13:
```suggestion
```

ampa-odoo:
I've added few comments and different approach to implement with. @iemu-odoo

sale_product_kit/__manifest__.py
```diff
@@ -0,0 +1,30 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+{
+    'name': "Sale Product Kit",
+    'category': 'Sales',
```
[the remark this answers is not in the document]
iemu-odoo: Yup. That's how it should be.

sale_product_kit/models/product_template.py
iemu-odoo: In my updated code, I used any() as i found that more clear to read than filtered().

sale_product_kit/__manifest__.py
```diff
@@ -0,0 +1,30 @@
...
+    'summary': "Create and sell products of type kit",
+    'description': """
+This bridge module adds the ability to create and sell products of type kit.
+    """,
+    'depends': [
+        'product',
```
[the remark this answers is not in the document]
iemu-odoo: Understood. I thought that we needed to include all those modules in depends whose models, views, etc. we inherited in our module. But that's incorrect.

sale_product_kit/models/product_template.py
iemu-odoo: Understood. Will use the inbuilt methods such as mapped, filtered, etc. as much as possible to avoid compromising the performance of code.

sale_product_kit/models/product_template.py
iemu-odoo: Sure. At the time of writing the code, I wanted to make the logic of removing sub-products reusable. But my mistake is that i didn't extract the logic in a business method and wrote it in this onchange method directly. I'll fix that.

sale_product_kit/models/sale_order_line.py
iemu-odoo: You're right. It would be much better if there was a field here to store all linked sub product order lines.

sale_renting_deposit/views/sale_renting_deposit_templates.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <template id="product_inherit_deposit" inherit_id="website_sale.product">
+        <xpath expr="//div[@id='o_wsale_cta_wrapper']" position="after">
+            <t t-if="product.rent_ok and product.is_deposit_required">
+                <div id="rental_deposit_amount_container" class="my-4">
+                    <strong class="fw-bold">Deposit Required:</strong>
+                    <t t-esc="product.currency_id.symbol"/>
```
ampa-odoo, line 8: Use t-out instead as it would give a deprecation warning.
```suggestion
                    <t t-out="product.currency_id.symbol"/>
```

ampa-odoo (changes requested):

PR #369, opened 2025-02-24
[ADD] event_ticket: add module for limiting event tickets per registr…

mkap-odoo:
…ation

In this commit:
 - inherit "event.event.ticket" model from event module in event_ticket.py file and add a integer field named "tickets_per_registration"
 - inherit list view of event.event.ticket from event module and add ticket field after the description field
 - inherit template containing modal for ticket registration from website_event module in event_ticket module
 - add new variable "seats_max_limit" and set its value by evaluating conditions in single and multi ticket registration form using xpath
 - if the "tickets_per_registration" field is set to 0 then the user will have no limit and can select the default upto 9 tickets
 - if the "tickets_per_registration" field is set to more than 0 then the user can register the minimum from the maximum tickets assigned to that event, tickets_per_registration and maximum tickets assigned to that ticket type

### DIFF
--- a/event_ticket/views/event_templates_page_registration.xml
+++ b/event_ticket/views/event_templates_page_registration.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <template id="event_ticket_registration" name="Modal for tickets registration" inherit_id="website_event.modal_ticket_registration">
+        <xpath expr="//t[@t-set='seats_max']" position="before">
+            <t t-set="seats_max_limit" t-value="(ticket.tickets_per_registration > 0 and ticket.tickets_per_registration + 1) or 10"/>
+        </xpath>
+
+        <xpath expr="//t[@t-set='seats_max']" position="attributes">
+            <attribute name="t-value">min(seats_max_ticket, seats_max_event , seats_max_limit) </attribute>
+        </xpath>
+
+        <xpath expr="//select[@class='d-inline w-auto form-select']//t[@t-set='seats_max']" position="before">
+            <t t-set="seats_max_limit" t-value="(tickets.tickets_per_registration > 0 and tickets.tickets_per_registration + 1) or 10"/>
+        </xpath>
+
+        <xpath expr="//select[@class='d-inline w-auto form-select']//t[@t-set='seats_max']" position="attributes">
+            <attribute name="t-value">min(seats_max_ticket, seats_max_event , seats_max_limit) if tickets else seats_max_event </attribute>
+        </xpath>
+    </template>
+</odoo>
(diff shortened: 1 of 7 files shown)


sjai-odoo:
We are improving current functionality so we should use [IMP] tag here.
- We avoid using technical names of model or fields in the commit message.
- Commit message should give an overview of what functionality changes not how(technically)
- In short commit message should be easy to understand even for non technical person.

event_ticket/views/event_ticket_views.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="view_event_event_ticket_tree_inherit_event_ticket" model="ir.ui.view">
+        <field name="name">event.event.ticket.list.inherit.event.ticket</field>
+        <field name="model">event.event.ticket</field>
+        <field name="inherit_id" ref="event.event_event_ticket_view_tree_from_event"/>
+        <field name="arch" type="xml">
+            <field name="description" position="after">
+                <field name="tickets_per_registration" width="180px"/>
```
sjai-odoo, line 9:
```suggestion
                <field name="tickets_per_registration"/>
```

event_ticket/views/event_templates_page_registration.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <template id="event_ticket_registration" name="Modal for tickets registration" inherit_id="website_event.modal_ticket_registration">
+        <xpath expr="//t[@t-set='seats_max']" position="before">
+            <t t-set="seats_max_limit" t-value="(ticket.tickets_per_registration > 0 and ticket.tickets_per_registration + 1) or 10"/>
+        </xpath>
+
```
sjai-odoo, line 7:
```suggestion
```
You have done the same below in the same file so why we need this and also accessing through ticket instead of tickets

event_ticket/views/event_templates_page_registration.xml
```diff
@@ -0,0 +1,20 @@
...
+        <xpath expr="//t[@t-set='seats_max']" position="attributes">
+            <attribute name="t-value">min(seats_max_ticket, seats_max_event , seats_max_limit) </attribute>
+        </xpath>
+
```
sjai-odoo, line 11: even this is also duplicate why have you added the same things twice with just different xpth expr?

sjai-odoo (changes requested):

event_ticket/views/event_templates_page_registration.xml
mkap-odoo: I have done this because the first two xpath are related with the multi ticket registrations (Standard, VIP, Free etc) and there is a foreach loop running for each multi ticket so I have used "ticket" there.
The next two xpath are related with single ticket registration so there we have to use "tickets" instead of ticket.
There is different block of code for both of these registration (single and multi) so we can't optimize it more and compulsorily set the value "seats_max" in both block of code using four xpath tags.
Also I have to add the functionality that if the ticket_per_registration field is set to 0, then the default dropdown the user will see will be upto 9 tickets and that I have to set for both single and multi registrations.

sjai-odoo:
LGTM,
you can close this PR.

PR #377, opened 2025-02-24
[IMP] add_pricelist_price:  Added Book Price

siha-odoo:
-Created Book Price named new module to display original pricelist price, 
-Added computed field book_price in sale_order_line and account move_line to calculate original price based on currently selected pricelist price, product
 and quantity,
-Displayed book price in sale order line view and account move line view.


book_price/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
rare-odoo, line 1: Please remove extra line

book_price/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "category": "Sales/Book Price",
+    "summary": "Added book price",
+    "description": """
+        Added book price in sales order line and account move line
+        so that customer can see difference between Pricelist price and other modified price.""",
+    "depends": ["account", "sale"],
+    "data": [
+        "views/sale_order_line_views.xml", 
+        "views/account_move_line_views.xml"
+        ],
```
rare-odoo, line 13:
```suggestion
    "data": [
        "views/sale_order_line_views.xml", 
        "views/account_move_line_views.xml"
     ],
```

book_price/models/account_move_line.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    book_price = fields.Float(
+        string="Book Price",
+        compute="_compute_book_price",
+        readonly=True,
+        store=True,
```
rare-odoo, line 11: Could you please clarify the reason for setting it to store true?

book_price/models/account_move_line.py
```diff
@@ -0,0 +1,33 @@
...
+    def _compute_book_price(self):
+        for record in self:
+            if record.move_id and record.move_id.invoice_line_ids:
+                product_pricelist = record.move_id.invoice_line_ids.mapped(
+                    "sale_line_ids.order_id.pricelist_id"
+                )
+
+                if (
+                    not product_pricelist
+                    or not record.product_id
```
rare-odoo, line 25: IMO, this should be checked first.
Additionally, please verify if it's possible 
to remove any unnecessary computations.

book_price/models/sale_order_line.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    book_price = fields.Float(
+        string="Book Price",
+        compute="_compute_book_price",
+        readonly=True,
+        store=True,
```
rare-odoo, line 11: same

rare-odoo (changes requested):
Hi @siha-odoo,
Great Work :clap: 
I have left some change requests in the review for your consideration.

book_price/__init__.py
siha-odoo: okay

book_price/models/account_move_line.py
siha-odoo: In order to store it in database and instead of recalculating it each time.

book_price/models/account_move_line.py
siha-odoo: Okay i will give it a look.

book_price/models/account_move_line.py
rare-odoo: In this case, you may opt out from storing it in the database

rare-odoo (changes requested):

rare-odoo:
Additionally, Please improve the Commit/PR message!

With this commit
=============
- The changes you have done...

book_price/models/sale_order_line.py
```diff
@@ -0,0 +1,30 @@
...
+        - Otherwise, compute the price using `_get_pricelist_price()` and multiply by quantity.
+        """
+        for record in self:
+            if not record.product_id:
+                record.book_price = 0.0
+                continue
+            if not record.pricelist_item_id:
+                record.book_price = record.product_id.lst_price * record.product_uom_qty
+                continue
+            record.book_price = record._get_pricelist_price() * record.product_uom_qty
```
rare-odoo, line 30: The current code structure can be optimized by replacing 
the multiple if conditions with an if, elif, else sequence. 
This would eliminate the need for the continue!

rare-odoo (changes requested):

book_price/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "Sale Book Price",
+    "version": "1.0",
+    "category": "Sales/Book Price",
+    "summary": "Added book price",
+    "description": """
+        Added book price in sales order line and account move line
+        so that customer can see difference between Pricelist price and other modified price.""",
+    "depends": ["account", "sale"],
```
rare-odoo, line 9: Is it necessary to include 'account' in the dependencies?

book_price/models/account_move_line.py
```diff
@@ -0,0 +1,33 @@
...
+        string="Book Price",
+        compute="_compute_book_price",
+        readonly=True,
+        help="A relevant pricelist price based on currently selected pricelist, product and quantity.",
+    )
+
+    @api.depends("move_id.invoice_line_ids", "quantity", "product_id")
+    def _compute_book_price(self):
+        """
+        Compute `book_price` based on the linked sale order's pricelist.
```
rare-odoo, line 17:
```suggestion
        """ Compute `book_price` based on the linked sale order's pricelist.
```

book_price/models/sale_order_line.py
```diff
@@ -0,0 +1,29 @@
...
+        string="Book Price",
+        compute="_compute_book_price",
+        readonly=True,
+        help="A relevant pricelist price based on currently selected pricelist, product and quantity.",
+    )
+
+    @api.depends("product_id", "pricelist_item_id", "product_uom_qty")
+    def _compute_book_price(self):
+        """
+        Compute `book_price` based on the selected pricelist item, product, and quantity.
```
rare-odoo, line 17:
```suggestion
        """ Compute `book_price` based on the selected pricelist item, product, and quantity.
```

rare-odoo (changes requested):
Hi @siha-odoo,

Just a few more changes, and then we’ll be good to go.

book_price/models/sale_order_line.py
siha-odoo: Got it. I have optimized the code.

book_price/__manifest__.py
siha-odoo: No. It is not necessary to add it because sale module has account_payment as it's dependencies so account module will be installed as sales modules will be installed.

book_price/models/account_move_line.py
siha-odoo: Corrected.

book_price/models/sale_order_line.py
siha-odoo: Corrected.

book_price/models/account_move_line.py
```diff
@@ -0,0 +1,31 @@
...
+        
+        for record in self:
+            if record.product_id and record.move_id and record.move_id.invoice_line_ids:
+                product_pricelist = record.move_id.invoice_line_ids.mapped("sale_line_ids.order_id.pricelist_id")
+                if product_pricelist:
+                    record.book_price = product_pricelist[0]._get_product_price(record.product_id, record.quantity, record.product_uom_id)* record.quantity
+                else:
+                    record.book_price = 0.0
+            else:
+                record.book_price = 0.0
```
rare-odoo, line 30: This seems redundant

rare-odoo (changes requested):
Hi @siha-odoo,

Just one small change, but aside from that, this PR LGTM!

book_price/models/account_move_line.py
```diff
@@ -0,0 +1,32 @@
...
+          - Compute the price using `_get_product_price()` and multiply by quantity.
+        - Otherwise, set `book_price` to 0.0."""
+        
+        for record in self:
+            if record.product_id and record.move_id and record.move_id.invoice_line_ids:
+                product_pricelist = record.move_id.invoice_line_ids.mapped("sale_line_ids.order_id.pricelist_id")
+                record.book_price = (
+                    product_pricelist[0]._get_product_price(record.product_id, record.quantity, record.product_uom_id) * record.quantity
+                    if product_pricelist
+                    else 0.0
```
rare-odoo, line 28: What makes you believe this approach is better than the previous one?
Additionally, I hadn't noticed earlier, Why is indexing used here?

rare-odoo (changes requested):

book_price/models/account_move_line.py
siha-odoo: i am checking the condition before setting book price and not after checking is there any pricelist available or not.
I was trying the use case there by indexing that how it will affect because as i have mapped there i thought is there any chance of mapping returning multiple pricelists or not. which will not be the case so yeah that's me forgot to correct that by not using indexing.

book_price/models/account_move_line.py
```diff
@@ -0,0 +1,27 @@
...
+        """Compute `book_price` based on the linked sale order's pricelist.
+        - If a valid `product_id`, `move_id`, and `invoice_line_ids` exist:
+          - Fetch the pricelist from the associated sale order.
+          - Compute the price using `_get_product_price()` and multiply by quantity.
+        - Otherwise, set `book_price` to 0.0."""
+        
+        for record in self:
+            record.book_price = 0.0
+            if record.product_id and record.move_id and record.move_id.invoice_line_ids:
+                product_pricelist = record.move_id.invoice_line_ids.mapped("sale_line_ids.order_id.pricelist_id")
```
rare-odoo, line 25: Since you are using `record.move_id` multiple times consider assigning it to a variable for reuse.
 Additionally, use `_read_group` to group lines by pricelist, avoiding loops for better efficiency.

rare-odoo (changes requested):

book_price/models/account_move_line.py
rare-odoo: Additionally, Could you please clarify why the data is being accessed through move_id.invoice_line_ids? 
Would it not be possible to directly retrieve it from record.sale.line_ids instead?

PR #379, opened 2025-02-24
[ADD] product_kit: added a new product type

ypat-odoo:
In this commit :
 - A new field added for product type kits and added many2many widget for adding subproducts.
 - added a button for adding sub products which a wizard form is there for adding products and that product will added in order line and but will invisible if quotation is confirmed

### DIFF
--- a/product_kit/__manifest__.py
+++ b/product_kit/__manifest__.py
@@ -1,0 +1,20 @@
+{
+    "name": "Product Type",
+    "version": "1.0",
+    "depends": ["sale_management","point_of_sale"],
+    "application": True,
+    "installable": True,
+    "data": [
+        "security/ir.model.access.csv",
+        "wizard/sub_products.xml",
+        "views/product_view.xml",
+        "views/sale_order_view.xml", 
+        "report/report_order_lines.xml",
+    ],
+    'assets': {
+        'point_of_sale._assets_pos': [
+            'product_kit/static/src/**/*',
+        ],
+    },
+    "license": "AGPL-3"
+}
--- a/product_kit/models/product_product.py
+++ b/product_kit/models/product_product.py
@@ -1,0 +1,11 @@
+from odoo import api,models,fields
+
+
+class ProductProduct(models.Model):
+    _inherit = ['product.product']
+
+    @api.model
+    def _load_pos_data_fields(self, config_id):
+        params = super()._load_pos_data_fields(config_id) 
+        params += ['is_kit','sub_product_ids']
+        return params
--- a/product_kit/models/product_template.py
+++ b/product_kit/models/product_template.py
@@ -1,0 +1,14 @@
+from odoo import api,models,fields
+
+
+class ProductTemplate(models.Model):
+    _inherit = ['product.template']
+
+    is_kit = fields.Boolean(default=False,string="Is Kit")   
+    sub_product_ids = fields.Many2many('product.product', relation='sub_products_rel', column1='base_id', column2='sub_id')
+
+    @api.onchange('is_kit')
+    def _onchange_is_kit(self):
+
+        if not self.is_kit:
+            self.sub_product_ids=[(6,0,[])]
--- a/product_kit/models/sale_order.py
+++ b/product_kit/models/sale_order.py
@@ -1,0 +1,7 @@
+from odoo import api,models,fields
+
+
+class SaleOrder(models.Model):
+    _inherit = ['sale.order']
+
+    print_in_report = fields.Boolean(default=False)
--- a/product_kit/models/sale_order_line.py
+++ b/product_kit/models/sale_order_line.py
@@ -1,0 +1,21 @@
+from odoo import api,models,fields,Command
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    parent_line_id = fields.Many2one(comodel_name='sale.order.line', ondelete="cascade")
+    is_kit = fields.Boolean(related='product_template_id.is_kit', store=True)
+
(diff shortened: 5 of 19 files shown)


product_kit/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "data": [
+        "security/ir.model.access.csv",
+        "wizard/sub_products.xml",
+        "views/product_view.xml",
+        "views/sale_order_view.xml", 
+        "report/report_order_lines.xml",
+    ],
+    "demo": [
+    ],
+    'assets': {
```
sjai-odoo, line 16:
```suggestion
    'assets': {
```

product_kit/models/product_product.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = ['product.product']
+
+    @api.model
+    def _load_pos_data_fields(self, config_id):
+        params = super()._load_pos_data_fields(config_id) 
+        params += ['is_kit','sub_product_ids']
+        return params
```
sjai-odoo, line 11: TRy after moving this in product_template.py

product_kit/models/product_template.py
```diff
@@ -0,0 +1,16 @@
...
+    sub_product_ids = fields.Many2many('product.product', relation='sub_products_rel', column1='base_id', column2='sub_id')
+
+    @api.onchange('is_kit')
+    def _onchange_is_kit(self):
+        if self.is_kit:
+            pass
+        if not self.is_kit:
+            self.sub_product_ids=[(6,0,[])]
+            pass
+        return {}
```
sjai-odoo, line 16:
```suggestion
        if not self.is_kit:
            self.sub_product_ids=[(6,0,[])]
```

product_kit/models/sale_order.py
```diff
@@ -0,0 +1,6 @@
+from odoo import api,models,fields
+
+
+class SaleOrder(models.Model):
+    _inherit = ['sale.order']
+    print_in_report = fields.Boolean(default=False)
```
sjai-odoo, line 6:
```suggestion
    _inherit = ['sale.order']

    print_in_report = fields.Boolean(default=False)
```

product_kit/models/sale_order_line.py
```diff
@@ -0,0 +1,23 @@
+from odoo import api,models,fields,Command
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+    parent_line_id = fields.Many2one(comodel_name='sale.order.line', ondelete="cascade")
```
sjai-odoo, line 6:
```suggestion
    _inherit = 'sale.order.line'

    parent_line_id = fields.Many2one(comodel_name='sale.order.line', ondelete="cascade")
```

product_kit/models/sale_order_line.py
```diff
@@ -0,0 +1,23 @@
...
+    is_kit = fields.Boolean(related='product_template_id.is_kit', store=True)
+
+    def unlink(self):
+        for line in self:
+            if line.product_id.is_kit:
+                linked_lines = self.env['sale.order.line'].filtered(lambda line: line.parent_line_id.id == self.parent_line_id.id)
```
sjai-odoo, line 12: Insteadof filtering in `self.env['sale.order.line'].` filter only on lines of that orderline's order.
This may takemuch more time when we have thousands of orders and may be lakhs of sols.

product_kit/static/src/models/pos_order.js
```diff
@@ -0,0 +1,22 @@
+import { patch } from '@web/core/utils/patch';
+import { PosOrder } from '@point_of_sale/app/models/pos_order';
+
+patch(PosOrder.prototype, {
+    setup(vals) {
+        super.setup(vals);
+    },
```
sjai-odoo, line 7:
```suggestion
```
not added value here

product_kit/static/src/models/pos_order.js
```diff
@@ -0,0 +1,22 @@
...
+        let lines = [];
+
+        for (let i = 0; i < sub_products.length; i++) {
+            let matchingLines = this.lines.filter((x) => x.product_id.id == sub_products[i].id);
+            lines.push(...matchingLines);
+        }
+
+        for (let line of lines) {
+            super.removeOrderline(line)
+        }
```
sjai-odoo, line 19: TRy to filter and unlink by filtering in not just a single subproduct something like in any of these ids, in a way you don't need to iterate all lines subproduct's length time, just one time

product_kit/static/src/product_screen/product_kit_screen.js
```diff
@@ -0,0 +1,32 @@
...
+import { ConfirmationDialog } from "@web/core/confirmation_dialog/confirmation_dialog";
+import { usePos } from "@point_of_sale/app/store/pos_hook";
+import { ProductScreen } from '@point_of_sale/app/screens/product_screen/product_screen';
+
+patch(ProductScreen.prototype, {
+    setup() {
+        super.setup()
+        this.pos = usePos();
+        this.dialogService = useService("dialog");
+    },
```
sjai-odoo, line 13:
```suggestion
```
all these are unnecessary already there in parent component

sjai-odoo (changes requested):

product_kit/models/product_product.py
ypat-odoo: we have to keep in this model because before the 18.0 version we cannot add from product template where as if we load in product_product then we can access in pos

sjai-odoo:
LGTM
Close your branch.

PR #383, opened 2025-02-25
[ADD] supplier_portal: allow supplier invoice uploads and draft bills

mkap-odoo:
In this commit:
 - Introduce a supplier portal where suppliers can log in and submit their invoices.
 - Add a form for suppliers to select their organization and upload PDF/XML invoice files.
 - Implement backend logic to create a draft bill upon submission, associating the uploaded files.
 - Ensure supplier details are pre-filled in the bill while products remain manually configurable.
 - Improve supplier experience by streamlining invoice submission and reducing manual data entry.

Task Id: 4598783

### DIFF
--- a/supplier_portal/controllers/supplier_portal_website.py
+++ b/supplier_portal/controllers/supplier_portal_website.py
@@ -1,0 +1,47 @@
+from datetime import datetime
+import base64
+
+from odoo import http
+from odoo.http import request
+
+
+class SupplierPortal(http.Controller):
+    @http.route(["/supplier/upload"], auth="user", type="http", website=True)
+    def supplier_login(self):
+        companies = request.env.user.company_ids
+        return request.render("supplier_portal.supplier_upload_page", {"companies": companies})
+
+    @http.route(["/supplier/submit"], auth="user", type="http", methods=["POST"], website=True)
+    def supplier_invoice_submit(self, **post):
+        try:
+            supplier = request.env.user.partner_id
+            company_id = int(post.get("company"))
+            pdf_invoice = post.get("pdf_invoice")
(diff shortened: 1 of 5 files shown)


supplier_portal/controllers/supplier_portal_website.py
```diff
@@ -0,0 +1,59 @@
...
+        try:
+            supplier = request.env.user.partner_id
+            company_id = int(post.get("company"))
+            pdf_invoice = post.get("pdf_invoice")
+            xml_invoice = post.get("xml_invoice")
+
+            if not pdf_invoice or not xml_invoice:
+                return request.redirect("/supplier/upload?error=missing_files")
+
+            # Create Vendor Bill
```
sjai-odoo, line 26:
```suggestion

            # Create Vendor Bill
```
they are set as required one so we will be able to submit only if we have those fields

supplier_portal/controllers/supplier_portal_website.py
```diff
@@ -0,0 +1,59 @@
...
+                            "res_id": bill.id,
+                            "type": "binary",
+                            "datas": base64.b64encode(file.read()),
+                            "mimetype": file.content_type,
+                        }
+                    )
+                )
+
+            attach_file(pdf_invoice, pdf_invoice.filename)
+            attach_file(xml_invoice, xml_invoice.filename)
```
sjai-odoo, line 54: Instead of creating a different function you can put a simple loop where  you can create and attach them, in that way there will be no need of calling attach_file externally.

sjai-odoo (changes requested):

sjai-odoo:
LGTM,
you can close this PR.

PR #385, opened 2025-02-25
18.0-custom-report-drat

drat-odoo:
[ADD] new_mrp: Implement new.mrp module

This commit includes:
- Added mo_delivery_note report and views (similar to Delivery Slip report but excludes Kit Products).

Task name: Custom Picking report  
Task link:  
https://link.excalidraw.com/readonly/caLMnqPQHAnC8yMdfgVW

### DIFF
--- a/new_mrp/__manifest__.py
+++ b/new_mrp/__manifest__.py
@@ -1,0 +1,11 @@
+{
+    'name': 'New Mrp',
+    'version': '1.0',
+    'description': 'Custom MO Delivery Note report without Kit Products(in Inventory)',
+    'depends': ['mrp', 'sale_management'],
+    'data': [
+        'report/report_mo_delivery_note_views.xml',
+        'report/report_mo_delivery_note.xml',
+    ],
+    'license': 'LGPL-3',
+}
--- a/new_mrp/report/report_mo_delivery_note_views.xml
+++ b/new_mrp/report/report_mo_delivery_note_views.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <template id="report_delivery_document" inherit_id="stock.report_delivery_document" primary="True">
+        <xpath expr="//table[@name='stock_move_table']/tbody/t" position="replace">
+            <t t-set="lines" t-value="o.move_ids.filtered(lambda x: x.product_uom_qty and (not x.bom_line_id or x.bom_line_id.bom_id.type != 'phantom'))" />
+        </xpath>
+    </template>
+    <template id="report_mo_delivery_note">
+        <t t-foreach="docs" t-as="o">
+            <t t-call="new_mrp.report_delivery_document" t-lang="o._get_report_lang()"/>
+        </t>
+    </template>
+</odoo>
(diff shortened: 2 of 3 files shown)


new_mrp/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+{
+    'name': 'New Mrp',
+    'version': '1.0',
+    'description': 'Custom MO Delivery Note report without Kit Products(in Inventory)',
+    'depends': ['mrp', 'stock' , 'sale_management'],
```
sjai-odoo, line 8:
```suggestion
    'depends': ['mrp', 'sale_management'],
```
stock is already installed if mrp is installed as stock is in depends of mrp module.

new_mrp/report/report_mo_delivery_note_views.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <template id="report_mo_delivery_note">
+        <t t-foreach="docs" t-as="o">
+            <t t-call="new_mrp.report_delivery_document" t-lang="o._get_report_lang()"/>
+        </t>
+    </template>
+    <template id="report_delivery_document" inherit_id="stock.report_delivery_document" primary="True">
```
sjai-odoo, line 8: why is this primary=True needed?

sjai-odoo, line 8:
```suggestion
    <template id="report_delivery_document" inherit_id="stock.report_delivery_document" primary="True">
        <t t-foreach="docs" t-as="o">
```
try this way to avoid 2 templates calling one from another.

new_mrp/report/report_mo_delivery_note_views.xml
```diff
@@ -0,0 +1,46 @@
...
+                    </tr>
+                </t>
+                <t t-else="">
+                    <tr>
+                        <td colspan="3">
+                            <h2 style="margin: 10px; color: red;">There is no such product without Kit</h2>
+                        </td>
+                    </tr>
+                </t>
+            </tbody>
```
sjai-odoo, line 43:
```suggestion
 <xpth ....
```
I have noticed you just replaced entire tbody but just changed one condition , this not a good ractice at all to directly replace entire code and write it again just change as less as possible.
xpath only for the needed things. REmove all these.

sjai-odoo (changes requested):

new_mrp/report/report_mo_delivery_note_views.xml
drat-odoo: okay sir!!

drat-odoo: Because i want copy of this template, If i not use primary=True it will change and modify specific parts of the original template, so that's why i use primary=true (Full Replacement Inheritance)

new_mrp/__manifest__.py
drat-odoo: okay noted!!

sjai-odoo:
LGTM,
You can now close this PR.

PR #386, opened 2025-02-25
[ADD] sale_renting_deposit: add Deposit Product feature to rental module

hash-odoo:
- Added an option of "Deposit Product" in product template options to mark products as deposit products.
- Added configuration to set a default deposit product for sales orders.
- Introduced new fields: Require Deposit and Deposit Amount in Product Template to set deposits for rental products.
- When a rental product is added to a sales quotation, the corresponding deposit product is automatically added.
- Displayed Deposit Price on the product page of `website_sale.product` and in the cart.

Task ID: 4599914


reka-odoo (changes requested):
Hello
Some changes

Deposit quantity is not working properly. can you check?

PR #398, opened 2025-02-26
[ADD] product_pricelist: Book price added in Sale Order Line

ppch-odoo:
- Models and views are inherited of sale order line and account move line
- Defined _compute_book_price function which will be depends on product_id,
  quantity and pricelist whenever mentioned field will change on that time 
  based on condition book_price will be calculated
- Book price is only visible in customer invoices.
- In account_move_line set book_price as related field
- Instead of displayed book_price for all quantity changed to for only 1 quantity

task - 4602859

### DIFF
--- a/product_pricelist/__manifest__.py
+++ b/product_pricelist/__manifest__.py
@@ -1,7 +1,7 @@
 {
     'name': "Book Price (Pricelist Price)",
     'version': '1.0',
-    'depends': ['base', 'sale_management'],
+    'depends': ['sale_management'],
     'author': "ppch",
     'category': 'Category',
     'description': """
--- a/product_pricelist/__manifest__.py
+++ b/product_pricelist/__manifest__.py
@@ -1,11 +1,12 @@
 {
-    'name': "Book Price Pricelist",
+    'name': "Book Price (Pricelist Price)",
     'version': '1.0',
-    'depends': ['base', 'sale_management', 'account'],
+    'depends': ['base', 'sale_management'],
     'author': "ppch",
     'category': 'Category',
     'description': """
-    Book Price on Sales order line and accunt move line
+    Book Price (Pricelist Price) on sales order lines and invoice lines
+    which will be used to compare between Book Price (Pricelist Price) and manually adjusted price on the lines
     """,
     'license': "LGPL-3",
     'data': [
--- a/product_pricelist/__manifest__.py
+++ b/product_pricelist/__manifest__.py
@@ -1,0 +1,16 @@
+{
+    'name': "Book Price Pricelist",
+    'version': '1.0',
+    'depends': ['base', 'sale_management', 'account'],
+    'author': "ppch",
+    'category': 'Category',
+    'description': """
+    Book Price on Sales order line and accunt move line
+    """,
+    'license': "LGPL-3",
+    'data': [
+        'views/sale_order_line_views.xml',
+        'views/account_move_line_views.xml',
+    ],
+    'installable': True,
+}
--- a/product_pricelist/__manifest__.py
+++ b/product_pricelist/__manifest__.py
@@ -6,7 +6,8 @@
     'category': 'Category',
     'description': """
     Book Price (Pricelist Price) on sales order lines and invoice lines
-    which will be used to compare between Book Price (Pricelist Price) and manually adjusted price on the lines
+    which will be used to compare between Book Price (Pricelist Price) and
+    manually adjusted price on the lines.
     """,
     'license': "LGPL-3",
     'data': [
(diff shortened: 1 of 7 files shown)


product_pricelist/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+{
+    'name': "Book Price (Pricelist Price)",
+    'version': '1.0',
+    'depends': ['base', 'sale_management'],
+    'author': "ppch",
+    'category': 'Category',
+    'description': """
+    Book Price (Pricelist Price) on sales order lines and invoice lines
+    which will be used to compare between Book Price (Pricelist Price) and manually adjusted price on the lines
+    """,
```
bhra-odoo, line 10: make the format proper.

product_pricelist/models/account_move_line.py
```diff
@@ -0,0 +1,22 @@
+from odoo import api, fields, models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    book_price = fields.Float(string="Book Price", compute="_compute_book_price", store=True)
```
bhra-odoo, line 7:
```suggestion
    book_price = fields.Float(string="Book Price", compute='_compute_book_price', store=True)
```
why store=True

product_pricelist/views/account_move_line_views.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="account_move_form_inherit_book_price" model="ir.ui.view">
+        <field name="name">account.move.form.inherite.book.price</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//notebook//page//field[@name='invoice_line_ids']//list//field[@name='quantity']" position="before">
+                <field name="book_price" readonly="True"/>
```
bhra-odoo, line 9: book price is only visible for customer invoices but it is visible for vendor bills also
image

product_pricelist/models/sale_order_line.py
```diff
@@ -0,0 +1,22 @@
+from odoo import api, fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    book_price = fields.Float(string="Book Price", compute="_compute_book_price", store=True)
```
bhra-odoo, line 7: same here

bhra-odoo (changes requested):

product_pricelist/views/account_move_line_views.xml
ppch-odoo: corrected

product_pricelist/models/account_move_line.py
ppch-odoo: store=True not needed bcz we are only displaying it

product_pricelist/models/sale_order_line.py
```diff
@@ -0,0 +1,22 @@
...
+        for record in self:
+            if not record.product_id:
+                record.book_price = 0.0
+                continue
+
+            pricelist = record.order_id.pricelist_id if record.order_id else None
+            if pricelist:
+                record.book_price = pricelist._get_product_price(
+                    record.product_id, record.product_uom_qty
+                ) * record.product_uom_qty
```
bhra-odoo, line 20: why multiply again with uom_qty?

product_pricelist/models/account_move_line.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    book_price = fields.Float(string="Book Price", compute='_compute_book_price')
+
+    @api.depends('product_id', 'quantity', 'move_id.invoice_origin')
+    def _compute_book_price(self):
+        sale_orders = {so.name: so for so in self.env['sale.order'].search([('name', 'in', self.mapped('move_id.invoice_origin'))])}
```
bhra-odoo, line 11: instead of sale order you can get price;ist

bhra-odoo (changes requested):

product_pricelist/models/sale_order_line.py
ppch-odoo: updated
now it will display only for 1 quantity

product_pricelist/models/account_move_line.py
ppch-odoo: updated
instead of compute field, book_price is set as related field so no any compute method

bhra-odoo (approved):
LGTM.
good work!!

PR #411, opened 2025-02-26
[ADD] webshop_extension: added new multilingual html extended description field

vpan-odoo:
### Steps to reproduce

1. For Internal Users

- Install webshop_extension module (depends on website_sale)
- Open Website > eCommerce > Products
- From Kanban View select any one product it will open the form view of that product 
- In the form view, visit the Sales page under there is one group Extended Description
- In that field you can add description as long as you want
- To enable multi language support User > Preferences > Languages > click on globe icon then add languages you want
- After adding languages, you are able to see EN on clicking it, it will open a pop-up to add description in particular language

**For Portal User it will conditionally render, if the particular product has extended description it will display else it will not display**

 ### Purpose of Changes

- Our webshop contains many products, including technical ones that require extensive descriptions.
- The existing description fields (description_sale and description_ecommerce) are not suitable for long descriptions.

### Changes

- To address this, we are introducing a new extended_description field that:

1. Displays on the product page (for portal users) without affecting the layout.
2. It is editable from the backend (internal users view) for better content management.
3. Supports multiple languages (Danish and Swedish) 
4. This field is also included in import/export functionality

Task-4605780

### DIFF
--- a/webshop_extension/models/product_template.py
+++ b/webshop_extension/models/product_template.py
@@ -1,0 +1,7 @@
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    ecommerce_extended_description = fields.Html(string="Ecommerce Extended Description", translate=True)
--- a/webshop_extension/views/product_views.xml
+++ b/webshop_extension/views/product_views.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="product_template_form_view" model="ir.ui.view">
+        <field name="name">product.template.form.view.inherit</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//page[@name='sales']//group[@name='sale']" position="after">
+                <group name="ecommerce_extended_description" string="Ecommerce Extended Description">
+                    <field name="ecommerce_extended_description" nolabel="1"/>
+                </group>
+            </xpath>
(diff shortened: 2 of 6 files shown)


webshop_extension/models/product_template.py
```diff
@@ -0,0 +1,7 @@
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    extended_description = fields.Html(string="Extended Description", translate=True)
```
suth-odoo, line 7: Try to give proper field name, as we already have `Quotation Description` and `Ecommserce Description` and it is a bit confusing which extended description ...Give something like `ecommerce_extended_description`.

webshop_extension/views/product_views.xml
```diff
@@ -0,0 +1,14 @@
...
+        <field name="inherit_id" ref="product.product_template_form_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//page[@name='sales']//group[@name='sale']" position="after">
+                <group name="extended_description" string="Extended Description">
+                    <field name="extended_description"/>
+                </group>
+            </xpath>
+        </field>
+    </record>
+</odoo>
```
suth-odoo, line 14: add xml declarations...refer codebase for more info!!

suth-odoo (changes requested):
Hello @vpan-odoo 
I have added some suggestions..Have a look into it whenever you have time
Thank you :innocent:

webshop_extension/models/product_template.py
vpan-odoo: field name changed to `ecommerce_extended_description`

PR #412, opened 2025-02-26
[ADD] book_price: Display 'Book Price' on Sales Order and Invoice Lines

praj-odoo:
- Created a new module 'book_price' to display the original pricelist price ('Book Price').
- Added a computed field to Sales Order Lines and Account Move Lines to fetch the relevant pricelist price based on the selected pricelist, product, and quantity.
- Updated the Invoice Lines view to display the 'Book Price' field.
- Updated the Sales Order Lines view to include the 'Book Price' field for price comparison.


book_price/models/sale_order_line.py
```diff
@@ -0,0 +1,19 @@
...
+    _inherit = 'sale.order.line'
+
+    book_price = fields.Float(string="Book Price", compute='_compute_book_price', store=True)
+
+    @api.depends('product_id', 'product_uom_qty', 'order_id.pricelist_id')
+    def _compute_book_price(self):
+        for line in self:
+            if line.product_id and line.order_id.pricelist_id:
+                price, _ = line.order_id.pricelist_id._get_product_price_rule(
+                    line.product_id, line.product_uom_qty
```
ampa-odoo, line 15: Here, `_get_product_price_rule` accepts some options param as well which are currency, uom, date which can be used for computation of advance pricelist rules like set particular price between given date period.

book_price/views/account_move_line.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <record id="acount_move_form_inherit" model="ir.ui.view">
+        <field name="name">account.move.form.inherit</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref    ="account.view_move_form"/>
```
ampa-odoo, line 6:
```suggestion
        <field name="inherit_id" ref="account.view_move_form"/>
```

book_price/views/account_move_line.xml
```diff
@@ -0,0 +1,13 @@
...
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='invoice_line_ids']/list/field[@name='quantity']" position="before">
+                <field name="book_price" readonly="1" />
```
ampa-odoo, line 9: image
Check specs properly.

book_price/views/sale_order_line.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <record id="view_order_form_inherit" model="ir.ui.view">
+        <field name="name">sale.order.form.inherit</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='order_line']/list/field[@name='price_unit']"
+                position="before">
```
ampa-odoo, line 9:
```suggestion
            <xpath expr="//field[@name='order_line']/list/field[@name='price_unit']" position="before">
```

book_price/models/account_move_line.py
```diff
@@ -0,0 +1,24 @@
...
+            line.book_price = 0.0 
+            if not (line.product_id and line.move_id.invoice_origin):
+                continue
+
+            sale_order = self.env['sale.order'].search([('name', "=", line.move_id.invoice_origin)], limit=1)
+            if sale_order:
+                sale_order_line = self.env['sale.order.line'].search([
+                    ('order_id', "=", sale_order.id),
+                    ('product_id', "=", line.product_id.id)
+                ], limit=1)
```
ampa-odoo, line 22:
```suggestion
            line.book_price = 0.0
            sale_order_line = line.sale_line_ids.filtered(lambda sol: sol.product_id == line.product_id)[:1]
```
We already have a relation between `account.move.line` and `sale.order.line`. Maybe, can we just directly use that? What do you say? :thinking: 
In both approaches, if there is same product in a sale order but in 2 or more different SOL, what would be the scenario?

book_price/models/account_move_line.py
```diff
@@ -0,0 +1,24 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import api, fields, models
+
+class AccountMoveLine(models.Model):
+    _inherit = 'account.move.line'
+
+    book_price = fields.Float(string='Book Price', compute='_compute_book_price', store=True)
+
+    @api.depends('product_id', 'quantity', 'move_id.invoice_origin')
```
ampa-odoo, line 10:
```suggestion
    @api.depends('product_id', 'quantity', 'sale_line_ids')
```

book_price/models/account_move_line.py
```diff
@@ -0,0 +1,24 @@
...
+    def _compute_book_price(self):
+        for line in self:
```
ampa-odoo, line 12: Here, you can iterate the loop only for customer invoice lines i.e `move_type == 'out_invoice'` and  invoice state is`state == 'draft'`

ampa-odoo (changes requested):

ampa-odoo:
Add the task ID in commit message and PR description.

book_price/views/account_move_line.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <record id="acount_move_form_inherit" model="ir.ui.view">
+        <field name="name">account.move.form.inherit</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form"/>
+        <field name="arch" type="xml">
+             <xpath expr="//field[@name='product_id']" position="after">
+                <field name="book_price" column_invisible="parent.move_type in ['in_invoice', 'in_refund', 'in_receipt']" />
```
ampa-odoo, line 9: What about other move_type? It would be better to give a not condition rather than giving multiple values in `in` condition.

book_price/models/account_move_line.py
ampa-odoo: Why have you resolved this without any implementation? :thinking:

book_price/views/sale_order_line.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <record id="view_order_form_inherit" model="ir.ui.view">
+        <field name="name">sale.order.form.inherit</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='order_line']/list/field[@name='price_unit']"
+        position="before">
```
ampa-odoo, line 9:
```suggestion
            <xpath expr="//field[@name='order_line']/list/field[@name='price_unit']" position="before">
```

book_price/models/account_move_line.py
praj-odoo: -Yes, we already have a relation between account.move.line and sale.order.line through sale_line_ids, so directly using it is the better approach instead of manually searching for sale.order and sale.order.line.
-In the current search() approach, if a sale order has multiple lines for the same product, only the first matching SOL is considered (limit=1), which may not always be correct.
By filtering sale_line_ids, we ensure we are selecting the correct sale order line based on product_id, quantity, and product_uom.

ampa-odoo: It would be better if we pass this field value in `_prepare_invoice_line` method instead of finding the related sale.order.line.

book_price/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+{
+    'name': "Book Price",
+    'category': 'Sales/Sales',
+    'description':  """
+This is the add-on module that enables 'Book Price' field into order lines of Sales Orders and Invoicing to let the client know of actual price and adjusted price.
+""",
+    'depends':['sale_management', 'account'],
```
ampa-odoo, line 9: The account module is in the hierarchical depends of sale_management. Do we need to add it again? :thinking:

ampa-odoo (changes requested):

book_price/__manifest__.py
praj-odoo: No sir , if the account module is already included in the hierarchical dependencies of sale_management, then there is no need to explicitly add it again.

book_price/views/account_move_line.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <record id="acount_move_form_inherit" model="ir.ui.view">
+        <field name="name">account.move.form.inherit</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form"/>
+        <field name="arch" type="xml">
+             <xpath expr="//field[@name='product_id']" position="after">
+                <field name="book_price" column_invisible="parent.move_type not in ['out_invoice']" />
```
ampa-odoo, line 9:
```suggestion
                <field name="book_price" column_invisible="parent.move_type != 'out_invoice'" />
```
Initially, you were adding multiple values to check that is why maybe we were using `in` operator.

ampa-odoo (changes requested):

book_price/views/account_move_line.xml
praj-odoo: Yes sir, earlier we were checking multiple values, which is why i used the in operator. Now, we only need to check a single specific value ('out_invoice'), so using != is more straightforward and efficient.

odoo_self_order_details/static/src/product_detail_screen/product_detail_screen.xml
```diff
@@ -0,0 +1,29 @@
...
+<templates id="template" xml:space="preserve">
+    <t t-name="odoo_self_order_details.ProductDetailScreen" t-inherit="pos_self_order.ProductInfoPopup" t-inherit-mode="primary">
+        <xpath expr="//div[@class='modal-content rounded']" position="replace">
+            <div class="modal-content rounded">
+                <button type="button" class="btn-close position-absolute top-0 end-0 m-2" 
+                    t-on-click.stop="() => this.props.close()"></button>
+                <div class="product-image-container text-center p-3">
+                    <img t-if="props.product.image_128"
+                        class="o_self_order_item_card_image img-fluid w-75 rounded"
+                        t-attf-src="/web/image/product.product/{{ props.product.id }}/image_512?unique={{props.product.write_date}}"
```
ampa-odoo, line 11:
```suggestion
                        t-attf-src="/web/image/product.product/{{ props.product.id }}/image_512?unique={{ props.product.write_date }}"
```

ampa-odoo:
Change the module name `odoo_self_order_details` :-1:

supplier_portal/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+{
+    'name': "Supplier Portal",
+    'description':  """
+Allows suppliers to upload invoices (PDF & XML) and create draft vendor bills.
+""",
+    'depends':['website'],
```
ampa-odoo, line 8:
```suggestion
    'depends': ['website'],
```

supplier_portal/controller/supplier_portal.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @http.route('/supplier_portal/submit', type='http', auth='user', website=True, methods=['POST'])
+    def supplier_portal_submit(self, **kwargs):
+        company_id = int(kwargs.get('company'))
+        pdf_file = kwargs.get('pdf_file')
+        xml_file = kwargs.get('xml_file')
+        user = request.env.user
+
+        if user.company_id.id != company_id:
+            user.company_id = company_id
```
ampa-odoo, line 22: Can you explain me the need of doing this? :thinking:

supplier_portal/controller/supplier_portal.py
```diff
@@ -0,0 +1,55 @@
...
+                    'res_id': bill.id,
+                    'res_model': 'account.move',
+                    'name': file_name,
+                    'datas': base64.b64encode(file_content),
+                    'type': 'binary',
+                    'mimetype': mime_type,
+                })
+
+        if attachments:
+            request.env['ir.attachment'].sudo().create(attachments)
```
ampa-odoo, line 51: Can we use `self` over `request`?

supplier_portal/views/supplier_portal_templates.xml
```diff
@@ -0,0 +1,56 @@
...
+        <t t-call="website.layout">
+            <div class="d-flex justify-content-center">
+                <div class="">
+                    <h2>Message</h2>
+                    <p t-out="message"></p>
+                </div>
+            </div>
+            <div class="text-center mt-4">
+                <a href="/my" class="btn btn-primary">
+                        <i class="fa fa-home"></i> Back to Dashboard
```
ampa-odoo, line 51:
```suggestion
                    <i class="fa fa-home"></i> Back to Dashboard
```
Fix indentation accordingly!

ampa-odoo (changes requested):

supplier_portal/controller/supplier_portal.py
praj-odoo: This code is used for in a multi-company scenario where a user is assigned to both Company A and Company B. If the user is currently logged in to Company A but wants to upload files for Company B (e.g., to create an invoice in Company B) 
 If the user's company ID (user.company_id.id) is different from the company ID coming from the supplier portal (company_id), it temporarily sets the user's current company to the one received from the supplier portal. This ensures that the operations are performed within the correct company

supplier_portal/controller/supplier_portal.py
praj-odoo: I used request.env in my code because it's inside a controller that handles web requests. In controllers, we don't have access to self like we do in models. request.env allows me to interact with the database and perform operations like creating attachments.

ampa-odoo: FYI: https://github.com/odoo/odoo/pull/174062/files

praj-odoo:
> Change the module name `odoo_self_order_details` 👎

The task specifications mention that the module name should be odoo_self_order_details.

PR #417, opened 2025-02-26
[IMP] Sale : Add Zero Stock Blockage Field

soza-odoo:
[IMP] sale: zero stock blockage boolean filed added

In this commit
1.  sale order can be confirm if zero stock blockage field is set
2.  this field is readonly for all groups except administrator
3.  add one filed to handle this condition

task-4605851

### DIFF
--- a/sale_zero_stock_blockage/models/sale_order.py
+++ b/sale_zero_stock_blockage/models/sale_order.py
@@ -1,0 +1,19 @@
+from odoo import api, exceptions, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    zero_stock_approval = fields.Boolean(string="Zero Stock Approval")
+    group_check = fields.Boolean(compute='_group_check')
+
+    def _group_check(self):
+        if self.env.user.has_group('sales_team.group_sale_manager'):
+            self.group_check = True
+        else:
+            self.group_check = False
+
+    def action_confirm(self):
+        if not self.zero_stock_approval:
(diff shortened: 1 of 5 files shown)


sale_zero_stock_blockage/models/sale_order.py
```diff
@@ -0,0 +1,19 @@
+from odoo import api, exceptions, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    zero_stock_approval = fields.Boolean(string="Zero Stock Approval")
+    group_check = fields.Boolean(compute='_group_check')
```
bhra-odoo, line 8:
```suggestion
    is_sales_manager = fields.Boolean(compute='_compute_is_sales_manager')
```

sale_zero_stock_blockage/models/sale_order.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def _group_check(self):
```
bhra-odoo, line 10:
```suggestion
    def _compute_is_sales_manager(self):
```

sale_zero_stock_blockage/models/sale_order.py
```diff
@@ -0,0 +1,19 @@
...
+        if self.env.user.has_group('sales_team.group_sale_manager'):
+            self.group_check = True
+        else:
+            self.group_check = False
```
bhra-odoo, line 14: why not for loop?
this if - else really needed?

bhra-odoo (changes requested):

bhra-odoo (approved):
LGTM. now you can close the pr.

PR #421, opened 2025-02-26
[ADD] pos_product_uom: enable product sales in secondary unit of meas…

rsbh-odoo:
- Added `second_uom_ids` field in `product.template` to manage secondary UOMs.
- Extended POS to allow selecting and converting quantities using secondary UOM.
- Enhanced POS UI to support seamless unit conversions during sales.


pos_product_uom/models/product_template.py
```diff
@@ -0,0 +1,19 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    first_uom_category_id = fields.Many2one(comodel_name='uom.category', related='uom_id.category_id')
+    second_uom_id = fields.Many2one(comodel_name='uom.uom', string="Second Unit Of Mesure", domain="[('category_id', '=', first_uom_category_id), ('id', '!=', uom_id)]", help="List of secondary UOMs that can be used for this product")
```
ampa-odoo, line 10: Some NIT picking first :wink: 
```suggestion
    second_uom_id = fields.Many2one(comodel_name='uom.uom', string="Second Unit Of Measure", domain="[('category_id', '=', first_uom_category_id), ('id', '!=', uom_id)]", help="List of secondary UOMs that can be used for this product")
```

pos_product_uom/static/src/component/control_buttons/control_buttons.js
```diff
@@ -0,0 +1,54 @@
...
+            ? productPrice / baseUom.factor_inv
+            : productPrice * factorInv;
+        selectedOrder.set_unit(secondUomRecord);
+        selectedOrder.product_id.is_use_second_uom = true;
+    },
+
+    onClick() {
+        const selectedOrder = this.pos.get_order()?.get_selected_orderline();
+        if (!selectedOrder) {
+            this.notification.add("No product? No problem! Oh wait... actually, it is a problem. 😅", {
```
ampa-odoo, line 38: The strings should be auto-translated everywhere.

pos_product_uom/static/src/component/quantity_dialog/quantity_dialog.js
```diff
@@ -0,0 +1,29 @@
...
+export class QuantityDialog extends Component {
+    static template = 'pos_product_uom.QuantityDialog';
+    static components = { Dialog };
+
+    setup() {
+        this.notification = useService('notification');
+    }
+
+    confirmAction() {
+        if (this.props.quantity == ''){
```
ampa-odoo, line 16:
```suggestion
        if (this.props.quantity == '') {
```

pos_product_uom/static/src/component/quantity_dialog/quantity_dialog.js
```diff
@@ -0,0 +1,29 @@
...
+           this.notification.add("Oops! Numbers first, please! Let's try that again. 😊", {
+            type: 'info'
```
ampa-odoo, line 18: Fix indentation.

ampa-odoo (changes requested):

pos_product_uom/models/product_template.py
```diff
@@ -0,0 +1,19 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    first_uom_category_id = fields.Many2one(comodel_name='uom.category', related='uom_id.category_id')
```
ampa-odoo, line 9: we have a field `uom_category_id` in relation with uom.category. Do we have a need of adding a new field? :thinking:

pos_product_uom/models/product_template.py
```diff
@@ -0,0 +1,19 @@
...
+    second_uom_id = fields.Many2one(comodel_name='uom.uom', string="Second Unit Of Mesure", domain="[('category_id', '=', first_uom_category_id), ('id', '!=', uom_id)]", help="List of secondary UOMs that can be used for this product")
+    is_use_second_uom = fields.Boolean(default=False)
```
ampa-odoo, line 11:
```suggestion
    is_use_second_uom = fields.Boolean()
```
Is there any need of explicitly setting its default value to False :thinking: ?

pos_product_uom/static/src/component/control_buttons/control_buttons.js
```diff
@@ -0,0 +1,54 @@
...
+import { patch } from "@web/core/utils/patch";
+import { useService } from "@web/core/utils/hooks";
+import { usePos } from "@point_of_sale/app/store/pos_hook";
+import { ControlButtons } from "@point_of_sale/app/screens/product_screen/control_buttons/control_buttons";
+import { QuantityDialog } from "../quantity_dialog/quantity_dialog";
+
+patch(ControlButtons.prototype, {
+    setup() {
+        super.setup();
+        this.notification = useService('notification');
```
ampa-odoo, line 12: IMO you are using the patch and the notification service is already used in the ControlButtons component, do we need to write it again?
```suggestion
```

pos_product_uom/static/src/component/product_screen/product_screen.js
```diff
@@ -0,0 +1,29 @@
+/** @odoo-module */
+
+import { patch } from "@web/core/utils/patch";
+import { usePos } from "@point_of_sale/app/store/pos_hook";
+import { ProductScreen } from "@point_of_sale/app/screens/product_screen/product_screen";
+
+patch(ProductScreen.prototype, {
+    setup() {
+        super.setup();
+        this.pos = usePos();
```
ampa-odoo, line 10: Do we need this LOC?

ampa-odoo (changes requested):

pos_product_uom/models/product_template.py
rsbh-odoo: Yes, we have the category field, but it cannot be directly used in a domain with (.) notation. To resolve this, we create first_uom_category_id as a related field, allowing seamless domain usage

product_warranty_management/models/warranty_configuration.py
```diff
@@ -0,0 +1,19 @@
...
+from odoo import fields, models
+
+
+class WarrantyConfiguration(models.Model):
+    _name = 'warranty.configuration'
+    _description = 'Warranty Configuration'
+    _rec_name = 'duration_unit'
+    
+    duration_unit =  fields.Integer(string="Duration Value", required=True, default=1)
+    duration_type = fields.Char(string="Duration Unit", required=True, default="Year")
```
ampa-odoo, line 12: What is the need of adding this character field as the end date is computed in years only. If someone changes the string to months or days but the computation is done on years only?

`warranty_line.end_date = fields.Date.today() + relativedelta(years=duration_unit)`

product_warranty_management/views/sale_order_views.xml
```diff
@@ -0,0 +1,17 @@
...
+      <field name="name">view.order.form.inherited</field>
+      <field name="model">sale.order</field>
+      <field name="inherit_id" ref="sale.view_order_form" />
+      <field name="arch" type="xml">
+         <div name="so_button_below_order_lines" position="inside">
+            <button 
+               name="product_warranty_management.product_warranty_wizard_action"
+               type="action"
+               string="Add warranty"
+               invisible="not order_line"/>
```
ampa-odoo, line 13: We can check more specific condition that if any warranty having product is present in order_line then only this button should be visible. What do you say? :thinking:

product_warranty_management/models/warranty_configuration.py
ampa-odoo, line 12:
```suggestion
```

product_warranty_management/models/product_template.py
```diff
@@ -0,0 +1,9 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+     
+    is_warranty_available = fields.Boolean(string="Is Warranty Available", default=False)
```
ampa-odoo, line 9:
```suggestion
    is_warranty_available = fields.Boolean(string="Is Warranty Available")
```
Need of explicitly defining the default value? :thinking:

ampa-odoo (changes requested):

pos_product_uom/static/src/component/control_buttons/control_buttons.js
rsbh-odoo: Yes, it is not necessary because it is already accessible to every inherited component when defined in the parent component.

pos_product_uom/static/src/component/product_screen/product_screen.js
rsbh-odoo: No, it is not needed because usePos() is already defined in the parent component. 😅

product_warranty_management/views/sale_order_views.xml
rsbh-odoo: In my opinion, we can make it more specific by showing the button only when the added order line has a warranty product.

product_warranty_management/models/warranty_configuration.py
rsbh-odoo: Yes, right now this field is not needed since the final computation is always based on years. This field would be useful if we provided an option to choose between years and months for the warranty.

pos_product_uom/models/product_template.py
rsbh-odoo: No, it's not needed since Boolean fields default to False. 😅

PR #422, opened 2025-02-26
[IMP] rental_deposit: added deposit in rental

ppch-odoo:
Created new module name "estate"

### DIFF
--- a/rental_deposit/models/product_template.py
+++ b/rental_deposit/models/product_template.py
@@ -1,0 +1,8 @@
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    require_deposit = fields.Boolean(string="Require Deposit")
+    deposit_amount = fields.Float(string="Deposit Amount")
--- a/rental_deposit/__init__.py
+++ b/rental_deposit/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/rental_deposit/__manifest__.py
+++ b/rental_deposit/__manifest__.py
@@ -1,0 +1,14 @@
+{
+    'name': "Rental Deposit",
+    'version': '1.0',
+    'depends': ['website_sale_renting'],
+    'author': "ppch",
+    'category': 'Category',
+    'description': """
+    """,'license': "LGPL-3",
+    'data': [
+        'views/res_config_settings_views.xml',
+        'views/product_template_views.xml',
+    ],
+    'installable': True,
+}
--- a/rental_deposit/models/__init__.py
+++ b/rental_deposit/models/__init__.py
@@ -1,0 +1,2 @@
+from . import res_config_settings
+from . import product_template
--- a/rental_deposit/models/res_config_settings.py
+++ b/rental_deposit/models/res_config_settings.py
@@ -1,0 +1,11 @@
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = 'res.config.settings'
+
+    deposit_product_id = fields.Many2one(
+        'product.product',
+        string="Deposit Product",
+        config_parameter='rental_deposit.deposit_product_id'
+    )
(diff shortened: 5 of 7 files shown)


rental_deposit/models/product_template.py
```diff
@@ -0,0 +1,14 @@
...
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    require_deposit = fields.Boolean(string="Require Deposit")
+    deposit_amount = fields.Float(string="Deposit Amount")
+
+    @api.onchange('require_deposit')
+    def _onchange_require_deposit(self):
+        self.write({
+            'deposit_amount': 0 if self.require_deposit else 0
```
bhra-odoo, line 13: why?

rental_deposit/models/sale_order.py
```diff
@@ -0,0 +1,22 @@
...
+        return self.order_line.filtered(lambda line: line.product_id.id != int(deposit_product_id))
+
+    def copy_data(self, default=None):
+        """Override copy_data to prevent deposit lines from being duplicated."""
+        default = dict(default or {})
+        default.setdefault('order_line', [])
+        vals_list = super().copy_data(default=default)
+        for order, vals in zip(self, vals_list):
+            vals['order_line'] = [
+                Command.create(line)
```
bhra-odoo, line 19: where have you define `line`?

rental_deposit/models/sale_order_line.py
```diff
@@ -0,0 +1,69 @@
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        """Create deposit lines automatically for products that require a deposit."""
+        lines = super().create(vals_list)
+
+        deposit_product_id = self.env['ir.config_parameter'].sudo().get_param('rental_deposit.deposit_product_id')
+        deposit_product = self.env["product.product"].browse(int(deposit_product_id))
+
+        if not deposit_product_id:
+            raise UserError(_("Please select deposit product from configuration"))
```
bhra-odoo, line 17: i guess you can throw usererror before browse so if there is no id, you can save browse call.

rental_deposit/models/sale_order_line.py
```diff
@@ -0,0 +1,69 @@
...
+            if line.product_id and line.product_id.product_tmpl_id.require_deposit:
+                deposit_price = line.product_id.product_tmpl_id.deposit_amount if line.product_id.product_tmpl_id.deposit_amount else deposit_product.lst_price
+                self.create([{
+                    "order_id": line.order_id.id,
+                    "product_id": deposit_product.id,
+                    "name": f"Deposit for {line.product_id.name}",
+                    "product_uom_qty": line.product_uom_qty,
+                    "price_unit": deposit_price,
+                    "linked_line_id": line.id,
+                }])
```
bhra-odoo, line 29: Instead of that you can create all together by creating list.

rental_deposit/models/sale_order_line.py
```diff
@@ -0,0 +1,69 @@
...
+                    })
+                else:
+                    self.create({
+                        "order_id": line.order_id.id,
+                        "product_id": deposit_product.id,
+                        "name": f"Deposit for {line.product_id.name}",
+                        "product_uom_qty": line.product_uom_qty,
+                        "price_unit": deposit_price,
+                        "linked_line_id": line.id,
+                    })
```
bhra-odoo, line 64: same create here

bhra-odoo (changes requested):

rental_deposit/models/product_template.py
ppch-odoo: This is not needed
I was checking that if someone after setting a deposit untick then retick on that time it should be zero.
But it i forgot to remove it
so i will remove it

rental_deposit/models/sale_order.py
ppch-odoo: below line 19, in for loop

Command.create(line)
for line in order._get_copiable_order_lines_without_deposit().copy_data()

bhra-odoo (changes requested):

PR #435, opened 2025-02-27
[ADD] event_ticket_limit:  restricts the maximum number of tickets a user can book per registration.

aych-odoo:
- Introduced a new field to store the maximum number of tickets a person can book per registration.
- This allows for dynamic configuration of the ticket limit instead of a fixed value which was 9.
Task: [4573739](https://www.odoo.com/odoo/project/18468/tasks/4573739)

### DIFF
--- a/event_ticket_limit/models/event_ticket.py
+++ b/event_ticket_limit/models/event_ticket.py
@@ -1,0 +1,21 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import fields, models
+
+
+class EventTemplateTicket(models.Model):
+    _inherit = "event.type.ticket"
+
+    max_tickets_per_registration = fields.Integer(
+        string="Max Tickets Per Registration",
+        default=9,
+        help="Define the maximum number of tickets allowed per registration. Between 1-9"
+    )
+
+    _sql_constraints = [
+        (
+            "check_max_tickets_per_registration",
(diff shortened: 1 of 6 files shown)


maad-odoo:
Good to go :+1:

event_ticket_limit/models/event_ticket.py
```diff
@@ -0,0 +1,21 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import fields, models
+
+
+class EventTemplateTicket(models.Model):
+    _inherit = "event.type.ticket"
```
maad-odoo, line 7: the inherited model is `event.type.ticket` and the  file name is `event_ticket` ?

event_ticket_limit/views/event_templates_page_registration.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <template inherit_id="website_event.modal_ticket_registration" id="limit_ticket_registrations">
```
maad-odoo, line 3: put id before inherit_id

event_ticket_limit/models/event_ticket.py
aych-odoo: The model `event.type.ticket` is defined in the file `event_ticket.py`. 
Which name would be preferred for naming this file: the name of the file from which I am inheriting the model, or the model's name?

PR #436, opened 2025-02-27
[ADD] *: practice tasks implementation

sbbh-odoo:


supplier_portal/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+{
```
aktr-odoo, line 1: It's good to follow the sequence odoo follow to define keys in some certain order

supplier_portal/controllers/main.py
```diff
@@ -0,0 +1,52 @@
...
+        # create a new vendor bill with selected company
+        bill = request.env["account.move"].sudo().create({
+            'move_type': 'in_invoice',
+            'partner_id': supplier.id,
+            'company_id': company_id,
+        })
+
+        files = []
+        if pdf_file:
+            files.append(("invoice.pdf", post.get("pdf_invoice"), "application/pdf"))
```
aktr-odoo, line 37: Any reason to do `post.get("pdf_invoice")` again when we have already have `pdf_file`?

supplier_portal/controllers/main.py
```diff
@@ -0,0 +1,52 @@
...
+        if xml_file:
+            files.append(("invoice.xml", post.get("xml_invoice"), "application/xml"))
```
aktr-odoo, line 39: Any reason to do `post.get("xml_invoice")` again when we have already have `xml_file`?

supplier_portal/controllers/main.py
```diff
@@ -0,0 +1,52 @@
...
+
+        if not post:
+            raise NotFound()
+        
+        supplier = request.env.user.partner_id
+        company_id = int(post.get("company_id"))
+        pdf_file = post.get("pdf_invoice")
+        xml_file = post.get("xml_invoice")
+
+        if not pdf_file and not xml_file:
```
aktr-odoo, line 25: How about validating file types too?

aktr-odoo (changes requested):
:wave: 

I have left a few comments. Could you please have a look?

supplier_portal/controllers/main.py
sbbh-odoo: forgot to change, I'll update it.

supplier_portal/controllers/main.py
sbbh-odoo: forgot to change, I'll update it too!

sbbh-odoo:
@aktr-odoo Requested changes have been made can you please have a look?

deposit_rental/models/sale_order_line.py
```diff
@@ -0,0 +1,51 @@
...
+                if not line.is_deposit_line:
+                    deposit_line = self.search([
+                        ("order_id", "=", line.order_id.id),
+                        ("deposit_link_id", "=", line.id),
+                    ])
+
+                    if deposit_line:
+                        deposit_line.write({
+                            "product_uom_qty": vals["product_uom_qty"],
+                            "price_unit": line.product_template_id.deposit_amount,
```
aktr-odoo, line 49: Do we need to update `price_unit`?

aktr-odoo (changes requested):
:wave: 

I have left a few comments. Could you please have a look?

sbbh-odoo:
@aktr-odoo Requested changes have been made can you please have a look?

aktr-odoo:
Give proper commit message

deposit_rental/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+        'sale_renting',
+        'website_sale'
+    ],
+    'data': [
+        'views/res_config_settings_views.xml',
+        'views/product_template_views.xml',
+        'views/website_sale_rental_templates.xml',
+    ],
+    'installable': True,
+    'application': False,
```
aktr-odoo, line 15: Do we need to give `'application': False` ?

deposit_rental/models/sale_order_line.py
```diff
@@ -0,0 +1,54 @@
+from odoo import api, fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    is_deposit_line = fields.Boolean(string="Is Deposit Line", default=False)
+    deposit_link_id = fields.Many2one(
```
aktr-odoo, line 8: We already have this relation? Any reason to define this again, and not using the original one?

deposit_rental/models/sale_order_line.py
```diff
@@ -0,0 +1,54 @@
...
+                    "product_uom_qty": line.product_uom_qty,
+                    "price_unit": line.product_template_id.deposit_amount,
+                    "is_deposit_line": True,
+                    "sequence": line.sequence + 1,
+                    "deposit_link_id": line.id,
+                    "name": f"Deposit for {line.name}",
+                })
+        return lines
+
+    def write(self, vals):
```
aktr-odoo, line 35: This could be simplified if we use the available relation of `sale.order.line`

aktr-odoo (changes requested):
:wave: 

I have left a few comments. Could you please have a look?

deposit_rental/__manifest__.py
sbbh-odoo: No, I don't want to add this module as an application so I change it to False in manifest initially, but forgot to remove entire line while pushing the code.

sbbh-odoo:
The requested changes have been made. Could you please review them?

deposit_rental/models/res_config_settings.py
```diff
@@ -0,0 +1,24 @@
...
+        string="Deposite",
+        config_parameter="sale_renting.deposit_product_id",
+        domain=[("type", "=", "service")],
+        help="Product used as a deposit when renting."
+    )
+
+    def set_values(self):
+        old_deposit_product_id = self.env["ir.config_parameter"].sudo().get_param("sale_renting.deposit_product_id")
+        super().set_values()
+        new_deposit_product_id = self.deposit_product_id.id if self.deposit_product_id else False
```
aktr-odoo, line 18: can we use `or` here?

deposit_rental/models/res_config_settings.py
```diff
@@ -0,0 +1,24 @@
...
+
+        if old_deposit_product_id != new_deposit_product_id:
+            products = self.env["product.template"].sudo().search([
```
aktr-odoo, line 21: Why sudo?

deposit_rental/models/sale_order_line.py
```diff
@@ -0,0 +1,46 @@
...
+        res = super().write(vals)
+
+        if "product_uom_qty" in vals:
+
+            deposit_lines = self.search([
+                ("is_deposit_line", "=", True),
+            ])
+            for line in self:
+                if not line.is_deposit_line:
```
aktr-odoo, line 39: We already have `deposit_lines` so do we need `if not line.is_deposit_line:` or it can be achieved in some other way?

deposit_rental/tests/test_deposit_rental.py
```diff
@@ -0,0 +1,81 @@
...
+            "name": "Projectore Rental",
+            "type": "consu",
+            "require_deposit": True,
+            "list_price": 100.0
+        })
+
+        self.env["ir.config_parameter"].sudo().set_param("sale_renting.deposit_product_id", self.deposit_product.id)
+        self.partner = self.env["res.partner"].create({"name": "Test Customer"})
+
+    def test_deposit_product_config(self):
```
aktr-odoo, line 28: This is odoo's default behaviour. No need to test it.
Just test what functionality what you have added

deposit_rental/tests/test_deposit_rental.py
```diff
@@ -0,0 +1,81 @@
...
+        """Test that the deposit product setting is saved correctly."""
+        config = self.env["res.config.settings"].create({
+            "deposit_product_id": self.deposit_product.id
+        })
+        config.set_values()
+        deposit_product_id = self.env["ir.config_parameter"].sudo().get_param("sale_renting.deposit_product_id")
+        self.assertEqual(int(deposit_product_id), self.deposit_product.id, "Deposit product config is incorrect!")
+
+    def test_rental_product_requires_deposit(self):
```
aktr-odoo, line 37: Same here.
No need to test this

deposit_rental/tests/test_deposit_rental.py
```diff
@@ -0,0 +1,81 @@
...
+    def test_sale_order_deposit_creation(self):
+        sale_order = self.env["sale.order"].create({"partner_id": self.partner.id})
+        sale_order.write({
+            "order_line": [
+                Command.create({"product_id": self.rental_product.id, "product_uom_qty": 2}),
+            ]
+        })
+        deposit_line = sale_order.order_line.filtered(lambda l: l.is_deposit_line)
+        
+        self.assertEqual(len(deposit_line), 1, "Deposit line was not created!")
```
aktr-odoo, line 51: Checking the length dosen't seem appropriate to me.
I mean what if line is created due to some other reason?

deposit_rental/tests/test_deposit_rental.py
aktr-odoo, line 1: All the best cases are included in one file.
The test cases must be in a separate file, code-wise.

deposit_rental/tests/test_deposit_rental.py
```diff
@@ -0,0 +1,81 @@
...
+                Command.create({"product_id": self.rental_product.id, "product_uom_qty": 1}),
+            ]
+        })
+        
+        copied_order = sale_order.copy()
+        
+        self.assertNotEqual(copied_order.order_line, sale_order.order_line, "Lines should not be directly copied!")
+        self.assertEqual(len(copied_order.order_line.filtered(lambda line: line.is_deposit_line)), 1, "Deposit should be recreated!")
+
+    def test_website_cart_deposit_display(self):
```
aktr-odoo, line 70: What is this test doing?
You are setting some product, then checking whether it is set or not.

deposit_rental/models/sale_order_line.py
sbbh-odoo: ok, understood. we can give directly `if not in deposit lines:`

deposit_rental/tests/test_deposit_rental.py
sbbh-odoo: okay, i understood! there is no need to recheck odoo functionality as we discussed.

deposit_rental/models/res_config_settings.py
sbbh-odoo: there is no need to bypass access rights while using search, i will remove it.

deposit_rental/models/sale_order_line.py
sbbh-odoo: since we have deposit_lines so we can remove deposit lines from entering in for loop by (self - deposit_lines)
image

deposit_rental/models/res_config_settings.py
sbbh-odoo: yes, instead of if else we can improve it by simplified syntax using or,
image

deposit_rental/tests/test_deposit_rental.py
sbbh-odoo: Instead of only checking the length (len(deposit_line)), we can make the test more specific to confirm that:
A deposit line exists
It is linked to the correct product
It has the expected quantity and price

sbbh-odoo:
All the requested changes have been made, can you have a look?

sbbh-odoo:
All the requested changes have been made, can you please have a look?

deposit_rental/models/res_config_settings.py
aktr-odoo: No need to add screenshots.
I can see your code once you have pushed.

barcode_catalog_handler/static/src/product_catalog/kanban_controller.js
```diff
@@ -0,0 +1,100 @@
...
+                ["order_id", "=", this.orderId],
+                ["product_id", "=", product.id]
+            ],
+            ["id", orderLineQty]
+        )
+
+        let newQty = 1;
+        const orderLine = orderLines?.[0];
+
+        if(orderLine){
```
kcv-odoo, line 77: What if there is more then one line with same product?

barcode_catalog_handler/static/src/product_catalog/kanban_controller.js
```diff
@@ -0,0 +1,100 @@
...
+            const newQty = orderLine[orderLineQty] + 1;
+
+            await rpc("/product/catalog/update_order_line_info", {
+                order_id: this.orderId,
+                product_id: product.id,
+                quantity: newQty,
+                res_model: this.orderResModel,
+            })
```
kcv-odoo, line 85: This could move outside if else with little change about qty computation

barcode_catalog_handler/static/src/product_catalog/kanban_model.js
```diff
@@ -0,0 +1,66 @@
+import { patch } from "@web/core/utils/patch";
+import { getFieldsSpec } from "@web/model/relational_model/utils";
```
kcv-odoo, line 2: What does `getFieldsSpec` return?

barcode_catalog_handler/static/src/product_catalog/kanban_model.js
```diff
@@ -0,0 +1,66 @@
...
+
+patch(ProductCatalogKanbanModel.prototype, {
+
+    /**
+     * Overrides "_loadUngroupedList" to fetch all products and sort them 
+     * based on their quantity in the current order.
+     *
+     * @returns {Object} - Paginated product list sorted by quantity in the order.
+     */
+    async _loadUngroupedList(config) {
```
kcv-odoo, line 14: What if there is groupBy added don't we want to show recent ordered product in that case?

barcode_catalog_handler/static/src/product_catalog/kanban_model.js
```diff
@@ -0,0 +1,66 @@
...
+        const orderId = config.context.order_id;
+        const orderResModel = config.context.product_catalog_order_model;
+        const activeModel = config.context.active_model;
+
+        const kwargs = {
+            specification: getFieldsSpec(config.activeFields, config.fields),
+            context: { bin_size: true, ...config.context },
```
kcv-odoo, line 21: What is bin_size?

barcode_catalog_handler/static/src/product_catalog/kanban_controller.js
```diff
@@ -0,0 +1,100 @@
...
+            const message = hasVendorFilter 
+                ? _t("Try removing the top filter to broaden your search.")
+                : _t("Product not found with this barcode!");
+            
+            this.notification.add(message, { type: "danger" });
+            return;
+        }
+
+        await this.orderLineHandler(product);
+        this.model.load();
```
kcv-odoo, line 51: Why we need to do this?

kcv-odoo (changes requested):
Thank you for your work!! :duck: 

Added some comments and questions regarding choices :)

Have a nice day!

barcode_catalog_handler/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+{
+    'name': "Barcode Catalog Handler",
+    'version': "1.0",
+    'description': """This module used to scans barcodes in product catalog kanban to add/increase quantity in Sale/Purchase Orders.""",
+    'depends': [
+        "barcodes",
+        "sale_management",
+        "purchase"
+    ],
```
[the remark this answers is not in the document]
sbbh-odoo: Okay understood, I'll create separate bridge modules for integrating with Sale and Purchase so the functionality remains modular and doesn't force-install both applications. This way, users can benefit from the catalog handler even if they're using only one of the apps.

barcode_catalog_handler/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+        "web.assets_backend": [
+            "barcode_catalog_handler/static/src/product_catalog/*"
+        ],
+        "web.assets_unit_tests": [
+            "barcode_catalog_handler/static/tests/**/*"
+        ],
+    },
+    'license': "LGPL-3",
+    'application': False,
+    'installable': True,
```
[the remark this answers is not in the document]
sbbh-odoo: You're right! Earlier, I had both sale and purchase in depends, so setting auto_install would have triggered unwanted installs. But now that it's only depending on barcodes, I’ll update it to auto_install: True. Thanks!

barcode_catalog_handler/static/src/product_catalog/kanban_model.js
```diff
@@ -0,0 +1,66 @@
...
+
+import { ProductCatalogKanbanModel } from "@product/product_catalog/kanban_model";
+
+patch(ProductCatalogKanbanModel.prototype, {
+
+    /**
+     * Overrides "_loadUngroupedList" to fetch all products and sort them 
+     * based on their quantity in the current order.
+     *
+     * @returns {Object} - Paginated product list sorted by quantity in the order.
```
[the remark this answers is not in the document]
sbbh-odoo: By `paginated`, I meant products list returned with pagination data (offset, limit), only 40 products will be visible from 0th-39th for the 1st page, from 40th will be on the next page.

barcode_catalog_handler/static/src/product_catalog/kanban_model.js
sbbh-odoo: It returns fields specification object, which contains all the fields that we are going to use (blueprint). It’s passed in `kwargs.specification` to ensure the model knows what to expect.

barcode_catalog_handler/static/src/product_catalog/kanban_model.js
sbbh-odoo: In that case we have to override `_loadGroupedList()` method too, Which keep the group structure (recent ordered product). But sort products inside each group by quantity in the order.

barcode_catalog_handler/static/src/product_catalog/kanban_controller.js
sbbh-odoo: To refresh or reload the data from the model and update the view to reflect the latest state. After scanning a barcode, we update the quantity and re-sort the product cards, so reloading the model ensures the view displays the correct order.

barcode_catalog_handler/static/src/product_catalog/kanban_model.js
sbbh-odoo: `bin_size` allows us to fetch only the size of a binary field (like an image) instead of its full content. It's used for performance reasons when we don't immediately need the actual data.

sbbh-odoo:
@kcv-odoo Requested changes have been made can you please have a look?

PR #439, opened 2025-02-27
[IMP] event_ticket: maximum tickets per registration allowed

assk-odoo:
Implemented a restriction using tickets_per_registration to limit the number of 
tickets a user can book in a single registration.Updated the event registration 
modal to enforce this limit dynamically based on ticket and event seat 
availability.Modified QWeb template of website_event to ensure proper validation
at the fronted, preventing users from exceeding the defined ticket limit.


event_ticket/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+{
+    'name': "Event Tickets",
+    'version': "1.0",
+    'category': "Events",
+    'summary': "Manage event tickets and registrations",
+    'description': "The event module allows event organizers to dynamically allot the number of tickets per registration for the events.",
+    'author': "sujal_asodariya",
+    'depends': ["website_event","event"],
```
divy-odoo, line 8: Do we need `event` as depends here? why?

event_ticket/models/event_ticket.py
```diff
@@ -0,0 +1,20 @@
+from odoo import api, fields, models
+from odoo.exceptions import ValidationError
+
+class EventTicket(models.Model):
+    _inherit = "event.event.ticket"
+
+    tickets_per_registration = fields.Integer(
+        string="Tickets per Registration",
+        help="Number of tickets per registration",
+        width = "150px",
```
divy-odoo, line 10: It won't work!

divy-odoo (commented):
Add a common field in event.event model for tickets per registration which will apply to all categories of ticket.

event_ticket/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    'description': "The event module allows event organizers to dynamically allot the number of tickets per registration for the events.",
+    'author': "sujal_asodariya",
+    'depends': ["website_event"],
+    'data': [
+        "views/event_register_website_views.xml",
+        "views/event_ticket_views.xml",
+    ],
+    'installable': True,
+    'application': True,
+    'application':  True,
```
divy-odoo, line 15: Two times same key?

event_ticket/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    'auto_install': False,
```
divy-odoo, line 16: what's the default value of `auto_install`, if i don't write the key here?

event_ticket/models/event_ticket.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @api.constrains('tickets_per_registration')
+    def _check_ticket_limit(self):
+        for record in self:
+            if record.tickets_per_registration < 1:
+                raise ValidationError("Maximum tickets per registration must be at least 1.")
+            for line in record.event_ticket_ids:
+                if record.tickets_per_registration > line.seats_max:
+                    raise ValidationError(f"You cannot book more than {line.seats_max} tickets.")
+            if record.tickets_per_registration > record.seats_max:
```
divy-odoo, line 21: What if `seats_max` is not set or it's not limited registration event?

image

divy-odoo (commented):
Here is the review for the `event_tickets`!!

event_ticket/models/event_ticket.py
assk-odoo: if we can not set seats_max when we raise one ticket its value 0 so it raise and error message that you can not book ticket more than 0 so after that we change the our seats_max field

divy-odoo: If `seats_max` field is not set, event will be considered for the unlimited registration. but in your case it's raising t he value 0. 
At the end, if `seat_max` is not set, the tickets per registration can be any positive number.

assk-odoo: done

project_deliverydate_linked/models/project_project.py
```diff
@@ -0,0 +1,41 @@
...
+            ('account_id.internal_group', 'in', ['income', 'expense']),
+            ('parent_state', '=', 'posted'),
+            ('date_reloaded', '!=', self.date_start)
+        ])
+
+        if not move_lines:
+            raise UserError("No journal items need to be recognized.")
+
+        action = self.env.ref('account.account_automatic_entry_wizard_action').read()[0]
+        print(self.env.context,"jhgf")
```
divy-odoo, line 30: Remove print statements from code.

project_deliverydate_linked/models/project_project.py
```diff
@@ -0,0 +1,41 @@
...
+        ctx = self.env.context.copy()
+        ctx.update({
+        'active_ids': move_lines.ids,
+        'active_model': 'account.move.line',
+        'default_action': 'change_period',
+        'default_date': self.date_start
```
divy-odoo, line 36: indentation

PR #440, opened 2025-02-27
[ADD] event_ticket_limit: allow configurable ticket limit per registration

nmak-odoo:
Added `max_tickets_per_registration` field to define a per-registration ticket limit.
Replaced the fixed limit of 9 with a dynamic limit based on this field.
Updated ticket selection logic to enforce the new limit. 
Modified the `website_event` UI to display available ticket options accordingly.

task-4589683

### DIFF
--- a/event_ticket_limit/__init__.py
+++ b/event_ticket_limit/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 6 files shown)


event_ticket_limit/models/event_ticket.py
```diff
@@ -0,0 +1,19 @@
+from odoo import models, fields
+
+
+class EventTemplateTicket(models.Model):
+    _inherit = 'event.type.ticket'
```
maad-odoo, line 5: the inherited model is event.type.ticket and the file name is event_ticket ?

PR #472, opened 2025-02-28
[ADD] deposit_rental: deposit for rental products

nmak-odoo:
New configuration option to define a deposit product in rental settings. 
'Require Deposit' checkbox on rental products with an amount field. 
Deposit line appears in sales orders when a rental product  with deposit amount is added to the order line.
Deposit amount field added on webshop product page and order overview.

Task ID: 4605707

### DIFF
--- a/deposit_rental/models/product_template.py
+++ b/deposit_rental/models/product_template.py
@@ -1,0 +1,21 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    deposit_product = fields.Boolean(
+        string="Deposit Product", help="Make it a deposit product."
+    )
+    require_deposit = fields.Boolean(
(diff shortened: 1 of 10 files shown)


deposit_rental/models/product_template.py
```diff
@@ -0,0 +1,26 @@
...
+        string="Deposit Product", help="Make it a deposit product."
+    )
+
+    require_deposit = fields.Boolean(
+        string="Require Deposit", help="Enable if this product requires deposit."
+    )
+
+    deposit_amount = fields.Float(
+        string="Amount", help="This specifies deposit for 1 unit of this product."
+    )
```
maad-odoo, line 19:
```suggestion
    deposit_product = fields.Boolean(
        string="Deposit Product", help="Make it a deposit product."
    )
    require_deposit = fields.Boolean(
        string="Require Deposit", help="Enable if this product requires deposit."
    )
    deposit_amount = fields.Float(
        string="Amount", help="This specifies deposit for 1 unit of this product."
    )
```
No need to leave extra lines between field declaration

PR #583, opened 2025-03-10
[ADD] quantity_on_hand: ui changes of product view

amga-odoo:
- added quantity on hand field editable if not multi-locations in product general information with update button invisible if not multi-locations and visible on hover, link to On Hand report
- Remove stat button, replace by update besides qty on hand in the form
- Change the layout of forecast button, red if negative
- Remove descriptions for goods

### DIFF
--- a/quantity_on_hand/__init__.py
+++ b/quantity_on_hand/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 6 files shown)


divy-odoo (commented):
Good to go solution, Just don't forgot to mention the task id in commit message. 
Thanks : )

PR #589, opened 2025-03-10
[ADD] book_price: adding book price at sales order line and at invoice.

pbch-odoo:


new_product_type/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import models
+from . import wizard
```
divy-odoo, line 2: Add new line at the end of each file.

new_product_type/models/product_template.py
```diff
@@ -0,0 +1,6 @@
+from odoo import models, fields
+
+class ProductKit(models.Model):
+    _inherit = "product.template"
+    is_kit = fields.Boolean(string="Is Kit")
+    kit_product_ids = fields.Many2many("product.product", "kit_product_m2m", "kit_id", "product_id")
```
divy-odoo, line 6: 1. Field string must be meaning full name.
2. What is `kit_id`?

new_product_type/models/sale_order.py
```diff
@@ -0,0 +1,6 @@
+from odoo import models, fields, api
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    print_subproducts = fields.Boolean(string="Print Subproducts in Report", default=True)
```
divy-odoo, line 6: string = `Print in report?`

new_product_type/models/sales_order_inherit.py
```diff
@@ -0,0 +1,39 @@
+from odoo import models, fields, api
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    is_kit_product = fields.Boolean(string="Is Kit Product", compute="_compute_is_kit", store=True)
+    kit_component_ids = fields.One2many("sale.order.line", "parent_line_id", string="Kit Components")
+    parent_line_id = fields.Many2one("sale.order.line", string="Parent Kit Line", ondelete="cascade")
+    display_price_unit = fields.Float(compute="_compute_display_price_unit", readonly=False)
+    display_price_subtotal = fields.Float(compute="_compute_display_price_subtotal")
```
divy-odoo, line 10: Why & where are these field used?

new_product_type/models/sales_order_inherit.py
```diff
@@ -0,0 +1,39 @@
+from odoo import models, fields, api
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    is_kit_product = fields.Boolean(string="Is Kit Product", compute="_compute_is_kit", store=True)
+    kit_component_ids = fields.One2many("sale.order.line", "parent_line_id", string="Kit Components")
+    parent_line_id = fields.Many2one("sale.order.line", string="Parent Kit Line", ondelete="cascade")
```
divy-odoo, line 8: From where this value are being assigned?

new_product_type/wizard/kit_wizard.xml
```diff
@@ -0,0 +1,26 @@
...
+                            <field name="product_id" readonly="0" />
+                            <field name="quantity" />
+                            <field name="price" />
+                        </list>
+                    </field>
+                </group>
+                <footer>
+                    <button string="Confirm" type="object" name="action_confirm" class="btn-primary" />
+                    <button string="Cancel" class="btn-secondary" special="cancel" />
+                </footer>
```
divy-odoo, line 22: Is it necessary to define these buttons in view?

new_product_type/wizard/kit_wizard.py
```diff
@@ -0,0 +1,161 @@
...
+    product_id = fields.Many2one("product.product", required=True, string="Component")
+    quantity = fields.Float(string="Quantity", default=1.0)
+    price = fields.Float(string="Price")
+    order_line_id = fields.Many2one("sale.order.line", string="Existing Order Line")
+
+    @api.model_create_multi
+    def create(self, vals):
+        print("INSIDE OVERRIDE CREATE")
+        print(self.wizard_id)
+        print(vals)
```
divy-odoo, line 160: Remove prints & no need of create method here.

new_product_type/wizard/kit_wizard.py
```diff
@@ -0,0 +1,161 @@
...
+
+    @api.model
+    def default_get(self, fields_list):
+        """Pre-fill the wizard with existing sub-products or default kit components."""
+        res = super().default_get(fields_list)  
+        sale_order_line_id = self.env.context.get("default_sale_order_line_id")
+        if not sale_order_line_id:
+            return res
+        
+        sale_order_line = self.env["sale.order.line"].browse(sale_order_line_id)
```
divy-odoo, line 20: sale_order_lines

new_product_type/wizard/kit_wizard.py
```diff
@@ -0,0 +1,161 @@
...
+        components = []
+        # Check if wizard has previously stored prices
+        prev_wizard = self.search([("sale_order_line_id", "=", sale_order_line_id)], order="id desc", limit=1)
+        if prev_wizard:
+            for component in prev_wizard.component_ids:
```
divy-odoo, line 25: I didn't get this check, so we really need this if condition check? can we narrow down this if else logic.

new_product_type/wizard/kit_wizard.py
```diff
@@ -0,0 +1,161 @@
+from odoo import fields, models, api, Command
+from odoo.exceptions import UserError
+
+class KitWizard(models.TransientModel):
+    _name = "kit.wizard"
+    _description = "Kit Wizard"
+
+    sale_order_line_id = fields.Many2one("sale.order.line", string="Sale Order Line", required=True)
```
divy-odoo, line 8: why this field? can't we directly get the`line_ids` from the current active `sale order`?

new_product_type/wizard/kit_wizard.py
```diff
@@ -0,0 +1,161 @@
...
+        if not sale_order_line_id:
+            return {"type": "ir.actions.act_window_close"}
+
+        sale_order_line = self.env["sale.order.line"].browse(sale_order_line_id)
+        sale_order = sale_order_line.order_id
+
+        original_main_product_price = sale_order_line.product_id.lst_price
+
+        existing_components = {line.product_id.id: line for line in sale_order_line.kit_component_ids}
+        updated_components = []
```
divy-odoo, line 92: Change the variable name with meaning full name. the one should get the idea about what value this variable will be holding.

new_product_type/wizard/kit_wizard.py
```diff
@@ -0,0 +1,161 @@
...
+        return super(SaleOrderLine, self).unlink()
+    
+    def action_confirm(self):
+        """Confirm the selection of sub-products and update sale order lines."""
+        sale_order_line_id = self.sale_order_line_id.id
+
+        if not sale_order_line_id:
+            return {"type": "ir.actions.act_window_close"}
+
+        sale_order_line = self.env["sale.order.line"].browse(sale_order_line_id)
```
divy-odoo, line 86: Why we need to get particular sale_order_line id?

new_product_type/wizard/kit_wizard.py
```diff
@@ -0,0 +1,161 @@
...
+                
+        # Delete the main product line itself
+        return super(SaleOrderLine, self).unlink()
+    
+    def action_confirm(self):
+        """Confirm the selection of sub-products and update sale order lines."""
+        sale_order_line_id = self.sale_order_line_id.id
+
+        if not sale_order_line_id:
+            return {"type": "ir.actions.act_window_close"}
```
divy-odoo, line 84: ```            return { "type": "ir.actions.act_window_close" }```

new_product_type/models/product_template.py
pbch-odoo: Sir here `kit_product_m2m` is a relational table name
Sir here kit_id is the `column_name` which is links to the current `model`
 and `product_id` is also a `column_name` which is links to the `product.product` model

new_product_type/models/sales_order_inherit.py
pbch-odoo: These fields are used for displaying the `price_unit` and `price_subtotal` in the sale order line.

`price_unit` represents the price per unit of the product.
`price_subtotal` represents the total price for that product.
We define them as computed fields to ensure that if a sale order line is a kit component (i.e., it has a `parent_line_id`), then:

The sub-product's price is set to `0.0`, so it does not affect the total.
Only the main kit product retains and displays the actual price.
If there is no `parent_line_id`, these fields will display the `line.price_unit` and `line.price_subtotal` as usual.

new_product_type/wizard/kit_wizard.xml
pbch-odoo: Yes

new_product_type/wizard/kit_wizard.py
pbch-odoo: The `if prev_wizard:` check ensures that if a previous wizard exists for the same `sale_order_line_id`, it reuses the sub-products from that wizard instead of fetching them again from the sale order line or product template. This helps retain any previous modifications made in the wizard.

To optimize, we can **remove redundant checks** and simplify the logic by only checking `prev_wizard`. If a previous wizard exists, it already contains sub-products, so there's no need to check `sale_order_lines.product_id.is_kit`. Instead, we directly fall back to fetching from the product template if no previous wizard exists:
`sub_products = []

prev_wizard = self.search([("sale_order_line_id", "=", sale_order_line_id)], order="id desc", limit=1)

if prev_wizard:
    sub_products = [
        Command.create({
            "product_id": subproduct.product_id.id,
            "quantity": subproduct.quantity,
            "price": subproduct.price,
            "order_line_id": subproduct.order_line_id.id,
        }) for subproduct in prev_wizard.sub_product_ids
    ]
elif sale_order_lines.product_id.is_kit:
    sub_products = [
        Command.create({
            "product_id": kit_product.id,
            "quantity": 1.0,
            "price": kit_product.lst_price,
        }) for kit_product in sale_order_lines.product_id.kit_product_ids
    ]`
This improves readability, removes unnecessary condition checks, and ensures we always load the most relevant sub-products

new_product_type/wizard/kit_wizard.py
pbch-odoo: Yes, we can directly use `self.sale_order_line_id`, but using a variable is better for:
🔹 Performance (avoids multiple lookups).
🔹 Readability (cleaner, easier to modify).
🔹 Error Prevention (avoids unnecessary access on None).

new_product_type/wizard/kit_wizard.py
pbch-odoo: The field `sale_order_line_id` is used in the **Kit Wizard** to ensure that the wizard operates on a **specific sale order line,** rather than fetching all lines from the active sale order `(order_id.line_ids)`.

Using `sale_order_line_id` ensures that the wizard works on only one specific sale order line, preventing unintended modifications to other lines in the order.

new_product_type/wizard/kit_wizard.py
pbch-odoo: `return {"type": "ir.actions.act_window_close"}` is a special Odoo action that closes the wizard or popup when there’s no valid sale order line.

new_product_type/wizard/kit_wizard.py
```diff
@@ -7,105 +7,76 @@ class KitWizard(models.TransientModel):
...
+        # Pre-fill the wizard with existing sub-ptoducts.
+        res = super().default_get(fields_list)
         sale_order_line_id = self.env.context.get("default_sale_order_line_id")
         if not sale_order_line_id:
             return res
         
-        sale_order_line = self.env["sale.order.line"].browse(sale_order_line_id)
-        components = []
-        # Check if wizard has previously stored prices
+        sale_order_lines = self.env["sale.order.line"].browse(sale_order_line_id)
```
divy-odoo, line 20: No need to fetch, if you already created a field for that. ( although creating a new field here is not  preferable.)

PR #608, opened 2025-03-12
[IMP] one_time_sale: allow one-time sale/purchase for subscription products

pcbe-odoo:
### One-time purchase support for subscription products

- Exclude one-time sale order lines from recurring plan enforcement.
- Ensure subscription plans follow a proper order (weekly → monthly → yearly).
- Provide both one-time and subscription-based purchase options on the website for selling products in both ways.
- Use selling price for one-time purchases and plan-based pricing for subscriptions.
- Apply the sale price when no plan is selected in a quotation; otherwise, use the pricelist price.
- Remove the price per frequency column for consumable products and add a discount column based on the one-time
  sale price or the highest recurring plan price on the website.
- Ensure one-time purchase plans appear at the top of the comparison list for consumable products on the website.
- Display discount comparison based on the selected subscription plan on the subscription purchase option.
- Prevent showing the selling price and discount in cases where no discount is available.
- Restrict cart to only allow compatible product types based on product purchase.
- Hide non-selected purchase options when adding a product to the cart for one-time purchase product.
- Ensure one-time purchases are treated as simple products.
- Correctly differentiate between one-time and subscription purchases in cart from website.
- Add guided tours for creating and purchasing one-time sale products.
- Implement test cases to validate functionalities of one time sale.
- Validate modified `_get_pricelist_price` in test cases to ensure one-time purchases in sale orders apply the correct pricing.
- Restrict adding consumable products to the cart if they do not allow one-time sales when there is already one-time purchase product available in cart.
- Add translations for Finnish, French, and Gujarati.
- Handle cases where a product’s type changes from consumable to service/combo with `accept_one_time` disabled.


one_time_sale/models/product_template.py
```diff
@@ -0,0 +1,86 @@
...
+        help="Enable the product to sale/purchase for one-time.")
+
+    @staticmethod
+    def has_one_time_purchase(sale_order):
+        """Check if the sale order contains a one-time purchase product."""
+        return any(
+            not line.order_id.plan_id
+            and line.product_id.recurring_invoice
+            and line.product_id.accept_one_time
+            and line.price_unit == line.product_id.list_price
```
kpt-odoo, line 22: Can you please explain me why we need this condition?

one_time_sale/models/sale_order.py
```diff
@@ -0,0 +1,35 @@
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.depends('order_line', 'order_line.recurring_invoice')
+    def _compute_has_recurring_line(self):
```
kpt-odoo, line 8: it is bad idea to make depandancy on `order_line` can not use field `accept_one_time` instead of whole order_line can you check if its working?

one_time_sale/models/sale_order.py
```diff
@@ -0,0 +1,35 @@
...
+        """Disable validation of plan_id for one-time product purchase."""
+        recurring_orders = self.filtered(lambda order: order.order_line.filtered(
+            lambda line: line.product_id.recurring_invoice and not line.product_id.accept_one_time
+        ))
+        recurring_orders.has_recurring_line = True
+        (self - recurring_orders).has_recurring_line = False
```
kpt-odoo, line 14: Always try to make code readable unless the code you written is improve the performance here is suggestion 
```suggestion
        for order in self:
            order.has_recurring_line = any(
                line.product_id.recurring_invoice and not line.product_id.accept_one_time
                for line in order.order_line
            )
```
             
its more readable but i haven't tested it could you check it and apply it  accordingly?

one_time_sale/models/sale_order.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.depends('order_line', 'order_line.recurring_invoice')
+    def _compute_has_recurring_line(self):
+        """Disable validation of plan_id for one-time product purchase."""
+        recurring_orders = self.filtered(lambda order: order.order_line.filtered(
+            lambda line: line.product_id.recurring_invoice and not line.product_id.accept_one_time
```
kpt-odoo, line 11: Also see when order_line have one time product and we add plan on order then it'll break a flow can you check it too?

one_time_sale/models/sale_subscription_plan.py
```diff
@@ -0,0 +1,16 @@
+from odoo import api, fields, models
+
+
+class SaleSubscriptionPlan(models.Model):
+    _inherit = 'sale.subscription.plan'
+    _order = 'sequence, billing_period_value'
+
+    sequence = fields.Integer(compute='_compute_sequence', store=True)
```
kpt-odoo, line 8: Why we need to add sequence here this field and compute could you explain it ? i see its not used anywhere except in tours :roll_eyes:

kpt-odoo (changes requested):
I've reviewed your code and found a few issues during testing:
1.)  image
2.)  image
The second spec point is not working. When selecting a price from the dropdown, the price does not update.

- Adding a one-time product to the cart is also not working as expected.
I've added few small suggestion and few questions.

Thanks for work!
Good work here :+1:

one_time_sale/static/src/js/tours/one_time_sale.js
```diff
@@ -0,0 +1,130 @@
+"use_strict";
```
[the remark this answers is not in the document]
pcbe-odoo: It is not mandatory, but using this JavaScript directive enables strict mode, which helps make the code more robust and prevent the unexpected behavior.

one_time_sale/models/product_template.py
pcbe-odoo: I need this condition to determine whether a sales order (SO) contains a product that allows one-time sales. Based on this, I want to dynamically show or hide certain views in the template.

one_time_sale/models/sale_subscription_plan.py
pcbe-odoo: I added the sequence field and its corresponding compute method because the specification requires the recurring plans to be displayed in a specific order, such as months appearing after weeks, and years appearing after months.

This ensures the plans are presented in user-friendly sequence.

one_time_sale/models/sale_order.py
pcbe-odoo: I rewrote this function as it is already defined in the `sale_subscription` module (in `sale_order.py`). It is unnecessary to use the `@api.depends` decorator in this case, as the parent method is already decorated with `@api.depends`. This means that it will be automatically triggered when any of the dependent fields change, so explicitly defining dependencies in the child function is redundant.

one_time_sale/static/src/js/tours/one_time_sale.js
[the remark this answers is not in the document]
kpt-odoo: I see you are only using in tours but not in public widget files is there any reason we are not using this those js file and only in tours?

kpt-odoo (approved):

kpt-odoo:
@pcbe-odoo can you also fix this small styling issues - https://runbot.odoo.com/runbot/build/77903037

PR #623, opened 2025-03-17
vval - technical onboarding

vval-odoo:


estate/models/estate_property.py
```diff
@@ -0,0 +1,43 @@
...
+            ('new', 'New'),
+            ('offer_received', 'Offer Received'),
+            ('offer_accepted', 'Offer Accepted'),
+            ('sold', 'Sold'),
+            ('cancelled', 'Cancelled'),
+        ],
+        copy=False,
+        default='new',
+        required=True
+    )
```
dbeguin, line 43:
```suggestion
    state = fields.Selection([
        ('new', 'New'),
        ('offer_received', 'Offer Received'),
        ('offer_accepted', 'Offer Accepted'),
        ('sold', 'Sold'),
        ('cancelled', 'Cancelled'),
    ], copy=False, default='new', required=True, string="Status")
```
just a matter of coding style. to gain some lines without being unreadable :)

estate/models/estate_property.py
```diff
@@ -0,0 +1,43 @@
+from odoo import fields, models
+
+class EstateProperty(models.Model):
+    _name = "estate_property"
+    _description = "Real estate management"
+
+    active = True
```
dbeguin, line 7: what is it for ?
active is a "magical" field in odoo. it should be a real field (not a property)
so `active = fields.Boolean("Is Active?", default=T/F)`
And if False, record will not be returned by the ORM during search or accessing related model records.

estate/models/estate_property.py
```diff
@@ -0,0 +1,43 @@
...
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
```
dbeguin, line 10: Always set the label of the field. even if it's obvious and even if the orm will generate one for you using the field name.
`postcode = fields.Char("Postcode") or postcode = fields.Char(string="Postcode")`
-> better the first in this case, unless string is not the first param. (typically for relational fields)

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,82 @@
...
+                <field name="living_area" string="Living Area (sqm)"/>
+                <field name="expected_price"/>
+                <field name="selling_price"/>
+                <field name="date_availability" string="Available From"/>
+            </list>
+        </field>
+    </record>
+
+    <record id="estate_property_form_view" model="ir.ui.view">
+        <field name="name">real_estate.estate_property.form</field>
```
dbeguin, line 26: following guidelines : id should be the same as name, but using _ instead. (well there are some expections but this is documented in the guidelines :)

dbeguin (commented):
Good work already :)
some picky comments mainly regarding guidelines :)

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,77 @@
...
+                case 'accepted':
+                    raise exceptions.UserError('You cannot refuse an accepted offer')
+                case False:
+                    record.status = 'refused'
+        return True
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            self.env['estate.property'].browse(vals['property_id']).state = 'offer_received'
```
dbeguin, line 72: will crash if 'property_id' not in vals.
also, I would do that after the create.
so you can use the created record to access their properties

estate/models/res_users.py
```diff
@@ -0,0 +1,10 @@
+from odoo import fields, models
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    property_ids = fields.One2many('estate.property', 'salesperson_id',
+        string='Properties', domain=['|',
+            ('state', '=', 'new'),
+            ('state', '=', 'offer_received'),
```
dbeguin, line 9: state in [new, received]

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,12 @@
...
+<odoo>
+    <menuitem id="estate_menu_root" name="Real Estate">
+        <menuitem id="estate_menu_advertisements" name="Advertisements">
+            <menuitem id="estate_action_properties" action="action_estate_properties"/>
+        </menuitem>
+        <menuitem id="estate_menu_settings" name="Settings">
+            <menuitem id="estate_action_property_types" action="action_estate_property_types"/>
+            <menuitem id="estate_action_property_tags" action="action_estate_property_tags"/>
+        </menuitem>
+    </menuitem>
```
dbeguin, line 11: for the exercise: make one line per menuitems, not using encapsulation
but using parent_id instead.
this will help you when adding menuitems on bridge modules or modules depending on the module where you want to put a new menu item

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,43 @@
...
+<odoo>
+    <record id="action_estate_property_offer" model="ir.actions.act_window">
+        <field name="name">Property Offers</field>
+        <field name="res_model">estate.property.offer</field>
+        <field name="view_mode">list,form</field>
+        <field name="domain">[('property_type_id', '=', active_id)]</field>
+    </record>
+
+    <record id="estate_property_offer_view_list" model="ir.ui.view">
+        <field name="name">estate.property.offer.list</field>
```
dbeguin, line 11: estate.property.offer.view.list
same rule everywhere for views

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,43 @@
...
+                    <field name="price"/>
+                    <field name="partner_id"/>
+                    <field name="validity"/>
+                    <field name="date_deadline"/>
+                    <field name="status"/>
+                </group>
+            </form>
+        </field>
+    </record>
+</odoo>
```
dbeguin, line 43: missing empty line
check everywhere

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,149 @@
...
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area" filter_domain="[
+                        ('living_area', '>=', self),
+                ]"/>
+                <field name="facades" string="Facades"/>
+                <filter name="available" string="Availbale" domain="[
+                    '|',
+                        ('state', '=', 'new'),
+                        ('state', '=', 'offer_received'),
```
dbeguin, line 133: state in

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
...
+                fields.Command.create({
+                    'name': 'administrative fees',
+                    'quantity': 1,
+                    'price_unit': 100000,
+                })
+            ]
+            record.env['account.move'].create({
+                    'partner_id': record.buyer_id.id,
+                    'move_type': 'out_invoice',
+                    'invoice_line_ids': lines,
```
dbeguin, line 23: indentations

dbeguin (commented):
Zblip ! A set of remark for code polishing and commit message beautifulness
Enjoy!

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,77 @@
...
+    def action_accept_offer(self):
+        for record in self:
+            match record.property_id.state:
+                case 'new' | 'offer_received':
+                    for offer in record.property_id.offer_ids:
+                        offer.status = 'refused'
+                    record.property_id.buyer_id = record.partner_id
+                    record.property_id.selling_price = record.price
+                    record.property_id.state = 'offer_accepted'
+                    record.status = 'accepted'
```
dbeguin, line 53: so you write on that offer twice. as you set it to refuse, like all the others, then you reset it to accepted. 
We tend to not modifying twice the same thing during the same process.
Either filter offer_ids to remove the current one, either you don't do refuse here but make the satus a computed stored field, that depends on property.offer_ids. where they all become refused one one of them becomes accepted.

estate/models/estate_property.py
```diff
@@ -0,0 +1,125 @@
...
+
+    @api.onchange('garden')
+    def _onchange_garden(self):
+        self.garden_area = 10 if self.garden else 0
+        self.garden_orientation = 'north' if self.garden else None
+
+    def action_mark_canceled(self):
+        for record in self:
+            if record.state == 'sold':
+                raise exceptions.UserError('You cannot mark a sold property as canceled')
```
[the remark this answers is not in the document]
vval-odoo: Check placed in write instead of constrain

estate_account/models/estate_property.py
dbeguin, line 23: create a dict of values to create , then create outside of the loop.
Never search / create in loop (if possible of course)

estate/models/estate_property.py
```diff
@@ -0,0 +1,106 @@
...
+        self.garden_area = 10 if self.garden else 0
+        self.garden_orientation = 'north' if self.garden else None
+
+    def write(self, vals):
+        origin = self._origin.state
+        current = vals.get('state')
+        if origin == 'canceled' and current == 'sold':
+            raise exceptions.UserError('You cannot mark a canceled property as sold')
+        elif origin == 'sold' and current == 'canceled':
+            raise exceptions.UserError('You cannot mark a sold property as canceled')
```
dbeguin, line 89: user errors should be translated. use _("yourtext") to allow translations

PR #646, opened 2025-03-19
[IMP] pricelist_refactor: Support discounts and formulas in pricelists for subscription and rental products

vrgo-odoo:
**Description of the feature this PR addresses:**
- Enhances pricelist handling by introducing discount and formula support for subscription and rental products.

**Current behavior before PR:**
1.Subscription and rental products did not support discount and formula-based pricing in pricelists.
2.No dedicated pricing page existed for rental products in the product form view.

**Desired behavior after PR** 
1.Discount and formula is added for subscription and rental products in pricelists.
2.A new dedicated pricing page is added to the product form for rental and subscription pricing.

### DIFF
--- a/pricelist_refactor/models/product_pricelist.py
+++ b/pricelist_refactor/models/product_pricelist.py
@@ -29,7 +29,7 @@ class ProductPricelist(models.Model):
         ]
     )
 
-    rental_pricelist_ids = fields.One2many(
+    product_rental_pricelist_ids = fields.One2many(
         comodel_name='product.pricelist.item',
         inverse_name='pricelist_id',
         string="Rental Pricing Rules",
@@ -50,7 +50,6 @@ class ProductPricelist(models.Model):
         currency.ensure_one()
         if not products:
             return {}
-
         if not date:
             date = fields.Datetime.now()
         results = {}
@@ -80,7 +79,12 @@ class ProductPricelist(models.Model):
                     price, currency, self.env.company, date
                 ), pricelist_id.id
             elif product.recurring_invoice:
-                results[product.id] = self.env['product.pricelist.item']._get_first_suitable_recurring_pricing(product, plan=plan_id, pricelist=self)._compute_price(product, quantity, uom, date, plan_id=plan_id), self.env['product.pricelist.item']._get_first_suitable_recurring_pricing(product, plan=plan_id, pricelist=self).id
+                results[product.id] = self.env['product.pricelist.item']._get_first_suitable_recurring_pricing(
+                    product, plan=plan_id, pricelist=self
+                )._compute_price(product, quantity, uom, date, plan_id=plan_id), 
+                self.env['product.pricelist.item']._get_first_suitable_recurring_pricing(
+                    product, plan=plan_id, pricelist=self
+                ).id
         price_computed_products = self.env[products._name].browse(results.keys())
         return {
             **results,
--- a/pricelist_refactor/models/product_pricelist.py
+++ b/pricelist_refactor/models/product_pricelist.py
@@ -1,0 +1,97 @@
+from odoo import fields, models
+
+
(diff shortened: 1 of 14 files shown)


pricelist_refactor/models/product_pricelist.py
```diff
@@ -0,0 +1,112 @@
...
+    _inherit = 'product.pricelist'
+
+    item_ids = fields.One2many(
+        comodel_name='product.pricelist.item',
+        inverse_name='pricelist_id',
+        string="Pricing Rules",
+        domain=[
+            '&',
+            '|', ('product_tmpl_id', '=', None), ('product_tmpl_id.active', '=', True),
+            '|', ('product_id', '=', None), ('product_id.active', '=', True),
```
kpt-odoo, line 14: this line is repeated many times we could do something about it as its not good to repeat same code DRY principle.

pricelist_refactor/models/product_template.py
```diff
@@ -0,0 +1,94 @@
+import math
+from odoo import fields, models
+
+PERIOD_RATIO = {
```
kpt-odoo, line 4: this is repeated too could you find a way to not repeat this ?

pricelist_refactor/models/product_template.py
```diff
@@ -0,0 +1,94 @@
...
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    pricelist_rule_ids = fields.One2many(
+        string="Pricelist Rules",
+        comodel_name='product.pricelist.item',
+        inverse_name='product_tmpl_id',
+        domain=lambda self: [
+            '|',
+            ('product_tmpl_id', 'in', self.ids),
```
kpt-odoo, line 22: same repeated code

pricelist_refactor/models/product_pricelist.py
```diff
@@ -0,0 +1,112 @@
...
+                ('plan_id', '=', plan_id.id),
+                '|', ('categ_id', '=', False), ('categ_id', 'parent_of', products.categ_id.ids),
+                '|', ('product_tmpl_id', '=', False), templates_domain,
+                '|', ('product_id', '=', False), products_domain,
+                '|', ('date_start', '=', False), ('date_start', '<=', date),
+                '|', ('date_end', '=', False), ('date_end', '>=', date),
+            ]
+        return super()._get_applicable_rules_domain(products, date, **kwargs)
+
+    def _enable_rental_price(self, start_date, end_date):
```
kpt-odoo, line 110: If this method is already in sale_renting why not use that instead creating this one?

pricelist_refactor/models/product_pricelist.py
```diff
@@ -0,0 +1,112 @@
...
+
+        price_computed_products = self.env[products._name].browse(results.keys())
+        return {
+            **results,
+            **super()._compute_price_rule(
+                products - price_computed_products, quantity, currency=currency, date=date, **kwargs
+            ),
+        }
+
+    def _get_applicable_rules_domain(self, products, date, plan_id=None, **kwargs):
```
kpt-odoo, line 90: This method already exists in the product. Instead of repeating the whole code, why not just update the return values from super? We can simply insert ('plan_id', '=', plan_id.id), at the desired position in the returned list. Also check other methods also if they can be improved.

kpt-odoo (changes requested):
Hello, I have added my review.

- I noticed a lot of repeated code, which is not a good practice. It should be optimized to avoid duplication.
- Each module should have its own bridge instead of being tightly connected.
For example, if only the subscription module is installed, shouldn't we still allow adding discounts and formulas in the price list, even without rental?
- The eCommerce module seems to be broken. As i am unable to add a product to the cart. Can you check this issue?
- you seem to override whole method (code) instead it can be simply done if you need to add something in parent methods code check if they really necessary.

PR #654, opened 2025-03-20
[IMP] last_ordered_products: improved sale and purchase by adding this module

ppch-odoo:
- In product_product and product_template override method name_search to get
  last ordered time for selected customer.
- Implemented functions to get last_sold_products and last_purchased_products
  based on customer's invoice and vendor's bill.
- Added compute method to get last_order_time and time difference string.
- Updated context in sale_order, purchase_order and account_move to get
  order_type and partner_id in context
- In catalog, to re-organize displayed information inherited sale_order and
  sale_order_line, override some methods to get sale_uom of order line
- In catalog, to re-organize displayed information inherited
  product_view_kanban_catalog view
- Extend ProductCatalogOrderLine to get sale_uom in catalog
- Patch ProductCatalogKanbanRecord to get information as per the requirement
- Extend Autocomplete to get time difference string in product dropdown
- Patch ProductLabelSectionAndNoteFieldAutocomplete to override function
  which will be used to get time difference string
- In product catalog also products will be sorted based on invoice creation date
- testcases added

### DIFF
--- a/last_ordered_products/models/product_product.py
+++ b/last_ordered_products/models/product_product.py
@@ -1,0 +1,153 @@
+from datetime import datetime
+from odoo import api, fields, models
+from odoo.osv import expression
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    last_order_time = fields.Datetime(compute='_compute_last_order_time')
+    last_date_str = fields.Char(compute='_compute_last_order_time')
+
+    @api.depends_context('order_id')
+    def _compute_last_order_time(self):
+        """Compute the last order time for each product based on the latest sale or purchase."""
+
+        order_type = False
+        if self.env.context.get('active_model') == 'sale.order.line':
+            partner_id = self.env['sale.order'].browse(self.env.context.get('order_id')).partner_id.id
+            order_type = 'sale'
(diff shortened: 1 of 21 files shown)


last_ordered_products/models/product_product.py
```diff
@@ -0,0 +1,132 @@
...
+            partner_id = self.env['purchase.order'].browse(self.env.context.get('order_id')).partner_id.id
+            order_type = 'purchase'
+        elif self.env.context.get('active_model') == 'account.journal':
+            active_id = self.env.context.get('active_id')
+            if active_id:
+                order_type = self.env['account.journal'].browse(active_id).type
+            partner_id = self.env.context.get('partner_id') or self.env.context.get('default_partner_id')
+        else:
+            partner_id = self.env.context.get('partner_id')
+            order_type = self.env.context.get('order_type')
```
kcv-odoo, line 30: We should not rely on context too much instead add something robust to compute these fields

kcv-odoo (changes requested):
Thank you for your work!!

Code look fragile and can be easily break. We should not rely that much on context instead we should add something robust and context should be last option and we should not make feature that can be used individually by single app in dependent of multiple main apps

Have a nice day!

PR #659, opened 2025-03-20
[ADD] lead_form: integrate lead form extensions for lead generation

pdso-odoo:
After this commit, lead form extensions from Google & Facebook Ads will be integrated via webhooks to auto-generate leads in the CRM module. Leads will be created based on incoming webhook data with source tracking. Lead data will be stored and attached to the chatter for reference.


pdso-odoo:
Landed in https://leads.odoo.co.in.

PR #710, opened 2025-04-14
[ADD] product_kit: Add option to set as kit

sris-odoo:


dhrs-odoo:
Hello @sris-odoo 
given changes

PR #711, opened 2025-04-14
[ADD] sale: Add `Zero Stock Approvalexchttps` field to allow SO confi…

sris-odoo:
…rmation

In this commit
-Added a new boolean field in Sale Order.
-If set to true, it allows sales users to confirm the sale order. -Field is read-only for users in the Sales/User group. -Editable only by users in the Sales/Administrator group.

### DIFF
--- a/zero_stock_blockage/__init__.py
+++ b/zero_stock_blockage/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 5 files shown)


dhrs-odoo:
Hello @sris-odoo 
Given changes and PR title should be proper

PR #713, opened 2025-04-16
[ADD] event: Limit max ticket per registration feature

sris-odoo:


dhrs-odoo:
Hello @sris-odoo 
Given changes and update PR message

PR #771, opened 2025-05-19
18.0 training gaso

SouGab:


estate/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+{
+    'name': "estate",
+
+    'description': """
+        Module to deal with estate sales"
+    """,
+
+    'author': "Gaso",
```
mathcoutant, line 9: Just a random comment to let you know that you can't do that in the official repo.
```suggestion
    'author': 'Odoo S.A.',
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @api.depends("garden_area", "living_area")
+    def _compute_area(self):
+        for record in self:
+            record.total_area = record.garden_area + record.living_area
+
+    @api.depends("property_offer_ids.price")
+    def _compute_best_offer(self):
+        for record in self:
+            record.best_offer_price = max(p.price for p in record.property_offer_ids)
```
mathcoutant: What if there are no offer?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @api.depends("validity")
+    def _compute_deadline(self):
+        for record in self:
+            computation_date = record.create_date if record.create_date else fields.date.today()
+            record.date_deadline = computation_date + relativedelta(days=record.validity)
+
+    def _inverse_deadline(self):
+        for record in self:
+            record.date_deadline = (record.date_deadline - record.create_date).days
```
mathcoutant: You should also check `if record.create_date` there in case the inverse method is called before creation.

mathcoutant (commented):
Very good! Only small changes.

Also your commit titles don't follow the [guidelines](https://www.odoo.com/documentation/18.0/fr/contributing/development/git_guidelines.html). They should have a specific title specifying the module concerned by the change. In this case it's always `estate` but it's a good habit you will have to learn.

And consider checking the style errors raised by the runbot to fix them.

estate/models/estate_property.py
mathcoutant: You just fixed it...

mathcoutant (commented):
Left some comments, some important, some less.
I didn't propose the correct diff, you have time to search for examples in odoo repo ;)

I noticed you sometimes changed the functional behavior the tutorial asks for, I'm ok with it, but I did some comments before realizing.

PR #773, opened 2025-05-19
[ADD] estate: create the Real State application which is a part of training.

mma1377:
This is a part of on-boarding's training. Chapter 2: A new Application and the goal is to make a estate module.
This module is developed to implement the real estate functionality. Using this app, one can create an estate to sell, configure properties e.g. expected price; then be able to add offers on the estate, and choose the buyer from the list of offers.

### DIFF
--- a/estate/__init__.py
+++ b/estate/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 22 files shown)


barracudapps (changes requested):
Good job globally!
Some nitpicking comments though 😉

PR #776, opened 2025-05-19
[ADD] Tutorial estate model

prdu-odoo:


prdu-estate/models/EstateModel.py
```diff
@@ -0,0 +1,22 @@
+from odoo import fields, models
+from dateutil.relativedelta import relativedelta
+
+
+class EstateModel(models.Model):
+    _name = "test_estate_model"
```
amah-odoo, line 6: we always keep the class name and the _name consistent and the class name is camel case while _name is snake case with `.` instead of `_`
```suggestion
class EstateProperty(models.Model):
    _name = "estate.property"
```

PR #778, opened 2025-05-19
[ADD] Estate: Adding the new Real Estate Module's directory.

karygauss03:
In this PR, I initialized the basic structure of the new Estate module directory (manifest and init files).
This work is related to the second chapter of the server framework 101 official documentation by Odoo.


estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,83 @@
+<?xml version="1.0"?>
```
amah-odoo, line 1: For the xmlids it just miss the word view before the view type.
https://www.odoo.com/documentation/18.0/contributing/development/coding_guidelines.html#security-view-and-action

estate/models/estate_property.py
```diff
@@ -0,0 +1,34 @@
+from odoo import models, fields
+from dateutil.relativedelta import relativedelta
+
+
+class EstateModel(models.Model):
```
amah-odoo, line 5: try to keep python class name consistent with the `_name` as it would be used in future
```suggestion
class EstateProperty(models.Model):
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,34 @@
...
+    _name = "estate_property"
```
amah-odoo, line 6: in the `_name` actually we use `.` instead of `_`

amah-odoo (commented):
Great job!! Thanks for your work 🚀
Just couple of tiny comments

estate/views/estate_property_views.xml
karygauss03: Fixed everywhere.

estate/models/estate_property.py
karygauss03: Fixed.

estate/models/estate_property.py
karygauss03: Fixed everywhere.

amah-odoo (commented):
Great work 🚀
one tiny nitpicking which is the xml files does not follow an defined order like search, list, form, action.

estate/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+from . import (
+    estate_property,
+    estate_property_offer,
+    estate_property_tag,
+    estate_property_type,
+    res_users,
+)
```
amah-odoo, line 7: I found that some apps are using this import style but its more common with tests package only.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,72 @@
...
+            if property.state == 'new':
+                property.state = 'offer received'
+            if any(offer.price >= vals['price'] for offer in property.offer_ids):
+                raise UserError("Cannot create offer: there is already an offer with an equal or higher price.")
+        return super().create(vals_list)
+
+    def _inverse_date_deadline(self):
+        for offer in self:
+            create_date = offer.create_date or fields.Datetime.now()
+            offer.validity = (offer.date_deadline - create_date.date()).days if offer.date_deadline else 0
```
amah-odoo, line 46: I think its better to keep `_inverse` function after its `_compute`

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,17 @@
...
+
+    name = fields.Char(string='Property Type', required=True)
+    sequence = fields.Integer('Sequence', default=1, help='Used to order offer types. Lower is better.')
+    property_ids = fields.One2many('estate.property', 'property_type_id', string='Properties')
+    offer_ids = fields.One2many('estate.property.offer', 'property_type_id', string='Properties offers')
+    offer_count = fields.Integer(compute='_compute_offer_count', string='Offers Count')
+
+    def _compute_offer_count(self):
+        for record in self:
+            record.offer_count = len(record.offer_ids)
```
amah-odoo, line 17: I think you forgot the `api.depends` here

estate/models/estate_property.py
```diff
@@ -0,0 +1,98 @@
...
+    description = fields.Char('Description')
+    post_code = fields.Char('Post Code')
+    state = fields.Selection(
+        string="State",
+        selection=[('new', 'New'), ('offer received', 'Offer Received'), ('offer accepted', 'Offer Accepted'), ('sold', 'Sold'), ('canceled', 'Canceled')],
+        help="Estate current state",
+        required=True,
+        default='new',
+        copy=False
+    )
```
amah-odoo, line 26: I prefer this format more and its the common one for selection list. Also, we use snack_case for the selection values
```suggestion
        selection=[
            ('new', 'New'), 
            ('offer_received', 'Offer Received'), 
            ('offer_accepted', 'Offer Accepted'), 
            ('sold', 'Sold'), 
            ('canceled', 'Canceled')],
        help="Estate current state", required=True, default='new', copy=False)
```

estate/views/res_users_views.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="users_form_inherit_estate_view_form" model="ir.ui.view">
```
amah-odoo, line 3: This xmlid does not follow the guideline

estate_account/models/estate_account.py
```diff
@@ -0,0 +1,6 @@
+from odoo import models
+
+
+class EstateAccount(models.Model):
+    _name = 'estate.account'
+    _description = 'Estate Account Model (for invoinving).'
```
amah-odoo, line 6: If it does not have fields So why we created it?

amah-odoo (commented):
Thanks for your work 🤩
Great job just some tiny comments

PR #781, opened 2025-05-19
test

maedekey:
pr


estate_invoice/models/estate_invoice_model.py
```diff
@@ -0,0 +1,32 @@
...
+                        'quantity': 0.06,
+                        'price_unit': selling_price
+                    }),
+                    Command.create({
+                        'name': 'Administrative fees',
+                        'quantity': 1,
+                        'price_unit': 100.00
+                    })
+                ]
+            })
```
ushyme, line 31:
```suggestion
        for property in self.filtered("buyer"):
            accepted_offer = property.offer_ids.filtered(lambda o: o.status == "Accepted")
            selling_price = max(property.expected_price, accepted_offer.price) if accepted_offer else property.expected_price

            self.env['account.move'].create({
                'partner_id': property.buyer.id,
                'move_type': 'out_invoice',
                'invoice_line_ids': [
                    Command.create({
                        'name': '6% of Selling price',
                        'quantity': 0.06,
                        'price_unit': selling_price,
                    }),
                    Command.create({
                        'name': 'Administrative fees',
                        'quantity': 1,
                        'price_unit': 100.00,
                    }),
                ]
            })```

File: Estate/models/real_estate_model.py [hunk 2]
```diff
@@ -0,0 +1,102 @@
...
+        if self.offer_ids:
+            self.state = 'offer_received'
+        else:
+            self.state = 'new'
+
+    @api.ondelete(at_uninstall=False)
+    def _ondelete(self):
+        for record in self:
+            if record.state != 'new' and record.state != 'cancelled':
+                raise UserError("Only new and cancelled properties can be deleted.")
```
ushyme, line 86:
```suggestion
                raise UserError(self.env._("Only new and cancelled properties can be deleted."))
```
In Odoo, translatable strings in "data"-type content are automatically exported for translation. However, with other user-facing strings, Odoo cannot automatically know whether they are translatable. To ensure these strings are exported for translation, you need to explicitly mark them. This is achieved by wrapping the literal string with the `recordset.env._()` function. The `_()` function signals to Odoo's translation system that the string should be included in the translation catalog.
More info here: https://www.odoo.com/documentation/18.0/developer/howtos/translations.html#explicit-exports

Note: I'm only making the suggestion here, but this applies to other places in your code

ushyme (changes requested):
Good work overall. Here are a few general remarks:
* The module name should be in lowercase and use underscores to separate words. Rename `Estate` directory to `estate`.
* for model file names, no need to add the `_model` suffix. By convention, we use the snake_case version of the model name. ex: `estate.property` => `estate_property.py`

PR #788, opened 2025-05-19
[ADD] estate module

thha-odoo:


ushyme (changes requested):
Good work overall 👍🏾. Here are a few things that can be improved:
* IDE-specific configurations are user-dependent and should not be included in the codebase. So your `.idea` folder should not be here.
* we generally use different quote signs to distinguishes between technical strings (single quotes) and user-facing strings (double quotes).
* be mindful of your commit messages; more specifically, the commit title needs to follow a [specific structure](https://www.odoo.com/documentation/18.0/contributing/development/git_guidelines.html)
* Remember to leave an empty line at the end of files. While it seems trivial, it ensures compatibility with many tools and version control systems.

PR #804, opened 2025-05-30
18.0 owl tutorials abye

A-Yehia19:


awesome_owl/static/src/card/card.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
+<templates xml:space="preserve">
+
+    <t t-name="awesome_owl.card">
+
+        <div class="card d-inline-block m-2" style="width: 18rem;">
```
barracudapps, line 6: Still avoid hardcoding styles 😉

awesome_owl/static/src/counter/counter.js
```diff
@@ -0,0 +1,25 @@
...
+
+    setup() {
+        this.count = useState({ value: 0 });
+    }
+
+    increment() {
+        this.count.value += 1;
+        if (this.props.callback) {
+            this.props.callback();
+        }
```
barracudapps, line 23:
```suggestion
        this.props.callback?.();
```
This is an easier way to write down the check before the call if exists.

PR #811, opened 2025-06-06
18.0 tutorial importable modules aras

abdrahmanrashed:


estate/data/estate_property.xml
```diff
@@ -0,0 +1,191 @@
...
+        <field name="ttype">many2one</field>
+        <field name="relation">res.partner</field>
+    </record>
+
+    <record id="field_estate_property_user_id" model="ir.model.fields">
+        <field name="model_id" ref="estate.model_estate_property" />
+        <field name="name">x_user_id</field>
+        <field name="field_description">Salesperson</field>
+        <field name="ttype">many2one</field>
+        <field name="relation">res.partner</field>
```
vava-odoo, line 144: Salesperson is a user, no?

estate/data/estate_property.xml
```diff
@@ -0,0 +1,191 @@
...
+    </record>
+
+    <record id="field_estate_property_garden_orientation_west" model="ir.model.fields.selection">
+        <field name="field_id" ref="field_estate_property_garden_orientation" />
+        <field name="name">West</field>
+        <field name="value">west</field>
+    </record>
+
+    <record id="field_estate_property_proprty_type_id" model="ir.model.fields">
+        <field name="model_id" ref="estate.model_estate_property" />
```
vava-odoo, line 122: `ref` to an id of the same module works without having to specify the complete xmlid
```suggestion
        <field name="model_id" ref="model_estate_property" />
```
This is not true when dealing with `env.ref`

estate/data/estate_property.xml
```diff
@@ -0,0 +1,191 @@
...
+        <field name="ttype">one2many</field>
+        <field name="relation">x_estate.property.offer</field>
+        <field name="relation_field">x_property_id</field>
+    </record>
+
+    <record id="field_estate_property_total_area" model="ir.model.fields">
+        <field name="model_id" ref="estate.model_estate_property" />
+        <field name="name">x_total_area</field>
+        <field name="field_description">Total Area</field>
+        <field name="ttype">float</field>
```
vava-odoo, line 168: Since both areas are integers, you may want to stick with integers
```suggestion
        <field name="ttype">integer</field>
```

estate/data/estate_property_offer.xml
```diff
@@ -0,0 +1,64 @@
...
+        <field name="model_id" ref="model_estate_property_offer"/>
+        <field name="name">x_action_accept_offer</field>
+        <field name="state">code</field>
+        <field name="code"><![CDATA[
+for offer in records:
+    offer.write({'x_status': 'accepted'})
+    offer.x_property_id.write({'x_selling_price': offer.x_price, 'x_partner_id': offer.x_partner_id})
+    for rejected_offer in offer.x_property_id.x_property_offer_ids:
+        if rejected_offer != offer:
+            rejected_offer.write({'x_status': 'refused'})
```
vava-odoo, line 61: Ok, but what if you accept two offers for the same property? It seems like only one will be accepted, but it is kind of magic, no? Shouldn't we take more care (and maybe raise in such cases)?

estate/data/estate_property.xml
```diff
@@ -0,0 +1,191 @@
...
+        <field name="name">x_country_id</field>
+        <field name="field_description">Buyer's Country</field>
+        <field name="ttype">many2one</field>
+        <field name="relation">res.country</field>
+        <field name="related">x_partner_id.country_id</field>
+    </record>
+
+    <record id="field_estate_property_api_published" model="ir.model.fields">
+        <field name="model_id" ref="estate.model_estate_property" />
+        <field name="name">x_api_published</field>
```
vava-odoo, line 187: Did you find out how to set a default value for this field?

estate/security/estate_security.xml
```diff
@@ -0,0 +1,20 @@
...
+        <field name="perm_write" eval="False"/>
+        <field name="active" eval="True"/>
+    </record>
+    
+    <record id="estate_property_public_read_record_rule" model="ir.rule">
+        <field name="name">Public user read rule</field>
+        <field name="model_id" ref="model_estate_property"/>
+        <field name="domain_force">[('x_api_published','=',True)]</field>
+        <field name="groups" eval="[(4, ref('base.group_public'))]"/>
+        <field name="perm_read" eval="True"/>
```
vava-odoo, line 17: By default the rule applies to all perm_* fields, so this one is not mandatory (and somehow misleading, because you are giving other rights too)
```suggestion
```

estate/static/src/js/tour.js
```diff
@@ -0,0 +1,10 @@
+import { registry } from "@web/core/registry";
+
+
+registry.category("web_tour.tours").add('estate_tour', {
+    url: "/web",
```
vava-odoo, line 5: url is now defined in the xml web_tour (and should point to /odoo). But since this is the default value, you don't have to :-)
```suggestion
```

estate/static/src/js/tour.js
```diff
@@ -0,0 +1,10 @@
...
+    steps: () => [{
+    trigger: '.o_app[data-menu-xmlid="estate.menu_root"]',
+    content: 'Start selling your properties from this app!',
```
vava-odoo, line 8: What if you want to translate the string here?

estate/data/estate_tour.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<odoo>
```
vava-odoo, line 3: User probably don't want this tour to show up anytime they upgrade, so setting this as noupdate seems appropriate

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,78 @@
...
+                                <field name="x_garden_area" invisible="not x_garden"/>
+                                <field name="x_garden_orientation" invisible="not x_garden"/>
+                                <field name="x_total_area"/>
+                            </group>
+                        </page>
+
+                        <page string="Offers">
+                            <group>
+                                <field name="x_property_offer_ids"/>
+                            </group>
```
vava-odoo, line 63: Probably better without a group here
```suggestion
                            <field name="x_property_offer_ids"/>
```

estate/data/estate_property.xml
```diff
@@ -0,0 +1,191 @@
...
+        <field name="name">x_property_offer_ids</field>
+        <field name="field_description">Offers</field>
+        <field name="ttype">one2many</field>
+        <field name="relation">x_estate.property.offer</field>
+        <field name="relation_field">x_property_id</field>
+    </record>
+
+    <record id="field_estate_property_total_area" model="ir.model.fields">
+        <field name="model_id" ref="estate.model_estate_property" />
+        <field name="name">x_total_area</field>
```
vava-odoo, line 166: Do you need to store that field?

PR #814, opened 2025-06-17
Chapter 2

baje-odoo:


estate/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
bso-odoo: We try to always have the return at the end of the last line of every file.
This way, when new lines are added later, the formerly last line does not show up as being impacted in the diff.

estate/models/estate_property.py
```diff
@@ -0,0 +1,23 @@
...
+    selling_price = fields.Float("Selling Price")
+    bedrooms = fields.Integer("Bedrooms")
+    living_area = fields.Integer("Living Area ")
+    facades = fields.Integer("Facades")
+    garage = fields.Boolean("Garage")
+    garden = fields.Boolean("Garden")
+    garden_area = fields.Integer("Garden Area")
+    garden_orientation = fields.Selection(
+        string='Garden Orientation',
+        selection=[('north','North'),('south','South'),('east','East'),('west','West')]
```
bso-odoo: On the python-side, we try to follow the following for quoting strings:
- single quote for technical strings
- double-quote for human-displayed strings

=> `('north', "North")`

Let's try to follow it during the training - even though you'll find many counter examples in existing code.

estate/models/estate_property.py
```diff
@@ -0,0 +1,23 @@
...
+    _order = "id"
+
+    name = fields.Char("Title", required=True)
+    description = fields.Text("Description")
+    postcode = fields.Char("Postcode")
+    date_availability = fields.Date("Available From")
+    expected_price = fields.Float("Expected Price", required=True)
+    selling_price = fields.Float("Selling Price")
+    bedrooms = fields.Integer("Bedrooms")
+    living_area = fields.Integer("Living Area ")
```
bso-odoo: Final space is not needed

bso-odoo (commented):
Here are a few cosmetic comments.

estate/__manifest__.py
```diff
@@ -2,7 +2,7 @@
     'name': "estate",
     'version': '0.1',
     'depends': ['base'],
-    'data':[
```
bso-odoo: As a git exercise, try to make the small typo changes inside the commits where the line was initially introduced.
You can look into the `git rebase -i` command to see how to rewrite your branch's history.
You will need to use `git push --force-with-lease` to send a rewritten version of your branch. (Try to avoid using `git push --force` to avoid accidents in situations where you work on a common branch with other people)

bso-odoo:
Actually, the `[ADD]` prefix in commit titles is only when adding a new module, we use the prefix `[IMP]` for improvement when adding stuff inside a module, and `[FIX]` when making a fix on an existing feature.

estate/__manifest__.py
```diff
@@ -3,9 +3,13 @@
     'version': '0.1',
     'depends': ['base'],
     'data':[
+<<<<<<< HEAD
         'security/ir.model.access.csv',
         'views/estate_property_views.xml',
         'views/estate_menus.xml',
+=======
+        'security/ir.model.access.csv'
+>>>>>>> 6f53f08 (Security)
```
bso-odoo: It seems there is an unresolved git conflict here.

estate/models/estate_property.py
```diff
@@ -13,11 +13,11 @@ class EstateProperty(models.Model):
     expected_price = fields.Float("Expected Price", required=True)
     selling_price = fields.Float("Selling Price", readonly=True, copy=False)
     bedrooms = fields.Integer("Bedrooms", default=2)
-    living_area = fields.Integer("Living Area")
+    living_area = fields.Integer("Living Area (sqm)")
     facades = fields.Integer("Facades")
     garage = fields.Boolean("Garage")
     garden = fields.Boolean("Garden")
-    garden_area = fields.Integer("Garden Area")
+    garden_area = fields.Integer("Garden Area (sqm)")
```
bso-odoo: Maybe these changes can be done in the commit where those lines were initially added.

estate/__manifest__.py
```diff
@@ -2,19 +2,15 @@
...
         ],
     'author': "baje",
     'category': 'Uncategorized',
     'description': """
     An app to manage a Real Estate Agency
     """,
     'application': True,
-}
\ No newline at end of file
+}
```
bso-odoo: Oh ok, it is fixed here. Let's try to make such fixes in their original commit - so that it looks like the problem just never happened.

estate/views/estate_property_views.xml
```diff
@@ -5,6 +5,48 @@
...
+        <field name="name">estate.property.form</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form>
+                <sheet>
+                    <div class="oe_title">
+                        <h1>
+                            <field name="name"/>
+                        </h1>
+                        <field name="tag_ids" widget="many2many_tags" options="{'color_field': 'color'}"/>
```
bso-odoo: The is no `tag_ids` field (yet) at this point.

estate/security/ir.model.access.csv
```diff
@@ -1,2 +1,3 @@
 id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
-access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
\ No newline at end of file
+access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
+access_estate_property_type,access_estate_property_type,model_estate_property_type,base.group_user,1,1,1,1
```
bso-odoo: Try to introduce a return on the last line on the first version of each file in order to avoid such diffs. (This change should typically only be about adding one new line)

estate/models/estate_property_type.py
```diff
@@ -7,4 +7,5 @@ class EstatePropertyType(models.Model):
     _description = "estate property type"
     _order = "id"
 
-    name = fields.Char("Name", required=True)
\ No newline at end of file
+    name = fields.Char("Name", required=True)
+
```
bso-odoo: The warning is still there... I guess your last line is not empty... maybe it contains a few spaces. (This happens in many files)

bso-odoo (commented):
Here are a few comments:

estate/models/estate_property.py
```diff
@@ -95,3 +95,8 @@ def _check_selling_price(self):
         for record in self:
             if not float_is_zero(record.selling_price, precision_rounding=0.01) and float_compare(record.selling_price, record.expected_price * 0.9, precision_rounding=0.01) < 0:
                 raise ValidationError("the selling price cannot be lower than 90% of the expected price")
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_if_state_is_new_or_cancelled(self):
+        if any(record.state not in ('new', 'canceled') for record in self):
+            raise UserError("Only new and canceled properties can be deleted.")
```
bso-odoo: ~~Those strings are not implicitly translated, they have to be explicitly translated using `_` (`from odoo import _`)~~

Oh, you actually do it in further commits. This is also a reason to include the fix inside the original commit. :wink:

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.languageServer": "None"
```
bso-odoo, line 2: Of course, on real PRs, such files wouldn't be included.

bso-odoo (commented):
Not much to say.
There are remaining `'` vs `"`, some extra unneeded files, etc... and fixes in late commits instead of "since the start".
But better continue with further training, if you have time later and want to practice more git cleanups, that's always possible too.

awesome_dashboard/static/src/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,46 @@
...
+    setup(){
+        this.canvasRef = useRef("canvas");
+        onWillStart (() => loadJS("/web/static/lib/Chart/Chart.js"));
+        onMounted(() => {
+            this.renderChart();
+        });
+        onWillUpdateProps(() => {
+            this.chart.destroy();
+            this.renderChart();
+        })
```
bso-odoo: At this stage, `this.props` are the old values. `onWillUpdateProps`'s function receives the new values as parameter.

PR #817, opened 2025-06-18
18.0 training mdoy

martiyen:


estate/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
bso-odoo: We try to include the return character on the last line in every file.
This way when adding more lines later, the previously last line is not impacted in the diff.

estate/models/estate_property.py
```diff
@@ -0,0 +1,22 @@
...
+    selling_price = fields.Float("Selling Price", readonly=True, copy=False)
+    bedrooms = fields.Integer("Bedrooms", default=2)
+    living_area = fields.Integer("Living Area (sqm)")
+    facades = fields.Integer("Facades")
+    garage = fields.Boolean("Garage")
+    garden = fields.Boolean("Garden")
+    garden_area = fields.Integer("Garden Area (sqm)")
+    garden_orientation = fields.Selection(
+        string='Type',
+        selection=[('north', 'North') ,('south', 'South'), ('east', 'East'), ('west', 'West')],
```
bso-odoo: In python, we try to follow the following convention for quoting strings:
- single quote for technical terms
- double quotes for displayed texts

=> `('north', "North")`

Let's try to apply it during this exercise, even though the existing codebase contains many counter examples.

estate/models/estate_property.py
```diff
@@ -2,8 +2,8 @@
```
bso-odoo, line 4: Commit messages are supposed to be truncated at 72 columns, so, this one is bit too long.
But after a blank line you can put as much text as you want.

Anyway, do not worry about that, instead let's take this opportunity for some git training.
Try to make these fixes directly within the commits where the lines were introduces.
Look at the `git rebase -i` command to rewrite the history on your branch, and then you'll need to `git push --force-with-lease` to send such a change to the repo. (Try not to use `git push --force` to avoid accidents when you will be working with other people on a common branch)

estate/models/__init__.py
```diff
@@ -1 +1 @@
-from . import estate_property
+from . import estate_property, estate_property_type, estate_property_tag, estate_property_offer
```
bso-odoo: We try to have imports in `__init__.py` files separated on distinct lines.
Always the same reason, to reduce the diff when making future changes.

estate/models/estate_property.py
```diff
@@ -55,3 +58,18 @@ def _onchange_garden(self):
         else:
             self.garden_area = 0
             self.garden_orientation = False
+
+    def action_set_sold(self):
+        for record in self:
+            if self.state == 'cancelled':
+                raise exceptions.UserError('Property is cancelled and cannot be sold.')
```
bso-odoo: Those strings are not implicitly translated, you have to explicitly use `_`. (`from odoo import _`)

estate/models/estate_property.py
```diff
@@ -62,14 +68,24 @@ def _onchange_garden(self):
...
-                raise exceptions.UserError('Property is cancelled and cannot be sold.')
+                raise exceptions.UserError("Property is cancelled and cannot be sold.")
 
             record.state = 'sold'
 
     def action_set_cancelled(self):
         for record in self:
             if self.state == 'sold':
-                raise exceptions.UserError('Sold properties cannot be cancelled.')
+                raise exceptions.UserError("Sold properties cannot be cancelled.")
```
bso-odoo: Also, try to make such changes inside the commit where the line initially appeared.
(See <https://github.com/odoo/tutorials/pull/817/commits/f22c1f0ff468d5b3c3eb8567429648fbec0f9c8d#r2154555307>)

estate/views/estate_property_views.xml
```diff
@@ -89,17 +89,43 @@
                             </group>
                         </page>
                     </notebook>
-                </sheet>
+                </sheet>
```
bso-odoo: Fishy end of line spaces

awesome_dashboard/static/src/dashboard/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,60 @@
...
+            type: Object,
+        },
+    };
+    
+    setup() {
+        onWillStart(() => loadJS("/web/static/lib/Chart/Chart.js"));
+        this.chartRef = useRef("chart");
+        onMounted(() => {
+            this.renderChart();
+        });
```
bso-odoo, line 19: Is this sufficient to get the updated data displayed in the chart ?

PR #821, opened 2025-06-18
[ADD] estate: New module to manage estates.

FedericoBraidi:


estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+{
+    'name': 'Estate',
```
aboo-odoo: In python, for quoting strings we try to follow the following convention:

single quote around technical terms
double quotes around display terms
=> ('north', "North")

Let's try to follow this during this tutorial, even though you'll find many counter examples in the existing code.

estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_property, estate_property_type, estate_property_tag, estate_property_offer, res_users
```
aboo-odoo: We tend to prefer one line per import, it's easier to manage when there is a lot of them :+1:

estate/models/estate_property.py
```diff
@@ -0,0 +1,87 @@
+from odoo import models, fields, api, exceptions
```
aboo-odoo: The [imports](https://www.odoo.com/documentation/17.0/contributing/development/coding_guidelines.html#imports) should be
1. External libraries (one per line sorted and split in python stdlib)
2. Imports of odoo
3. Imports from Odoo modules (rarely, and only if necessary)

Inside these 3 groups, the imported lines are **alphabetically** sorted. You can use RUFF to help you sort them, it's quite handy

estate/models/estate_property.py
```diff
@@ -0,0 +1,87 @@
...
+    _description = "Properties of the Estate app"
+
+    _sql_constraints = [('positive_expected_price', 'CHECK(expected_price > 0)', 'The expected price should be strictly positive.'),
+                      ('positive_selling_price', 'CHECK(selling_price > 0)', 'The selling price should be strictly positive')]
+    _order = "id desc"
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(copy=False, default=fields.Date.add(fields.Date.today(), months=3))
```
aboo-odoo: This default value will be computed once the server starts, if I start and let it run 3 months, then the default availability date will be today(the today 3 month ago) + 3 months = today aha

Like for the salesperson_id it should use a lambda function.

estate/models/estate_property.py
```diff
@@ -0,0 +1,87 @@
...
+    selling_price = fields.Float(readonly=True, copy=False)
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
+    )
```
```suggestion
    garden_orientation = fields.Selection(
        selection=[
            ('north', 'North'),
            ('south', 'South'),
            ('east', 'East'),
            ('west', 'West'),
        ]
    )
```
Easier to manage and read

estate/models/estate_property.py
```diff
@@ -0,0 +1,87 @@
...
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
```
aboo-odoo, line 47: Same as above :smile:

estate/models/estate_property.py
```diff
@@ -0,0 +1,87 @@
...
+        selection=[('new', 'New'), ('offer_received', 'Offer Received'), ('offer_accepted', 'Offer Accepted'), ('sold', 'Sold'), ('cancelled', 'Cancelled')],
+        required=True,
+        copy=False,
+        default='new'
+    )
+    property_type_id = fields.Many2one("estate.property.type", "Type")
+    salesman_id = fields.Many2one("res.users", "Salesman", default=lambda self: self.env.user)
+    buyer_id = fields.Many2one("res.partner", "Buyer", copy=False)
+    tag_ids = fields.Many2many("estate.property.tag")
+    offer_ids = fields.One2many("estate.property.offer", "property_id", "Offers")
```
aboo-odoo: Good use of `_id` and `_ids` suffix, well done :+1:

estate/models/estate_property.py
```diff
@@ -0,0 +1,87 @@
...
+    total_area = fields.Integer(compute="_compute_total_area")
+    best_price = fields.Float(compute="_compute_best_price")
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
```
aboo-odoo, line 72: For those methods, `self` could be a recordset, i.e. containing more than one object. The reason it could be a recordset is a bit advanced but in summary, it's optimization due to prefetching mecanisms of the data. Another simple example would be code that create 2 estate at the same time.

Currently, in said scenario, your method will raise an error. You need to loop over self `for estate in self:` ...

estate/models/estate_property.py
```diff
@@ -0,0 +1,87 @@
...
+        else:
+            self.garden_area = 0
+            self.garden_orientation = None
+
+    def sold_property_button_action(self):
+        for record in self:
+            if self.state != 'cancelled':
+                record.state = 'sold'
+            else:
+                raise exceptions.UserError("Cannot set cancelled property as sold.")
```
aboo-odoo: Some strings are automatically translated but this is not the case for error messages. To make it translatable, use `_()` method 👍

estate/models/estate_property.py
```diff
@@ -0,0 +1,87 @@
...
+            if self.state != 'sold':
+                record.state = 'cancelled'
+            else:
+                raise exceptions.UserError("Cannot set sold property as cancelled.")
+
+    @api.constrains("selling_price", "expected_price")
+    def _check_selling_90_perc(self):
+        for record in self:
+            if float_is_zero(record.selling_price, precision_digits=2):
+                break
```
aboo-odoo: This means that, in case you are checking several at once, you iterate over them. If the first estate's selling price is 0, you won't check the rest. Not sure this is what you want ahah

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,53 @@
...
+    date_deadline = fields.Datetime(compute="_compute_date_deadline", inverse="_inverse_date_deadline")
+    property_type_id = fields.Many2one(related="property_id.property_type_id", store=True)
+
+    @api.depends("validity")
+    def _compute_date_deadline(self):
+        for record in self:
+            if record.create_date:
+                record.date_deadline = fields.Datetime.add(record.create_date, days=record.validity)
+            else:
+                record.date_deadline = fields.Datetime.add(fields.Datetime.today(), days=record.validity)
```
aboo-odoo: This works too if you prefer
```suggestion
            record.date_deadline = fields.Datetime.add(record.create_date or fields.Datetime.today(), days=record.validity)
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def _inverse_date_deadline(self):
+        for record in self:
+            if record.create_date:
```
aboo-odoo: You can use the `or` here too

aboo-odoo:
i you want to change you editor for git https://stackoverflow.com/questions/2596805/how-do-i-make-git-use-the-editor-of-my-choice-for-editing-commit-messages

aboo-odoo:
Good work, well done !

aboo-odoo:
Now that your onboarding is done, could you close the PR ? :smile:

PR #849, opened 2025-07-09
[ADD] barcode_scan:  Added barcode scanning in SO/PO 

hapt-odoo:
On scanning the barcode of any product, if the product is available in
The catalog, then the quantity will be updated by 1 unit, and if the product
is not available, then the notification of warning will be shown to the user and
The product is added to the first page of the catalog based on the quantity.


barcode_scan/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+{
+    'name': "Barcode Scanning In SO/PO",
+    'category': '',
+    'version': '0.1',
+    'summary': 'Adds barcode scanning capabilities to product catalog',
+    'depends': [
+        'product',
+        'barcodes',
+        'sale_management',
+        'web',
```
bit-odoo: No need to add it in the depends.

barcode_scan/static/src/product_catalog/kanban_controller.js
```diff
@@ -0,0 +1,56 @@
...
+            console.error("Model not found");
+            return;
+        }
+
+        const orderLines = await this.orm.searchRead(
+            this.props.context.active_model,
+            [["order_id", "=", this.props.context.order_id],
+            ["product_id", "=", product.id]],
+            ["id", productQuantity, "product_id"]
+        );
```
bit-odoo, line 42: why not { limit: 1 } ?

barcode_scan/static/src/product_catalog/kanban_model.js
```diff
@@ -0,0 +1,36 @@
+import { ProductCatalogKanbanModel } from "@product/product_catalog/kanban_model";
+import { getFieldsSpec } from "@web/model/relational_model/utils";
+import { rpc } from "@web/core/network/rpc";
+
+export class CustomProductCatalogKanbanModel extends ProductCatalogKanbanModel {
+    async _loadUngroupedList(config) {
+        const totalProducts = await this.orm.search(config.resModel, config.domain);
```
bit-odoo: Use meaningful variable names like `ProductIds`

barcode_scan/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+        'stock_barcode',
+        'purchase'
+    ],
+    'sequence': 1,
```
bit-odoo: Why sequence 1?

barcode_scan/static/src/product_catalog/kanban_controller.js
```diff
@@ -0,0 +1,56 @@
...
+
+        let productQuantity;
+        if (this.resModel == "sale.order") {
+            productQuantity = "product_uom_qty";
+        } else if (this.resModel == "purchase.order") {
+            productQuantity = "product_qty";
+        } else {
+            console.error("Model not found");
+            return;
+        }
```
bit-odoo: You can also improve this. Because If there is more model then we have to write more else if.It will reduce code readability.

barcode_scan/__manifest__.py
hapt-odoo: Sequence is used to add this module at the first position in AppList to manually upgrade it easily.

barcode_scan/static/src/product_catalog/kanban_controller.js
hapt-odoo: If in any case there are more than 1 order line for any specific product, and if the limit: 1 is used, then the total quantity of the product will be shown wrong.

PR #851, opened 2025-07-11
[ADD] pos_receipt: created multiple layout styles for POS order receipts

arat-odoo:
In the POS application, multiple layout styles were created for the order 
receipt. Based on the selected style, the corresponding format is displayed in 
the order receipt PDF. The layout format can also be configured from the General 
Settings.

### DIFF
--- a/pos_receipt/static/src/pos_order_receipt.xml
+++ b/pos_receipt/static/src/pos_order_receipt.xml
@@ -124,6 +124,212 @@
                         <div id="order-date" t-esc="props.data.date" />
                     </div>
             </t>
+
+            <t t-if="order.config_id.receipt_layout=='boxes'">
+                    <ReceiptHeader data="props.data.headerData"/>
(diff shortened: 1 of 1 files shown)


pos_receipt/wizard/pos_receipt_layout.py
```diff
@@ -0,0 +1,60 @@
...
+        elif self.receipt_layout == "lined":
+            self.receipt_preview = self.env["ir.ui.view"]._render_template(
+                "pos_receipt.custom_receipt_lined",
+                {
+                    "header": self.receipt_header,
+                    "footer": self.receipt_footer,
+                    "logo": self.receipt_logo,
+                    "is_restaurant": self.is_restaurant,
+                },
+            )
```
bit-odoo, line 57: Repetition of code. Function is also not scalable.

bit-odoo (commented):
Hello,
Your commit message tag is different than PR tag.It should be same.

PR #858, opened 2025-07-23
[ADD] book_price: add book price to sale order and invoice lines

arkp-odoo:
It's important for users to see the original price of a product with applied pricelist when creating a sale order or invoice. Without this, it's hard to know if the price being charged is correct or if a discount has been applied. Showing the book price helps sales and accounting teams compare the price with pricelist with the charged one, making it easier to catch mistakes, explain price differences, and ensure consistency between sales and invoicing.

### DIFF
--- a/book_price/models/account_move_line.py
+++ b/book_price/models/account_move_line.py
@@ -1,0 +1,24 @@
+from odoo import api, fields, models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    book_price = fields.Float(string="Book Price", compute="_compute_book_price")
+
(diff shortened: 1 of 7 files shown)


book_price/models/account_move_line.py
```diff
@@ -0,0 +1,7 @@
+from odoo import fields, models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    book_price = fields.Float(related='sale_line_ids.book_price')
```
maad-odoo: Shouldn't  this be a compute field as well ?

maad-odoo:
The PR and commit message is not structured properly.
Also they are missing the mention of task id.

PR #867, opened 2025-07-24
18.0 technical training yabd

YassinWalid:


estate/models/estate_property.py
```diff
@@ -0,0 +1,92 @@
+from odoo import api, fields, models
+
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare
+
+from dateutil.relativedelta import relativedelta
```
Megaaaaaa: Almost got it, alphabetical order is nice but the blocks are actually 
- external exports
- everything else
- odoo.addons

so:

```suggestion
from dateutil.relativedelta import relativedelta

from odoo import api, fields, models
from odoo.exceptions import UserError, ValidationError
from odoo.tools.float_utils import float_compare
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,92 @@
...
+    selling_price = fields.Float(readonly=True, copy=False)
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection([('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')])
+    active = fields.Boolean(default=True)
+    state = fields.Selection([('new', 'New'), ('offer received', 'Offer Received'), ('offer accepted', 'Offer Accepted'), ('sold', 'Sold'), ('cancelled', 'Cancelled')], required=True, copy=False, default='new')
```
Megaaaaaa: For Selection fields, we often multiline their definition so it's a bit easier to read but since we don't really care about line length in Python, feel free to do it the way that you prefer :+1: (if you do multiline field definitions, don't forget a comma on the very last line and if you multiline a selection, don't forget the comma after the last tuple)

```suggestion
    garden_orientation = fields.Selection(
        selection=[
            ('north', 'North'),
            ('south', 'South'),
            ('east', 'East'),
            ('west', 'West'),
        ],
    )
    active = fields.Boolean(default=True)
    state = fields.Selection(
        [
            ('new', 'New'),
            ('offer received', 'Offer Received'),
            ('offer accepted', 'Offer Accepted'),
            ('sold', 'Sold'),
            ('cancelled', 'Cancelled'),
        ],
        required=True,
        copy=False,
        default='new',
    )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,92 @@
...
+        for record in self:
+            record.total_area = record.garden_area + record.living_area
+
+    @api.depends('offer_ids')
+    def _compute_best_offer(self):
+        for record in self:
+            if record.offer_ids:
+                record.best_price = max(record.offer_ids.mapped('price'))
+            else:
+                record.best_price = 0
```
Megaaaaaa: Since your PR is quite nice and there is not much to say, I'll give you a quick tip. When it's a simple 
```
if A:
    record.<field> = X
else:
    record.<field> = Y
```
, you can use a simple boolean operation to make it look a bit nicer :wink:
```
record.<field> = A and X or Y
```

```suggestion
        for record in self:
            record.best_price = record.offer_ids and max(record.offer_ids.mapped('price')) or 0
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @api.onchange('garden')
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = 'north'
+        else:
+            self.garden_area = 0
+            self.garden_orientation = ''
```
Megaaaaaa: When deassigning a value, we prefer using `False` regardless of the type of the field; It's just a bit more explicit. :+1:

estate/models/estate_property.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def action_set_sold(self):
+        for record in self:
+            if record.state == 'cancelled':
+                raise UserError("Cancelled properties cannot be sold")
+            else:
+                record.state = 'sold'
+                return True
```
Megaaaaaa: No need to force a return in actions :+1:

```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def action_set_cancelled(self):
+        for record in self:
+            if record.state == 'sold':
+                raise UserError("Sold properties cannot be cancelled")
+            else:
+                record.state = 'cancelled'
+                return True
```
Megaaaaaa: Same here :smile:

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,62 @@
+from odoo import api, fields, models
+
+from odoo.exceptions import UserError
+
+from dateutil.relativedelta import relativedelta
```
Megaaaaaa: I realize I've lready put a comment on the imports order but I haven't included the doc. [Here](https://www.odoo.com/documentation/18.0/contributing/development/coding_guidelines.html#imports) it is.

```suggestion
from dateutil.relativedelta import relativedelta

from odoo import api, fields, models
from odoo.exceptions import UserError
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,62 @@
...
+         'The offer price for a property must be strictly positive.'),
+    ]
+
+    @api.depends('create_date', 'validity')
+    def _compute_date_dealine(self):
+        for record in self:
+            if record.create_date:
+                record.date_deadline = record.create_date + relativedelta(days=record.validity)
+            else:
+                record.date_deadline = fields.Date.today() + relativedelta(days=record.validity)
```
Megaaaaaa: You can apply the same semantics here to have a cleaner method: :+1:

```suggestion
            create_date = record.create_date and record.create_date or fields.Date.today()
            record.date_deadline = create_date + relativedelta(days=record.validity)
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,62 @@
...
+                raise UserError("An offer has already been accepted")
+
+    def action_refuse(self):
+        for record in self:
+            record.status = 'refused'
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            property_id = self.env['estate.property'].browse(vals['property_id'])
```
Megaaaaaa: Same here, this is just an advanced thing because there isn't much else to say:

In opposition to using directly field attributes that use the cache and are really quick, any `search()` or `browse()` or other similar methods that imply a query cna have quite an impact on performance when ran inside a loop. Therefore, each time you'll run one of those inside a loop, any reviewer will ask to put it out of the loop in a variable and use the data you want in that variable in the loop. It would like something like this: :+1:

```suggestion
        property_ids = self.env['estate.property'].browse([vals['property_id'] for vals in vals_list])
        for vals in vals_list:
            property_id = property_ids.filtered(lambda p: p.id == vals['property_id'])
```

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property,estate.property,model_estate_property,base.group_user,1,1,1,1
+access_estate_property_type,estate.property.type,model_estate_property_type,base.group_user,1,1,1,1
+access_estate_property_tag,estate.property.tag,model_estate_property_tag,base.group_user,1,1,1,1
+access_estate_property_offer,estate.property.offer,model_estate_property_offer,base.group_user,1,1,1,1
```
Megaaaaaa: Missing EOL (End Of Line character (linebreak)). This is something we do mostly to ensure clean editor history (if someone add something to the file and you didn't put the last linebreak, he will override you as the last editor of the previous (your) last line) but also because some compiler may face issues if this is not present so all the files in odoo codebase always end with an empty line. :+1:

```suggestion
access_estate_property_offer,estate.property.offer,model_estate_property_offer,base.group_user,1,1,1,1

```

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <menuitem id="home_menu_root" name="Estate">
+        <menuitem id="estate_property_menu_action" action="estate_property_action"/>
+        <menuitem id="settings" name="Settings">
+            <menuitem id="estate_property_type_menu_action" action="estate_property_type_action"/>
+            <menuitem id="estate_property_tag_menu_action" action="estate_property_tag_action"/>
+        </menuitem>
+    </menuitem>
+</odoo>
```
Megaaaaaa: Same here :smile:

Many ther instances of this lower in the diff :wink:

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,44 @@
...
+
+    <record id="estate_property_offer_view_list" model="ir.ui.view">
+        <field name="name">estate.property.offer.list</field>
+        <field name="model">estate.property.offer</field>
+        <field name="arch" type="xml">
+            <list string="Offers" editable="bottom" decoration-danger="status == 'refused'" decoration-success="status == 'accepted'">
+                <field name="price" string="Price"/>
+                <field name="validity" string="Validity"/>
+                <field name="date_deadline" string="Deadline"/>
+                <field name="partner_id" string="Partner"/>
```
Megaaaaaa: You can also put all the strings on the field definition in Python so that you XML looks a bit cleaner but that also depends on what you want to achieve so I let you decide :+1:

And there are ofc multiple instances of this found lower in the diff but it's your call, do what you want :wink:

Megaaaaaa (changes requested):
Hello 👋

Really great PR, not much to say tbh :clap:

There are very few important stuff that repeat themselves and then I slipped in some advanced tips since there wasn't much to say :+1: 

Keep in mind that a review is not the utlimate solution, it's only someone seeing things he would have done differently and offering his alternative. You're always free to answer with an other alternative or even say you disagree and bring your arguments.

What I would suggest you do and continue to do while you're not the most comfortable with odoo's structure is to not apply all the changes at once. Some changes might have a cascade of impact and if you change everything at once, you will probably encounter errors when running you database. Splitting the review in multiple steps and trying to run your db in between those can save you a lot of time in the long run. Even if no comment looks too big, a typo can be a hell to find if you've changed tens or hundreds of lines at once.

Also a quick tip for the methodology when applying reviews changes: What I like to do is put a reaction like a thumbsup on a message once I have made the change locally. Then when everything is marked with something, I push, go through my diff and mark as resolved the things that are now outdated so nothing is left forgotten.

If you have any question, don't hesitate to ask them!

estate/models/estate_property.py
YassinWalid: Shouldn't line 52 be:
`record.best_price = record.offer_ids and max(record.offer_ids.mapped('price')) or 0` ?

Megaaaaaa: Yup you're right, typo duo to writing everything in the github text editor, mb :sweat_smile:

estate/views/estate_property_offer_views.xml
YassinWalid: :+1: I believe this is cleaner, since if a field might be displayed multiple times in the XML file, then we would repeat defining its string each time.

PR #872, opened 2025-07-24
18.0 technical training kmah

Mahmoudk3m:


estate/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+from . import (
+    estate_property,
+    estate_property_type,
+    estate_property_tag,
+    estate_property_offer,
+    res_users,
+)
```
Megaaaaaa: I know it looks better but it's required to write `from . import x` on each line. We only use `()` in tests folders.

Also while at it, in `__init__.py`, we order everthing alphabetically :+1:

```suggestion
from . import estate_property
from . import estate_property_offer
from . import estate_property_tag
from . import estate_property_type
from . import res_users
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,154 @@
+from odoo import models, fields, api
+from odoo.exceptions import UserError, ValidationError
```
Megaaaaaa: You can find the imports guidelines [here](https://www.odoo.com/documentation/18.0/contributing/development/coding_guidelines.html#imports). There are a bunch of rules that are good to know but here the only thing I want to point out is that we order the imported elements alphabetically. :+1:

```suggestion
from odoo import api, fields, models
from odoo.exceptions import UserError, ValidationError
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,154 @@
...
+    _name = "estate.property"
+    _description = "Real Estate Property"
+    _order = "id desc"
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        copy=False, default=fields.Date.add(fields.Date.today(), months=3)
+    )
```
Megaaaaaa: This one is short enough to be one lines imo but if you want to multi-line it, then go all-in :smile:

Also, when you multiline a field definition or a list, you need to end your last line with a comma so that if someone adds a new line, he doesn't modify the author of the previous (your) last line. :+1:

```suggestion
    date_availability = fields.Date(
        copy=False,
        default=fields.Date.add(fields.Date.today(), months=3),
    )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,154 @@
...
+        default="new",
+        string="Status",
+        required=True,
+        copy=False,
+    )
+    active = fields.Boolean(default=True)
+    property_type_id = fields.Many2one("estate.property.type", string="Property Type")
+    salesperson_id = fields.Many2one(
+        "res.users", string="Salesperson", default=lambda self: self.env.user
+    )
```
Megaaaaaa: All-in :wink: (If you multi-line and en up putting everything on one line, maybe you shouldn't have multi-lined in the first place but I like to multiline my fields definitions too so I get it :smile:)

```suggestion
    salesperson_id = fields.Many2one(
        "res.users",
        string="Salesperson",
        default=lambda self: self.env.user,
    )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,154 @@
...
+    buyer_id = fields.Many2one("res.partner", string="Buyer", copy=False)
+    tag_ids = fields.Many2many("estate.property.tag", string="Tags")
+    offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers")
+
```
Megaaaaaa: It's not flagged by ci/style but we tend to not space the fields list at all excecpt if there is a group of fields that are linked together.

```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,154 @@
...
+            record.buyer_id = record.buyer_id or self.env.user.partner_id
+
+        # Force form reload
+        return {
+            "type": "ir.actions.act_window",
+            "res_model": "estate.property",
+            "res_id": self.id,
+            "view_mode": "form",
+            "target": "current",
+        }
```
Megaaaaaa: You shouldn't have to do that for the data to be refreshed. Is there a particular reason for it ? :thinking:

estate/models/estate_property.py
```diff
@@ -0,0 +1,154 @@
...
+            record.buyer_id = False
+
+        # Force form reload
+        return {
+            "type": "ir.actions.act_window",
+            "res_model": "estate.property",
+            "res_id": self.id,
+            "view_mode": "form",
+            "target": "current",
+        }
```
Megaaaaaa: Same here but I suppose it's the same reason ?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,122 @@
+from odoo import models, fields, api
+from odoo.exceptions import UserError
+from datetime import timedelta
```
Megaaaaaa: Imports order :+1:

```suggestion
from datetime import timedelta

from odoo import api, fields, models
from odoo.exceptions import UserError
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,122 @@
...
+
+class EstatePropertyOffer(models.Model):
+    _name = "estate.property.offer"
+    _description = "Real Estate Property Offer"
+    _order = "price desc"
+
+    price = fields.Float()
+    status = fields.Selection(
+        [("accepted", "Accepted"), ("refused", "Refused")], copy=False
+    )
```
Megaaaaaa: As said earlier, either oneline or multiline but pick a side :smile:

```suggestion
    status = fields.Selection(
        selection=[
            ("accepted", "Accepted"),
            ("refused", "Refused"),
        ],
        copy=False,
    )
```

Multiple instances of this lower in the diff :wink:

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,122 @@
...
+        compute="_compute_date_deadline", inverse="_inverse_date_deadline"
+    )
+
+    _sql_constraints = [
+        ("check_price_positive", "CHECK(price >= 0)", "The price must be positive!"),
+        (
+            "check_validity_positive",
+            "CHECK(validity > 0)",
+            "The validity must be positive!",
+        ),
```
Megaaaaaa: There is no better option between onelining both or multilining both but one thing that will be asked from your PR is consistency. When there's a choice to be made and there is no best option, do what you want but the same everywhere in your diff. That's something that can be applied to multiple things like single quotes and doubles quotes for example.

Here I would recomment to multiline both but if you want to oneline both, it's totally valid too. :+1:

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,122 @@
...
+        for record in self:
+            if record.create_date:
+                record.date_deadline = record.create_date.date() + timedelta(
+                    days=record.validity
+                )
+            else:
+                # Fallback for when create_date is not set yet
+                record.date_deadline = fields.Date.today() + timedelta(
+                    days=record.validity
+                )
```
Megaaaaaa: This looks very repetitive imo :thinking:  I would simplify that doing this instead:

```suggestion
            create_date = (record.create_date and record.create_date.date()) or fields.Date.today()
            record.date_deadline = create_date + timedelta(days=record.validity)
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def _inverse_date_deadline(self):
+        for record in self:
+            if record.date_deadline and record.create_date:
+                record.validity = (
+                    record.date_deadline - record.create_date.date()
+                ).days
+            elif record.date_deadline:
+                # Fallback for when create_date is not set yet
+                record.validity = (record.date_deadline - fields.Date.today()).days
```
Megaaaaaa: Same here but again, it's only for the looks so your call :+1:

```suggestion
            create_date = (record.create_date and record.create_date.date()) or fields.Date.today()
            if record.date_deadline:
                record.validity = (record.date_deadline - create_date).days
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,122 @@
...
+    def action_accept(self):
+        for offer in self:
+            # Check if another offer is already accepted for this property
+            existing_accepted = self.search(
+                [
+                    ("property_id", "=", offer.property_id.id),
+                    ("status", "=", "accepted"),
+                    ("id", "!=", offer.id),
+                ]
+            )
```
Megaaaaaa: Okay so first of al, even when you're searching on the model of self, you don't want to do `self.search()` because you will use the access rights and the context of the current recordset. Always do `self.env['<model>'].search()`

Then, in terms of performance, doing a `search()` **always** implies a query so it's **always** faster to make your way trough the related fields than searching the model. It doesn't make a lot of difference when it for one record but here you are inside a `for reocrd in self:` type of loop so depending on the size of `self` you will potentially do a lot of queries. To avoid that, we always put the searches outside of loops and group the result by record but here there's an even easier solution using the related fields.

```suggestion
            existing_accepted = offer.property_id.offer_ids.filtered(lambda o: o.status == 'accepted' and o.id != offer.id)
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,122 @@
...
+            if existing_accepted:
+                raise UserError(
+                    "Another offer has already been accepted for this property!"
+                )
```
Megaaaaaa: And actually since you are doing that only to check the existence of a record in that subset, you can even do: (the check on `id != offer.id` is useless since `offer.status` is not `accepted` yet)

```suggestion
            if any(o.status == 'accepted' for o in offer.property_id.offer_ids):
                raise UserError(
                    "Another offer has already been accepted for this property!"
                )
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,122 @@
...
+
+            # Cancel all other offers for this property
+            other_offers = self.search(
+                [
+                    ("property_id", "=", offer.property_id.id),
+                    ("id", "!=", offer.id),
+                    ("status", "!=", "refused"),
+                ]
+            )
+            other_offers.write({"status": "refused"})
```
Megaaaaaa: Then you will have:

```suggestion
            other_offers = offer.property_id.offer_ids.filtered(lambda o: o.id != offer.id)
            other_offers.status = 'refused'
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,122 @@
...
+                [
+                    ("property_id", "=", offer.property_id.id),
+                    ("status", "=", "accepted"),
+                ]
+            )
+
+            if not remaining_accepted:
+                offer.property_id.state = "offer_received"
+                offer.property_id.buyer_id = False
+                offer.property_id.selling_price = 0.0
```
Megaaaaaa: Try to use `any()` here too :+1:

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,18 @@
+from odoo import models, fields
```
Megaaaaaa: imports order :smile:

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,30 @@
+from odoo import models, fields, api
```
Megaaaaaa: imports order too :smile:

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class EstatePropertyType(models.Model):
+    _name = "estate.property.type"
+    _description = "Real Estate Property Type"
+    _order = "sequence, name"
+
+    name = fields.Char(required=True)
+    sequence = fields.Integer(default=1, help="Used to order types manually")
+    property_ids = fields.One2many(
```
Megaaaaaa, line 14: Oneline or multiline. Other instances lower in the file :+1:

estate/models/res_users.py
```diff
@@ -0,0 +1,12 @@
+from odoo import models, fields
```
Megaaaaaa: Imports order

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,53 @@
...
+                                    <field name="state"/>
+                                </list>
+                            </field>
+                        </page>
+                    </notebook>
+                </sheet>
+            </form>
+        </field>
+    </record>
+</odoo>
```
Megaaaaaa: This icon signals that you are missing the EOL (End Of Line character (line break)) on the last line of you file. We always put it even on the last line so that if someone was to add something to your file, they would not modify the last editor of the previous (your) last line. Some compiler may also face issues when files re missing it but it's mostly for clean history purposes :+1: Every file in Odoo must end with an empty line. Some are signaled by github, some are not. You just have to take the habit of adding it or have our linter do it for you (I know some people do but I don't so don't ask me how :laughing:)

```suggestion
</odoo>

```

estate/views/res_users_views.xml
```diff
@@ -0,0 +1,23 @@
...
+                            <field name="state"/>
+                            <field name="property_type_id"/>
+                            <field name="date_availability"/>
+                        </list>
+                    </field>
+                </page>
+            </notebook>
+        </field>
+    </record>
+</odoo>
```
Megaaaaaa: missing EOL :smile:

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,30 @@
+from odoo import models, Command
```
Megaaaaaa: Small tip about this, it's not because c < m in alphabetical order, it's actually because UPPERCASE < lowercase. (It's not an actual alphabetical order but an [ASCIIbetical](https://en.wiktionary.org/wiki/ASCIIbetical#English) order :face_in_clouds: )

```suggestion
from odoo import Command, models
```

Megaaaaaa (changes requested):
Hello 👋

First of all, very nice PR, here's a first review for you 😄

It might seem like there's a lot but don't worry about it, it's a lot of the same details that repeat themselves.

You can go through all my comments and apply what I suggested or bring your own alternative. Keep in mind that a review is not the utlimate solution, it's only someone seeing things he would have done differently and offering his alternative. You're always free to answer with an other alternative or even say you disagree and bring your arguments.

What I would suggest you do and continue to do while you're not the most comfortable with odoo's structure is to not apply all the changes at once. Some changes might have a lot of impact and if you change everything at once, you will probably have a lot of errors when running you database. Splitting the review in multiple steps and trying to run your db in between those can save you a lot of time in the long run. Even if no comment looks too big, a typo can be a hell to find if you've changed many lines at once.

Also a quick tip for the methodology when applying reviews changes: What I like to do is put a reaction like a thumbsup on a message once I have made the change locally. Then when everything is marked with something, I push, go through my diff and mark as resolved the things that are now outdated so nothing is left forgotten.

If you have any question, don't hesitate to ask them!

estate/models/estate_property.py
Mahmoudk3m: Actually no, it wasn't reloading due to a bug on my end and I thought I needed this.

PR #878, opened 2025-07-24
[ADD] estate: Create New Estate Tutorial App, [IMP] awesome_owl, awesome_dashboard: Web Framework Tutorial

Ghost8345:
Create a new Estate App where Odoo Users can sell, track real estate properties.
Improve both awesome_owl, awesome_dashboard for tutorial


Ghost8345:
squash

Ghost8345:
robodoo squash

estate/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+{
+    'name': "Real Estate",
+    'version': '1.0',
+    'category': 'Sales/RealEstate',
+    'summary': 'Track and deal with real estate properties.',
+    'author': "Meniem",
```
aboo-odoo: Unfortunately, all your work now belongs to 'Odoo S.A.'

estate/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    'description': """
+    Description text
+    """,
+    'depends': ['base'],
+    'installable': True,
+    'application': True,
+    'auto_install': False,
```
aboo-odoo: It is False by default
```suggestion
```

estate/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    'data': [
+        'security/ir.model.access.csv',
+        'views/estate_property_views.xml',
+        'views/estate_property_offer_views.xml',
+        'views/estate_property_type_views.xml',
+        'views/estate_property_tag_views.xml',
+        'views/estate_menus.xml',
+    ],
+    'license': 'LGPL-3'
```
aboo-odoo:
```suggestion
    'license': 'LGPL-3',
```
General tip, always add a comma after the last dict/list element (when not inline) to reduce the GitHub diff next time someone adds something at the end of your dict/list

estate/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+from . import (estate_property,
+               estate_property_type,
+               estate_property_tag,
+               estate_property_offer)
```
aboo-odoo: We tend to do it this way
```suggestion
from . import estate_property
from . import estate_property_type
from . import estate_property_tag
from . import estate_property_offer
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+    _name = "estate.property"
+    _description = "Estate Property"
+    _order = "id desc"
+
+    _orientation_list = [('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
+    _states_list = [('new', 'New'),
+                    ('offer_received', 'Offer Received'),
+                    ('offer_accepted', 'Offer Accepted'),
+                    ('sold', 'Sold'),
+                    ('cancelled', 'Cancelled')]
```
aboo-odoo: Not really necessary: it does not save space and makes it harder to read in the code as you have to jump to here to get the states. Furthermore, `_attribute` are for PSQL attribute; I don't think PSQL needs the list of states as attributes

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
+from odoo import fields, models, api, exceptions
+from odoo.tools.date_utils import add
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate Property"
+    _order = "id desc"
```
aboo-odoo: Try to use single quote `'` for strings that will not be displayed in the UI, i.e. dict keys and double quote `"` for UI strings 

This comment applies to other places too

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+
+    _orientation_list = [('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
+    _states_list = [('new', 'New'),
+                    ('offer_received', 'Offer Received'),
+                    ('offer_accepted', 'Offer Accepted'),
+                    ('sold', 'Sold'),
+                    ('cancelled', 'Cancelled')]
+
+    name = fields.Char('Title', required=True)
```
```suggestion
    name = fields.Char("Title", required=True)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+    description = fields.Text("Description")
+    postcode = fields.Char("Postcode")
+    date_availability = fields.Date("Available From", copy=False, default=add(fields.Date.today(), months=3))
```
aboo-odoo: This default value will be computed once the server starts, if I start and let it run 3 months, then the default availability date will be today(the today 3 month ago) + 3 months = today aha. Like for the salesperson_id it should use a `lambda` function so that it gets recomputed everytime it is called.

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+        help="Orientation of the garden, if it exists."
+    )
+    active = fields.Boolean("Active", default=True)
+    state = fields.Selection(
+        string="Status",
+        selection=_states_list,
+        default='new',
+        help="State of the property.",
+        copy=False,
+        required=True
```
```suggestion
        required=True,
```
General tip, always add a comma after the last dict/list element (when not inline) to reduce the GitHub diff next time someone adds something at the end of your dict/list

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+    property_type_id = fields.Many2one(
+        string="Property Type",
+        comodel_name="estate.property.type",
+        help="Type of the property."
```
```suggestion
        help="Type of the property.",
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+    buyer_id = fields.Many2one(
+        string="Buyer",
+        comodel_name="res.partner",
+        copy=False,
+        help="Buyer of the property."
```
```suggestion
        help="Buyer of the property.",
```
and so on

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+    state = fields.Selection(
+        string="Status",
+        selection=_states_list,
+        default='new',
+        help="State of the property.",
+        copy=False,
+        required=True
+    )
+    property_type_id = fields.Many2one(
+        string="Property Type",
```
aboo-odoo: By default, `string` is the name of the field minus the `_id(s)` suffix. So it is not necessary here

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+         'The expected price of a property should be strictly positive.'),
+        ('check_selling_price_positive', 'CHECK(selling_price >= 0)',
+         'The selling price of a property should be positive or zero.'),
+    ]
+
+    @api.constrains('expected_price', 'selling_price')
+    def _check_prices(self):
+        for record in self:
+            if ((not float_is_zero(record.selling_price, precision_rounding=0.01))
+                    and float_compare(record.selling_price, 0.9 * record.expected_price, precision_rounding=0.01) < 0):
```
aboo-odoo: The indent makes it hard to read
```suggestion
            if (
                not float_is_zero(record.selling_price, precision_rounding=0.01)
                and float_compare(record.selling_price, 0.9 * record.expected_price, precision_rounding=0.01) < 0
            ):
```

Furthermore, currently, if I put 0.0, the condition is False and thus goes through

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends('offer_ids.price')
+    def _compute_best_offer(self):
+        for record in self:
+            if not record.offer_ids:
+                record.best_offer = 0.0
+            else:
+                record.best_offer = max(record.offer_ids.mapped('price'))
```
```suggestion
            record.best_offer = max(record.offer_ids.mapped('price'), default=0.0)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+    _sql_constraints = [
+        ('check_expected_price_strictly_positive', 'CHECK(expected_price > 0)',
+         'The expected price of a property should be strictly positive.'),
+        ('check_selling_price_positive', 'CHECK(selling_price >= 0)',
+         'The selling price of a property should be positive or zero.'),
+    ]
+
+    @api.constrains('expected_price', 'selling_price')
+    def _check_prices(self):
+        for record in self:
```
aboo-odoo: General comments for loops like this: 
```suggestion
        for estate in self:
```
`record` is a bit too generic, it's fine for small method like this. But when you'll work in huuuge methods, you'll be happy to have a name a bit more meaningful trust me

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+            if ((not float_is_zero(record.selling_price, precision_rounding=0.01))
+                    and float_compare(record.selling_price, 0.9 * record.expected_price, precision_rounding=0.01) < 0):
+                raise exceptions.ValidationError("The selling price cannot be lower than 90% of the expected price.")
```
aboo-odoo: Some strings are automatically translated but this is not the case for error messages. To make it translatable, use `_()` method 👍

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
+from odoo import fields, models, api, exceptions
+from odoo.tools.date_utils import add
+from odoo.tools.float_utils import float_compare, float_is_zero
```
aboo-odoo: The [imports](https://www.odoo.com/documentation/17.0/contributing/development/coding_guidelines.html#imports) should be
1. External libraries (one per line sorted and split in python stdlib)
2. Imports of odoo
3. Imports from Odoo modules (rarely, and only if necessary)

Inside these 3 groups, the imported lines are **alphabetically** sorted. You can use RUFF to help you sort them, it's quite handy

estate/models/estate_property_offer.py
aboo-odoo, line 1: Mind that the [convention](https://www.odoo.com/documentation/17.0/contributing/development/coding_guidelines.html#symbols-and-conventions) for model ordering is:

1. Private attributes (_name, _description, _inherit, _sql_constraints, …)
3. Field declarations
4. Compute, inverse and search methods in the same order as field declaration
5. Selection method (methods used to return computed values for selection fields)
6. Constrains methods (@api.constrains) and onchange methods (@api.onchange)
7. CRUD methods (ORM overrides)
8. Action methods
9. And finally, other business methods.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,90 @@
...
+    _sql_constraints = [
+        ('check_price_strictly_positive', 'CHECK(price > 0)',
+         'The price of an offer should be strictly positive.'),
+    ]
+
+    @api.depends('validity')
+    def _compute_date_deadline(self):
+        for record in self:
+            if record.create_date:
+                record.date_deadline = add(fields.Date.from_string(record.create_date), days=record.validity)
```
aboo-odoo: from_string ?! Isn't `create_date` a Datetime ?
```suggestion
                record.date_deadline = add(fields.Date.to_date(record.create_date) or fields.Date.today(), days=record.validity)
```

Python allows for the handy `and` and `or` usage like so :+1:

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,90 @@
...
+            record.property_id.state = 'offer_accepted'
+            record.property_id.buyer_id = record.partner_id
+            record.property_id.selling_price = record.price
+
+        self.search(
+            [('property_id', '=', record.property_id.id),
+             ('status', '=', False)]
+        ).write({'status': 'refused'})
+
+        return True
```
aboo-odoo: As I showed you, this cannot work since `self` could be a recordset. Currently, you would only update the last property in the recordset:

1. Loop over all record in `self` until `self` is empty
2. exit the loop; at that moment `record == self[-1]`
3. `record.property_id.id = self[-1].property_id.id`; it does not include the others
Furthermore, currently, I'm allowed to accept 2 offers for the same property.

```suggestion
    def action_accept(self):
        for record in self:
            if record.property_id.state != 'offer_accepted':
                record.status = 'accepted'
                record.property_id.state = 'offer_accepted'
                record.property_id.buyer_id = record.partner_id
                record.property_id.selling_price = record.price

        self.env['estate.offer'].search([
            ('property_id', 'in', self.property_id.ids),
            ('status', '=', False),
        ]).write({'status': 'refused'})

        return True
```
`self.property.ids` does the trick as relational fields can be read in set and ids can be used on recordset too to return the list of IDs in the set

_not tested, this may need to be tweeked_

estate/security/__init__.py
aboo-odoo, line 1: To delete

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_user,estate.user,model_estate_property,base.group_user,1,1,1,1
+access_estate_prop_type_user,estate.property.type.user,model_estate_property_type,base.group_user,1,1,1,1
+access_estate_prop_tag_user,estate.property.tag.user,model_estate_property_tag,base.group_user,1,1,1,1
+access_estate_prop_offer_user,estate.property.offer.user,model_estate_property_offer,base.group_user,1,1,1,1
```
aboo-odoo: Even though runbot does not check it, it's better to always add an EOL in every files

estate/views/__init__.py
aboo-odoo, line 1: to delete

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <menuitem id="estate_menu_root" name="Real Estate">
```
aboo-odoo: ID for a menu: `<model_name>_menu`, or `<model_name>_menu_do_stuff` for submenus.

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="estate_prop_offer_view_form" model="ir.ui.view">
```
aboo-odoo: ID for a view: `<model_name>_view_<view_type>`, where view_type is kanban, form, list, search

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,55 @@
...
+                                </list>
+                            </field>
+                        </page>
+                    </notebook>
+                </sheet>
+            </form>
+        </field>
+    </record>
+
+    <record id="estate_prop_type_action" model="ir.actions.act_window">
```
aboo-odoo: ID for an action: the main action respects `<model_name>_action`. Others are suffixed with `_<detail>`

Ghost8345:
> Since the onboarding is over, can you close this PR ? Thanks :smile:  

Sure.

PR #904, opened 2025-07-31
[ADD] pos_receipt: allow custom receipt layout with preview

niyp-odoo:
Before this commit:
POS receipts had only one fixed layout.
Logo, header, and footer fields were not in one place.

After this commit:
Users can choose from 3 receipt layouts: light, lined, and boxes.
Header, footer, and logo can be updated from POS settings.
A live preview shows how the receipt will look.

Impact:
Easier receipt setup for users.
Helps businesses match their branding.
Reduces errors by showing preview before saving.

### DIFF
--- a/pos_receipt/static/src/order_receipt_inherit.js
+++ b/pos_receipt/static/src/order_receipt_inherit.js
@@ -1,0 +1,22 @@
+import { OrderReceipt } from "@point_of_sale/app/screens/receipt_screen/receipt/order_receipt";
+import { patch } from "@web/core/utils/patch";
+import { usePos } from "@point_of_sale/app/store/pos_hook";
+
+patch(OrderReceipt, {
+    template: "pos_receipt.order_receipt_inherited"
+});
+
+patch(OrderReceipt.prototype, {
+    setup(){
+        super.setup();
+        this.pos = usePos();
+    },
+
+    get orderQuantity() {
+        return this.props.data.orderlines.reduce((acc, line) => acc + parseFloat(line.qty), 0);
+    },
+
+    get order() {
+        return this.pos.get_order();
+    }
+});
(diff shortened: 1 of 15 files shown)


pos_receipt/wizard/pos_receipt_wizard.py
```diff
@@ -0,0 +1,54 @@
...
+    receipt_layout = fields.Selection(related='pos_config_id.receipt_layout', readonly=False, required=True)
+    receipt_logo = fields.Binary(related='pos_config_id.receipt_logo', readonly=False)
+    receipt_header = fields.Text(related='pos_config_id.receipt_header', readonly=False)
+    receipt_footer = fields.Text(related='pos_config_id.receipt_footer', readonly=False)
+    is_restaurant = fields.Boolean(related='pos_config_id.module_pos_restaurant')
+    receipt_preview = fields.Html(compute='_compute_receipt_preview')
+
+    @api.depends('receipt_layout', 'receipt_header', 'receipt_footer', 'receipt_logo')
+    def _compute_receipt_preview(self):
+        if self.receipt_layout == "lined":
```
deso-odoo, line 22: Shouldn't this loop over self? If not why?

pos_receipt/static/src/order_receipt_inherit.js
```diff
@@ -0,0 +1,22 @@
...
+        super.setup();
+        this.pos = usePos();
+    },
+
+    get orderQuantity() {
+        return this.props.data.orderlines.reduce((acc, line) => acc + parseFloat(line.qty), 0);
+    },
+
+    get order() {
+        return this.pos.get_order()
```
deso-odoo:
```suggestion
        return this.pos.get_order();
```

PR #908, opened 2025-08-05
[ADD] pos_formate: add option for select and edit receipt formate

gasa-odoo:
In this task,
Add new module pos format in which select receipt type (ex. lined, light, boxed) also add header, footer and logo and can see the preview of receipt. for this task use wizards, js file for report and add three template for preview.

### DIFF
--- a/pos_formate/tests/test_pos_configure_receipt.py
+++ b/pos_formate/tests/test_pos_configure_receipt.py
@@ -36,8 +36,8 @@ class TestPOSConfigureReceipt(TransactionCase):
         wizard.action_apply_configuration()
 
         self.assertEqual(self.pos_config.receipt_layout, 'lined')
-        self.assertEqual(self.pos_config.receipt_header, '<p>Header</p>')
-        self.assertEqual(self.pos_config.receipt_footer, '<p>Footer</p>')
+        self.assertEqual(self.pos_config.receipt_header, 'Header')
+        self.assertEqual(self.pos_config.receipt_footer, 'Footer')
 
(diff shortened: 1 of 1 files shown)


pos_formate/static/src/order_receipt_patch.js
```diff
@@ -0,0 +1,44 @@
...
+  template: "pos_formate.order_receipt_inherited",
+});
+
+console.log("OrderReceipt Template Patch Applied");
+
+patch(OrderReceipt.prototype, {
+  setup() {
+    super.setup();
+    this.pos = usePos();
+    console.log("POS Receipt Patch Loaded");
```
bit-odoo: This should not be pushed.

pos_formate/static/src/order_receipt_patch.js
```diff
@@ -0,0 +1,44 @@
...
+  },
+
+  get receiptLogo() {
+    const logo = this.pos.config?.receipt_logo;
+    console.log("Receipt Logo Data:", logo ? "Logo Found" : "No Logo");
```
bit-odoo: This should not be pushed.

pos_formate/static/src/order_receipt_patch.js
```diff
@@ -0,0 +1,44 @@
...
+    return logo ? `data:image/png;base64,${logo}` : false;
+  },
+
+  get receiptLayout() {
+    const layout = this.pos.config?.receipt_layout || "light";
+    console.log("Receipt Layout:", layout);
```
bit-odoo: This should not be pushed.

PR #941, opened 2025-09-15
19.0 tutorial admac

macocianradu:


estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
+from odoo import models, fields
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate Property"
+
+    name = fields.Char("Estate name", required=True, translate=True)
+    description = fields.Text("Description", translate=True)
```
cgun-odoo: "Description" is already the default string here (it capitalizes the field name). So no need to write it again
```suggestion
    description = fields.Text()
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
...
+    postcode = fields.Char("Postcode")
```
```suggestion
    postcode = fields.Char()
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
...
+    date_availability = fields.Date("Available From")
+    expected_price = fields.Float("Expected Price")
```
```suggestion
    expected_price = fields.Float()
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
...
+    selling_price = fields.Float("Actual Price")
+    bedrooms = fields.Integer("Bedrooms")
+    living_area = fields.Integer("Living Area")
+    facades = fields.Integer("Facades")
+    garage = fields.Boolean("Garage")
+    garden = fields.Boolean("Garden")
+    garden_area = fields.Integer("Garden Area")
```
cgun-odoo: same

estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
...
+    _name = "estate.property"
+    _description = "Estate Property"
+
+    name = fields.Char("Estate name", required=True, translate=True)
+    description = fields.Text("Description", translate=True)
+    postcode = fields.Char("Postcode")
+    date_availability = fields.Date("Available From")
+    expected_price = fields.Float("Expected Price")
+    selling_price = fields.Float("Actual Price")
+    bedrooms = fields.Integer("Bedrooms")
```
```suggestion
    bedrooms = fields.Integer()
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
...
+    # Bedrooms
+    # Living area
+    # Expected price
+    # selling price
+    # available from
+    # Description
+    # facades
+    # garage yes/no
+    # garden yes/no
+    # total area
```
cgun-odoo: Don't forget to delete the comments

.gitignore
```diff
@@ -109,6 +109,8 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+# Used by pyright linter
+pyrightconfig.json
```
cgun-odoo: Don't push .gitignore to the repo. You can exclude it from git.

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0"?>
+<odoo>
+    <menuitem id="estate_menu_root" name="Estate">
+        <menuitem id="estate_first_level_menu" name="Advertisments">
```
cgun-odoo:
```suggestion
        <menuitem id="estate_first_level_menu" name="Advertisements">
```
You can install an extension like Code Spell Checker. It always helps :smile:

estate/models/estate_property.py
```diff
@@ -8,15 +9,30 @@ class EstateProperty(models.Model):
     name = fields.Char("Estate name", required=True, translate=True)
     description = fields.Text()
     postcode = fields.Char()
-    date_availability = fields.Date("Available From")
+    date_availability = fields.Date(
+        "Available From", default=date.today() + timedelta(days=91), copy=False
```
cgun-odoo: Right now you're giving a value as default. This will run once when you boot your server up, calculate the value and use that value every time a new record is created.
What you need to do here instead is to pass a function as default so that every time a record is created the function will be executed again to calculate a new value.

estate/models/estate_property.py
```diff
@@ -0,0 +1,47 @@
...
+    garden_area = fields.Integer()
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
+        selection=[
+            ("new", "New"),
+            ("offer_received", "Offer Received"),
+            ("offer_accepted", "Offer Accepted"),
+            ("sold", "Sold"),
+            ("cancelled", "Cancelled"),
+        ],
```
```suggestion
        [
            ("new", "New"),
            ("offer_received", "Offer Received"),
            ("offer_accepted", "Offer Accepted"),
            ("sold", "Sold"),
            ("cancelled", "Cancelled"),
        ],
```
This is just a suggestion what you did is correct. For selection fields the first element is always the selection values, you don't have to use the keyword.

estate/views/estate_property_views.xml
```diff
@@ -5,4 +5,89 @@
...
+                <filter
+                    string="Available"
+                    name="available"
+                    domain="['|',
+                        ('state', '=', 'new'),
+                        ('state', '=', 'offer_received')]" />
+                <filter
+                    string="Postcode"
+                    name="postcode"
+                    context="{'group_by':'postcode', 'residual_visible':True}"/>
```
cgun-odoo: I think you should have this inside a group tag. Does it look correct without it?

estate/views/estate_property_views.xml
```diff
@@ -5,4 +5,89 @@
...
+            <form string="Estate Property">
+                <sheet>
+                    <div class="oe_title p-3">
+                        <div class="oe_edit_only">
+                            <label for="name"/>
+                        </div>
+                        <h1>
+                            <field name="name"/>
+                        </h1>
+                        <field name="active" invisible="1"/>
```
cgun-odoo: I would move the invisible fields on top of the sheet to separate it from the other fields. But it doesn't matter what you did is also perfect

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="estate_property_tag_action" model="ir.actions.act_window">
+        <field name="name">Property Tag</field>
+        <field name="res_model">estate.property.tag</field>
+        <field name="view_mode">list,form</field>
+    </record>
```
cgun-odoo, line 7: Where are the views though :rofl:

estate/security/ir.model.access.csv
```diff
@@ -1,3 +1,4 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,0
 estate.access_estate_property_type,access_estate_property_type,estate.model_estate_property_type,base.group_user,1,1,1,0
+estate.access_estate_property_tag,access_estate_property_tag,estate.model_estate_property_tag,base.group_user,1,1,1,0
```
```suggestion
access_estate_property_tag,access_estate_property_tag,model_estate_property_tag,base.group_user,1,1,1,0
```
No need to repeat the module name since we are in the same module.

estate/models/estate_property.py
```diff
@@ -45,3 +45,4 @@ class EstateProperty(models.Model):
     property_type_id = fields.Many2one("estate.property.type", string="Property Type")
     buyer_id = fields.Many2one("res.partner", string="Buyer")
     seller_id = fields.Many2one("res.users", string="Salesman")
+    tag_ids = fields.Many2many("estate.property.tag", string="Tags")
```
cgun-odoo: This is an R&D practice don't blame me if it doesn't make sense:

All user facing strings are in double quotes and all technical strings are in single quotes so this one would be

`tag_ids = fields.Many2many('estate.property.tag', string="Tags")`

```suggestion
    tag_ids = fields.Many2many("estate.property.tag")
```
The default name for this field should be Tags.

cgun-odoo:
The commit messages should also include the name of the module that you're changing like:
[ADD] estate: add onchange trigger for garden

estate/models/estate_property.py
```diff
@@ -2,6 +2,9 @@
 from datetime import date, timedelta
 
 from odoo.exceptions import UserError
+from odoo.tools import float_compare, float_is_zero
+
+from odoo.orm.utils import ValidationError
```
```suggestion
from odoo.exceptions import ValidationError
```

estate/models/estate_property.py
```diff
@@ -50,9 +53,32 @@ class EstateProperty(models.Model):
...
+
+    _check_unique_property_name = models.Constraint(
+            'UNIQUE(name)',
+            "The property name must be unique")
+
+    @api.constrains('selling_price', 'expected_price')
+    def _check_selling_price(self):
+        for record in self:
+            if not float_is_zero(record.selling_price,
+                                 precision_digits=2):
```
cgun-odoo, line 73: Why do you need to check this?

macocianradu: The exercise mentions it: 
"Add a constraint so that the selling price cannot be lower than 90% of the expected price.

Tip: the selling price is zero until an offer is validated. You will need to fine tune your check to take this into account."

Tbf I didn't check if it works without the check

macocianradu: Just checked without it and it won't let me create new properties without the check

estate/__manifest__.py
```diff
@@ -5,9 +5,10 @@
     "data": [
         "security/ir.model.access.csv",
         "views/estate_property_views.xml",
-        "views/estate_property_type_views.xml",
         "views/estate_property_offer_views.xml",
+        "views/estate_property_type_views.xml",
         "views/estate_property_tag_views.xml",
         "views/estate_menus.xml",
+        "views/user_views.xml"
```
cgun-odoo:
```suggestion
        "views/res_user_views.xml"
```
That's how it's named everywhere else

estate/models/estate_property.py
```diff
@@ -112,3 +114,10 @@ def action_cancelled(self):
                 raise UserError("Cannot cancel a sold property")
             record.state = 'cancelled'
         return True
+
+    @api.ondelete(at_uninstall=False)
+    def _delete_check(self):
+        for record in self:
+            if record.state != 'new' and record.state != 'cancelled':
+                raise UserError("Cannot delete a property that received offers"
+                                "Please first cancel the property")
```
cgun-odoo: Crud methods should be above action methods https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions

estate/models/estate_property.py
```diff
@@ -112,3 +114,10 @@ def action_cancelled(self):
                 raise UserError("Cannot cancel a sold property")
             record.state = 'cancelled'
         return True
+
+    @api.ondelete(at_uninstall=False)
+    def _delete_check(self):
+        for record in self:
+            if record.state != 'new' and record.state != 'cancelled':
```
```suggestion
            if record.state not in ['new', 'cancelled']
```

estate/models/estate_property_offer.py
```diff
@@ -43,3 +52,16 @@ def action_accept(self):
     def action_refuse(self):
         for record in self:
             record.status = 'refused'
+
+    @api.model
+    def create(self, vals_list):
+        for val in vals_list:
```
cgun-odoo:
```suggestion
        for vals in vals_list:
```
it goes like vals_list -> vals -> val
vals_list refers to a list of dictionaries, every dictionary consists of multiple values.

estate/models/estate_property.py
```diff
@@ -112,3 +114,10 @@ def action_cancelled(self):
...
+                raise UserError("Cannot delete a property that received offers"
```
cgun-odoo: You can make these error messages translatable. 
Also you can use """ """ for multi line strings

estate/models/estate_property_offer.py
```diff
@@ -43,3 +52,16 @@ def action_accept(self):
...
+            if 'property_id' in val:
+                property_id = val['property_id']
+                property = self.env['estate.property'].browse(property_id)
```
cgun-odoo: Here you're executing browse (therefore running a select query), in every iteration. But you can already access the property ids before the loop. You can try creating the list (or maybe a map) of properties before the loop and use that inside the loop.

estate/models/estate_property_type.py
```diff
@@ -1,8 +1,19 @@
-from odoo import models, fields
+from odoo import models, fields, api
```
cgun-odoo: These should be in alphabetical order for runbot

estate/models/estate_property_type.py
```diff
@@ -1,8 +1,19 @@
...
 
 
 class EstatePropertyType(models.Model):
     _name = 'estate.property.type'
     _description = "Estate Property Type"
 
+    _order = 'sequence, name'
+
     name = fields.Char("Estate Property Type", required=True)
+    sequence = fields.Integer(default=1)
```
cgun-odoo: Is there any specific reason for default value? Doesn't it function the same way if the default value is 0?

estate/views/estate_property_type_views.xml
```diff
@@ -21,8 +22,27 @@
         <field name="model">estate.property.type</field>
         <field name="arch" type="xml">
             <form string="Property Type">
+                <header>
+                    <button
+                        name="%(estate.estate_property_offer_action)d"
+                        type="action"
+                        string="offers" />
```
cgun-odoo, line 30: You can also put an icon on this smart button to make it pretty

estate/views/estate_property_views.xml
```diff
@@ -96,13 +116,22 @@
...
+                                    invisible="not garden"
+                                    name="garden_orientation"/>
                                 <field name="total_area"/>
                             </group>
                         </page>
                         <page class="p-3" string="Offers">
-                            <field name="offer_ids" />
+                            <field
+                                name="offer_ids"
+                                readonly="state == 'offer_accepted' or
```
cgun-odoo, line 128: state in [...]

estate/views/estate_property_views.xml
```diff
@@ -72,14 +87,19 @@
                         <h1>
                             <field name="name"/>
                         </h1>
-                        <field name="tag_ids" widget="many2many_tags"/>
+                        <field
+                            name="tag_ids"
+                            widget="form.many2many_tags"
```
cgun-odoo: was it not working without `form` :open_mouth:

estate/views/estate_property_views.xml
```diff
@@ -72,14 +87,19 @@
...
+                            options="{'color_field': 'color'}"/>
                     </div>
                     <div class="row p-3">
                         <group class="col">
-                            <field name="property_type_id"/>
+                            <field
+                                name="property_type_id"
+                                options="{'no_create': true}"
+                                can_write="false" />
```
cgun-odoo, line 97: What is this for? I've never seen it before :laughing:

estate/views/estate_property_views.xml
```diff
@@ -59,8 +65,17 @@
...
-                    <button name="action_cancelled" type="object" string="Cancel" />
+                    <button
+                        name="action_sold"
+                        type="object"
+                        invisible="state == 'sold' or state == 'cancelled'"
+                        string="Sold" />
+                    <button
+                        name="action_cancelled"
+                        type="object"
+                        invisible="state == 'sold' or state == 'cancelled'"
```
cgun-odoo: state in ['sold', 'cancelled'] same as the one above

estate/views/estate_property_views.xml
```diff
@@ -4,20 +4,25 @@
...
+            decoration-muted="state == 'sold'"
+            string="estate_property_list">
                 <field name="name"/>
                 <field name="postcode"/>
                 <field name="bedrooms"/>
                 <field name="living_area" string="Living Area (sqm)"/>
                 <field name="expected_price"/>
                 <field name="selling_price" string="Selling Price"/>
-                <field name="date_availability"/>
+                <field name="date_availability" optional="True"/>
```
cgun-odoo, line 25: iirc this doesn't work unless it's optional="hidden". Let me know if that changes anything or if it was already working

estate/views/estate_property_views.xml
```diff
@@ -4,20 +4,25 @@
...
     <record id="estate_property_view_list" model="ir.ui.view">
         <field name="name">estate.property.list</field>
         <field name="model">estate.property</field>
         <field name="arch" type="xml">
-            <list string="estate_property_list">
+            <list
+            decoration-success="state == 'offer_received' or state == 'offer_accepted'"
+            decoration-bf="state == 'offer_accepted'"
+            decoration-muted="state == 'sold'"
+            string="estate_property_list">
```
cgun-odoo, line 18: The string shouldn't be technical. It will be represented to the user.

estate/models/estate_property_offer.py
macocianradu: But I'm taking the property_id from val, which is different in every loop iteration. 

I tried doing vals_list['property_id'], so no loop would be necessary (like in the tutorial), but it throws an exception because vals_list is a list of dicts.

estate/views/estate_property_views.xml
macocianradu: I don't remember why I added it. But the no_create is to not allow the creation of new property types in the estate form. They can only be created in the property type list.

The can write is so they can't be edited in the form.

estate/__manifest__.py
```diff
@@ -1,4 +1,6 @@
 {
+    "author": "admac@odoo.com",
```
cgun-odoo: For everything you do at work the author is Odoo :smiling_imp:

estate/models/estate_property.py
```diff
@@ -101,23 +100,26 @@ def _onchange_garden(self):
             self.garden_area = 0
             self.garden_orientation = ''
 
+    @api.ondelete(at_uninstall=False)
+    def _delete_check(self):
```
cgun-odoo: ondelete methods usually start with `_unlink`
You should also give it a more specific name like `_check_status`

estate/models/estate_property_offer.py
```diff
@@ -55,13 +56,13 @@ def action_refuse(self):
...
+        properties = self.env['estate.property'].browse(property_ids)
+        for property in properties:
+            if property.state == 'new':
                 property.state = 'offer_received'
-                for offer in property.offer_ids:
-                    if offer.price > self.price:
-                        raise UserError("Cannot create new offer when other "
-                                        "higher offers exist")
+            for offer in property.offer_ids:
+                if offer.price > self.price:
```
cgun-odoo: Since this is the create method you don't have self yet (it should be an empty recordset)
Instead of iterating over properties you should iterate over vals_list and get property with respect to the vals.
Then, you can use vals['price'] here

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,39 @@
...
+        if not self.env['account.move'].check_access('create'):
+            try:
+                self.check_access('write')
+            except AccessError:
+                return self.env['account.move']
+
+        partner_id = self.buyer_id
+        journal = self.env['account.journal'].search([
+                *self.env['account.journal']._check_company_domain(self.env.company),
+                ('type', '=', 'sale')], limit=1)
```
cgun-odoo: Really nice

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,39 @@
...
+        property_name = self.name
```
cgun-odoo: that doesn't make it more concise or clear imo.

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,39 @@
...
+
+        invoice_vals = {'partner_id': partner_id.id,
+                        'move_type': 'out_invoice',
+                        'journal_id': journal.id,
```
cgun-odoo: It also gets the default sale journal if you don't give it a value :smile:

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,39 @@
...
+
+class EstateProperty(models.Model):
+    _inherit = 'estate.property'
+
+    def action_sold(self):
+        if not self.env['account.move'].check_access('create'):
+            try:
+                self.check_access('write')
+            except AccessError:
+                return self.env['account.move']
```
cgun-odoo: So, what you're doing in this method is: If the user has write access on account.move they are able to create account moves without having create access.
Is that what you want? Snippets like this one are always a bit sus, you should usually let ORM do it's job when it comes to access rights

estate/models/estate_property_offer.py
macocianradu: but that's how it was in the first place and you said I should change it so that I don't run the query inside the loop and instead to loop over the properties 😄

estate/models/estate_property_offer.py
```diff
@@ -58,11 +58,13 @@ def action_refuse(self):
...
+                    lambda p: p.id == vals['property_id'],
+                    properties), None)
             if property.state == 'new':
                 property.state = 'offer_received'
             for offer in property.offer_ids:
-                if offer.price > self.price:
-                    raise UserError(_("""Cannot create new offer when other
-                                    higher offers exist"""))
+                if offer.price > vals['price']:
+                    raise UserError(_("""Cannot create new offer when other higher offers exist"""))
```
cgun-odoo, line 69: Well done, nice, amazing :goat:

cgun-odoo (commented):
:clap:

PR #944, opened 2025-09-15
trcaz - Tutorial PR

trcazier:


estate/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from . import models
```
cgun-odoo: Add an empty line at the end of every file

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+        'base',
+    ],
+    'installable': True,
+    'application': True,
+    'data': [
+        'security/ir.model.access.csv',
+        'views/estate_property_views.xml',
+        'views/estate_menus.xml', # depends on estate_property_views.xml
+    ]
+}
```
cgun-odoo: Add an empty line at the end of every file

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+
```
cgun-odoo:
```suggestion
```

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    'name': 'Estate',
+    'depends': [
+        'base',
+    ],
+    'installable': True,
+    'application': True,
+    'data': [
+        'security/ir.model.access.csv',
+        'views/estate_property_views.xml',
+        'views/estate_menus.xml', # depends on estate_property_views.xml
```
cgun-odoo: Comments like this can be nice but this is not maintainable because the menus will depend on every view you add so you would need to update this comment every time you add a new view :smile: 

I think if the dependency is not easy to understand adding a comment is important but in this case not really necessary.

estate/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from . import estate_property
```
cgun-odoo: Add an empty line at the end of every file

estate/models/estate_property.py
```diff
@@ -0,0 +1,29 @@
+from odoo import fields, models
+from datetime import date, timedelta
+
+class Property(models.Model):
```
cgun-odoo, line 7: Runbot might complain that you should have 2 empty lines between imports and class signature 
```suggestion

class Property(models.Model):
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,29 @@
...
+    garden_area        = fields.Integer()
+    garden_orientation = fields.Selection(
+        string='Orientation',
+        selection=[("north","North"), ("south", "South"), ("east","East"), ("west", "West")])
+
+    active             = fields.Boolean(default=True)
+    state              = fields.Selection(
+        string='State',
+        selection=[("new", "New"), ("offerreceived", "Offer Received"), ("offeraccepted", "Offer accepted"), ("sold", "Sold"), ("cancelled", "Cancelled")],
+        default="new"
```
cgun-odoo: The way you structured it looks nice imo, but it's not how it's done in other parts of odoo. And likely the runbot is going to complain about the unnecessary white spaces.

estate/models/estate_property.py
```diff
@@ -0,0 +1,29 @@
...
+
+class Property(models.Model):
+    _name = "estate.property"
+    _description = "Test description for estate.property model"
+
+    name               = fields.Char(required=True)
+    expected_price     = fields.Float(required=True)
+    description        = fields.Text()
+    postcode           = fields.Char()
+    date_availability  = fields.Date(copy=False, default=date.today() + timedelta(days=90))
```
cgun-odoo: Here you're passing default a value. This value is interpreted once by the ORM when your module is loaded as `today + 90 days`. So, the value you're giving here will never change throughout the runtime of your application.
What you want to do here is to pass default a function that returns `today + 90 days`.

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,1
```
cgun-odoo: Add one empty line at the end of the file

```suggestion
access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
```
Since you're already in the estate module, you don't need to use the module name to refer to its records.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,90 @@
...
+            <form string="Property form test string?">
+                <sheet>
+                    <div class="oe_title">
+                        <div class="oe_edit_only">
+                            <label for="name"/>
+                        </div>
+                        <h1 class="mb32">
+                            <field name="name" class="mb16"/>
+                        </h1>
+                        <field name="active" invisible="1"/>
```
cgun-odoo: If invisible fields are not used only at one place I would prefer putting them on top of the form separate from all other fields. But it's not a problem this is perfect

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,90 @@
...
+        <field name="arch" type="xml">
+            <form string="Property form test string?">
+                <sheet>
+                    <div class="oe_title">
+                        <div class="oe_edit_only">
+                            <label for="name"/>
+                        </div>
+                        <h1 class="mb32">
+                            <field name="name" class="mb16"/>
+                        </h1>
```
cgun-odoo, line 75: Show me how this looks on your page it looks a bit weird

estate/models/estate_property.py
```diff
@@ -0,0 +1,29 @@
...
+    garage             = fields.Boolean()
+    garden             = fields.Boolean()
+    garden_area        = fields.Integer()
+    garden_orientation = fields.Selection(
+        string='Orientation',
+        selection=[("north","North"), ("south", "South"), ("east","East"), ("west", "West")])
+
+    active             = fields.Boolean(default=True)
+    state              = fields.Selection(
+        string='State',
```
cgun-odoo: The default string is already State. No need to specify it again

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,90 @@
...
+                <field name="living_area" />
+                <field name="facades" />
+                <filter string="Available" name="available" domain="['|', ('state', '=', 'new'), ('state', '=', 'offerreceived')]"/>
+                <group expand="0" string="Group By">
+                    <filter string="Postcode" name="group_by_postcode" domain="[]" context="{'group_by':'postcode'}"/>
+                </group>
+            </search>
+        </field>
+    </record>
+</odoo>
```
cgun-odoo: Add one empty line at the end of the file

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,90 @@
...
+        <field name="arch" type="xml">
+            <search string="Estate property search string?">
+                <field name="name" />
+                <field name="postcode" />
+                <field name="expected_price" />
+                <field name="bedrooms" />
+                <field name="living_area" />
+                <field name="facades" />
+                <filter string="Available" name="available" domain="['|', ('state', '=', 'new'), ('state', '=', 'offerreceived')]"/>
+                <group expand="0" string="Group By">
```
cgun-odoo: Wait does this work :open_mouth: 
We had a problem with lecan in the morning and it was caused by the attributes in group

estate/models/estate_property.py
```diff
@@ -0,0 +1,29 @@
...
+        selection=[("new", "New"), ("offerreceived", "Offer Received"), ("offeraccepted", "Offer accepted"), ("sold", "Sold"), ("cancelled", "Cancelled")],
```
```suggestion
        selection=[("new", "New"), ("offer_received", "Offer Received"), ("offer_accepted", "Offer accepted"), ("sold", "Sold"), ("cancelled", "Cancelled")],
```
use snake case for values with multiple words

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,90 @@
...
+                    <filter string="Postcode" name="group_by_postcode" domain="[]" context="{'group_by':'postcode'}"/>
```
cgun-odoo:
```suggestion
                    <filter string="Postcode" name="group_by_postcode" context="{'group_by':'postcode'}"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,90 @@
...
+        </field>
+    </record>
+
+    <record id="estate_property_view_form" model="ir.ui.view">
+        <field name="name">estate.property.view.form</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form string="Property form test string?">
+                <sheet>
+                    <div class="oe_title">
```
cgun-odoo, line 69: Why is everything inside an oe_title div :smile:

estate/__manifest__.py
trcazier: I was doing that for myself in order to remember it since this is a tutorial. Should I delete the comments?

cgun-odoo: No of course not :smile: I'm trying to get you used to the nitpicking that will happen on your PR's after the onboarding.

estate/models/estate_property.py
```diff
@@ -7,6 +7,12 @@ class Property(models.Model):
 
     name               = fields.Char(required=True)
     expected_price     = fields.Float(required=True)
+
+    _check_expected_price = models.Constraint(
+        "CHECK(expected_price > 0)",
+        "The expected price must be strictly positive"
+    )
```
cgun-odoo: https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions
Sql constraints should come after field declarations

estate/models/estate_property.py
```diff
@@ -56,4 +68,20 @@ def _compute_total_area(self):
...
+                record.state = "cancelled"
+        return True
+
+    def action_property_sold(self):
+        for record in self:
+            if record.state == "cancelled":
+                raise exceptions.UserError("A cancelled property cannot be sold.")
+            else:
+                record.state = "sold"
+        return True
```
cgun-odoo: Have an empty line at the end of the file

estate/models/estate_property.py
```diff
@@ -1,4 +1,4 @@
-from odoo import api, fields, models
+from odoo import api, exceptions, fields, models
```
cgun-odoo: You can do
`from odoo.exceptions import UserError`
since you're only using UserError in this class

estate/models/estate_property.py
```diff
@@ -56,4 +68,20 @@ def _compute_total_area(self):
...
\ No newline at end of file
+            record.best_price = max([o.price for o in record.offer_ids]) if len(record.offer_ids) > 0 else 0
+
+    def action_property_cancel(self):
+        for record in self:
+            if record.state == "sold":
+                raise exceptions.UserError("A sold property cannot be cancelled.")
+            else:
+                record.state = "cancelled"
+        return True
```
cgun-odoo, line 79: I'm just writing this to have a discussion please comment your thoughts about it.

What this method mainly does is cancelling properties, so the happy path of the method should end with `record.state = 'cancelled'`. Hiding the happy path under else might make the method less readable (at least for longer methods than this one).
What do you think?

estate/models/estate_property_offer.py
```diff
@@ -1,18 +1,31 @@
...
+
+    _check_price = models.Constraint(
+        "CHECK (price > 0)",
+        "The price must be strictly positive"
+    )
+
+    @api.constrains("price")
+    def _check_price_2(self):
+        for record in self:
+            if float_compare(record.price, 0.90*record.property_id.expected_price, precision_digits=2) == -1:
```
cgun-odoo:
```suggestion
            if float_compare(record.price, 0.9 * record.property_id.expected_price, precision_digits=2) == -1:
```
Also what if record.property_id is None seems like you would get an error have you thought about it?

estate/models/estate_property.py
```diff
@@ -56,4 +68,20 @@ def _compute_total_area(self):
...
     def _compute_best_price(self):
         for record in self:
-            record.best_price = max(map(lambda o: o.price, record.offer_ids)) if len(record.offer_ids) > 0 else 0
\ No newline at end of file
+            record.best_price = max([o.price for o in record.offer_ids]) if len(record.offer_ids) > 0 else 0
+
+    def action_property_cancel(self):
+        for record in self:
+            if record.state == "sold":
+                raise exceptions.UserError("A sold property cannot be cancelled.")
```
cgun-odoo: I guess it's not yet mentioned in the tutorial but we have a method  `_()` that makes strings translatable. Rule of thumb is to always call this method when you're raising an error.
```suggestion
                raise exceptions.UserError(_("A sold property cannot be cancelled."))
```

estate/models/estate_property.py
```diff
@@ -56,4 +68,20 @@ def _compute_total_area(self):
     @api.depends("offer_ids")
     def _compute_best_price(self):
         for record in self:
-            record.best_price = max(map(lambda o: o.price, record.offer_ids)) if len(record.offer_ids) > 0 else 0
\ No newline at end of file
+            record.best_price = max([o.price for o in record.offer_ids]) if len(record.offer_ids) > 0 else 0
```
```suggestion
            record.best_price = max([o.price for o in record.offer_ids]) if record.offer_ids else 0
```

A list in python (or a recordset) is falsy if it's empty and truthy if it's not. So you don't have to check the length like you would do in other languages.

estate/models/estate_property.py
```diff
@@ -56,4 +68,20 @@ def _compute_total_area(self):
...
+
+    def action_property_cancel(self):
+        for record in self:
+            if record.state == "sold":
```
```suggestion
            if record.state == 'sold':
```
R&D string rule

estate/models/estate_property_offer.py
```diff
@@ -1,18 +1,31 @@
...
 
     price         = fields.Float()
+
+    _check_price = models.Constraint(
+        "CHECK (price > 0)",
+        "The price must be strictly positive"
+    )
+
+    @api.constrains("price")
+    def _check_price_2(self):
```
cgun-odoo: _check_price_2 is an awful name

estate/models/estate_property_offer.py
```diff
@@ -21,7 +34,23 @@ def _computed_date_deadline(self):
...
+                raise exceptions.UserError("The selling price cannot be lower than 90% of the expected price")
+            else:
+                record.status = "accepted"
+                record.property_id.buyer_id = record.partner_id
+                record.property_id.selling_price = record.price
+                
+   
+    def action_offer_refuse(self):
+        for record in self:
+            record.status = "refused"
```
cgun-odoo: Since you're assigning every record to the same value you can just do this:
```suggestion
        self.status = 'refused'
```

cgun-odoo: + line at the end of file

estate/models/estate_property_tag.py
```diff
@@ -4,4 +4,9 @@ class PropertyTag(models.Model):
...
     _description = "Test description for estate.property.tag model"
 
-    name               = fields.Char(required=True)
\ No newline at end of file
+    name = fields.Char(required=True)
+
+    _check_name = models.Constraint(
+        "UNIQUE (name)",
+        "Property tag name must be unique"
+    )
```
cgun-odoo: EOF line

estate/models/estate_property_type.py
```diff
@@ -4,4 +4,9 @@ class PropertyType(models.Model):
     _name = "estate.property.type"
     _description = "Test description for estate.property.type model"
 
-    name               = fields.Char(required=True)
\ No newline at end of file
+    name               = fields.Char(required=True)
```
cgun-odoo:
```suggestion
    name = fields.Char(required=True)
```

estate/models/estate_property_type.py
```diff
@@ -4,4 +4,9 @@ class PropertyType(models.Model):
...
+
+    _check_name = models.Constraint(
+        "UNIQUE (name)",
+        "Property type name must be unique"
+    )
```
cgun-odoo: EOF line

estate/models/estate_property.py
trcazier: Very sad because it's way easier to read :(

estate/views/estate_property_views.xml
trcazier: I'll change it in my next commit, tell me if I understood correctly

estate/views/estate_property_views.xml
trcazier: I copied it from somewhere (I don't remember)
It looks like this
image

estate/views/estate_property_views.xml
trcazier: It did not indeed, and we solved that together yesterday

estate/views/estate_property_views.xml
trcazier: Don't ask me, I did what I found in the code base example 😅

estate/models/estate_property.py
trcazier: Now that I got Ruff working properly, the code has changed, I applied the style warnings and this code has change, let me know what you think.

estate/models/estate_property_offer.py
trcazier: (This constraint doesn't seem to work for me anyway 😅)
Can a property offer have no property id? Isn't it populated from the one2many automatically? When creating an offer from the property form?

estate/models/estate_property.py
trcazier: It says it's undefined and I can't find how to import it

estate/models/estate_property.py
trcazier: I come from a strongly typed language and JS and Python scare me precisely because of stuff like this, if something wrong happens in the code, like say offer_ids becomes an integer, then the if will succeed because of the implicit conversions instead of just raising an error... this seems kind of dangerous to me

estate/models/estate_property.py
trcazier: For all the strings or just for "values" like that?

estate/models/estate_property.py
cgun-odoo: `from odoo import _`

estate/models/estate_property_offer.py
cgun-odoo: Yes as you said it should always be filled, so no problem here as long as property_id is a required field.
Let's check the constraint if it's still not working

estate/models/estate_property.py
cgun-odoo: here if offer_ids becomes an integer you'll still get an error as you can't have something like `for o in record.offer_ids`
you're right to be concerned but you're the one writing the method you shouldn't convert a recordset to an integer

estate/models/estate_property.py
cgun-odoo: I thought I explained it in your PR as well sorry.
This is an R&D Practice. They put technical strings in single quotes and user facing strings in double quotes.

estate/models/estate_property.py
```diff
@@ -79,16 +65,22 @@ def _onchange_garden(self):
             self.garden_area = None
             self.garden_orientation = None
 
+    @api.ondelete(at_uninstall=False)
+    def _prevent_deletion_unless_new_or_cancelled(self):
```
```suggestion
    def _unlink_prevent_deletion_unless_new_or_cancelled(self):
```
couldn't find a rule for that in the guidelines but most ondelete methods start with unlink

estate/models/estate_property_offer.py
```diff
@@ -39,19 +39,26 @@ def _inverse_date_deadline(self):
...
-                raise exceptions.UserError("The selling price cannot be lower than 90% of the expected price")
+                raise exceptions.UserError(_("The selling price cannot be lower than 90% of the expected price"))
             record.status = "accepted"
             record.property_id.buyer_id = record.partner_id
             record.property_id.state = 'offer_accepted'
             record.property_id.selling_price = record.price
 
+    @api.model_create_multi
+    def create(self, vals_list):
+        for record in vals_list:
```
cgun-odoo:
```suggestion
        for vals in vals_list:
```
These are not records yet so the name can be confusing

estate/models/estate_property_offer.py
```diff
@@ -39,19 +39,26 @@ def _inverse_date_deadline(self):
...
+            property = self.env['estate.property'].browse(record['property_id'])
+            property.state = 'offer_received'
```
cgun-odoo: Here you're executing `browse` in every iteration.
Instead you can browse for the ids before the loop and even do everything without the loop.

estate/models/res_users.py
```diff
@@ -0,0 +1,7 @@
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    property_ids = fields.One2many("estate.property", "salesperson_id", string="Properties")
```
cgun-odoo: Properties might be the default string here. But double check

estate/models/estate_property_offer.py
trcazier: But I'm using `vals/record` to browse and it changes with every iteration...?

estate/models/estate_property.py
```diff
@@ -5,21 +5,21 @@
...
-        selection=[("new", "New"), ("offer_received", "Offer Received"), ("offer_accepted", "Offer accepted"), ("sold", "Sold"), ("cancelled", "Cancelled")],
-        default="new",
+        selection=[('new', "New"), ('offer_received', "Offer Received"), ('offer_accepted', "Offer accepted"), ('sold', "Sold"), ('cancelled', "Cancelled")],
+        default='new',
     )
     description = fields.Text()
     postcode = fields.Char()
     selling_price = fields.Float(copy=False, readonly=True)
-    date_availability = fields.Date(copy=False, default=date.today() + timedelta(days=90))
+    date_availability = fields.Date(copy=False, default=lambda _: date.today() + timedelta(days=90))
```
cgun-odoo: Using '_' here for the unused parameter name is correct for python conventions. However, in odoo there is a commonly used function _() to translate strings. So you almost never see `lambda _:` used.

estate/models/estate_property.py
```diff
@@ -28,59 +28,59 @@ class Property(models.Model):
...
-        selection=[("north", "North"), ("south", "South"), ("east", "East"), ("west", "West")])
+        selection=[('north', "North"), ('south', "South"), ('east', "East"), ('west', "West")])
     active = fields.Boolean(default=True)
-    buyer_id = fields.Many2one("res.partner", string="Buyer")
-    salesperson_id = fields.Many2one("res.users", string="Salesperson", copy=False, default=lambda self: self.env.user)
-    tags_ids = fields.Many2many("estate.property.tag", string="Tags")
-    offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers")
-    total_area = fields.Float(compute="_compute_total_area")
-    best_price = fields.Float(compute="_compute_best_price")
+    buyer_id = fields.Many2one('res.partner', string="Buyer")
```
cgun-odoo: default string

estate/models/estate_property.py
```diff
@@ -28,59 +28,59 @@ class Property(models.Model):
...
+    salesperson_id = fields.Many2one('res.users', string="Salesperson", copy=False, default=lambda self: self.env.user)
```
cgun-odoo: default string

estate/models/estate_property.py
```diff
@@ -28,59 +28,59 @@ class Property(models.Model):
...
+    tags_ids = fields.Many2many('estate.property.tag', string="Tags")
```
```suggestion
    tag_ids = fields.Many2many('estate.property.tag')
```
ids is enough to make it plural

estate/models/estate_property.py
```diff
@@ -28,59 +28,59 @@ class Property(models.Model):
...
+    offer_ids = fields.One2many('estate.property.offer', 'property_id', string="Offers")
```
cgun-odoo: default string

estate/models/estate_property.py
```diff
@@ -28,59 +28,59 @@ class Property(models.Model):
...
             self.garden_area = 10
-            self.garden_orientation = "north"
+            self.garden_orientation = 'north'
         else:
             self.garden_area = None
             self.garden_orientation = None
 
     @api.ondelete(at_uninstall=False)
-    def _prevent_deletion_unless_new_or_cancelled(self):
+    def _unlink_prevent_deletion_unless_new_or_cancelled(self):
```
```suggestion
    def _unlink_check_status(self):
```
try to be more concise

estate/models/estate_property_offer.py
```diff
@@ -38,19 +35,30 @@ def _inverse_date_deadline(self):
...
+            record.property_id.buyer_id = record.partner_id
+            record.property_id.state = 'offer_accepted'
+            record.property_id.selling_price = record.price
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            property = self.env['estate.property'].browse(vals['property_id'])
+            property.state = 'offer_received'
+        return super().create(vals_list)
```
cgun-odoo: CRUD method should be above action method

estate/models/estate_property_type.py
```diff
@@ -1,12 +1,23 @@
...
-    _name = "estate.property.type"
+    _name = 'estate.property.type'
     _description = "Test description for estate.property.type model"
+    _order = 'name'
 
-    name               = fields.Char(required=True)
+    name = fields.Char(required=True)
+    sequence = fields.Integer('Sequence', default=1, help="Used to order types.")
+    property_ids = fields.One2many('estate.property', 'property_type_id', string="Properties")
+    offers_ids = fields.One2many('estate.property.offer', 'property_type_id', string="Offers")
```
cgun-odoo:
```suggestion
    offer_ids = fields.One2many('estate.property.offer', 'property_type_id', string="Offers")
```
one plural is enough

estate/models/estate_property_type.py
```diff
@@ -1,12 +1,23 @@
...
-    )
\ No newline at end of file
+        'UNIQUE (name)',
+        "Property type name must be unique",
+    )
+
+    @api.depends('offers_ids')
+    def _compute_offer_count(self):
+        for record in self:
+            record.offer_count = len(record.offers_ids or [])
```
cgun-odoo:
```suggestion
            record.offer_count = len(record.offers_ids)
```
record.offer_ids is always a recordset. No need to be afraid of it being `None`

estate/views/estate_property_type_views.xml
```diff
@@ -8,22 +8,39 @@
...
         <field name="model">estate.property.type</field>
         <field name="arch" type="xml">
             <form string="Property type form test string?">
                 <sheet>
+                    <div class="oe_button_box" name="button_box">
+                        <button type="action"
+                            class="oe_stat_button"
+                            name="%(estate.action_estate_property_offer_model)d" 
+                            icon="fa-bars"
+                            invisible="offer_count == 0"
```
cgun-odoo, line 41: Nice touch, didn't see this in other PR's :goat:

estate/views/estate_property_type_views.xml
```diff
@@ -8,22 +8,39 @@
...
+                            string="Offers"/>
```
cgun-odoo, line 42:
```suggestion
                            string="Offer(s)"/>
```
It can be one offer :rock:

estate/views/estate_property_type_views.xml
```diff
@@ -32,6 +49,13 @@
                             <field name="name" class="mb16"/>
                         </h1>
                     </div>
+                    <field name="property_ids" string="Title">
+                        <list>
+                            <field name="name" string="Title"/>
+                            <field name="expected_price" string="Expected Price"/>
```
cgun-odoo: No need for the string it's the default value

estate/views/estate_property_views.xml
```diff
@@ -4,22 +4,52 @@
...
+                        </div>
+                        <div>
+                            <div t-if="record.selling_price.raw_value > 0">
+                                Selling price: <field name="selling_price"/>
+                            </div>
+                            <div t-if="record.best_price.raw_value != 0">
+                                Best offer: <field name="best_price"/>
+                            </div>
+                            <div>
+                                <field name="tags_ids"/>
```
cgun-odoo: I missed this above, rename the field to tag_ids. No need to have plural_plural

estate/views/estate_property_views.xml
```diff
@@ -29,9 +59,11 @@
         <field name="model">estate.property</field>
         <field name="arch" type="xml">
             <form string="Property form test string?">
+                <field name="active" invisible="1"/>
                 <header>
-                    <button name="action_property_cancel" type="object" string="Cancel" />
-                    <button name="action_property_sold" type="object" string="Sold" />
+                    <button name="action_property_cancel" type="object" string="Cancel" invisible="state == 'cancelled' or state == 'sold'"/>
```
cgun-odoo:
```suggestion
                    <button name="action_property_cancel" type="object" string="Cancel" invisible="state in ['cancelled', 'sold']"/>
```
You can have any python expression here

estate/views/estate_property_views.xml
```diff
@@ -41,17 +73,14 @@
                         <h1>
                             <field name="name" class="mb16"/>
                         </h1>
-                        <field name="tags_ids" widget="many2many_tags"/>
+                        <field name="tags_ids" widget="form.many2many_tags" options="{'color_field': 'color'}"/>
```
cgun-odoo: Why do you need form. here was it not working?

estate_account/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+
+{
+    'name': 'Estate Account',
+    'depends': [
+        'base',
+        'estate',
+        'account',
+    ],
+    'version': '19.0.0.0',
+    'installable': True,
```
cgun-odoo, line 11: Why is installable True

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,27 @@
...
+                    'quantity': 1,
+                    'price_unit': 0.06 * self.selling_price,
+                }),
+                Command.create({
+                    'name': "Administrative fees",
+                    'quantity': 1,
+                    'price_unit': 100,
+                }),
+            ],
+        }
```
cgun-odoo: Really great job with this method. Just one thing to add, since I'm in Ps-Tech seeing unextendable methods annoys me.
I think it's best to create a method like `get_sold_property_invoice_values` that returns this object. If you do that I can create a module named `estate_account_romania` in the future and add a specific romanian tax to the account move. 
If you don't create a new method I would need to rewrite this method and copy everything to the module I create. And then I would need to maintain that module also the customers need to pay for loc in their customization.

estate/models/estate_property.py
trcazier: So what should I name it instead?

estate/models/estate_property.py
trcazier: That name is a bit vague
Idk I've always been told that code should be self-commenting and that names should be meaningful even if too long

estate/views/estate_property_type_views.xml
trcazier: Ewww

estate/views/estate_property_views.xml
trcazier: I don't remember :innocent:

estate_account/__manifest__.py
trcazier: It was in the other manifests I looked at

estate/models/estate_property.py
cgun-odoo: anything 😆 most people use `lambda self`

estate/tests/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from . import (
+    common,
+    test_estate_property,
+    test_estate_property_offer,
+)
```
cgun-odoo:
```suggestion
from . import common
from . import test_estate_property
from . import test_estate_property_offer
```

estate/tests/common.py
```diff
@@ -0,0 +1,79 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        cls.partner = cls.env['res.partner'].create({
+            'name': 'Test Buyer',
+            'email': 'buyer@example.com',
+        })
+
+        cls.property_type = cls.env['estate.property.type'].search([('name', '=', "House")])
```
cgun-odoo: What if you run the test on an empty database?
Make sure your test can be run in any environment.
You should always assume that your test does not have more than what you create.

estate/tests/common.py
```diff
@@ -0,0 +1,79 @@
...
+                'tag_ids': [(6, 0, [cls.tag2.id])],
+            },
+        ])
+
+        # Create offers for only 2 properties (leave one without offers)
+        cls.offers = cls.env['estate.property.offer'].create([
+            {
+                'price': 260000,
+                'partner_id': cls.partner.id,
+                'property_id': cls.properties[0].id,  # Cozy Cottage
```
cgun-odoo: Instead of referring to with `properties[0]` and adding a comment every time, I would prefer just creating the properties separately like:
`cls.cozy_cottage = cls.env['...'].create(...)`

cgun-odoo: This would also help me navigate through your tests if I'm extending them. I could just ctrl +click on `self.cozy_cottage` and see its value

estate/tests/test_estate_property.py
```diff
@@ -0,0 +1,28 @@
...
+
+        self.properties[0].action_property_sold()
+
+        self.assertRecordValues(self.properties, [
+           {'name': "Cozy Cottage", 'state': 'sold'},
+           {'name': "Modern Apartment", 'state': 'offer_received'},
+           {'name': "Beachfront Villa", 'state': 'new'},
+        ])
+
+        with self.assertRaises(UserError, msg="A property with no accepted offer cannot be sold"):
```
cgun-odoo, line 20: 🤔 I think including the message here doesn't really help the tests and it makes your code harder to maintain. Up for discussion of course, what do you think?

estate/tests/test_estate_property.py
```diff
@@ -0,0 +1,28 @@
+from odoo.exceptions import UserError
+from odoo.tests import tagged
+
+from .common import EstateTestCommon
+
+
+# The CI will run these tests after all the modules are installed,
+# not right after installing the one defining it.
+@tagged('post_install', '-at_install')
+class EstateTestProperty(EstateTestCommon):
```
cgun-odoo: Weird that the file name is test_estate_property but class name is EstateTestProperty

estate/tests/test_estate_property_offer.py
```diff
@@ -0,0 +1,31 @@
...
+# The CI will run these tests after all the modules are installed,
+# not right after installing the one defining it.
+@tagged('post_install', '-at_install')
+class EstateTestPropertyOffer(EstateTestCommon):
+    def test_create(self):
+        """Test that an offer cannot be created for a sold property."""
+
+        # Accept one offer to be able to sell the property
+        self.offers[0].action_offer_accept()
+        # Sell the property to create an offer on it
```
cgun-odoo: Isn't the offer already created? 🫨

cgun-odoo (commented):
One suggestion for you: Create unit_test for estate_account where you override test_action_property_sold and confirm that an account move is created.

estate/tests/test_estate_property_offer.py
trcazier: Yes but the test is to create an offer on a sold property so I have to 1. create the property 2. create an offer for said property 3. accept it 4. sell the property 5. create an offer on the sold property to see if it rightfully creates the error

estate/tests/test_estate_property.py
trcazier: The file I took as example must not have respected the conventions

estate/tests/test_estate_property.py
trcazier: The documentation of assertRaises is a bit lackluster: "An optional keyword argument 'msg' can be provided when assertRaises is used as a context object." 
I thought it meant that the exception had to have that message to be caught, I don't know if that's really the case, but in any case, yeah maybe it's harder to maintain, but if it works like I think it does it makes the test more reliable since only an exception with that message will be caught and not the other one for example, which reduces the chance the test gives a false positive

estate/tests/common.py
trcazier: Then I don't know how to do it because when I create it gives an error because it's already in the db

estate/tests/test_estate_property_offer.py
cgun-odoo: Ah okay I thought you were calling the sold method so that it automatically creates an offer.

cgun-odoo: Maybe this is a sign to update the comment so that it's *clearer*

estate/tests/test_estate_property.py
cgun-odoo: I don't think there is a convention it just looked weird to me

estate/tests/test_estate_property.py
cgun-odoo: Yeah that's also right if your method doesn't have any side effects and if it calls other methods that also raise `UserError` having the message should help :game_die:

estate/tests/common.py
cgun-odoo: Create it for the empty db, and give it a name like `test_house_type` so that it doesn't touch your constraint

estate_account/tests/test_estate_property.py
```diff
@@ -0,0 +1,17 @@
+from odoo.fields import Domain
+from odoo.tools import float_compare
+
+from odoo.addons.estate.tests.test_estate_property import TestEstateProperty
+
+
+class TestEstateAccountProperty(TestEstateProperty):
+    def test_action_property_sold(self):
```
cgun-odoo, line 8: You might want to change the name

How we usually run tests on local is with --test-tags .[TEST_NAME] if two tests have the same name it will only run the one from the module that was loaded later.

estate_account/tests/test_estate_property.py
```diff
@@ -0,0 +1,17 @@
...
+        """Test that everything behaves like it should when selling a property."""
+
+        self._sell_cozy_cottage()
+
+        self.invoice = self.env['account.move'].search(Domain('partner_id', '=', self.cozy_cottage.buyer_id.id) & Domain('move_type', '=', 'out_invoice'))
+
+        self.assertTrue(self.invoice)
+        self.assertEqual(float_compare(self.invoice.invoice_line_ids[0].price_unit, 0.06 * self.cozy_cottage.selling_price, precision_digits=2), 0)
```
cgun-odoo: I wouldn't trust the order in which `invoice_line_ids` are fetched. You probably will never have an issue with it but it would be more robust if you obtained the lines with something *you* defined.

cgun-odoo: So what can happen (not likely but possible in theory) is the cache might have one of the invoice_line_ids and not have the other one. If that happens the one in cache will be the first one ([0]).

cgun-odoo: What I would do in this case is sort the lines according to their `price_unit` -> because you already determine that the tax will be 6% of the other one. After sorting you can use it.

cgun-odoo: This is reaaaaally unlikely to happen but keep it in mind when you access unsorted records in a recordset

cgun-odoo (commented):
Nice job 💣

estate_account/tests/test_estate_property.py
trcazier: Well I named it like that because you told me I had to override it :rofl:

cgun-odoo: :sleeping:

PR #965, opened 2025-09-22
[IMP] awesome_owl: Technical onboarding, chapter 1, step 1 to 5

fmalfroid:
Technical onboarding Part 2: Discover the web framework

### DIFF
--- a/awesome_owl/static/src/counter/counter.js
+++ b/awesome_owl/static/src/counter/counter.js
@@ -1,0 +1,14 @@
+import { Component, useState } from "@odoo/owl";
+
+export class Counter extends Component {
+    static template = "awesome_owl.counter";
+    static props = {};
+
+    setup() {
+        this.state = useState({ value: 0 });
+    }
+
+    increment() {
+        this.state.value++;
+    }
+}
--- a/awesome_owl/static/src/card/card.js
+++ b/awesome_owl/static/src/card/card.js
@@ -1,0 +1,6 @@
+import { Component } from "@odoo/owl";
+
+export class Card extends Component {
+    static template = "awesome_owl.card";
+    static props = { title: String, content: String };
+}
--- a/awesome_owl/static/src/card/card.xml
+++ b/awesome_owl/static/src/card/card.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<templates xml:space="preserve">
+
+    <t t-name="awesome_owl.card">
+        <div class="card d-inline-block m-2" style="width: 18rem;">
+            <div class="card-body">
+                <h5 class="card-title"><t t-out="props.title"/></h5>
+                <p class="card-text"><t t-out="props.content"/></p>
+            </div>
+        </div>
+    </t>
+
+</templates>
--- a/awesome_owl/static/src/counter/counter.xml
+++ b/awesome_owl/static/src/counter/counter.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<templates xml:space="preserve">
+
+    <t t-name="awesome_owl.counter">
+        <div class="m-2 p-2 border d-inline-block">
+            <span class="me-2">Counter: <t t-esc="state.value"/></span>
+            <button class="btn btn-primary" t-on-click="increment">Increment</button>
+        </div>
+    </t>
+
+</templates>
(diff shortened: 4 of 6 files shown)


awesome_dashboard/static/src/dashboard/dashboard_item/dashboard_item.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+
+    <t t-name="awesome_dashboard.DashboardItem">
+        <div class="card m-2 d-inline-block" t-att-style="'width: ' + (18 * (props.size || 1)) + 'rem;'">
```
cgun-odoo: For the default value you can also use defaultProps = { size: 1} in the js file

awesome_dashboard/static/src/dashboard/number_card/number_card.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+
+    <t t-name="awesome_dashboard.NumberCard">
+        <div class="card-header">
+            <h6 t-esc="props.title"></h6>
+        </div>
+        <div class="card-body d-flex align-items-center justify-content-center">
+            <h1 class="text-success" t-esc="props.value || 0"></h1>
```
cgun-odoo: here from what you have 'value' is a required prop. So you don't need to have the ||

awesome_dashboard/static/src/dashboard/pie_chart_card/pie_chart_card.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+
+    <t t-name="awesome_dashboard.PieChartCard">
+        <div class="card-header">
+            <h6 t-esc="props.title"></h6>
+        </div>
+        <div class="card-body d-flex align-items-center justify-content-center">
+            <PieChart data="props.value || {}"/>
```
cgun-odoo: Same again, value is required if you don't put optional: true

awesome_dashboard/static/src/dashboard/dashboard.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    openCustomers = () => {
+        this.action.doAction("base.action_partner_form");
+    }
+
+    openLeads = () => {
+        this.action.doAction({
+            type: 'ir.actions.act_window',
+            name: 'Leads',
```
cgun-odoo: You can make this translatable as well

awesome_owl/static/src/counter/counter.js
```diff
@@ -0,0 +1,19 @@
...
+        onChange: Function
+    };
+
+    setup() {
+        this.state = useState({ value: 0 });
+    }
+
+    increment() {
+        this.state.value++;
+        if ("onChange" in this.props) {
```
cgun-odoo:
```suggestion
        if (this.props.onChange) {
```
Since props is never undefined you can do it in this shorter way

awesome_owl/static/src/todo_list/todo_list.js
```diff
@@ -0,0 +1,38 @@
...
+    static components = { TodoItem };
+    static props = {};
+
+    setup() {
+        this.todos = useState([]);
+        useAutofocus("todoInput");
+    }
+
+    addTodo(ev) {
+        if (ev.key === "Enter" && ev.target.value.trim() !== "") {
```
cgun-odoo, line 16: nice touch with the trim() 🐐

awesome_owl/static/src/todo_list/todo_list.js
```diff
@@ -0,0 +1,38 @@
...
+                isCompleted: false
+            });
+            ev.target.value = "";
+        }
+    }
+
+    setTodoCompletion(todo, ev) {
+        if (ev.target.checked != todo.isCompleted) {
+            todo.isCompleted = ev.target.checked;
+        }
```
cgun-odoo: Why do you need the 'if'? If they are the same they will stay the same 😆

awesome_owl/static/src/utils.js
```diff
@@ -0,0 +1,9 @@
+import { useRef, useEffect } from "@odoo/owl";
+
+export function useAutofocus(name) {
+    let ref = useRef(name);
+    useEffect(
+        (el) => el && el.focus(),
+        () => [ref.el]
+    );
```
cgun-odoo: So if i'm not mistaken what this will do is it will focus on the element everytime you changed the referenced element. Is it possible to change the referenced element 🤔. If you don't see a case for it, I don't think we need to cover it and imo this should be done with onMount instead for simplicity.

PR #972, opened 2025-09-23
[ADD] todo: add new todo, card and counter components

macocianradu:


awesome_owl/static/src/card/card.js
```diff
@@ -0,0 +1,15 @@
...
+export class Card extends Component {
+    static template = "awesome_owl.card";
+
+    static props = ['title', 'slots']
+
+    setup() {
+        this.state = useState({ hidden: false })
+    }
+
+    hide() {
```
cgun-odoo: this method can also show the Card :thinking: Perhaps it needs another name

awesome_owl/static/src/counter/counter.js
```diff
@@ -0,0 +1,21 @@
+import { Component, useState } from "@odoo/owl";
+
+export class Counter extends Component {
+    static template = "awesome_owl.counter";
+    static props = {
+        onChange: {
+            type: Function,
+            optional: true}
```
cgun-odoo:
```suggestion
            optional: true
        }
```

awesome_owl/static/src/playground.js
```diff
@@ -1,7 +1,27 @@
...
 export class Playground extends Component {
     static template = "awesome_owl.playground";
+    static components = { TodoList, Counter, Card }
+
+    setup() {
+        this.state = useState({sum: 2})
+    }
+
+    incrementSum() {
+        console.log('call')
```
cgun-odoo: :call_me_hand:

awesome_owl/static/src/todo/todo.js
cgun-odoo, line 1: Nice :1st_place_medal:

awesome_owl/static/src/todo/todo_item.js
```diff
@@ -0,0 +1,18 @@
+import { Component } from "@odoo/owl";
+import { Todo } from "./todo";
+import { Card } from "../card/card";
+
+export class TodoItem extends Component {
+    static template = "awesome_owl.todo_item";
+    static components = { Card };
+
+    static props = { todo: { type: Todo }, onRemove: { type: Function } }
```
cgun-odoo: make sure it doesn't let you leave onRemove empty `i'm not sure if by default it's required`

awesome_owl/static/src/todo/todo_item.js
```diff
@@ -0,0 +1,18 @@
...
+
+    toggle() {
+        this.props.todo.isCompleted = !this.props.todo.isCompleted;
+    }
+
+    remove() {
+        this.props.onRemove(this.props.todo.id)
```
cgun-odoo: or check `if (this.props.onRemove)` here if it's not required

awesome_owl/static/src/todo/todo_list.js
```diff
@@ -0,0 +1,31 @@
...
+import { TodoItem } from "./todo_item";
+import { Todo } from "./todo";
+import { Card } from "../card/card";
+import { useAutofocus } from "../utils";
+
+export class TodoList extends Component {
+    static template = "awesome_owl.todo_list";
+    static components = { Card, TodoItem };
+
+    ids = 0
```
cgun-odoo: :eyes: Does this work when it's outside of setup?

awesome_owl/static/src/todo/todo_list.js
```diff
@@ -0,0 +1,31 @@
...
+
+    setup() {
+        this.state = useState({ todos: [], input: '' })
```
cgun-odoo: I always read todos as if it's spanish :rofl: I think todo_list might be better

awesome_owl/static/src/todo/todo_list.js
```diff
@@ -0,0 +1,31 @@
...
+        useAutofocus('todo_input')
+    }
+
+    keyup(event) {
+        if (event.keyCode === 13) {
+            this.state.todos.push(new Todo(this.ids++, event.srcElement.value, false))
```
cgun-odoo: is event.srcElement.value same with event.target.value?

awesome_dashboard/static/src/dashboard/dashboard_item.js
```diff
@@ -0,0 +1,13 @@
+import { Component } from "@odoo/owl";
+
+export class DashboardItem extends Component {
+    static template = "awesome_dashboard.DashboardItem"
+
+    static props = {'size': {type: Number, optional: true }}
+
+    setup() {
+        if(!this.props.size) {
```
cgun-odoo:
```suggestion
        if (!this.props.size) {
```

awesome_dashboard/static/src/dashboard/dashboard_item.js
```diff
@@ -0,0 +1,13 @@
...
+            this.props.size = 1
```
cgun-odoo:
```suggestion
            this.props.size = 1;
```

awesome_dashboard/static/src/dashboard.js
```diff
@@ -1,17 +1,23 @@
...
+import { rpc } from "@web/core/network/rpc";
 
 class AwesomeDashboard extends Component {
     static template = "awesome_dashboard.AwesomeDashboard";
-    static components = { Layout };
+    static components = { Layout, DashboardItem };
 
     setup() {
         this.action = useService("action");
+        this.state = useState({stats: []})
```
cgun-odoo: This will be an object after the rpc request. So why not have it as an empty object `{}` instead of an empty array

awesome_dashboard/static/src/dashboard/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,27 @@
...
+    static template = "awesome_dashboard.PieChart";
+
+    static props = {'data':{optional: false}, 'size': {type: Number, optional: true}}
+
+    setup() {
+        onWillStart(() => loadJS(["/web/static/lib/Chart/Chart.js"]));
+        this.ctx = useRef('pieChart')
+        this.chart = null
+        useEffect(() => {
+            if(this.chart) {
```
cgun-odoo:
```suggestion
            if (this.chart) {
```

awesome_dashboard/static/src/dashboard/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,27 @@
...
+
+export class PieChart extends Component {
+    static template = "awesome_dashboard.PieChart";
+
+    static props = {'data':{optional: false}, 'size': {type: Number, optional: true}}
+
+    setup() {
+        onWillStart(() => loadJS(["/web/static/lib/Chart/Chart.js"]));
+        this.ctx = useRef('pieChart')
+        this.chart = null
```
cgun-odoo:
```suggestion
        this.ctx = useRef('pieChart');
        this.chart = null;
```

awesome_dashboard/static/src/dashboard/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,27 @@
+import { Component, onWillStart, useEffect, useRef } from "@odoo/owl"
+import { loadJS } from "@web/core/assets"
+
+export class PieChart extends Component {
+    static template = "awesome_dashboard.PieChart";
+
+    static props = {'data':{optional: false}, 'size': {type: Number, optional: true}}
```
cgun-odoo: Should be split into multiple lines

awesome_dashboard/static/src/dashboard/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,27 @@
...
+                this.chart.destroy()
+            }
+            const dataset = {
+                labels: Object.keys(this.props.data),
+                datasets: [{
+                    label: 'T-Shirt Size Distribution',
+                    data: Object.values(this.props.data)
+                }]
+            }
+            this.chart = new Chart(this.ctx.el, {type: 'pie', data: dataset})
```
cgun-odoo: Why didn't you initialize it inside setup() instead of using useEffect(). Since you don't give it any dependencies this useEffect will also only run when the component is mounted. It is okay since you want this only once but you could maybe use something else

awesome_owl/static/src/todo/todo_item.js
macocianradu: yea optional is set by default to false

awesome_owl/static/src/todo/todo_list.js
macocianradu: it does, but I will move it 🫠

awesome_owl/static/src/todo/todo_list.js
macocianradu: kinda, but I think target is the standard so I'll change

awesome_dashboard/static/src/dashboard/pie_chart/pie_chart.js
macocianradu: I'm using lazy loading for the chart. If I extract the code directly in the setup() then Chart won't be loaded yet and it will run into a "Chart is undefined" error

macocianradu: I changed it to onMounted

awesome_dashboard/static/src/dashboard/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,31 @@
+import { Component, onWillStart, useRef, onMounted } from "@odoo/owl";
+import { loadJS } from "@web/core/assets";
+
+export class PieChart extends Component {
+    static template = "awesome_dashboard.PieChart";
+
+    static props = {
+        data: { optional: false },
```
cgun-odoo: optional: false is already the default value. But I assume you added this just because of the props validation.

awesome_dashboard/static/src/dashboard/configuration_dialog.js
```diff
@@ -0,0 +1,34 @@
+import { Component } from "@odoo/owl";
+import { Dialog } from "@web/core/dialog/dialog";
+import { CheckBox } from "@web/core/checkbox/checkbox";
+
+export class ConfigurationDialog extends Component {
+    static template = "awesome_dashboard.ConfigurationDialog";
+    static components = { Dialog, CheckBox };
+    static props = {
+        items: [Array],
```
cgun-odoo:
```suggestion
        items: Array,
```
You only need a list when you allow more than one type e.g. [Array, Function],

awesome_dashboard/static/src/dashboard/configuration_dialog.js
```diff
@@ -0,0 +1,34 @@
...
+        close: [{}],
+        onApply: [Function],
```
cgun-odoo: Don't need array

awesome_dashboard/static/src/dashboard/configuration_dialog.js
```diff
@@ -0,0 +1,34 @@
...
+            .filter((item) => !item.enabled)
+            .map((item) => item.element.id);
+    }
+
+    onChange(id) {
+        let index = this.disabledElements.indexOf(id);
+        if (index === -1) {
+            this.disabledElements.push(id);
+            return;
+        }
```
cgun-odoo, line 30: What your function does is: 
- If item is already disabled, remove it from the list
- If item is not disabled add it to disabled list
Thus, this if condition is actually half of your functionality. But when you return inside the block it looks like it's a corner case and for most records what will execute is `this.disabledElements.splice(index, 1);` which is not true.

In this case I would prefer to have
```suggestion
        if (index === -1) {
            this.disabledElements.push(id);
        }
        else {
            do other thing
        }
```
so what your method does is clearer.
This is of course up for debate. What do you think?

awesome_dashboard/static/src/dashboard/dashboard.js
```diff
@@ -0,0 +1,71 @@
...
+    setup() {
+        this.action = useService("action");
+        this.statistics = useService("awesome_dashboard.statistics");
+        this.state = useState({ items: [], stats: this.statistics });
+        this.dialog = useService("dialog");
+        this.updateDashboard();
+    }
+
+    updateDashboard() {
+        let disabledElements = localStorage.getItem("disabledElements") || "";
```
cgun-odoo:
```suggestion
        const disabledElements = localStorage.getItem("disabledElements") || "";
```
I'd rather keep this a const and create another variable for the array

awesome_dashboard/static/src/dashboard/dashboard.js
```diff
@@ -0,0 +1,71 @@
...
+        if (disabledElements.includes(",")) {
+            disabledElements = disabledElements.split(",");
+        } else {
+            disabledElements = [disabledElements];
+        }
+        this.state.items = registry
+            .category("awesome_dashboard")
+            .getAll()
+            .filter((el) => !disabledElements.some((e) => e == el.id));
```
cgun-odoo, line 38: you can check `.includes()`

awesome_dashboard/static/src/dashboard/dashboard.js
```diff
@@ -0,0 +1,71 @@
...
+            views: [
+                [false, "list"],
+                [false, "form"],
+            ],
+        });
+    }
+
+    customizeDashboard() {
+        let items = [];
+        let all = registry.category("awesome_dashboard").getAll();
```
cgun-odoo:
```suggestion
        const all = registry.category("awesome_dashboard").getAll();
```
also a find a better name because `all` is not descriptive

awesome_dashboard/static/src/dashboard/dashboard.js
```diff
@@ -0,0 +1,71 @@
...
+        all.forEach((el) => {
+            const enabled = this.state.items.some((item) => item.id == el.id);
+            items.push({
+                element: el,
+                enabled: enabled,
+            });
```
cgun-odoo: Could you also use disabledElements here and never need the items state variable 🤔

awesome_dashboard/static/src/dashboard/dashboard.xml
```diff
@@ -0,0 +1,26 @@
...
+            <t t-foreach="state.items" t-as="item" t-key="item.id">
+                <DashboardItem size="item.size || 1">
+                    <t t-set="itemProp" t-value="item.props
+                           ? item.props(state.stats)
+                           : {'data': state.stats}" />
+                    <t t-component="item.Component" t-props="itemProp" />
+                </DashboardItem>
+            </t>
+        </div>
+    </Layout>
```
cgun-odoo: indent

awesome_dashboard/static/src/dashboard/dashboard_item.js
```diff
@@ -0,0 +1,16 @@
...
+    static props = {
+        size: { type: Number, optional: true },
+        slots: {},
+    };
+
+    setup() {
+        if (!this.props.size) {
+            this.props.size = 1;
+        }
+    }
```
cgun-odoo: static defaultProps = { size: 1 }

awesome_dashboard/static/src/dashboard/number_card.js
```diff
@@ -0,0 +1,12 @@
+import { Component, xml } from "@odoo/owl";
+
+export class NumberCard extends Component {
+    static template = xml`
+        <p><t t-esc="props.title" /></p>
+        <h1><b><t t-esc="props.value" /></b></h1>
+    `;
+    static props = {
+        title: [String],
```
cgun-odoo: don't need list

awesome_dashboard/static/src/dashboard/pie_chart_card.js
```diff
@@ -0,0 +1,15 @@
+import { Component, xml } from "@odoo/owl";
+import { PieChart } from "./pie_chart/pie_chart";
+
+export class PieChartCard extends Component {
+    static components = { PieChart };
+    static props = {
+        title: [String],
+        value: [Object],
```
cgun-odoo: don't need list

awesome_dashboard/static/src/dashboard/configuration_dialog.js
macocianradu: I guess it's a matter of preference.
I usually prefer to do early returns instead of else statements, so that if the function gets more complicated, I don't need to nest so much.
In this case it's probably the same.

PR #993, opened 2025-10-20
QUCOL Onboarding

qucol-odoo:


estate/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+from . import estate_properties
+from . import estate_property_types
+from . import estate_property_tags
+from . import estate_property_offers
```
alan-odoo: They should be listed alphabetically.

estate/models/estate_property.py
```diff
@@ -0,0 +1,96 @@
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools import float_compare
+
+DEFAULT_GARDEN_AREA = 10
+DEFAULT_GARDEN_ORIENTATION = "north"
+
```
alan-odoo: That's the right way to write constants but I am not sure to understand why it is required. They are used only in _update_garden_area_and_orientation, can they be defined directly in that method?

estate/models/estate_property.py
```diff
@@ -0,0 +1,96 @@
...
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
+    )
+    total_living_area = fields.Integer(compute="_compute_total_area")
+
```
alan-odoo: No need of that blank line 😃.

estate/models/estate_properties.py
```diff
@@ -0,0 +1,96 @@
...
+    tag_ids = fields.Many2many("estate.property.tag")
+    offer_ids = fields.One2many("estate.property.offer", "property_id")
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_living_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids")
+    def _get_highest_price(self):
```
alan-odoo: As this method is used to compute the best_price, it should be named _compute_best_offer 😃, that way we know easily 1. it computes field and 2 which field. 

FYI, _get or get methods should return an object and should not modify the database. For example, _get_res_user should only return a res_user and do nothing else.

estate/models/estate_property.py
```diff
@@ -0,0 +1,96 @@
...
+        for record in self:
+            record.total_living_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids")
+    def _get_highest_price(self):
+        for record in self:
+            record.best_offer = max(record.offer_ids.mapped("price")) if record.offer_ids else 0
+
+    @api.onchange("garden")
+    def _update_garden_area_and_orientation(self):
```
alan-odoo: Same here.

estate/models/estate_property.py
```diff
@@ -0,0 +1,96 @@
...
+        for record in self:
+            if record.garden:
+                record.garden_area = DEFAULT_GARDEN_AREA
+                record.garden_orientation = DEFAULT_GARDEN_ORIENTATION
+            else:
+                record.garden_area = 0
+                record.garden_orientation = None
+
+    def mark_as_sold(self):
```
alan-odoo: As this method is called by a button, the name should start by action. You can name it action_mark_as_sold for example.

estate/models/estate_property.py
```diff
@@ -0,0 +1,96 @@
...
+        for record in self:
+            if record.state == "cancelled":
+                raise UserError("A cancelled property cannot be set as sold.")
+            record.state = "sold"
+        return True
+
+    def mark_as_cancelled(self):
```
alan-odoo: Same here.

estate/models/estate_properties.py
```diff
@@ -0,0 +1,96 @@
...
+            if record.state == "sold":
+                raise UserError("A sold property cannot be set as cancelled.")
+            record.state = "cancelled"
+        return True
+
+    @api.constrains("selling_price", "expected_price")
+    def _check_selling_price(self):
+        for record in self:
+            if record.selling_price and float_compare(record.selling_price, record.expected_price * .9, 0) == -1:
+                raise ValidationError(r"The selling price cannot be lower than 90% of the expected price.")
```
alan-odoo: Is it mandatory to add a raw string ? Just asking

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,42 @@
...
+    @api.depends("validity")
+    def _compute_deadline(self):
+        for record in self:
+            record.date_deadline = fields.Date.add((record.create_date or fields.Datetime.now()), days=record.validity)
+
+    def _inverse_deadline(self):
+        for record in self:
+            record.validity = (record.date_deadline - fields.Date.to_date(record.create_date)).days if record.date_deadline else record.validity
+
+    def accept_offer(self):
```
alan-odoo: Same as for the other action methods, the name should start by action 😃.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,42 @@
...
+        for record in self:
+            record.status = "accepted"
+            record.property_id.selling_price = record.price
+            record.property_id.buyer = record.partner_id
+
+    def refuse_offer(self):
```
alan-odoo: Same here.

estate/models/estate_property.py
```diff
@@ -0,0 +1,102 @@
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools import float_compare
+
+DEFAULT_GARDEN_AREA = 10
+DEFAULT_GARDEN_ORIENTATION = "north"
+
+
+class PropertyModel(models.Model):
```
alan-odoo, line 6: No need of Model inside the class name of a model, but I agree that it could have been. It is an Odoo convention.

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,48 @@
...
+        <field name="view_mode">list,form</field>
+    </record>
+
+    <record id="estate_property_type_form_view" model="ir.ui.view">
+        <field name="name">estate.property.type.form</field>
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <form string="Property type">
+                <sheet>
+                    <button name="%(estate.estate_property_offer_action)d" string="Offers" type="action" title="Offers list" invisible="not offer_count > 0"/>
```
alan-odoo: If you follow the exercise (you are not obliged to do so.), you should make a button with the class oe_stat_button. That's up to you but that way it will be displayed in the header of your form.

estate/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+from . import estate_property
+from . import estate_property_type
+from . import estate_property_tag
+from . import estate_property_offer
+from . import res_user
```
alan-odoo, line 6: Should be sorted alphabetically, that's a python convention:
```suggestion
from . import estate_property
from . import estate_property_offer
from . import estate_property_tag
from . import estate_property_type
from . import res_user
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,63 @@
...
+            record.date_deadline = fields.Date.add((record.create_date or fields.Datetime.now()), days=record.validity)
+
+    def _inverse_deadline(self):
+        for record in self:
+            record.validity = (record.date_deadline - fields.Date.to_date(record.create_date)).days if record.date_deadline else record.validity
+
+    @api.model
+    def create(self, vals_list: list[dict]):
+        for val in vals_list:
+            estate_property = self.env["estate.property"].browse(val["property_id"])
```
alan-odoo: You shouldn't do sql request inside a loop 😃. It will do a lot of requests to your db as the create method can be used to create multiple records. Get first all the needed ids and then make a request with all of them with_prefetch(), which will allow you to get records by their id with browse inside the loop without doing any requests to the data base.  Here also you can see that I used EstateProperties instead of estate_properties, that's an Odoo convention. When a variable represent a model we have to use the same convention than for the name of a python class. Sorry for that one, I have been a little hard on that one 😅. Just remember no SQL queries inside loops.
```suggestion
       EstateProperties = self.env["estate.property"].with_prefetch(
            [vals["property_id"] for vals in vals_list]
        )
        for val in vals_list:
            estate_property = EstateProperties.browse(val["property_id"])
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,63 @@
...
+            if any(offer.price > val["price"] for offer in estate_property.offer_ids):
```
alan-odoo:
```suggestion
            if val["price"] < min(estate_property.offer_ids.mapped('price')):
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,103 @@
...
+            ("cancelled", "Cancelled")
+        ],
+        string="Status",
+        required=True,
+        copy=False,
+        default="new"
+    )
+    property_type_id = fields.Many2one("estate.property.type")
+    buyer = fields.Many2one("res.partner", copy=False)
+    salesperson = fields.Many2one("res.users", default=lambda self: self.env.user)
```
alan-odoo: Should contain _id in the name as they are relationship fields.
```suggestion
    buyer_id = fields.Many2one("res.partner", copy=False)
    salesperson_id = fields.Many2one("res.users", default=lambda self: self.env.user)
```

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,48 @@
...
+        <field name="view_mode">list,form</field>
+    </record>
+
+    <record id="estate_property_type_view_form" model="ir.ui.view">
+        <field name="name">estate.property.type.form</field>
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <form string="Property type">
+                <sheet>
+                    <button name="%(estate.estate_property_offer_action)d" string="Offers" type="action" title="Offers list" invisible="not offer_count > 0"/>
```
alan-odoo:
```suggestion
                    <button name="%(estate.estate_property_offer_action)d" string="Offers" type="action" title="Offers list" invisible="offer_count == 0"/>
```

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,48 @@
...
+                    <h1>
+                        <field name="name"/>
+                    </h1>
+                    <group>
+                        <field name="offer_count"/>
```
alan-odoo, line 28: Maybe you can test to add it inside your oe_stat_button ? (not a priority). You can find a solution [here](/home/odoo/Documents/workspaces/odoo_workspace/odoo/community/addons/account/views/account_move_views.xml) (line 847)

estate/views/estate_res_user_views.xml
```diff
@@ -0,0 +1,21 @@
+<odoo>
+    <record id="estate_res_user_view_form" model="ir.ui.view">
```
alan-odoo: You can call it as its parent. Other modules will refer to this one as estate.res_user_view_form. So here we know that the main form of the user has been modified and this is not a new one.
```suggestion
    <record id="res_user_view_form" model="ir.ui.view">
```

estate/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+{
+    'name': 'Estate',
+    'depends': [
+        'base'
+    ],
+    'data': [
+        'security/ir.model.access.csv',
+
```
alan-odoo: No need of this blank line. Do not forget that a lot of developers are working on the project and we must avoid unnecessary changes.

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def _generate_invoice_name(self):
+        taken_ids = set()
+        for invoice in self.env["account.move"].search([]):
+            if (match := re.match(r"Invoice\s+(\d+)", invoice.name)):
+                taken_ids.add(int(match.group(1)))
+        i = 1
+        while i in taken_ids:
+            i += 1
+        return f"Invoice {i}"
```
alan-odoo: Do you want to select the first number missing in the sequence or the one above the maximum ?
If it is the last case, I would do something like that to avoid too many loops.
```suggestion
        invoice = self.env["account.move"].search([("name", "like", "Invoice %")], order='id desc', limit=1)
        if invoice and (match := re.match(r"Invoice (\d+)", invoice.name)):
            return f"Invoice {match.group(1)}"
        return "Invoice 1"
```

alan-odoo (commented):
Hi @qucol-odoo 👋,

Here is a review of your code. You went through the tutorials without much difficulty so I've been a little bit more touchy on your PR 😃.

Your branch is very clean! And we're not going to dirty it. Then can you do the corrections in the right commits ? In the end, your branch shouldn't contain new commit. If you meet some issue doing it, ping me 😃. 

Thanks for your job and for helping your teammates!

estate/tests/test_estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+            "state": "new",
+            "expected_price": 4000.0
+        }])
+        cls.env["estate.property.offer"].create([{
+            "partner_id": ADMINISTRATOR_ID,
+            "property_id": cls.sold_property.id,
+            "status": "accepted",
+            "price": 4000.0
+        }])
+        cls.sold_property.state = "sold"
```
alan-odoo: You should create functions/methods only if they are used several times or overridden. I must say that before Odoo I liked creating functions because according to me, it made the code more readable. But they asked me to stop 😅 because it creates a ping pong effect and you should be able to read as much code as possible at once.

estate/tests/test_estate_property.py
```diff
@@ -0,0 +1,116 @@
+from odoo.tests.common import TransactionCase
+from odoo.exceptions import UserError
+from odoo.tests import tagged, Form
+
+
+ADMINISTRATOR_ID = 3
```
alan-odoo: Data created in xml/csv have an external_id. This is the id of the xml record. It can be easily found. For example here you can search `model="res.partner"' in all your xml files and then you can check the demo data. 

When you need to cache a record by its id you can used the ref method of self.env. It contacts the ir_model_data table which contains all the external ids and with the other information of the record in that table its able to find where is the record asked. 

Also we usually do not write global constants. So you could just put it inside the setUpClass.

```suggestion
ADMINISTRATOR_ID = self.env.ref("base.partner_admin").id
```

estate/tests/test_estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+            {"partner_id": ADMINISTRATOR_ID, "property_id": self.accepted_property.id, "price": 3000},
+            {"partner_id": ADMINISTRATOR_ID, "property_id": self.cancelled_property.id, "price": 5000},
+        ]
+        with self.assertRaises(UserError, msg="Creating an offer for a sold property should raise a UserError but it did not."):
+            self.env["estate.property.offer"].create([offer_for_sold_property])
+        for offer in correct_offers:
+            try:
+                self.env["estate.property.offer"].create([offer])
+            except UserError:
+                self.fail()
```
alan-odoo: No need to catch the exception 😃, or maybe I'm missing something. If an exception occurs the test doesn't pass. As it can make a code a bit less readable, do not hesitate to write why you do that. Do not forget always why and not how as every developer should be able to read your code, but sometimes the purpose may not be clear.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,135 @@
...
+    <record id="estate_property_action_filter" model="ir.ui.view">
+        <field name="name">estate.property.search</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <search string="Search Properties">
+                <field name="name"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area" filter_domain="['|', ('living_area', '=', self), ('living_area', '&gt;', self)]"/>
```
alan-odoo: Could you do this?
```suggestion
                <field name="living_area" filter_domain="[('living_area', '&gt;=', self)]"/>
```

alan-odoo (commented):
Hi @qucol-odoo 😃,

Just a quick review. Those changes are not a priority as we are arriving at the end of the onboarding.  It's just to let you know about other good practices.

About the title of your PR, its pattern is the same as the one of the commits. [IMP/FIX/ADD/...]: applications: what the PR is doing (starting with a verb). Here for the applications you could write estate{_account}. We use {} to avoid estate, estate_account. We squash them. And if there are to much app, you just write * in the title and in the description you write * = all the names of the apps.  
And for the description you should summarize the content of your commits.  The description must mainly explain why you are doing your modifications and not how you reached the goal. 
Once again no need to change it now, it's just for you to know.

Cheers !

PR #994, opened 2025-10-20
YAGHO Onboarding

yagho-odoo:


dery-odoo (commented):
Good job keeping your branch green on the runbot ! 
Maybe you could try to squash your linting commits to keep the commit history clean ?

PR #1008, opened 2025-10-27
HKAFE - JS Tutorial

HashemKhaled:


awesome_dashboard/static/src/dashboard/dashboard.xml
```diff
@@ -0,0 +1,24 @@
...
+                    <DashboardItem size="item.size || 1">
+                        <t t-set="itemProp" t-value="item.props ? item.props(stats) : {'data': stats}"/>
+                        <t t-component="item.component" t-props="itemProp"/>
+                    </DashboardItem>
+                </t>
+            </div>
+        </Layout>
+    </t>
+
+</templates>
```
dery-odoo: The empty line is missing here, I've seen it in other files as well, make sure to check everything

awesome_dashboard/static/src/dashboard/dashboard_item/dashboard_item.js
```diff
@@ -0,0 +1,16 @@
+import { Component } from "@odoo/owl";
+
+export class DashboardItem extends Component {
+    static template = "awesome_dashboard.DashboardItem";
+    static props = {
+        size: {
+            type: Number,
+            default: 1,
```
dery-odoo: This is not the way to define a default value for the props, you can check this : https://github.com/odoo/owl/blob/master/doc/reference/props.md#default-props

awesome_dashboard/static/src/dashboard/dashboard.xml
```diff
@@ -0,0 +1,24 @@
...
+            <button type="button" class="btn btn-primary" t-on-click="openCustomers">
+                Customers
+            </button>
+            <div class="divider"/>
+            <button type="button" class="btn btn-primary" t-on-click="openLeads">
+                Leads
+            </button>
+            <div class="d-flex flex-wrap" t-if="stats.isLoaded">
+                <t t-foreach="items" t-as="item" t-key="item.id">
+                    <DashboardItem size="item.size || 1">
```
dery-odoo: If you define a default value in the component there is no need to define a value here if there's no item.size

awesome_dashboard/static/src/dashboard/statistics_service.js
```diff
@@ -0,0 +1,25 @@
+import { registry } from "@web/core/registry";
+import { rpc } from "@web/core/network/rpc";
+import { reactive } from "@odoo/owl";
+
+const statisticsService = {
+    async: ["loadStatistics"],
```
dery-odoo, line 6: Is this line necessary for the service ?

awesome_owl/__manifest__.py
```diff
@@ -35,7 +35,7 @@
             ('include', 'web._assets_bootstrap'),
             ('include', 'web._assets_core'),
             'web/static/src/libs/fontawesome/css/font-awesome.css',
-            'awesome_owl/static/src/**/*',
+            'awesome_owl/static/src/**/**',
```
dery-odoo: There's is no need to update the manifest here, the /**/* pattern match every folder and file recursively so even when creating new folder it will keep loading everything

awesome_dashboard/static/src/dashboard/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,52 @@
...
+            }
+        });
+    }
+
+    renderChart() {
+        const labels = Object.keys(this.props.data);
+        const values = Object.values(this.props.data);
+        const color = labels.map((_, idx) => getColor(idx * 2));
+        if (this.chart) {
+            this.chart.destroy();
```
dery-odoo: Have you seen a case where you have to destroy it here ? Owl should handle it in the onWillUnmount before rerendering it.

dery-odoo (changes requested):
Hey! I added a few comments. Overall, really solid work! Just double-check the empty line at the end of the file, but everything else looks great.

PR #1024, opened 2025-11-18
ngtpn - Technical Training

nguyenntph:


estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
+from odoo import fields, models
+from dateutil.relativedelta import relativedelta
```
clbr-odoo: Please follow our [coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#imports) regarding order of imports.

estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class Property(models.Model):
```
```suggestion
class EstateProperty(models.Model):
```
It won't have big incidence here but we generally put the name of the file here, in CamelCase.

estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
...
+    _name = "estate.property"
+    _description = "All properties"
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        copy=False,
+        default=fields.Datetime.today() + relativedelta(months=+3),
```
```suggestion
        default=fields.Date.add(fields.Date.today(), months=3),
```
1. Your code is (almost, see 2) correct, just know that we have some built-in helper for that.
It allows to avoid the import of relativedelta :)
2. You should use `fields.Date.today()` instead of Datetime since you have a `Date' type field. Otherwise you trigger an extra cast needlessly.

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,7 @@
+<odoo>
+  <menuitem id="test_menu_root" name="Real Estate">
```
clbr-odoo: I don't think there is a generic "Odoo rule" for that, but in my team we always indent with 4 spaces, even in XMLs.
And I guess the `test_` is not really necessary in those `xmlids`, otherwise you'll end up prepending `test` everywhere in this tutorial 😆

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+{
+    'name': "Real Estate",
+    'license': 'LGPL-3',
+    'version': '1.0',
+    'depends': ['base'],
+    'author': "ngtpn",
```
clbr-odoo: No problem in this tutorial, just know that when developping an Odoo official module, we must use `'author': 'Odoo S.A.',`.

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    'category': 'Category',
+    'description': """
+    Real Estate Advertisement module
+    """,
+    'data': [
```
clbr-odoo, line 12:
```suggestion
    'application': True,
    'data': [
```
You probably want to have application=True for this module. See the [manifest reference](https://www.odoo.com/documentation/19.0/developer/reference/backend/module.html)

clbr-odoo (changes requested):
Nice to see that you know of to rewrite your git history & force-push.
I advice to use the `--force-with-lease` option if you haven't, as a general security.

Very good job so far 👍 
Props for the clean indentation of the Python file 🥳

estate/models/estate_property.py
nguyenntph: Thanks for that. I was looking into the helper but missed this `fields.Date.add()`. This is now updated.

estate/__manifest__.py
nguyenntph: Good point. Thanks for sharing that.

estate/models/estate_property.py
nguyenntph: This is now updated. I just want to check if we have a convention for this. As in our [Chapter 3](https://www.odoo.com/documentation/19.0/developer/tutorials/server_framework_101/03_basicmodel.html) example [CRM Recurring Plan](https://github.com/odoo/odoo/blob/e80911aaead031e7523173789e946ac1fd27c7dc/addons/crm/models/crm_recurring_plan.py#L1-L9), we omit the add on name and just use the model for class name, so I was inferring that we have a specific naming convention, like add-on-name_model-name for file, and model-name only for class name.

estate/views/estate_menus.xml
nguyenntph: Good point. I notice I used mixed 2 and 4 spaces. This is now updated.

estate/models/estate_property.py
clbr-odoo: No it's not a strong official convention, more an employee tacit thing, and it's indeed not always the case.
Always remember that some part of the Odoo code are 10 years old, when there was 20 devs :wink:

estate/models/estate_property.py
```diff
@@ -0,0 +1,120 @@
...
+    def _compute_total_area(self):
+        for record in self:
+            living_area = record.living_area or 0
+            garden_area = record.garden_area or 0
+            record.total_area = living_area + garden_area
+
+    @api.depends('offer_ids')
+    def _compute_best_price(self):
+        for record in self:
+            record.best_price = max(record.offer_ids.mapped('price')) if self.offer_ids else 0
```
```suggestion
            record.best_price = max(record.offer_ids.mapped('price'), default=0.0)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,120 @@
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_is_zero, float_compare
+
+
+class EstateProperty(models.Model):
+    _name = 'estate.property'
+    _description = 'All properties'
```
```suggestion
    _description = 'Estate Properties'
```
Nitpick :innocent: 
Also description should always be in title case

estate/models/estate_property.py
```diff
@@ -0,0 +1,120 @@
...
+        for record in self:
+            record.best_price = max(record.offer_ids.mapped('price')) if self.offer_ids else 0
+
+    @api.onchange('garden')
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = 'north'
+        else:
+            self.garden_area = None
```
```suggestion
            self.garden_area = 0
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,71 @@
+from odoo import api, fields, models
+from odoo.exceptions import UserError
+
+
+class EstatePropertyOffer(models.Model):
+    _name = 'estate.property.offer'
+    _description = 'All offers'
```
yoba-odoo: Same as estate property :slightly_smiling_face:

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,71 @@
...
+        string='Status',
+        selection=[
+            ('accepted', 'Accepted'),
+            ('refused', 'Refused'),
+        ],
+    )
+    partner_id = fields.Many2one('res.partner', string='Partner')
+    property_id = fields.Many2one('estate.property', string='Property')
+    validity = fields.Integer(default=7)
+    date_deadline = fields.Date(compute='_compute_deadline', inverse="_inverse_deadline")
```
yoba-odoo:
```suggestion
    date_deadline = fields.Date(compute='_compute_date_deadline', inverse="_inverse_date_deadline")
```
When comes to naming the compute and inverse method it is always better to use the whole field name

estate/models/estate_property.py
```diff
@@ -0,0 +1,121 @@
...
+    @api.ondelete(at_uninstall=False)
+    def _unlink_if_new_or_cancelled(self):
+        for record in self:
+            if record.state not in ['new', 'cancelled']:
+                raise UserError(f"Can't delete property in {record.state} state")
+
+    def action_mark_as_sold(self):
+        for record in self:
+            if record.state == 'cancelled':
+                raise UserError('Property is already cancelled')
```
```suggestion
                raise UserError('Cannot sell cancelled properties')
```
For errors it is better to be precise about why can't we do the action not just that we can't do it

estate/models/estate_property.py
```diff
@@ -0,0 +1,121 @@
...
+
+            record.state = 'sold'
+
+    def action_mark_as_cancelled(self):
+        for record in self:
+            if record.state == 'sold':
+                raise UserError('Property is already sold')
```
yoba-odoo: Same :point_up: :smile:

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @api.model
+    def create(self, val_lists):
+        for vals in val_lists:
+            linked_property = self.env['estate.property'].browse(vals['property_id'])
+            if vals['price'] < linked_property.best_price:
+                raise UserError('Offer price cannot be lower than existing offers')
+
+            linked_property.state = 'offer_received'
+        return super().create(val_lists)
```
yoba-odoo, line 50: We have guidelines also about the order of the methods in a file you can check it here https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions

Also when we have a compute and inverse method for a field we tend to put the compute then right after it the inverse

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,71 @@
...
+        for record in self:
+            created_date = record.create_date.date() or fields.Date.today()
+            record.validity = (record.date_deadline - created_date).days
+
+    def action_mark_as_accepted(self):
+        for record in self:
+            if (
+                record.property_id.state != 'new'
+                and record.property_id.state != 'offer_received'
+            ):
```
yoba-odoo, line 56:
```suggestion
            if (
                record.property_id.state not in ('new', 'offer_received')
            ):
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,71 @@
...
+            record.status = 'accepted'
+            record.property_id.selling_price = record.price
+            record.property_id.buyer_id = record.partner_id
+            record.property_id.state = 'offer_accepted'
+
+    def action_mark_as_refused(self):
+        for record in self:
+            if record.status == 'accepted':
+                record.property_id.state = 'offer_received'
+                record.property_id.selling_price = None
```
yoba-odoo:
```suggestion
                record.property_id.selling_price = 0.0
```

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,15 @@
+from odoo import fields, models
+
+
+class EstatePropertyTag(models.Model):
+    _name = 'estate.property.tag'
+    _description = 'All property tags'
```
yoba-odoo: Surprise!! Same as above :smile:

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,22 @@
+from odoo import api, fields, models
+
+
+class EstatePropertyType(models.Model):
+    _name = 'estate.property.type'
+    _description = 'All property types'
```
yoba-odoo: You know already :sweat_smile:

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,71 @@
...
+                        </group>
+                        <group>
+                            <field name="status"/>
+                        </group>
+                        <group>
+                            <field name="validity" string="Validity (days)"/>
+                        </group>
+                        <group>
+                            <field name="date_deadline" string="Deadline"/>
+                        </group>
```
yoba-odoo: You don't have to put each field alone in a group. Am I missing something?

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,35 @@
...
+        <field name="model">estate.property.tag</field>
+        <field name="arch" type="xml">
+            <form string="Property Tag">
+                <sheet>
+                    <group>
+                        <field name="name"/>
+                    </group>
+                    <group>
+                        <field name="color" widget="color_picker"/>
+                    </group>
```
yoba-odoo, line 28: same

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,191 @@
...
+                <field
+                    name="living_area"
+                    filter_domain="[('living_area', '>=', self)]"
+                />
+                <field name="facades"/>
+                <field name="property_type_id"/>
+
+                <separator/>
+
+                <filter string="Available" name="available" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"/>
```
yoba-odoo:
```suggestion
                <filter string="Available" name="available" domain="[('state', 'in', ('new', 'offer_received'))]"/>
```

yoba-odoo (changes requested):
Nice work :clap: 
Left some non-blocking comments

estate/models/estate_property.py
nguyenntph: Good point. All updated for all models :smiling_face_with_tear:

estate/views/estate_property_offer_views.xml
nguyenntph: Seems like we don't. This is the code block we used as example in [Chapter 6](https://www.odoo.com/documentation/19.0/developer/tutorials/server_framework_101/06_basicviews.html#form), using group for each fields. 

Codeblock

I tried having the fields in 1 group and everything works fine.

estate/models/estate_property.py
nguyenntph: Thank you. That's a good point. Error messages updated.

estate/views/estate_property_tag_views.xml
nguyenntph: Same as above as well.

PR #1026, opened 2025-11-18
19.0 tutorial roole

roole-odoo:
# [roole] chapter 2

Finished Chapter 2 for roole.


estate/models/estate_model.py
clbr-odoo, line 1: We prevent from appending `_model` on every model. Since your model is `estate.property` it should be `estate_property.py`

estate/models/estate_model.py
```diff
@@ -0,0 +1,33 @@
+import datetime
+
+from odoo import fields, models
+
+
+class EstateModel(models.Model):
```
clbr-odoo:
```suggestion
class EstateProperty(models.Model):
```
Following previous comment.

estate/models/estate_model.py
```diff
@@ -0,0 +1,33 @@
...
+    _name = "estate.property"
+    _description = "Estate description"
+    _property = "property"
```
clbr-odoo:
```suggestion
```
This is doing nothing, remove. Also add an empty line between "class metadata" and the fields description.

estate/models/estate_model.py
```diff
@@ -0,0 +1,33 @@
...
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(copy=False, default=datetime.datetime.now() + datetime.timedelta(days=90))
```
clbr-odoo:
```suggestion
    date_availability = fields.Date(
        copy=False, 
        default=fields.Date.add(fields.Date.today(), months=3),
        string="Available from",
    )
```
1. Using datetime/relativedelta is not wrong (more or less, see next comment) but just so you know: there are helper that exists within Odoo for this, which avoids a new import.
2. Your field is a Date type, not Datetime, therefore using datetime.now() adds a cast that is easily avoidable.
3. We usually use the biggest unit possible: `days` -> `months`.
4. Also FYI, if not provided though the `string` parameter, Odoo deduce a label to display in the UI. In your case, it will deduce "Date Availability", which is not very user friendly 😉

estate/models/estate_model.py
```diff
@@ -0,0 +1,33 @@
...
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        string='Orientation',
+        selection=[('north', 'North'), ('south', 'South'), ('west', 'West'), ('east', 'East')])
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
+        string="Estate status",
+        selection=[('new', 'New'), ('offer received', 'Offer Received'), ('offer accepted', 'Offer Accepted'), ('sold', 'Sold'), ('cancelled', 'Cancelled')],
```
clbr-odoo:
```suggestion
        selection=[
            ('new', 'New'), 
            ('offer_received', 'Offer Received'), 
            ('offer_accepted', 'Offer Accepted'), 
            ('sold', 'Sold'), ('cancelled', 'Cancelled'),
        ],
```
1. It's generally a good idea to indent selection values when there is a possibility that the list would be extended in the future.
2. the first part of the tuple is a technical key, it's a bad idea to put spaces inside it them. `offer received -> offer_received`

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_user,estate.property,model_estate_property,base.group_user,1,1,1,1
```
clbr-odoo: 1. the second value is the name of the `ir.rule`, not the name of the model. Name it like you want be be specific, name it like the id works.
2. Missing EOL at the end of the file.

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<odoo>
+<menuitem id="estate_menu_root" name="Estate">
+    <menuitem id="estate_first_level_menu" name="First Level">
+        <menuitem id="estate_model_menu_action" action="estate_model_action"/>
+    </menuitem>
+</menuitem>
+</odoo>
```
clbr-odoo: EOL.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<odoo>
+<record id="estate_model_action" model="ir.actions.act_window">
+    <field name="name">Test properties</field>
```
```suggestion
    <field name="name">Properties</field>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,8 @@
...
+    <field name="res_model">estate.property</field>
+    <field name="view_mode">list,form</field>
+</record>
+</odoo>
```
clbr-odoo: EOL.

estate/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
+    'name': "My Awesome module Roole",
+    'version': '1.0',
+    'depends': ['base'],
+    'author': "roole",
+    'application': True,
+    'installable': True,
```
clbr-odoo: You don't really need that key `installable` since `True` is the default value.

estate/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
+    'name': "My Awesome module Roole",
+    'version': '1.0',
+    'depends': ['base'],
+    'author': "roole",
```
clbr-odoo: Fine in this tutorial but know that when developing an Odoo module, we must put "Odoo S.A.".

estate/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    'application': True,
+    'installable': True,
+    'category': 'Tutorials',
+    'description': """
+    Description text
```
clbr-odoo: Take care that triple quote do take spacing into consideration, usually the content must not be indented with the rest.

estate/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    """,
+    'license': 'GPL-3',
```
```suggestion
    'license': 'LGPL-3',
```
We use LGPL-3 as Open-Source/Community license by default. (And `OEEL-1` for Enterprise modules.)

clbr-odoo (changes requested):
Very good job, just some nitpicks to report 👍

estate/models/estate_offer.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @api.onchange('date_deadline')
+    def _inverse_deadline(self):
+        for record in self:
+            record.validity = (record.date_deadline - fields.Date.today()).days
+
+    def action_accept(self):
+        for record in self:
+            for offers in record.property_id.offer_ids:
+                offers.status = "refused"
```
clbr-odoo, line 44: You can actually do this since you're setting all records of the recordset to the same value.
`record.property_id.offer_ids.status = "refused"`

And take care of the plural misused:
```suggestion
            for offer in record.property_id.offer_ids:
                offer.status = "refused"
```

estate/models/estate_offer.py
```diff
@@ -0,0 +1,63 @@
...
+            record.status = "accepted"
+            record.property_id.state = "offer_accepted"
+            record.property_id.selling_price = record.price
+            record.property_id.buyer_id = record.partner_id
```
clbr-odoo, line 48:
```suggestion
            record.property_id.write({
                "state": "offer_accepted",
                "selling_price": record.price,
                "buyer_id": record.partner_id,
            })
```
You can also do this in one call. Tbh the ORM will most likely process all the write in one batch anyway but still, it's a good practice to do this.

estate/models/estate_offer.py
```diff
@@ -0,0 +1,63 @@
...
+        string='Deadline', compute="_compute_deadline", inverse="_inverse_deadline"
+    )
+
+    _check_price = models.Constraint(
+        'CHECK(price >= 0)', 'The offer price must be positive'
+    )
+
+    @api.depends('validity')
+    def _compute_deadline(self):
+        for record in self:
```
clbr-odoo, line 31:
```suggestion
        for estate in self:
```
We more and more try to avoid generic "record" naming in those cases to make it more readable.
This way even when reading a snippet we know what kind of records we're looping on.

estate/models/estate_property.py
```diff
@@ -0,0 +1,112 @@
...
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for record in self:
+            if len(record.offer_ids) > 0:
```
clbr-odoo, line 75:
```suggestion
            if record.offer_ids:
```
an empty recordset will return a Falsy value when evaluated like this, no need to check the lenght.

estate/models/estate_property.py
```diff
@@ -0,0 +1,112 @@
...
+                record.best_price = max(record.offer_ids.mapped("price"))
```
clbr-odoo, line 76: You can avoid the if_else and do a one liner like this if you want 😉 
```suggestion
                record.best_price = max(record.offer_ids.mapped("price"), default=0.0)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,112 @@
...
+                record.garden_orientation = ''
+                record.garden_area = 0
+
+    def action_sold(self):
+        for record in self:
+            if record.state == "canceled":
+                raise UserError("Canceled property can't be sold.")
+            else:
+                record.state = "sold"
+                return True
```
clbr-odoo, line 96: That return should not be in the else clause but at the very end of the method.
But tbh you don't need the else at all.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,101 @@
...
+            <field name="arch" type="xml">
+                <search string="Searching in estates">
+                    <field name="name"/>
+                    <field name="postcode"/>
+                    <field name="expected_price"/>
+                    <field name="bedrooms"/>
+                    <field name="living_area"/>
+                    <field name="facades"/>
+                    <separator/>
+                    <filter string="Available" name="active" domain="['|', ('state', '=', 'new' ),('state', '=', 'offer received' )]"/>
```
```suggestion
                    <filter string="Available" name="active" domain="[('state', 'in', ('new', 'offer received'))]"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,101 @@
...
+                    <filter string="Postco²de" name="postcode"  context="{'group_by':'postcode'}"/>
```
clbr-odoo: typo !

.gitignore
clbr-odoo, line 1: avoid pushing your personal config 😉

clbr-odoo (commented):
Very good job, keep up !

PR #1027, opened 2025-11-18
[ADD] estate: init commit for estate module

ybenbrai:


estate/security/ir.model.access.csv
clbr-odoo, line 1: This should be in a folder named `security` by convention, not `data`.

estate/models/estate_property.py
```diff
@@ -0,0 +1,35 @@
+from odoo import fields, models
+
+
+class Property(models.Model):
```
```suggestion
class EstateProperty(models.Model):
```
There is no big incidence here but we generally put the name of the file in CamelCase here.
It's usually good to be specific. `Property` is super generic and could be existing in other modules.

estate/models/estate_property.py
```diff
@@ -0,0 +1,35 @@
...
+    _name = "estate.property"
+    _description = "This is a estate property"
```
```suggestion
    _description = "Estate property"
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,35 @@
...
+    _order = "sequence"
```
clbr-odoo: That `_order` will define how the `search` will return you the records.
You don't have a `sequence` field on your `estate.property` model, leave it empty to keep the default (that order by ID) for now.

estate/models/estate_property.py
```diff
@@ -0,0 +1,35 @@
...
+
+    name = fields.Char('Name', required=True)
```
```suggestion
    name = fields.Char(string='Name', required=True)
```
Note that what you did is totally equivalent to this ☝️ . Some reviewers prefer that we name all parameters explicitly.
Also note that by default, Odoo will use the variable name to create a label if not provided.
In this case it will create "Name", so you don't really need to specify it.

estate/view/__init__.py
clbr-odoo, line 1: 1. You don't need to have an `__init__` file in your `view` folder as it does not contain Python files.
2. By convention, that folder should not be `view` but `views`. Please double-check the [coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#directories) for reference. Also don't hesitate to start to explore Odoo codebase and copy how it's done 🙂

estate/view/estate_property_views.xml
```diff
@@ -0,0 +1,18 @@
...
+
+    <menuitem id="estate_menu_advertisements" 
+              name="Advertisements" 
+              parent="estate_menu_root"/>
+
+    <menuitem id="estate_property_menu_action" 
+              action="estate_property_action" 
+              parent="estate_menu_advertisements"/>
+
+</odoo>
```
clbr-odoo: Missing EOL at the end of the file.

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    'summary': 'Track Real Estate',
+    'website': 'https://www.odoo.com/app/realestate',
+    'author': 'Odoo S.A.',
+    'license': 'LGPL-3',
+    'depends': [
+        'base',
+    ],
+     'data': [
+         'data/ir.model.access.csv',
+         'view/estate_property_views.xml',
```
```suggestion
         'security/ir.model.access.csv',
         'views/estate_property_views.xml',
```

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+{
+    'name': 'Real Estate',
+    'description': 'Real Estate !',
+    'version': '1.0',
+    'summary': 'Track Real Estate',
+    'website': 'https://www.odoo.com/app/realestate',
+    'author': 'Odoo S.A.',
+    'license': 'LGPL-3',
```
clbr-odoo, line 9:
```suggestion
    'application': True,
    'license': 'LGPL-3',
```
You probably want to have application=True for this module. See the [manifest reference](https://www.odoo.com/documentation/19.0/developer/reference/backend/module.html).

estate/view/estate_property_views.xml
```diff
@@ -0,0 +1,18 @@
+<odoo>
+    <record id="estate_property_action" model="ir.actions.act_window">
+        <field name="name">Properties</field>
+        <field name="res_model">estate.property</field>
+        <field name="view_mode">tree,form</field>
+    </record>
+
+    <menuitem id="estate_menu_root" name="Real Estate"/>
```
clbr-odoo: Note that you can also nest the module to avoid defining the `parent`, like this:
```xml
<menuitem ...>
    <menuitem .../>
</menuitem/>
```

clbr-odoo (commented):
Good job so far, just a few comments ! 

Your commit title are not fully correct: usually we put the `[ADD]` tag on the module init (where the manifest is). Following commits will be `[IMP](provements)` to the now existing module.
See [git guidelines](https://www.odoo.com/documentation/19.0/developer/reference/backend/module.html) for more details.

estate/models/estate_property.py
ybenbrai: > Note that what you did is totally equivalent to this ☝️ . Some reviewers prefer that we name all parameters explicitly. Also note that by default, Odoo will use the variable name to create a label if not provided. In this case it will create "Name", so you don't really need to specify it.

okay noted !, i changed it to `Title` in local now to match the screenshots

estate/views/estate_tree.xml
```diff
@@ -0,0 +1,17 @@
...
+                <field name="post_code"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="expected_price"/>
+                <field name="selling_price"/>
+                <field name="date_availability"/>
+            </list>
+        </field>
+    </record>
+</odoo>
```
antonrom1: There is a missing end of line here

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property,access.estate.property,model_estate_property,base.group_user,1,1,1,1
+access_estate_property_type,access.estate.property.type,model_estate_property_type,base.group_user,1,1,1,1
+access_estate_property_tag,access.estate.property.tag,model_estate_property_tag,base.group_user,1,1,1,1
+access_estate_property_offer,access.estate.property.offer,model_estate_property_offer,base.group_user,1,1,1,1
```
clbr-odoo: EOL missing at the end of the file.

estate/views/estate_list.xml
```diff
@@ -0,0 +1,18 @@
...
+                <field name="post_code"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="expected_price"/>
+                <field name="selling_price"/>
+                <field name="date_availability"/>
+            </list>
+        </field>
+    </record>
+</odoo>
```
clbr-odoo: EOL

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,11 @@
...
+    <menuitem id="estate_menu_root" name="Real Estate">
+        <menuitem id="estate_menu_advertisements" name="Advertisements">
+            <menuitem id="estate_property_menu_action" name="Properties" action="estate_property_action"/>
+        </menuitem>
+        <menuitem id="estate_menu_settings" name="Settings">
+            <menuitem id="estate_property_type_menu_action" name="Property Types" action="estate_property_type_action"/>
+            <menuitem id="estate_property_tag_menu_action" name="Property Tags" action="estate_property_tag_action" />
+        </menuitem>
+    </menuitem>
+</odoo>
```
clbr-odoo: EOL

estate/views/estate_property_form_views.xml
```diff
@@ -0,0 +1,26 @@
...
+            </form>
+        </field>
+    </record>
+
+    <record id="estate_property_offer_action" model="ir.actions.act_window">
+        <field name="name">Property Offers</field>
+        <field name="res_model">estate.property.offer</field>
+        <field name="view_mode">tree,form</field>
+    </record>
+    </odoo>
```
clbr-odoo: EOL

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,18 @@
...
+        </field>
+    </record>
+
+    <record id="estate_property_tag_action" model="ir.actions.act_window">
+        <field name="name">Property Tags</field>
+        <field name="res_model">estate.property.tag</field>
+        <field name="view_mode">list,form</field>
+    </record>
+
+</odoo>
```
clbr-odoo: EOL

estate/views/estate_view_form.xml
```diff
@@ -0,0 +1,73 @@
...
+                                    <field name="buyer_id"/>
+                                </group>
+                            </group>
+                        </page>
+                    </notebook>
+                </sheet>
+            </form>
+        </field>
+    </record>
+</odoo>
```
clbr-odoo: EOL

estate/views/estate_view_form.xml
```diff
@@ -0,0 +1,73 @@
...
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form string="Property">
+                <header>
+                    <field name="state" widget="statusbar"/>
+                    <button name="action_sold" string="SOLD" type="object" 
+                    invisible="state in ['action_solde']"
+                         class="oe_highlight"/>
+                    <button name="action_cancel" string="CANCEL" type="object" 
+                    invisible="state in ['action_cancel']"
```
clbr-odoo:
```suggestion
                    invisible="state == 'action_cancel"
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,97 @@
...
+    @api.depends('living_area', 'garden_area')
+    def _compute_total_area(self):
+        for rec in self:
+            rec.total_area = (rec.living_area or 0) + (rec.garden_area or 0)
+
+    @api.depends('offer_ids.price')
+    def _compute_best_price(self):
+        for rec in self:
+            prices = rec.offer_ids.mapped('price')
+            rec.best_price = max(prices) if prices else 0.0
```
clbr-odoo: That works, you can also do 
```suggestion
            rec.best_price = max(prices, default=0.0)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,97 @@
...
+    offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers")
+    total_area = fields.Integer("Total Area (sqm)", compute="_compute_total_area")
+    best_price = fields.Float(
+        "Best Offer", compute="_compute_best_price", readonly=True
+    )
+
+    @api.depends('living_area', 'garden_area')
+    def _compute_total_area(self):
+        for rec in self:
+            rec.total_area = (rec.living_area or 0) + (rec.garden_area or 0)
```
clbr-odoo, line 74: This compute is defined twice (see line 76).

estate/models/estate_property.py
```diff
@@ -0,0 +1,97 @@
...
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = False
+
+    def action_sold(self):
+        if "cancelled" in self.mapped("state"):
```
clbr-odoo: It works but it's not super readable. I'd do:
```suggestion
        if any([prop.state == "cancelled" for prop in self]):
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,97 @@
...
+            raise UserError("Canceled property cannot be sold !")
+        return self.write({"state": "sold"})
```
```suggestion
        self.state = 'sold'
        return True
```
write doesn't return anything, and this notation also works, even if the recordset has many records.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,48 @@
...
+    property_id = fields.Many2one("estate.property", string="Property", required=True)
+    validity = fields.Integer("Validity (days)", default=7)
+    date_deadline = fields.Date(
+        "Deadline", compute="_compute_date_deadline", store=True
+    )
+
+    @api.depends('validity')
+    def _compute_date_deadline(self):
+        for offer in self:
+            if offer._origin.validity:
```
clbr-odoo:
```suggestion
            if offer.validity:
```
validity will always be set, worst case to 0 so this if...else can be avoided.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,48 @@
...
+                )
+            else:
+                offer.date_deadline = False
+
+    def action_accept(self):
+        for offer in self:
+            offer.state = 'accepted'
+            offer.property_id.selling_price = offer.price
+            offer.property_id.buyer_id = offer.partner_id.id
+            offer.property_id.state = 'offer_accepted'
```
clbr-odoo:
```suggestion
            offer.property_id.write({
                'selling_price': offer.price,
                'buyer_id': offer.partner_id.id,
                'state': 'offer_accepted',
            })
```
It's better to do those three "write" in one :)

clbr-odoo (commented):
Don't forget to fix all missing EOL, keep up 👍

PR #1028, opened 2025-11-18
[ADD] Estate: init

haall-odoo:


estate/data/.gitkeep
clbr-odoo, line 1: This should not be there, setup your .gitignore correctly 🙂

estate/models/estate_property.py
```diff
@@ -0,0 +1,35 @@
+import datetime as dt
+
+from odoo import fields, models
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "An awesome estate module"
```
```suggestion
    _description = "Estate Property"
```
The description should describe the model only, not the whole module 😉

estate/models/estate_property.py
```diff
@@ -0,0 +1,35 @@
...
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(copy=False, default=dt.datetime.today() + dt.timedelta(days=90))
```
```suggestion
    date_availability = fields.Date(copy=False, default=fields.Date.add(fields.Date.today(), months=3)
```
1. You have a `Date` type, therefore using `dt.datetime` instead of `dt.date` just add some extra useless casting.
2. What you did is mostly correct, just know that we have some Odoo built-ins (`fields.Date.add/subtract`) if you prefer.
3. Usually we take the biggest unit (`days` -> `months`).
4. Last small note: when `string` parameter is not provided to the field, Odoo will extrapolate a label to display in the view, in this case it will be `Date Availability` which is not super user friendly. But you adapted it in the view directly, so it works 👍 Just know that if someone create another view of the model, it will not be "Available From" but "Date Availability" by default.

estate/models/estate_property.py
```diff
@@ -0,0 +1,35 @@
...
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        string='Orientation',
+        selection=[('north', 'North'), ('west', 'West'), ('south', 'South'), ('east', 'East')],
+        help="Choose the appropriate orientation of the garden"
+    )
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
+        string="Estate status",
+        selection=[('new', 'New'), ('offer received', 'Offer Received'), ('offer accepted', 'Offer Accepted'), ('sold', 'Sold'), ('cancelled', 'Cancelled')],
```
```suggestion
        selection=[
            ('new', 'New'), 
            ('offer_received', 'Offer Received'), 
            ('offer_accepted', 'Offer Accepted'), 
            ('sold', 'Sold'), ('cancelled', 'Cancelled'),
        ],
```
1. It's a good idea to indent selection values that might evolve in time so that a colleague can easily add a value without breaking git history. At some point lines are also just too long 😆 
2. the first item in the tuple is a technical key, it's a bad idea to let space inside them. (`offer received -> offer_received`).

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_model,estate_property,model_estate_property,base.group_user,1,1,1,1
+portal_access_estate_model,estate_property,model_estate_property,base.group_portal,1,0,0,0
```
clbr-odoo: 👍 Nice test with the portal user.
Note that the second item "name" is the name of the `ir.rule`, it's better if it's not the same for all rules.
Usually you can just put the same thing as the id. Which ideally contains both the model concerned and the group of users that will be affected.
But you forgot an EOL at the end of the file.

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<odoo>
+    <menuitem id="estate_menu_root" name="Estate">
+        <menuitem id="estate_first_level_menu" name="Advertisements">
+            <menuitem id="estate_model_menu_action" action="estate_model_action"/>
+        </menuitem>
+    </menuitem>
+</odoo>
```
clbr-odoo: Missing EOL.

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<odoo>
+    <menuitem id="estate_menu_root" name="Estate">
+        <menuitem id="estate_first_level_menu" name="Advertisements">
+            <menuitem id="estate_model_menu_action" action="estate_model_action"/>
```
clbr-odoo:
```suggestion
            <menuitem id="estate_property_menu_action" action="estate_property_action"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="estate_property_view_tree" model="ir.ui.view">
```
```suggestion
    <record id="estate_property_view_list" model="ir.ui.view">
```
tree was the old name of list view, now we try to only use list.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,80 @@
...
+        <field name="name">estate.property.name</field>
```
clbr-odoo: This field the the name of the view, so `estate.property.list` is probably more correct.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,80 @@
...
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <list string="Channel">
```
```suggestion
            <list string="Properties">
```
copy-pasta 🍝

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,80 @@
...
+                <field name="living_area" string="Living Area (sqm)"/>
+                <field name="expected_price"/>
+                <field name="selling_price"/>
+                <field name="date_availability" string="Available From"/>
+            </list>
+        </field>
+    </record>
+
+    <record id="estate_property_form_view" model="ir.ui.view">
+        <field name="name">estate.property.name</field>
```
clbr-odoo: Same comment, the name is not descriptive/correct.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,80 @@
...
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form string="My new house">
+                <sheet>
+                    <group>
+                        <h1><field name="name" string="" placeholder="My new house"/></h1>
```
```suggestion
                        <h1><field name="name" nolabel="1" placeholder="My new house"/></h1>
```
I think that works better.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,80 @@
...
+                            </group>
+                        </page>
+                    </notebook>
+                </sheet>
+            </form>
+        </field>
+    </record>
+
+    <record id="estate_property_search_view" model="ir.ui.view">
+        <field name="name">estate.property.name</field>
```
clbr-odoo: Same.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,80 @@
...
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <search string="TestSearch">
```
```suggestion
            <search string="Search Properties">
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,80 @@
...
+                <field name="name" string="Title"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area" string="Living Area (sqm)"/>
+                <field name="facades"/>
+                <separator/>
+                <filter name="Available" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer received')]"/>
```
```suggestion
                <filter name="Available" domain="[('state', 'in', ('new', 'offer_received'))]"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,80 @@
...
+                <filter name="Postcode" context="{'group_by': 'postcode'}"/>
+            </search>
+        </field>
+    </record>
+
+    <record id="estate_model_action" model="ir.actions.act_window">
+        <field name="name">Properties</field>
+        <field name="res_model">estate.property</field>
+        <field name="view_mode">list</field>
```
```suggestion
        <field name="view_mode">list,form</field>
```

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+{
+    'name': 'Estate',
+    'version': '0.0',
```
```suggestion
    'version': '1.0',
```
Default is usually starting at 1.0.

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    'depends': ['base'],
+    'author': 'haall-odoo',
```
clbr-odoo: Fine in this tutorial, just know that when developing an Odoo module, we must put `Odoo S.A.`.

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    'application': True,
+    'installable': True,
+    'category': '',
+    'description': '',
+    'license': 'GPL-3',
```
```suggestion
    'license': 'LGPL-3',
```
Odoo Community (Open-Source) default license is `LGPL-3`, and Enterprise is `OEEL-1`.

clbr-odoo (commented):
Very good job, continue like this ! 


The module name in the commit title should be lower case, it's the technical name of the module, which is equivalent to the folder name.

estate/data/.gitkeep
haall-odoo: What should not be there?

estate/models/estate_property.py
```diff
@@ -0,0 +1,108 @@
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate Property"
+    _order = "id desc"
+    _check_expected_price = models.Constraint(
```
clbr-odoo, line 10: Constraints should be just below field definition. Check the coding guidelines for the order of the methods and fields because it's not the only one misplaced :)

estate/models/estate_property.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @api.depends("garden_area", "living_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.garden_area + record.living_area
+
+    @api.depends("offer_ids")
+    def _compute_best_offer(self):
+        for record in self:
+            record.best_price = max(record.offer_ids.mapped("price")) if len(record.offer_ids) > 0 else 0.0
```
clbr-odoo, line 66:
```suggestion
            record.best_price = max(record.offer_ids.mapped("price"), default=0.0)
```
1. You never need to check the length if you want to check if a recordset is filled, just do `if record.offer_ids:`
2. You can simplify a bit like this ☝️

estate/models/estate_property.py
```diff
@@ -0,0 +1,108 @@
...
+        for record in self:
+            if not record.garden:
+                record.garden_area = 0
+                record.garden_orientation = ''
+            else:
+                record.garden_area = 10
+                record.garden_orientation = 'north'
+
+    def estate_property_action_sold(self):
+        self.__estate_property_action_sold_cancel('sold', "A cancelled property cannot be sold!", "This property is already sold!")
```
clbr-odoo, line 79:
```suggestion
        self._estate_property_action_sold_cancel('sold', "A cancelled property cannot be sold!", "This property is already sold!")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,108 @@
...
+    @api.constrains('selling_price')
+    def _check_price_constraint(self):
+        for record in self:
+            if record.selling_price and float_compare(record.selling_price, record.expected_price * 0.9, precision_digits=4) < 0:
+                raise ValidationError("The price cannot be les than 90% of the expected price")
+
+    @api.onchange("offer_ids")
+    def _onchange_offer_ids(self):
+        for record in self:
+            if len(record.offer_ids) > 0:
```
clbr-odoo, line 105:
```suggestion
            if record.offer_ids:
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @api.onchange("date_deadline")
+    def _inverse_date_deadline(self):
+        for record in self:
+            record.validity = (record.date_deadline - fields.Date.today()).days
+
+    def action_accept_offer(self):
+        for record in self:
+            if record.status == 'accepted':
+                return False
```
clbr-odoo, line 41: It's weird to have a method that returns False in the middle of the loop.
This will lead to inconsistencies.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def action_refuse_offer(self):
+        for record in self:
+            record.status = "refused"
+        return True
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_if_deleted(self):
+        for record in self:
+            record.property_id.selling_price = 0.0
```
clbr-odoo, line 58: You can actually "batch write" it self.property_id.selling_price = 0.0

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,107 @@
...
+        </field>
+    </record>
+
+    <record id="estate_property_form_view" model="ir.ui.view">
+        <field name="name">estate.property.form</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form string="My new house">
+                <header>
+                    <button name="estate_property_action_sold" type="object" string="Sold" invisible="state == 'sold' or state == 'cancelled'"/>
```
clbr-odoo, line 29:
```suggestion
                    <button name="estate_property_action_sold" type="object" string="Sold" invisible="state in ('sold', 'cancelled')"/>
```

clbr-odoo (commented):
Good job !

PR #1029, opened 2025-11-18
[ADD] estate: init

Sylsee:


estate/data/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,1
```
yoba-odoo: ~~1) We should always have an empty line at the end of each file~~ You already did this :ok_hand: 
2) The folder should be named `security` not `data`

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <menuitem id="estate_menu_root" name="Estate">
+        <menuitem id="estate_property_first_level_menu" name="Property">
+            <menuitem id="estate_property_menu_action" action="estate.estate_property_action"/>
```
yoba-odoo, line 5:
```suggestion
            <menuitem id="estate_property_menu_action" action="estate_property_action"/>
```
We don't have to call it with `estate.` since it is already in the same path so it should be read automatically

estate/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+{
+    'name': "Estate",
+    'author': "sypol",
+    'license': "LGPL-3",
```
yoba-odoo, line 8: would be nice if we added a description key in json to add a small description about the model
Check https://www.odoo.com/documentation/19.0/developer/reference/backend/module.html for more info about the keys we are using in manifest files

Also version key would be better to have it

yoba-odoo (changes requested):
Nice work :clap: 
Just some small comments, Also if you can squash the LINT and FIX comments with their parent IMP commit it would be better to keep a clean commit history.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,83 @@
...
+                </sheet>
+            </form>
+        </field>
+    </record>
+
+    <record id="estate_property_view_tree" model="ir.ui.view">
+        <field name="name">estate.property.list</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <list string="Channel">
```
yoba-odoo:
```suggestion
            <list string="Properties">
```
Would be more accurate :smile:

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,83 @@
...
+        <field name="arch" type="xml">
+            <search string="Search Properties">
+                <field name="name" />
+                <field name="postcode" />
+                <field name="expected_price" />
+                <field name="bedrooms" />
+                <field name="living_area" />
+                <field name="facades" />
+                <filter string="Available" name="state"
+                    domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]" />
```
yoba-odoo:
```suggestion
                <filter string="Available" name="state"
                    domain="[('state', 'in', ('new', 'offer_received'))]" />
```

yoba-odoo (changes requested):
Some more changes :innocent:

estate/views/estate_menus.xml
Sylsee: I've noticed we sometimes use it even within the same path and sometimes not. Is this legacy code, or does it depend on the developer?
Personally, I find it clearer to always specify the path, it's more explicit and helps avoid oversight errors. What do you think?

estate/models/estate_property.py
```diff
@@ -0,0 +1,117 @@
...
+            ('canceled', 'Canceled'),
+        ],
+        required=True,
+        copy=False,
+        default='new',
+    )
+    property_type_id = fields.Many2one('estate.property.type', string='Property Types')
+    property_tag_ids = fields.Many2many('estate.property.tag', string='Property Tags')
+    buyer = fields.Many2one('res.partner', string='Buyer', copy=False)
+    salesperson = fields.Many2one(
```
```suggestion
    buyer_id = fields.Many2one('res.partner', string='Buyer', copy=False)
    salesperson_id = fields.Many2one(
```
the naming convention for Many2one fields always should have `_id`

estate/models/estate_property.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @api.depends('garden_area', 'living_area')
+    def _compute_total_area(self):
+        for property in self:
+            property.total_area = property.garden_area + property.living_area
+
+    @api.depends('offer_ids.price')
+    def _compute_best_price(self):
+        for property in self:
+            if len(self.offer_ids):
```
```suggestion
            if self.offer_ids:
```
Yours is completely okay and correct but you don't have to call `len()`

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,1
+estate.access_estate_property_type,access_estate_property_type,estate.model_estate_property_type,base.group_user,1,1,1,1
+estate.access_estate_property_tag,access_estate_property_tag,estate.model_estate_property_tag,base.group_user,1,1,1,1
+estate.access_estate_property_offer,access_estate_property_offer,estate.model_estate_property_offer,base.group_user,1,1,1,1
```
yoba-odoo: Missing EOL

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,36 @@
...
+                <field name="partner_id" />
+                <field name="validity" />
+                <field name="date_deadline" />
+                <button name="action_accept" type="object" icon="fa-check" title="Accept" />
+                <button name="action_refuse" type="object" icon="fa-times" title="Refuse" />
+                <field name="status" />
+            </list>
+        </field>
+    </record>
+</odoo>
```
yoba-odoo: Missing EOL

yoba-odoo (changes requested):
Almost perfect :ok_hand: 
Just pay attention to the EOL at the end of every file :smile:

estate/models/estate_property.py
```diff
@@ -0,0 +1,117 @@
...
+                property.best_price = max(property.offer_ids.mapped('price'))
+            else:
+                property.best_price = 0
```
```suggestion
            property.best_price = max(record.offer_ids.mapped("price"), default=0.0)
```

estate/models/estate_property_offer.py
```diff
@@ -46,6 +46,21 @@ def _inverse_date_deadline(self):
                 offer.date_deadline - (offer.create_date.date() or fields.Date.today())
             ).days
 
+    @api.model
+    def create(self, offers_list):
```
yoba-odoo:
```suggestion
    def create(self, vals_list):
```
not very important but just FYI we usually name this as `vals_list`

estate/models/estate_property_offer.py
```diff
@@ -46,6 +46,21 @@ def _inverse_date_deadline(self):
...
+        for offer in offers_list:
+            linked_property = self.env['estate.property'].browse(offer['property_id'])
```
yoba-odoo:
```suggestion
            linked_property = self.env['estate.property'].browse(offer.get('property_id'))
```
When fetching something from a dictionary we tend to use get to avoid KeyError if the key doesn't exist. I know mostly the property_id will always be there but it is better safe than sorry :sweat_smile:

Sylsee: Just learn a new python thing ty 😂

Sylsee: By doing that I got this error:
image

I would do that to fix it :
```py
property_id = offer.get('property_id')
if not property_id:
    continue

linked_property = self.env['estate.property'].browse(property_id)
```

I just think I should raise an error instead of continuing but idk which type of error, I didn't saw any `InternalError` or something similar in the exceptions definitions. What would you do ?

yoba-odoo: Yup, my bad. When using get() and the key doesn't exist it will return None which might cause an Error as browse() is waiting for `int`

what you can do is fallback on a default value in get() like this:

```py
offer.get('property_id', 0)
```

Sylsee: But then if their is a property with an id of 0 it will not raise an error but change the wrong property no ?

yoba-odoo: Normally we don't have a record with id 0, our id sequences are 1-indexed. But to stick with the better-safe-than-sorry ideology we can do something like

```py
if offer.get('property_id'):
    linked_property = self.env['estate.property'].browse(offer['property_id'])
```

PR #1033, opened 2025-11-18
[ADD] estate : init

abdullahsherif:


estate/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+{
+    'name': "Real estate",
+    'depends': ['base'],
+    'application': True,
+    'installable': True,
+    'data': [
+        'security/ir.model.access.csv',
+        'views/estate_menus.xml',
+        'views/estate_property_views.xml'
```
antonrom1: We usually add a comma at the end of every line for multiline dicts, lists, arguments, tuples etc. 
This is to avoid the next person that wants to add a new line after yours would have to add a diff to your line (which will set him as author of the line when someone will run git blame)
```suggestion
        'views/estate_property_views.xml',
```

estate/__manifest__.py
```diff
@@ -0,0 +1,13 @@
...
+    ],
+    'license': 'LGPL-3',
+    'author': 'Abdallah'
```
antonrom1: Also missing trailing comma

estate/models/estate_property.py
```diff
@@ -0,0 +1,20 @@
+from odoo import fields, models
+
+
+class Property(models.Model):
+    _name = "estate.property"
+    _description = "estate property info"
```
antonrom1: The description should usually be title-cased
```suggestion
    _description = "Estate Property Info"
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,20 @@
...
+    date_availability = fields.Date()
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float()
+    bedrooms = fields.Integer()
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(string='garden_orientation', selection=[('North', 'north'), ('east', 'East'), ('West', 'west'), ('South', 'south')])
```
antonrom1: string is the name of the field that will be displayed to the user. Therefore it should be formatted as such, in title case

The selection list should be a list of tuples as such:
`[('database_value', "Display Value"), ...]`
usually we snake case the database value and title case the user-facing value. Here is an [example](https://github.com/odoo-dev/odoo/blob/19.0-peppol-rate-limit-ui-roan/addons/account/models/account_account_tag.py#L12)

```suggestion
    garden_orientation = fields.Selection(string='Garden Orientation', selection=[('north', 'North'), ('east', 'East'), ('west', 'West'), ('south', 'South')])
```

estate/__manifest__.py
antonrom1: You probably saw that there is an error with that view when you try to startup odoo (or the in runbot logs)

The problem is that the order of data imports in the manifest file matters. Indeed, the menu item references `estate_property_action`, which does not yet exist at the time of import of the `views/estate_menus.xml`, as it was imported first. You thus need to first create the views file, then the menus file, which in tern will be able to reference the views

```suggestion
        'views/estate_property_views.xml',
        'views/estate_menus.xml',
```

antonrom1 (changes requested):
Thank you! You're almost there, mostly little things.

I also invite you to read the [git guidelines](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html) for naming and writing a description to your commits, and the commit descriptions that you will have to respect in the future, as well as the [coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html) (mostly styleguides)

estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_property, estate_property_misc
```
clbr-odoo:
```suggestion
from . import estate_property
from . import estate_property_misc
```
We usually do one import per line in those init files 🙂

estate/models/estate_property.py
```diff
@@ -0,0 +1,113 @@
...
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids")
+    def _compute_best_price(self):
+        for record in self:
+            if record.offer_ids:
+                record.best_offer = max(record.offer_ids.mapped("price"))
+            else:
+                record.best_offer = 0
```
```suggestion
            record.best_offer = max(record.offer_ids.mapped("price"), default=0.0)
```
Your code works perfectly, just a one-liner suggestion.

estate/models/estate_property.py
```diff
@@ -0,0 +1,113 @@
...
+
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        self.garden_area = 10 if self.garden else 0
+        self.garden_orientation = "north" if self.garden else ""
```
```suggestion
        self.garden_orientation = "north" if self.garden else None
```
Your code works thanks to the ORM but it's a better idea to use None.

estate/models/estate_property.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def sell_property_action(self):
```
```suggestion
    def button_sell_property(self):
```
Just a personal approach, I always prepend those method that are called from button with `button_`.
"Action" in Odoo refers to a specific technical report (`ir.action`) that is used to redirect, to open views, etc...
You don't have to make the change here, just keep it in mind 😉

estate/models/estate_property.py
```diff
@@ -0,0 +1,113 @@
...
+        return True
+
+    def cancel_property_action(self):
+        for record in self:
+            if record.status == "sold":
+                raise exceptions.UserError("Property Sold")
+            record.status = "cancelled"
+        return True
+
+    _check_positive_expected_price = models.Constraint(
```
clbr-odoo: Constraints needs to go upper in the file, just after fields definition.
See https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions

estate/models/estate_property.py
```diff
@@ -0,0 +1,113 @@
...
+    _check_positive_selling_price = models.Constraint(
+        'CHECK(selling_price > 0)',
+        'Prices Must Be Positive',
+    )
+
+    @api.constrains("selling_price")
+    def _check_selling_price(self):
+        for record in self:
+            if (record.selling_price - 0.9 * record.expected_price >= 1e-5):
+                raise exceptions.ValidationError("The selling price is too low")
```
clbr-odoo: Missing EOL at the end of the file.

estate/models/estate_property.py
```diff
@@ -0,0 +1,113 @@
...
+
+    _check_positive_selling_price = models.Constraint(
+        'CHECK(selling_price > 0)',
+        'Prices Must Be Positive',
+    )
+
+    @api.constrains("selling_price")
+    def _check_selling_price(self):
+        for record in self:
+            if (record.selling_price - 0.9 * record.expected_price >= 1e-5):
```
clbr-odoo: 🤣 why `1e-5` ?
This is slightly too complex for the case; something like this works just as well, and is handling float correctly.
`float_compare(record.selling_price, record.expected_price * 0.90, 2) < 0`

estate/models/estate_property_misc.py
```diff
@@ -0,0 +1,82 @@
+from odoo import fields, models, api, exceptions
+
+
+class PropertyType(models.Model):
```
clbr-odoo, line 4: Ideally try to put one file per class.

estate/models/estate_property_misc.py
```diff
@@ -0,0 +1,82 @@
...
+        selection=[
+            ("accepted", "Accepted"),
+            ("refused", "Refused"),
+        ],
+    )
+    buyer_id = fields.Many2one("res.partner", string="Buyer", required=True)
+    property_id = fields.Many2one("estate.property", string="Property", required=True)
+
+    validity = fields.Integer(default=7)
+    date_deadline = fields.Date(compute="_compute_date", inverse="_inverse_date")
```
clbr-odoo:
```suggestion
    date_deadline = fields.Date(compute="_compute_date_deadline", inverse="_inverse_date_deadline")
```
Always be specific enough 🙂

estate/models/estate_property_misc.py
```diff
@@ -0,0 +1,82 @@
...
+
+    @api.depends("validity")
+    def _compute_date(self):
+        for record in self:
+            create_date = record.create_date
+            if not create_date:
+                create_date = fields.Date.today()
```
clbr-odoo:
```suggestion
            create_date = record.create_date or fields.Date.today()
```
a simple `or` works just as well

estate/models/estate_property_misc.py
```diff
@@ -0,0 +1,82 @@
...
+    )
+    buyer_id = fields.Many2one("res.partner", string="Buyer", required=True)
+    property_id = fields.Many2one("estate.property", string="Property", required=True)
+
+    validity = fields.Integer(default=7)
+    date_deadline = fields.Date(compute="_compute_date", inverse="_inverse_date")
+
+    @api.depends("validity")
+    def _compute_date(self):
+        for record in self:
```
clbr-odoo:
```suggestion
        for property in self:
```
In many Odoo teams, we try to avoid naming everything "record" for readability.
This way we never have to think to what type of records we are manipulating.

estate/models/estate_property_misc.py
```diff
@@ -0,0 +1,82 @@
...
+            record.date_deadline = fields.Date.add(
+                create_date,
+                days=record.validity,
+            )
+
+    def _inverse_date(self):
+        for record in self:
+            create_date = record.create_date.date()
+            if not create_date:
+                create_date = fields.Date.today()
```
clbr-odoo: Same as above.

estate/models/estate_property_misc.py
```diff
@@ -0,0 +1,82 @@
...
+            record.validity = (record.date_deadline - create_date).days
+
+    def accept_offer(self):
+        if self.property_id.status == "sold":
```
clbr-odoo:
```suggestion
        self.ensure_one()
        if self.property_id.status == "sold":
```
Calling the `self.ensure_one()` helper is a safeguard that indicates that the following code can only be applied to one record.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,136 @@
...
+		<field name="arch" type="xml">
+			<search string="Property">
+				<field name="name"/>
+				<field name="postcode"/>
+				<field name="expected_price"/>
+				<field name="bedrooms"/>
+				<field name="living_area"/>
+				<field name="facades"/>
+				<filter string="Archived" name="inactive" domain="[('active', '=', False)]"/>
+				<filter string="Available" name="available" domain="['|', ('status', '=', 'new'), ('status', '=', 'offer_received')]"/>
```
```suggestion
				<filter string="Available" name="available" domain="[('status', 'in', ('new', 'offer_received')]"/>
```

clbr-odoo (commented):
Good job, keep up !

PR #1035, opened 2025-11-18
19.0 tutorials taskv

tskv:


estate/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    'version': '1.0',
+    'depends': ['base'],
+    'author': "taskv",
+    'category': 'Category',
+    'description': """
+    Tutorial Project
+    """,
+    'data': [
+        'security/ir.model.access.csv',
+
```
yoba-odoo: No need for this empty line

estate/models/estate_property.py
```diff
@@ -0,0 +1,67 @@
...
+
+class EstateProperty(models.Model):
+    _name = 'estate.property'
+    _description = 'Estate Property'
+
+    name = fields.Char('Title', required=True, default='Unknown', translate='True')
+    active = fields.Boolean('Active', default=True)
+    description = fields.Text('Description')
+    postcode = fields.Char('Postcode')
+    date_availability = fields.Date('Available From', copy=False, default=fields.Datetime.now() + timedelta(days=90))
```
```suggestion
    date_availability = fields.Date('Available From', copy=False, default=fields.Date.add(fields.Date.today(), months=3)
```
What you are doing is correct as well but we have a helper in odoo to do the same so it is better to use it

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0"?>
+<odoo>
+    <menuitem
+        id="estate_menu_root"
+        name="Real Estate"/>
+    <menuitem
+        id="properties_menu"
+        name="Properties"
+        parent="estate_menu_root"
+        action="estate.estate_property_action"
```
yoba-odoo:
```suggestion
        action="estate_property_action"
```
Since we are in the same folder as the action we don't have to use estate. to call the action it should be read automatically

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,29 @@
...
+    <menuitem
+        id="settings_menu"
+        name="Settings"
+        parent="estate_menu_root"
+        sequence="2"/>
+    <menuitem
+        id="properties_type_menu"
+        name="Property Types"
+        parent="settings_menu"
+        action="estate.estate_property_type_action"
```
yoba-odoo:
```suggestion
        action="estate_property_type_action"
```
Same

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,29 @@
...
+        sequence="1"/>
+    <menuitem
+        id="properties_tag_menu"
+        name="Property Tags"
+        parent="settings_menu"
+        action="estate.estate_property_tag_action"
```
yoba-odoo:
```suggestion
        action="estate_property_tag_action"
```
Same

yoba-odoo (changes requested):
Impressive work :clap: 
Just some small nitpicks. Would also be great if we can squash the FIX commits to their parent IMP commits so we can have a clean commit history.

Keep up the good work :ok_hand:

estate/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
```
antonrom1: The odoo license in every file is not needed anymore. You can remove it!

estate/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+        'security/ir.model.access.csv',
+        'views/estate_property_views.xml',
+        'views/estate_property_offer_views.xml',
+        'views/estate_property_type_views.xml',
+        'views/estate_property_tag_views.xml',
+        'views/estate_menus.xml',
+        'views/res_users_views.xml',
+    ],
+    'demo': [
+    ],
```
antonrom1, line 22: you don't need to specify demo if it's empty

estate/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from . import estate_property
+from . import estate_property_type
+from . import estate_property_tag
+from . import estate_property_offer
+from . import res_users
```
antonrom1, line 5: You just need to sort all imports by alphabetical order like here
https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#imports
Screenshot 2025-11-21 at 10 39 33

estate/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
```
antonrom1: Also you can remove that license, it's not needed anymore for any new odoo files

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+    facades = fields.Integer('Facades')
+    garage = fields.Boolean('Garage')
+    garden = fields.Boolean('Garden')
+    garden_area = fields.Integer('Garden Area (sqm)')
+    garden_orientation = fields.Selection(string='Garden orientation',
+                                          selection=[('north', 'North'),
+                                                    ('south', 'South'),
+                                                    ('east', 'East'),
+                                                    ('west', 'West')]
+                                          )
```
antonrom1: This indentation is a little bit strange, I would do something like this
```suggestion
    garden_orientation = fields.Selection(
        string='Garden orientation',
        selection=[
            ('north', 'North'),
            ('south', 'South'),
            ('east', 'East'),
            ('west', 'West'),
        ],
    )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+    state = fields.Selection(string='State',
+                            selection=[('new', 'New'),
+                                    ('offer_received', 'Offer Received'),
+                                    ('offer_accepted', 'Offer Accepted'),
+                                    ('sold', 'Sold'),
+                                    ('cancelled', 'Cancelled')],
+                             default='new', required=True, copy=False)
```
antonrom1: Same, the indent is a little bit weird, it looks better to just have one argument per line here
```suggestion
    state = fields.Selection(
        string='State',
        selection=[
            ('new', 'New'),
            ('offer_received', 'Offer Received'),
            ('offer_accepted', 'Offer Accepted'),
            ('sold', 'Sold'),
            ('cancelled', 'Cancelled'),
        ],
        default='new',
        required=True,
        copy=False,
    )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+        for record in self:
+            record.total_area = record.garden_area + record.living_area
+
+    @api.depends('offer_ids.price')
+    def _compute_best_offer(self):
+        for record in self:
+            if record.offer_ids:
+                record.best_offer = max(record.offer_ids.mapped('price'))
+            else:
+                record.best_offer = 0
```
antonrom1: you can just do this
```suggestion
            record.best_offer = max(record.offer_ids.mapped('price'), default=0.0)
```

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,49 @@
...
+        <field name="arch" type="xml">
+            <list string="Estate Property Offer" editable="bottom"
+                  decoration-success="status == 'accepted'"
+                  decoration-danger="status == 'refused'">
+                <field name="price"/>
+                <field name="partner_id"/>
+                <field name="validity"/>
+                <field name="date_deadline"/>
+                <button name="action_accept" title="Accept" type="object" icon="fa-check" invisible="status"/>
+                <button name="action_refuse" title="Refuse" type="object" icon="fa-close" invisible="status"/>
```
antonrom1, line 37: these invisibles don't seem correct. I think it should be `status in ('accepted', 'refused')`

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,59 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
```
antonrom1, line 1: Same, you don't need to put the odoo license in every file anymore

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class EstatePropertyOffer(models.Model):
+    _name = 'estate.property.offer'
+    _description = 'Estate Property Offer'
+    _order = "price desc"
+
+    price = fields.Float(string='Price')
+    status = fields.Selection(string='Status', copy=False,
+                            selection=[('accepted', 'Accepted'), ('refused', 'Refused')])
```
antonrom1: Same, this indentation looks a little bit strange. When you have multiline arguments IMO it's better to do one argument per line
```suggestion
    status = fields.Selection(
        string='Status', 
        copy=False,
        selection=[('accepted', 'Accepted'), ('refused', 'Refused')]
    )
```

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,17 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import fields, models
+
+
+class EstatePropertyTag(models.Model):
+    _name = 'estate.property.tag'
+    _description = 'Estate Property Tag'
+    _order = "name"
```
antonrom1: little nitpick, although not everybody respects that at odoo, we have somewhat of a convention for using single vs double quotes. Everything that is user-facing (e.i. everything an end-user will see) is usually double-quoted. Everything else is single quoted (so mostly technical strings). 
```suggestion
    _description = "Estate Property Tag"
    _order = 'name'
```

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,49 @@
...
+            </list>
+        </field>
+    </record>
+
+    <record id="estate_property_offer_action" model="ir.actions.act_window">
+            <field name="name">Offers</field>
+            <field name="res_model">estate.property.offer</field>
+            <field name="view_mode">list,form</field>
+            <field name="domain">[('property_type_id','=',active_id)]</field>
```
antonrom1:
```suggestion
            <field name="domain">[('property_type_id', '=', active_id)]</field>
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,59 @@
...
+    @api.depends('validity')
+    def _compute_deadline(self):
+        for record in self:
+            record.date_deadline = (record.create_date if record.create_date else date.today()) + timedelta(days=record.validity)
+
+    def _inverse_deadline(self):
+        for record in self:
+            record.validity = (record.date_deadline - (record.create_date.date() if record.create_date else date.today())).days
+
+    @api.model
```
antonrom1: You already support a list of dicts, you can add @api.model_create_multi decorator

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,59 @@
...
+    def create(self, vals_list):
+        for vals in vals_list:
+            if 'property_id' in vals and vals.get('property_id'):
```
antonrom1: You are checking almost the same condition twice
```suggestion
            if vals.get('property_id'):
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,59 @@
...
+                current_property = self.env['estate.property'].browse(vals['property_id'])
+                if 'price' in vals and current_property.best_offer > vals.get('price', 0):
+                    raise exceptions.ValidationError(f'The offer must be higher than {current_property.best_offer}')
```
antonrom1: Not a problem for the tutorial, but with an actual module, we usually don't use fstrings for errors, as you cannot translate these

Here is an example of how we usually format translated strings with parameters
https://github.com/odoo/odoo/blob/fc658b2e3d9c5afbfa5d00b579bf8928a90dab87/addons/barcodes/models/barcode_rule.py#L41

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,17 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
```
antonrom1: No need license

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,148 @@
...
+            </kanban>
+        </field>
+    </record>
+
+    <record id="estate_property_action" model="ir.actions.act_window">
+        <field name="name">Estate Property</field>
+        <field name="res_model">estate.property</field>
+        <field name="view_mode">list,form,kanban</field>
+        <field name="search_view_id" ref="estate_property_view_search"/>
+        <field name="context">{'search_default_filter_available':1}</field>
```
antonrom1: Keep the spacing consistent in XMLs too
```suggestion
        <field name="context">{'search_default_filter_available': 1}</field>
```

estate_account/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    'depends': ['estate', 'account'],
+    'author': "taskv",
+    'category': 'Tutorials',
+    'description': """
+    Tutorial Project
+    """,
+    'data': [
+    ],
+    'demo': [
+    ],
```
antonrom1: Don't need to define these if no data or demo

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,31 @@
...
+
+from odoo import models, Command
+
+
+class EstateProperty(models.Model):
+    _inherit = 'estate.property'
+
+    def action_property_sold(self):
+        invoice_vals_list = []
+        for record in self:
```
antonrom1: In super().action_property_sold you call self.ensure_one(), and here you suppose that it's an iterable. You should pick one.

antonrom1 (commented):
Good job! Mostly small comments. Keep it going like this for JS next week :P

estate/views/estate_property_offer_views.xml
tskv: Let me explain why I've done like this. 
The "status" is a selection-field with two possible values ('accepted' or 'refused'). According to the tutorial task, if it is defined, we don't need to show these two buttons. We want to see them only if "status" is empty yet. 
So i thought in this case invisible="status" should also work fine

PR #1044, opened 2025-12-03
[ADD] sale_discount:  Added the recomputation logic of global discount

hapat-odoo:
- Updated sale order methods to recalculate discount on all line updates.
- Removed discount line when order becomes empty.
- Enforced single global discount line handling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,10 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+.eslintignore
+.eslintrc.json
+jsconfig.json
+package.json
+package-lock.json
+node_modules/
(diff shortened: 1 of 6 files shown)


.gitignore
bit-odoo, line 1: unneccary changes.

bit-odoo (commented):
Hello,
can you please update your commit message tiltle and add task id in end of commit message.

sale_discount/models/sale_order.py
```diff
@@ -0,0 +1,35 @@
...
+        if not product_lines:
+            self.order_line -= discount_lines
+            return
+
+        subtotal = 0
+        for line in product_lines:
+            subtotal += line.price_subtotal
+
+        for discount_line in discount_lines:
+            match = re.search(r"(\d+(?:\.\d+)?)%", discount_line.name)
```
bit-odoo, line 32: Is this required?

PR #1050, opened 2025-12-09
[ADD] sale_order_zero_stock_blockage: add manager approval on insufficient stock

danal-odoo:
Currently, a sales user can confirm a Quotation even if it contains no
products or if the products have insufficient stock compared to the
order quantity. This leads to the creation of invalid Sales Orders,
which negatively impacts inventory planning and invoicing.

This module introduces a validation to the Sales Order confirmation
process to prevent these issues.

The following checks are added:
1. Prevent confirmation if the order has no lines.
2. Prevent confirmation if the product stock is insufficient.

To bypass the insufficient stock restriction, a user with the 'Sales Manager'
group must enable the new 'Approval' (zero_stock_approval) field on the
order.

Technical details:
- Model `sale.order`: Added `zero_stock_approval` Boolean field.
- Method `fields_get`: Overridden to set `zero_stock_approval` as
  readonly for users who are not in the `sales_team.group_sale_manager` group.
- Method `action_confirm`: Added logic to raise a `UserError` if:
    - The order has no lines.
    - A product's `qty_available` is less than `product_uom_qty` (for
      goods) and the `zero_stock_approval` is not enabled.
- added TestCases.

Task-5382939

### DIFF
--- a/sale_order_zero_stock_blockage/__init__.py
+++ b/sale_order_zero_stock_blockage/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/sale_order_zero_stock_blockage/__manifest__.py
+++ b/sale_order_zero_stock_blockage/__manifest__.py
@@ -1,0 +1,18 @@
+{
+    "name": "Sale Order Zero Stock Blockage",
+    "description": """
+    Zero Stock Blockage is module to prevent order to confirm if product is out of stock.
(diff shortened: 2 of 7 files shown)


bit-odoo:
Hello @danal-odoo 
Can you please add a test for this?

PR #1055, opened 2025-12-15
vijan - Technical Training  

VincentJanssen-Code:
This pull request is for the [Server 101 tutorial](https://www.odoo.com/documentation/19.0/developer/tutorials/server_framework_101.html) of the on boarding of vijan


estate/models/estate_property.py
```diff
@@ -0,0 +1,84 @@
...
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Property"
+    _order = "id desc"
+
+    _positif_expected_price = models.Constraint("CHECK (expected_price > 0)","A price can't be negatif");
+    _positif_selling_price = models.Constraint("CHECK (selling_price > 0)","A price can't be negatif");
+
+    state = fields.Selection(selection = [("New","New"), ("Offer_Received","Offer Received") ,("Offer_Accepted","Offer Accepted"), ("Sold","Sold"), ("Cancelled","Cancelled")])
```
amah-odoo: for the selection fields I prefer to write them that way to make the choices clear while we look at the field. and for the values the first part of each tuple is the technical value for the field which should be written like a var name in snake case 🐍 
```suggestion
    state = fields.Selection([
        ("new","New"),
        ("offer_received","Offer Received"),
        ("offer_accepted","Offer Accepted"),
        ("sold","Sold"),
        ("cancelled","Cancelled")
    ], default="new")
```

estate/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+from . import estate_property
+from . import estate_property_type
+from . import estate_property_tag
+from . import estate_property_offer
```
amah-odoo: newline missed

estate/models/estate_property.py
```diff
@@ -0,0 +1,84 @@
...
+    active = fields.Boolean('Active',default=True)
+    name = fields.Char(required=True,default="Unkown")
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date("Available From",copy=False,default= fields.Datetime.today() + relativedelta(months=3))
```
amah-odoo: when you run the server for the first time the default values will be stored on the db so when you open the next day it will be 3 months from day it got evaluated. but with the lambda it will be expression that will be evaluated every time 
```suggestion
    date_availability = fields.Date("Available From",copy=False,default= lambda self: fields.Datetime.today() + relativedelta(months=3))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,84 @@
...
+    last_seen= fields.Date("Last Seen", default=fields.Datetime.now)
```
amah-odoo: same as `date_availability`

estate/models/estate_property.py
```diff
@@ -0,0 +1,84 @@
...
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float(readonly=True,copy=False)
+    best_price = fields.Float(string="Best Price",compute="_get_best_price")
```
amah-odoo: we tend to use the `_compute` prefix for all the compute methods 👀 
```suggestion
    best_price = fields.Float(string="Best Price",compute="_compute_best_price")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,83 @@
...
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Property"
+    _order = "id desc"
+
+    _positif_expected_price = models.Constraint("CHECK (expected_price > 0)", "A price can't be negatif")
+    _positif_selling_price = models.Constraint("CHECK (selling_price > 0)", "A price can't be negatif")
+
+    state = fields.Selection(selection=[("New", "New"), ("Offer_Received", "Offer Received"), ("Offer_Accepted", "Offer Accepted"), ("Sold", "Sold"), ("Cancelled", "Cancelled")])
```
amah-odoo: I prefer to write Selection that way to make the choices clear while reading the file. Also the tuples' first value should be written in snake_case 🐍 as they are technical names
```suggestion
    state = fields.Selection(selection=[
        ("new", "New"),
        ("offer_received", "Offer Received"),
        ("offer_accepted", "Offer Accepted"),
        ("sold", "Sold"),
        ("cancelled", "Cancelled")
    ], default="new")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,83 @@
...
+    active = fields.Boolean('Active', default=True)
+    name = fields.Char(required=True, default="Unkown")
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date("Available From", copy=False, default=fields.Datetime.today() + relativedelta(months=3))
```
amah-odoo: without the lambda this will be evaluated once the server is running and will not be re-evaluated with each new record which is not the case we want 🤔 
```suggestion
    date_availability = fields.Date("Available From", copy=False, default=lambda self: fields.Datetime.today() + relativedelta(months=3))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,83 @@
...
+    last_seen = fields.Date("Last Seen", default=fields.Datetime.now)
```
amah-odoo: same here as `date_availability` 👀

estate/models/estate_property.py
```diff
@@ -0,0 +1,83 @@
...
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float(readonly=True, copy=False)
+    best_price = fields.Float(string="Best Price", compute="_get_best_price")
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(selection=[('North', 'North'), ('South', 'South'), ('East', 'East'), ('West', 'West')])
```
amah-odoo: same here as `state`

estate/models/estate_property.py
```diff
@@ -0,0 +1,83 @@
...
+    state = fields.Selection(selection=[("New", "New"), ("Offer_Received", "Offer Received"), ("Offer_Accepted", "Offer Accepted"), ("Sold", "Sold"), ("Cancelled", "Cancelled")])
+    active = fields.Boolean('Active', default=True)
+    name = fields.Char(required=True, default="Unkown")
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date("Available From", copy=False, default=fields.Datetime.today() + relativedelta(months=3))
+    last_seen = fields.Date("Last Seen", default=fields.Datetime.now)
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float(readonly=True, copy=False)
+    best_price = fields.Float(string="Best Price", compute="_get_best_price")
```
amah-odoo: we tend to use '_compute' prefix for the compute methods

estate/models/estate_property.py
```diff
@@ -0,0 +1,83 @@
...
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(selection=[('North', 'North'), ('South', 'South'), ('East', 'East'), ('West', 'West')])
+    total_area = fields.Float(compute="_get_total_area", string="Total Area")
```
amah-odoo: same as `best_price`

estate/models/estate_property.py
```diff
@@ -0,0 +1,83 @@
...
+    property_type_id = fields.Many2one("estate.property.type", string="Type")
+    buyer_id = fields.Many2one("res.partner", string="Buyer")
+    seller_id = fields.Many2one("res.users", default=lambda self : self.env.user,string="Seller")
+    tag_ids = fields.Many2many("estate.property.tag", string="Tags")
+    offer_ids = fields.One2many("estate.property.offer", "property_id",string="Offers")
+
+    def sell_property(self):
+        for record in self:
+            if (record.state == "Cancelled"):
+                raise exceptions.UserError("Can't sell cancelled property.")
```
amah-odoo: you could import it like the `ValidationError` to not import the whole exceptions module

estate/models/estate_property.py
```diff
@@ -0,0 +1,83 @@
...
+        for record in self:
+            max: int = 0
+            if (len(record.offer_ids) == 0):
+                record.best_price = 0
+                continue
+
+            for offer in record.offer_ids:
+                if (offer.price >= max):
+                    max = offer.price
+                record.best_price = max
```
amah-odoo: the `.mapped('field_name')` method return a list with all the values for this field for this record set
```suggestion
            record.best_price = max(record.offer_ids.mapped('price'))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,83 @@
...
+    @api.depends('living_area', 'garden_area')
+    def _get_total_area(self):
+        for record in self:
+            record.total_area = record.garden_area + record.living_area
+
+    @api.depends('offer_ids')
+    def _get_best_price(self):
+        for record in self:
+            max: int = 0
+            if (len(record.offer_ids) == 0):
```
amah-odoo: the empty record set is a false value so you can just check for that 
```suggestion
            if not record.offer_ids:
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,83 @@
...
+                record.best_price = 0
+                continue
+
+            for offer in record.offer_ids:
+                if (offer.price >= max):
+                    max = offer.price
+                record.best_price = max
+
+    @api.onchange("garden")
+    def _garden_pre_fill(self):
```
amah-odoo: we have the `_on_change` prefix also 👀 
```suggestion
    def _on_change_garden(self):
```

amah-odoo (commented):
Huge Work 👏🏼 

let's go for those today... TO BE CONTINUED 👀

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property_model,access_estate_property_model,model_estate_property,base.group_user,1,1,1,1
+access_estate_property_type_model,access_estate_property_type_model,model_estate_property_type,base.group_user,1,1,1,1
+access_estate_property_tag_model,access_estate_property_tag_model,model_estate_property_tag,base.group_user,1,1,1,1
+access_estate_property_offer_model,access_estate_property_offer_model,model_estate_property_offer,base.group_user,1,1,1,1
```
amah-odoo: missing end of file newline 👀

estate/views/estate_menu_views.xml
```diff
@@ -0,0 +1,12 @@
...
+    <menuitem id="estate_menu_root" name="Estate">
+        <menuitem id="estate_property_menu" name="Property">
+            <menuitem id="estate_property_menu_action" action="estate_property_model_action"/>
+        </menuitem>
+        <menuitem id="estate_settings_menu" name="Settings">
+            <menuitem id="estate_property_type_menu_acion" action="estate_property_type_model_action"/>
+            <menuitem id="estate_property_tag_menu_acion" action="estate_property_tag_model_action"/>
+        </menuitem>
+    </menuitem>
+</odoo>
```
amah-odoo: missing end of file newline 👀

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,59 @@
...
+            record.validity = (datetime.combine(record.date_deadline, time()) - record.create_date).days
+
+    @api.model
+    def create(self, vals_list):
+        for record in vals_list:
+            property = self.env['estate.property'].browse(record['property_id'])
+            if (float_compare(record['price'], property.best_price, 2) == -1):
+                raise UserError("This offer is lower than what has already been offered.")
+            property.state = 'offer_received'
+        return super().create(vals_list)
```
amah-odoo, line 65: I think we can optimize this browse and write calls for the property model by getting all the properties before the loop and update all of them once after the creation. Also I think we should update the state at least after the offer creation or at the property model it self 🤔

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,45 @@
...
+                <field name="date_deadline"/>
+                <field name="validity"/>
+                <field name="property_type_id"/>
+                <button name="accept_offer" type="object" string="Accept" icon="fa-check" invisible="status == 'accepted' or status == 'refused'"/>
+                <button name="refused_offer" type="object" string="Refuse" icon="fa-times" invisible="status == 'accepted' or status == 'refused'"/>            
+            </list>
+        </field>
+    </record>
+
+</odoo>
```
amah-odoo: missing end of file newline 👀

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0"?>
+<odoo>
+
+    <record id="estate_property_tag_model_action" model="ir.actions.act_window">
+        <field name ="name">Property Tag</field>
+        <field name="res_model">estate.property.tag</field>
+        <field name="view_mode">list,form</field>
+    </record>
+
+    <record id="estate_property_tag_model_form" model="ir.ui.view">
```
amah-odoo: I think we do not need the form and the list as it is only the name and this is the default for list 🤔

amah-odoo (commented):
Huge work 💯 

with the translation this will be more than prefect module

PR #1058, opened 2025-12-15
auber - Technical Training

aurb9:
This provides an implementation for a module allowing people to manage their properties, people to place offers and alters the users' view to include the list of properties being sold by the person

### DIFF
--- a/estate/models/__init__.py
+++ b/estate/models/__init__.py
@@ -1,0 +1,5 @@
+from . import property
+from . import property_offer
+from . import property_tag
+from . import property_type
+from . import res_users
--- a/estate/models/property.py
+++ b/estate/models/property.py
@@ -1,0 +1,110 @@
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools import float_compare, float_is_zero
+
+
+class Property(models.Model):
+    _name = 'estate.property'
+    _description = "Estate property"
+    _order = 'id desc'
+
+    name = fields.Char(string="Title", required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        string="Available From",
+        default=lambda self: fields.Date.today() + relativedelta(months=3),
+        copy=False,
+    )
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float(readonly=True, copy=False)
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer(string="Living Area (sqm)")
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer(string="Garden Area (sqm)")
+    garden_orientation = fields.Selection(
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
+    )
+    state = fields.Selection(
+        default="new",
+        string="Status",
+        selection=[
+            ('new', 'New'),
+            ('offer_received', 'Offer Received'),
+            ('offer_accepted', 'Offer Accepted'),
+            ('sold', 'Sold'),
+            ('cancelled', 'Cancelled'),
+        ],
+        required=True,
+        copy=False,
+    )
+    active = fields.Boolean(default=True)
+    property_type_id = fields.Many2one('estate.property.type', string="Property Type")
+    partner_id = fields.Many2one('res.partner', string="Buyer", readonly=True)
+    user_id = fields.Many2one('res.users', string="Salesman", default=lambda self: self.env.user)
+    tag_ids = fields.Many2many('estate.property.tag', string="Property Tags")
+    offer_ids = fields.One2many('estate.property.offer', 'property_id', string="Offers")
+    total_area = fields.Integer(string="Total Area (sqm)", compute='_compute_total_area')
+    best_offer = fields.Float(default=0.0, compute="_compute_best_offer")
+
+    _expected_price_strictly_pos = models.Constraint(
+        "CHECK(expected_price > 0)", "The expected price must be strictly positive."
+    )
+    _selling_price_pos = models.Constraint(
+        "CHECK(selling_price >= 0)", "The selling price must be positive."
+    )
+
+    @api.depends('living_area', 'garden_area')
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends('offer_ids')
+    def _compute_best_offer(self):
+        for record in self:
+            record.best_offer = max(record.offer_ids.mapped('price')) if record.offer_ids else 0.0
+
+    @api.onchange('garden')
+    def _onchange_garden(self):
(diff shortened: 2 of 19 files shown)


barracudapps (changes requested):
Please review your commit message and its title to follow [Odoo's Git Guidelines](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html)

**FYI**
[REV] stands for *Revert*. Here, you didn't revert anything, you didn't "erase" something but changes/added. IMHO [FIX] would've been a better tag

estate/models/property.py
```diff
@@ -0,0 +1,48 @@
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models
+
+
+class Property(models.Model):
+    _name = 'estate.property'
+    _description = 'Estate property'
```
barracudapps: Please use double quotes for non-technical strings
```suggestion
    _description = "Estate property"
```

barracudapps (changes requested):
Hi @aurb9 
Can you please consider my nit and apply it wherever it suits ?
Thanks!

estate/models/property.py
```diff
@@ -65,13 +66,20 @@ def _compute_total_area(self):
...
         for record in self:
-            record.best_offer = max(record.offer_ids.mapped('price'))
+            record.best_offer = max(record.offer_ids.mapped('price')) if record.offer_ids else 0.0
 
     @api.onchange('garden')
     def _onchange_garden(self):
         self.garden_area = 10 if self.garden else 0
         self.garden_orientation = 'north' if self.garden else ''
 
+    @api.onchange("offer_ids")
```
barracudapps: This is a technical string 😉

estate/models/property.py
```diff
@@ -65,13 +66,20 @@ def _compute_total_area(self):
...
+    def _onchange_state(self):
+        if not self.offer_ids:
+            return
```
barracudapps: I think you might need to update `self.state` here even with no changes (`self.state = self.state`) for the ORM to work properly. To be checked though

estate/views/estate_property_views.xml
```diff
@@ -6,8 +6,11 @@
         <field name="arch" type="xml">
             <form string="Property">
                 <header>
-                    <button name="action_set_sold" type="object" string="SOLD" class="oe_highlight"/>
-                    <button name="action_set_cancelled" type="object" string="CANCEL" class="oe_highlight"/>
+                    <button name="action_set_sold" type="object" string="SOLD" class="oe_highlight"
+                            invisible="state == 'sold'"/>
+                    <button name="action_set_cancelled" type="object" string="CANCEL" class="oe_highlight"
+                            invisible="state == 'sold'"/>
```
barracudapps: Some logic question: do we want to be allowed to sell a property in a canceled state?

barracudapps (changes requested):
Quick review requiring some changes (nit)

estate/models/property.py
aurb9: It worked fine for me, but if that's best-practice then I will change it :)

estate/views/estate_property_views.xml
aurb9: No, but it was a task in the model itself to throw a `UserError` if we try to sell a cancelled property and vice-versa. Should the logic be in there instead?

barracudapps: To prevent the error to be raised, not allowing the user to click the button might lead to a better UX

barracudapps:
@aurb9 
You still have an issue on runbot with your XML file `estate_property_type_views.xml`. Try to make it green

aurb9:
> @aurb9 You still have an issue on runbot with your XML file `estate_property_type_views.xml`. Try to make it green

Solved -- was importing the views in the wrong order

estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import property, property_offer, property_tag, property_type, res_users
```
barracudapps: Didn't see this last time 🤔
Split this on multiple lines so that if someone has to add a new file and import it here, it doesn't break your git history

estate/models/property.py
```diff
@@ -0,0 +1,110 @@
...
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float(readonly=True, copy=False)
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer(string="Living Area (sqm)")
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer(string="Garden Area (sqm)")
+    garden_orientation = fields.Selection(
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
```
barracudapps, line 30: As discussed, here are some non-technical strings that should have double quotes (no need to spend time on it for the tutorial as I know you got the point)

estate/models/property.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @api.depends('offer_ids')
+    def _compute_best_offer(self):
+        for record in self:
+            record.best_offer = max(record.offer_ids.mapped('price')) if record.offer_ids else 0.0
+
+    @api.onchange('garden')
+    def _onchange_garden(self):
+        self.garden_area = 10 if self.garden else 0
+        self.garden_orientation = 'north' if self.garden else ''
```
barracudapps, line 74: I would prefer this version as it might lead to some extensions in the future but not a required change so far
```suggestion
        if self.garden:
            self.garden_area = 10
            self.garden_orientation = 'north'
        if self.garden:
            self.garden_area = 0
            self.garden_orientation = False
```

estate/models/property.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @api.constrains('selling_price', 'expected_price')
+    def _check_selling_price(self):
+        for record in self:
+            if (
+                not float_is_zero(record.selling_price, precision_digits=2) and
+                float_compare(record.selling_price, record.expected_price * 0.9, precision_digits=2) == -1
+            ):
```
barracudapps, line 82: Could have been only one condition as `expected_price` is strictly positive (via your SQL constraint), this leads `selling_price to be strictly positive as well (but when `expected_price` is very small, which is a functional stupid case).

BTW, you could also have added a SQL constraint like
```sql
(
    "check_selling_price",
    "CHECK(selling_price >= 0.9 * expected_price)",
    "Selling price must be at least 90% of expected price.",
),
```

estate_account/__manifest__.py
```diff
@@ -0,0 +1,7 @@
+{
+    'name': 'Real Estate Account',
+    'depends': ['base', 'estate', 'account'],
```
barracudapps: Avoid redundancies here as `estate` already relies on `base`
```suggestion
    'depends': ['estate', 'account'],
```

barracudapps (approved):
LGTM 👍
Please adapt your PR title and message to summarize your work.
Then, you can squash your commits and mark the PR as ready 🚀

barracudapps:
> mark the PR as ready 🚀

*This is for a common R&D flow having pre-review requirements, nothing to do for the training 😉*

PR #1062, opened 2025-12-15
Real Estate module (ibmah)

ibmah-odoo:

PR #1069, opened 2025-12-15
real estate module (ibmah)

ibmah-odoo:
closes #1062. This one is based on master instead of v19.0

### DIFF
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -9,4 +9,7 @@
     """,
     "application": True,
     "installable": True,
+    "data": [
+        "data/ir.model.access.csv",
+    ],
 }
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -16,5 +16,6 @@
         "views/estate_property_type.xml",
         "views/estate_property_tag.xml",
         "views/estate_menus.xml",
+        "views/res_users.xml",
     ],
 }
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -12,6 +12,9 @@
     "data": [
         "data/ir.model.access.csv",
         "views/estate_property_views.xml",
+        "views/estate_property_type.xml",
+        "views/estate_property_tag.xml",
+        "views/estate_property_offer.xml",
         "views/estate_menus.xml",
     ],
 }
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -1,0 +1,12 @@
+{
+    "name": "estate",
+    "version": "1.0",
+    "depends": ["base"],
+    "author": "Odoo S.A.",
+    "category": "Real Estate",
+    "description": """
+    Description text
+    """,
+    "application": True,
+    "installable": True,
+}
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -11,5 +11,7 @@
     "installable": True,
     "data": [
         "data/ir.model.access.csv",
+        "views/estate_property_views.xml",
+        "views/estate_menus.xml",
     ],
 }
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -18,4 +18,5 @@
         "views/estate_menus.xml",
         "views/res_users.xml",
     ],
+    'license': 'AGPL-3',
 }
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -12,9 +12,9 @@
     "data": [
         "data/ir.model.access.csv",
         "views/estate_property_views.xml",
+        "views/estate_property_offer.xml",
         "views/estate_property_type.xml",
         "views/estate_property_tag.xml",
-        "views/estate_property_offer.xml",
         "views/estate_menus.xml",
     ],
 }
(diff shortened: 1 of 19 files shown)


estate/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+{
+    'name': "estate",
+    'version': '1.0',
+    'depends': ['base'],
+    'author': "ibrahim (ibmah)",
```
amah-odoo: The author is always `'Odoo S.A.'` for the RD because to develop for odoo it is not out own apps 🤔

estate/models/estate_property.py
```diff
@@ -0,0 +1,26 @@
...
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        selection=[('north', 'North'),
+                   ('south', 'South'),
+                   ('east', 'East'),
+                   ('west', 'West')],
+        help = 'Indicates the direction the garden is facing with respect to the house')
+
```
amah-odoo, line 58: we tend to leave only 1 line at the end of the file

estate/models/estate_property.py
```diff
@@ -0,0 +1,26 @@
...
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        selection=[('north', 'North'),
+                   ('south', 'South'),
+                   ('east', 'East'),
+                   ('west', 'West')],
```
amah-odoo: A personal pref but want to show another way to structure this
```suggestion
    garden_orientation = fields.Selection([
        ('north', 'North'),
        ('south', 'South'),
        ('east', 'East'),
        ('west', 'West')
    ], help = "put it here")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,26 @@
+from odoo import fields, models
+
+class EstateProperty(models.Model):
+    _name = "estate_property"
+    _description = "Estate Property data model"
+    name = fields.Char(required=True)
```
amah-odoo, line 12: we can have a blank line to make clear the end of the metadata and fields

amah-odoo (commented):
Let's go with the first review 🎉

.gitignore
```diff
@@ -127,3 +127,4 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+pyrightconfig.json
```
amah-odoo: do not add your own `.gitignore` to the commits

amah-odoo (commented):
This is mainly about the commit msg let's try working to have the chapter number in the commit header (commit first line) and I want you to add some description for each commit body with what is done in this chapter

estate/models/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import (estate_property, estate_property_offer, estate_property_tag,
+               estate_property_type)
```
amah-odoo: we tend to use this structure in the imports
```suggestion
from . import estate_property
from . import estate_property_offer
from . import estate_property_tag
from . import estate_property_type
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    active = fields.Boolean(default=True)
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        copy=False, default=datetime.now() + relativedelta(month=3)
+    )
+    expected_price = fields.Float(required=True)
+    _check_expected_price = models.Constraint(
```
amah-odoo: can you move all the `models.Constraint` after all the fields are defined

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+                record.state = "sold"
+            else:
+                raise exceptions.UserError("can't sell a cancelled property")
+
+    def cancel_estate_property(self):
+        for record in self:
+            if record.state != "sold":
+                record.state = "cancelled"
+            else:
+                raise exceptions.UserError("can't cancel a sold property")
```
amah-odoo, line 97: it is not optimal to import all the module to have only 1 or 2 object so we use to
`from odoo.exceptions import UserError` and use it directly 
```suggestion
                raise UserError("can't cancel a sold property")
```

amah-odoo (commented):
Small comments about .py structure you can refer to this [DOC](https://www.odoo.com/documentation/18.0/contributing/development/coding_guidelines.html#symbols-and-conventions) to know about it all

estate/models/estate_property.py
ibmah-odoo: I usually prefer qualified imports since they prevent namespace pollution and name clashes on larger modules at the cost of being a tiny bit verbose.

It's a common misconception that importing a single object from a module doesn't cause python to import and cache the entire module. There's no performance hit for importing a module compared to an object in the module.

> The only difference between the two statements is what name is bound

[ref](https://softwareengineering.stackexchange.com/a/187471)

ibmah-odoo: If that's the convention in odoo codebase, I'd happily adopt it but thought I'd put this info out there

PR #1071, opened 2025-12-15
Master technical onboarding faavc

Faruk372742:


estate/models/estate_property.py
```diff
@@ -0,0 +1,20 @@
...
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float()
+    bedrooms = fields.Integer()
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(string="Orientation", selection=[("North", "North"), ("South", "South"), ("East", "East"), ("West", "West")])
+
```
amah-odoo: do not forget the new line at the end of file

estate/models/estate_property.py
```diff
@@ -0,0 +1,20 @@
...
+    date_availability = fields.Date()
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float()
+    bedrooms = fields.Integer()
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(string="Orientation", selection=[("North", "North"), ("South", "South"), ("East", "East"), ("West", "West")])
```
amah-odoo: I would prefer to right the selection options that way to make it more clearer Also the tuple first value represent the technical name so it should be in snake_case 🐍 
```suggestion
    garden_orientation = fields.Selection([
        ("north", "North"),
        ("south", "South"),
        ("east", "East"),
        ("west", "West")
    ], string="Orientation")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class EstateProperty(models.Model):
+    _name = "estate_property"
+    _description = "Estate Property"
+    name = fields.Char("Title", required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        copy=False, default=datetime.now() + relativedelta(months=3)
```
amah-odoo: for the default values it is get evaluated on server boot-up So without the `lambda` once you run the server it will take the date and after 1 week if you created new property it will take the 3 month from the server boot-up not 3 month from real today.
```suggestion
        copy=False, default=lambda self: datetime.now() + relativedelta(months=3)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,93 @@
...
+        copy=False,
+    )
+    type_id = fields.Many2one("estate_property_type", string="Property Type")
+    salesperson_id = fields.Many2one(
+        "res.users", string="Salesperson", default=lambda self: self.env.user
+    )
+    buyer_id = fields.Many2one("res.partner", string="Buyer", copy=False)
+    tag_ids = fields.Many2many("estate_property_tag")
+    offer_ids = fields.One2many("estate_property_offer", "property_id", string="Offers")
+    total_area = fields.Integer("Total Area (sqm)", compute="_compute_area")
```
amah-odoo: for compute method we tend to use the full field name if it is the only field computed in the method
```suggestion
    total_area = fields.Integer("Total Area (sqm)", compute="_compute_total_area")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,93 @@
...
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for record in self:
+            record.best_price = (
+                0
+                if len(record.offer_ids) == 0
+                else max(offer.price for offer in record.offer_ids)
+            )
```
amah-odoo: the `.mapped('field')` return a list with the all the field values for the recordset and the `or [0]` will replace the list of prices with list contain 0 if the list is empty list
```suggestion
            record.best_price = max(record.offer_ids.mapped('price') or [0])
```

amah-odoo (commented):
Huge work so far 🎉 

just small comment for now

estate/models/estate_property.py
```diff
@@ -0,0 +1,136 @@
...
+            ("new", "New"),
+            ("offer_received", "Offer Received"),
+            ("offer_accepted", "Offer Accepted"),
+            ("sold", "Sold"),
+            ("cancelled", "Cancelled"),
+        ],
+        default="new",
+        required=True,
+        copy=False,
+        store=True,
```
amah-odoo: we do not need this as this is not computed anymore 👀

estate/models/estate_property.py
```diff
@@ -0,0 +1,136 @@
...
+    best_price = fields.Float("Best Price", compute="_compute_best_price")
+
+    _check_expected_price = models.Constraint(
+        "CHECK(expected_price > 0)", "The expected price must be positive"
+    )
+    _check_selling_price = models.Constraint(
+        "CHECK(selling_price > 0)", "The selling price must be positive"
+    )
+
+    _order = "id desc"
```
amah-odoo: let's keep all the meta data at first before any field declaration

estate/models/estate_property.py
```diff
@@ -0,0 +1,136 @@
...
+    def _compute_best_price(self):
+        for record in self:
+            record.best_price = max(record.offer_ids.mapped("price") or [0])
+
+    # @api.depends("offer_ids")
+    # def _compute_state(self):
+    #     for record in self:
+    #         if len(record.offer_ids) == 1 and record.state == "new":
+    #             # record.state = "offer_received" OPEN here do not forget!!!
+    #             record.state = record.state
```
amah-odoo: let's drop this also 🤔

amah-odoo (commented):
small comments to keep structure 👀

amah-odoo (commented):
Finilizing you framework journey please adapt the code to follow those [Coding Guidelines](https://www.odoo.com/documentation/18.0/contributing/development/coding_guidelines.html#symbols-and-conventions) 🫡 

Thanks for your work 🎉

PR #1077, opened 2025-12-16
[ADD] estate: add real estate module (framework 101 tutorial)

robal-odoo:


estate/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
+    'name': 'Real Estate',
+    'description': 'Real Estate Advertising',
+    # 'version': '1.0',
+    'author': 'robal',
```
ushyme:
```suggestion
    'author': 'Odoo S.A.',
```
As an Odoo employee, you are creating this module in the name of Odoo

estate/models/estate_property.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate property"
+
+    name = fields.Char(string="Title", required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(string="Available From", default=fields.Date.today() + relativedelta(months=3), copy=False)
```
ushyme:
```suggestion
    date_availability = fields.Date(string="Available From", default=lambda self: fields.Date.context_today(self) + relativedelta(months=3), copy=False)
```


`date_availability`'s default was calculated only once at server launch. Wrapping it in a `lambda` fixes this by computing it dynamically for each new record. I also used `fields.Date.context_today(self)` for timezone correctness.

estate/models/estate_property.py
```diff
@@ -0,0 +1,112 @@
...
+        else:
+            self.garden_area = None
+            self.garden_orientation = None
+
+    # Public methods
+    def action_set_sold(self) -> bool:
+        for record in self:
+            if record.state == "sold":
+                error_msg = "This property has already been sold."
+                raise UserError(error_msg)
```
ushyme:
```suggestion
            if record.state == "sold":
                continue
```
In the `action_` methods, it's slightly more user-friendly to silently ignore a button click if the record is already in the target state, rather than raising an error.

estate/models/estate_property.py
```diff
@@ -0,0 +1,112 @@
...
+
+            if record.state == "cancelled":
+                error_msg = "Cancelled properties cannot be sold."
```
ushyme:
```suggestion
                error_msg = record.env._("Cancelled properties cannot be sold.")
```
User-facing strings in errors and constraints must be wrapped in `record.env._()` to be translatable.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # Constraints
+    _price_strict_positive = models.Constraint(
+        "CHECK(price > 0)",
+        "An offer's price must be strictly greater than 0.",
+    )
+
+
+def _get_reference_date(offer: EstatePropertyOffer) -> date:
+    return fields.Date.today() if not offer.create_date else offer.create_date.date()
```
ushyme: I don't know if this function is really helpful here. I think the code would overall be more readable if this logic was inlined directly in the compute/inverse method.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,75 @@
...
+        for record in self:
+            record.validity = (record.date_deadline - _get_reference_date(record)).days
+
+    # CRUD overrides
+    @api.model
+    def create(self, vals):
+        for val in vals:
+            estate_property = self.env["estate.property"].browse(val["property_id"])
+            best_price = estate_property.best_price or 0.0
+            if val["price"] < best_price:
```
ushyme: You should also use `float_compare()` here.

estate_account/__manifest__.py
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "Real Estate Accounting",
+    "description": "Real Estate Accounting (link module)",
+    "author": "robal",
```
ushyme:
```suggestion
    "author": "Odoo S.A.",
```
As an employee, you work in the name of Odoo.

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,21 @@
+from odoo import Command, models, _
+
+
+class EstateProperty(models.Model):
+    _inherit = "estate.property"
+
+    def action_set_sold(self) -> bool:
+        percent_description = _("6 percent of selling price")
+        admin_fee_description = _("Administrative fees")
```
ushyme: You don't need to explicitly call `_()` for field values. If the fields are marked as translatable, the system takes care of the rest.

estate/models/estate_property.py
```diff
@@ -0,0 +1,113 @@
...
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = None
+            self.garden_orientation = None
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_except_new_or_cancelled(self) -> None:
+        if any(record.state not in {"new", "cancelled"} for record in self):
+            raise UserError(_("Cannot delete properties unless they are new or cancelled."))
```
ushyme:
```suggestion
            raise UserError(self.env._("Cannot delete properties unless they are new or cancelled."))
```
It's best to access the translation method defined on the environment object. The older standalone `_()` function has a few issues that the newer `self.env._()` doesn't.

estate/models/estate_property.py
```diff
@@ -0,0 +1,113 @@
...
+        "CHECK(selling_price >= 0)",
+        "A property's selling price must be equal to or greater than 0.",
+    )
+
+    @api.constrains("expected_price", "selling_price")
+    def _restrict_selling_price(self) -> None:
+        for record in self:
+            # Selling price is zero when no offer has been accepted
+            if not float_is_zero(record.selling_price, precision_digits=2) and float_compare(record.selling_price, 0.9 * record.expected_price, precision_digits=2) < 0:
+                raise ValidationError(_("A property's selling price cannot be lower that 90 percent of its expected price."))
```
ushyme:
```suggestion
                raise ValidationError(_("A property's selling price cannot be lower than 90 percent of its expected price."))
```
small typo

estate/models/estate_property.py
```diff
@@ -0,0 +1,113 @@
...
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self) -> None:
+        for record in self:
+            record.best_price = max(record.offer_ids.mapped("price"), default=0)
+
+    # Methods that trigger on changes
+    @api.onchange("garden")
+    def _compute_garden_defaults(self) -> None:
```
ushyme:
```suggestion
    def _onchange_garden_defaults(self) -> None:
```
This is an `onchange`, not a `compute`. Should be named `_onchange_garden_defaults` or `_onchange_garden`.

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,21 @@
...
+        # Note: does not work if fiscal localization is not set
+        for record in self:
+            account_move = {
+                "partner_id": record.buyer_id.id,
+                "move_type": "out_invoice",
+                "invoice_line_ids": [
+                    Command.create({"name": percent_description, "quantity": 1, "price_unit": 0.06 * record.selling_price}),
+                    Command.create({"name": admin_fee_description, "quantity": 1, "price_unit": 100.0}),
+                ],
+                }
```
ushyme:
```suggestion
            }
```

estate/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "Real Estate",
+    "description": "Real Estate Advertising",
+    "author": "robal",
```
ushyme:
```suggestion
    "author": "Odoo S.A.",
```
As an employee, you work in the name of Odoo.

.gitignore
```diff
@@ -127,3 +127,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# VScode
+.vscode/
```
ushyme:
```suggestion
```
This is personal configuration that shouldn't be pushed to the general codebase.

ushyme (commented):
Good work so far. I just have a few suggestions.

estate/models/estate_property.py
robal-odoo: Is there a practical difference between using `record.env._()` and using `_()` imported with `from odoo import _`  or the one from `self.env._()` ?

estate/models/estate_property_offer.py
robal-odoo: Should I use `context_today` like in the other model ?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,72 @@
...
+    def _inverse_deadline(self) -> None:
+        for record in self:
+            ref_date = fields.Date.context_today(self) if not record.create_date else record.create_date.date()
+            record.validity = (record.date_deadline - ref_date).days
+
+    # CRUD overrides
+    @api.model
+    def create(self, vals):
+        for val in vals:
+            estate_property = self.env["estate.property"].browse(val["property_id"])
```
ushyme: If `vals` contains offers for multiple properties, this will trigger multiple singleton browses.
Extract all `property_ids` from `vals` first, browse them in one go, and then iterate.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,72 @@
...
+        # TODO double check validation
+        for record in self:
+            if record.property_id.offer_ids.filtered(lambda r: r.status == "accepted"):
+                raise UserError(record.env._("Cannot accept this offer because another offer has already been accepted for the property."))
+
+            record.status = "accepted"
+            record.property_id.selling_price = record.price
+            record.property_id.buyer_id = record.partner_id
+            record.property_id.state = "offer_accepted"
+        return True
```
ushyme, line 72: what if the property is already sold or cancelled ?

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,21 @@
...
+            account_move = {
+                "partner_id": record.buyer_id.id,
+                "move_type": "out_invoice",
+                "invoice_line_ids": [
+                    Command.create({"name": percent_description, "quantity": 1, "price_unit": 0.06 * record.selling_price}),
+                    Command.create({"name": admin_fee_description, "quantity": 1, "price_unit": 100.0}),
+                ],
+            }
+            self.env["account.move"].create(account_move)
+        return super().action_set_sold()
```
ushyme: Currently, you are calling `self.env["account.move"].create(account_move)` **inside** the loop. This triggers an `INSERT` and potentially multiple compute/inverse calls per record.

Also, what if the `buyer_id` is not set ?

ushyme (commented):
Looks good 👍🏾; Just a few remarks.

PR #1098, opened 2026-01-01
[ADD] real_estate: Create New Module

rugot-odoo:
Implemented the core business logic for the Real Estate module as part of Odoo technical training.
This PR sets up the module structure, base models, and initial views.
The configuration follows the official Odoo tutorial step by step.
It establishes a solid foundation for managing properties and offers.
Further features will be built on top of this base.


mash-odoo:
Hello @rugot-odoo 
For this PR, we only need the changes related to the `real_estate` module.
Could you please remove the extra files (like .idea and other unrelated diffs) and keep the commit limited to the module scope?

rugot-odoo:
Hello @mash-odoo 
I’ve cleaned up the PR by removing all unrelated files (including .idea)

real_estate/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+{
+    'name': "real.estate",
+    'summary': "summary of the real estate",
+    'description': """Rugot first module""",
+    'author': "Ruchita Gothi (Rugot)",
+    'category': 'Uncategorized',
+    'version': '0.1',
+    'depends': ['base'],
+
```
mash-odoo:
```suggestion
```

real_estate/models/real_estate_properties_model.py
```diff
@@ -0,0 +1,36 @@
+from datetime import date, timedelta
+
+from odoo import models, fields
+
+
+class real_estate(models.Model):
+    _name = 'real.estate'
+    _description = 'Real Estate Property'
+
+    name = fields.Char(string='Name', default="Unknown", required=True)
```
mash-odoo: You don't need to explicitly write the string if it is same as the field name.

real_estate/models/real_estate_properties_model.py
```diff
@@ -0,0 +1,36 @@
...
+    date_availability = fields.Datetime(string='Date Availability', default=date.today() + timedelta(days=90))
+    expected_price = fields.Float(string='Expected Price')
+    selling_price = fields.Float(string='Selling Price', default=1000)
+    bedrooms = fields.Integer(string='Bedrooms', default=2)
+    living_area = fields.Integer(string='Living Area')
+    bathrooms = fields.Integer(string='Bathrooms')
+    facades = fields.Integer(string='Facades')
+    garage = fields.Boolean(string='Garage')
+    garden = fields.Boolean(string='Garden')
+    garden_area = fields.Integer(string='Garden Area')
```
mash-odoo: You don't need to explicitly write the string if it is same as the field name

real_estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,4 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+real_estate,real_estate,model_real_estate,base.group_user,1,1,1,1
+real_estate_tag,real_estate_tag,model_real_estate_tag,base.group_user,1,1,1,1
+real_estate_property_offer,real_estate_property_offer,model_real_estate_property_offer,base.group_user,1,1,1,1
```
mash-odoo: Always add a line at the end of file

real_estate/views/properties_views.xml
```diff
@@ -0,0 +1,114 @@
+<odoo>
+    <data>
```
mash-odoo: I noticed the `<data>` tag here, what was the reason for adding it?

real_estate/views/properties_views.xml
```diff
@@ -0,0 +1,114 @@
...
+
+        <record model="ir.ui.view" id="real_estate_list">
+            <field name="name">real_estate_list</field>
```
mash-odoo: 1. Place `id` attribute before `model`
2. For naming a view: `<model_name>_view_<view_type>`, where view_type is `kanban`, `form`, `list`, `search`, …
```suggestion
        <record id="real_estate_view_list" model="ir.ui.view">
            <field name="name">real.estate.view.list</field>
```

real_estate/views/properties_views.xml
```diff
@@ -0,0 +1,114 @@
...
+                    <field name="date_availability"/>
+                    <field name="expected_price"/>
+                    <field name="selling_price"/>
+                    <field name="bedrooms"/>
+                    <field name="garden_orientation"/>
+                    <field name="tag_ids" widget="many2many_tags"/>
+                </list>
+            </field>
+        </record>
+
```
mash-odoo:
```suggestion
```
No need to leave 2 lines

real_estate/views/properties_views.xml
```diff
@@ -0,0 +1,114 @@
...
+
+        <record model="ir.ui.view" id="real_estate_form">
+            <field name="name">real_estate_form</field>
```
mash-odoo: Do the changes of xml naming as suggested above

real_estate/views/properties_views.xml
```diff
@@ -0,0 +1,114 @@
...
+        <record id="real_estate_search_view" model="ir.ui.view">
+            <field name="name">real_estate.search</field>
+            <field name="model">real.estate</field>
+            <field name="arch" type="xml">
+                <search>
+                    <field name="name"/>
+                    <filter
+                            string="Active"
+                            name="active"
+                            domain="[('active', '=', True)]"
```
mash-odoo: Indentation issue

real_estate/views/properties_views.xml
```diff
@@ -0,0 +1,114 @@
...
+                    />
+                    <group>
+                        <filter
+                                string="Orientation"
+                                name="garden_orientation"
+                                context="{'group_by': 'garden_orientation'}"
+                        />
```
mash-odoo: Indentation issue

real_estate/views/real_estate_properties_offer.xml
```diff
@@ -0,0 +1,48 @@
...
+            </field>
+        </record>
+
+        <record model="ir.actions.act_window" id="real_estate_property_offer_action">
+            <field name="name">Property Offer</field>
+            <field name="res_model">real.estate.property.offer</field>
+            <field name="view_mode">list,form</field>
+        </record>
+
+        <menuitem id="rreal_estate_property_offer_menu"
```
mash-odoo:
```suggestion
        <menuitem id="real_estate_property_offer_menu"
```

mash-odoo (changes requested):
Hello @rugot-odoo,
Good Work on the PR.
Please view my comments and apply the changes wherever needed.

real_estate/views/properties_views.xml
rugot-odoo: Here, the <data> tag is not required. However, it is useful when we want to control updates for a specific block of code. By using the <data> tag with the noupdate attribute, we can prevent Odoo from updating that particular data during module upgrades or not

real_estate/views/real_estate_property_views.xml
```diff
@@ -0,0 +1,139 @@
...
+                        class="btn-primary"
+                    />
+                    <button
+                        name="action_cancel"
+                        type="object"
+                        string="Cancel"
+                        class="btn-secondary"
+                    />
+                    <field name="stage" widget="statusbar"
+                           options="{'clickable': False}"
```
mash-odoo, line 58: What if this option is not written?

real_estate/models/real_estate_property_type.py
```diff
@@ -0,0 +1,17 @@
...
+    _description = 'Real Estate Property Type'
+
+    name = fields.Char(required=True)
+    property_ids = fields.One2many(
+        'real.estate',
+        'property_type_id',
+        string='Properties'
+    )
+    _unique_type_name = models.Constraint(
+        'UNIQUE(name)',
```
mash-odoo, line 23: Will this work for all the cases or is there any drawback?

real_estate/views/real_estate_properties_views.xml
```diff
@@ -0,0 +1,139 @@
...
+                <sheet>
+                    <group>
+                        <group>
+                            <field name="name"/>
+                            <field name="property_type_id"/>
+                            <field name="postcode"/>
+                            <field name="date_availability"/>
+                            <field name="living_area"/>
+                            <field name="garden_area"/>
+                            <field name="total_area" readonly="1"/>
```
mash-odoo: Do you think readonly is required here?

real_estate/models/real_estate_properties.py
```diff
@@ -0,0 +1,123 @@
...
+        else:
+            self.garden_area = 0
+            self.garden_orientation = False
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_if_accepted_offer(self):
+        for record in self:
+            for offer in record.offer_ids:
+                if offer.status == 'accepted':
+                    raise UserError("Can't delete an active record!")
```
mash-odoo: Can you optimize this?

real_estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,6 @@
...
+real_estate_property_type,real_estate_property_type,model_real_estate_property_type,base.group_user,1,1,1,1
+,,,,,,,
```
mash-odoo: Unnecessary diff

real_estate/views/real_estate_properties_views.xml
```diff
@@ -0,0 +1,139 @@
...
+                    </group>
+                    <notebook>
+                        <page string="Description">
+                            <group>
+                                <field name="description"/>
+                            </group>
+                        </page>
+                        <page string="Extra Info">
+                            <group>
+                                <group string="Email">
```
mash-odoo: give proper meaningful names, it shouldn't be random.

real_estate/views/real_estate_properties_views.xml
```diff
@@ -0,0 +1,139 @@
...
+                    name="active_record"
+                    domain="[('active', '=', True)]"
+                />
+                <filter
+                    string="Garden orientation"
+                    name="name"
+                    domain="['|','&amp;',('garden_orientation', '=' , 'north'),
+                    ('garden_orientation', '=' , 'south'),
+                    ('name', '=' , 'test')]"
+                />
```
mash-odoo: As `garden_orientation` is a selection field, this `AND` condition won’t ever match.
Should this be an `OR`?

real_estate/models/real_estate_property_offer.py
```diff
@@ -0,0 +1,64 @@
...
+            ('status', '=', 'accepted')
+        ])
+        if accepted_offer:
+            raise UserError(
+                "Only one offer can be accepted for a property."
+            )
+        self.status = 'accepted'
+        self.property_id.selling_price = self.price
+        self.property_id.stage = 'sold'
+        self.property_id.buyer_id = self.partner_id.id
```
mash-odoo: Can try to make this more optimized?

mash-odoo (changes requested):
Hello,
Good going on the PR,
Here are a few suggestions..

real_estate/models/real_estate_property_type.py
rugot-odoo: This will work, but there is one drawback.
The UNIQUE(name) constraint is case-sensitive, so values like "House" and "house" would still be allowed.

real_estate/views/real_estate_property_views.xml
rugot-odoo: The statusbar is read-only by default, so it is not clickable even if the option is omitted.
The clickable: False is added only for clarity, is not strictly required
If we want users to change the state manually from the UI, we can explicitly enable it using clickable: 1.

real_estate/views/real_estate_properties_views.xml
rugot-odoo: No, The attribute was added by mistake.
total_area is a computed field, which is already read-only by default.

real_estate/models/real_estate_properties.py
rugot-odoo: Yes, this can be optimized.
Instead of looping through all offers, we can directly filter the related recordset using filtered_domain.

def _unlink_if_accepted_offer(self):
    if self.offer_ids.filtered_domain([('status', '=', 'accepted')]):
        raise UserError("Can't delete an active record!")

real_estate/models/real_estate_property_offer.py
rugot-odoo: Yes, this can be optimized.
Instead of assigning each field individually, we can update the related property in a single write() call.

self.property_id.write({
    'selling_price': self.price,
    'stage': 'sold',
    'buyer_id': self.partner_id.id,
})

real_estate/views/real_estate_properties_views.xml
rugot-odoo: Got it.
The & was added for testing; this should be an OR condition.
You’re right, since garden_orientation is a selection field, it can’t be both values at the same time.

real_estate/models/real_estate_property.py
```diff
@@ -0,0 +1,150 @@
+from datetime import timedelta
+
+from odoo import models, fields, api
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools import float_is_zero, float_compare
+
+
+class real_estate(models.Model):
```
mash-odoo: Always have the class name in PascalCase, and keep it same as the model name
```suggestion
class RealEstate(models.Model):
```

real_estate/models/real_estate_property_offer.py
```diff
@@ -0,0 +1,149 @@
...
+    #         offer.property_id.stage = 'offer_received'
+    #     return offer
+
+    @api.depends('create_date', 'validity')
+    def _compute_date_deadline(self):
+        for offer in self:
+            if offer.create_date:
+                offer.date_deadline = offer.create_date.date() + timedelta(days=offer.validity)
+            else:
+                offer.date_deadline = fields.Date.today() + timedelta(days=offer.validity)
```
mash-odoo: Will `record.create_date` always be set?
Also, can you optimize this part of code, I don't think you really need the if-else loop!

real_estate/views/real_estate_property_views.xml
```diff
@@ -0,0 +1,167 @@
...
+                           statusbar_visible="new,offer_received,offer_accepted,sold,cancelled"/>
+                </header>
+                <sheet>
+                    <group>
+                        <group>
+                            <field name="name"/>
+                            <field name="property_type_id" options="{'no_create': True, 'no_create_edit': True}"/>
```
mash-odoo: What are these options used for?
What does `no_create_edit` do?

real_estate/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+        'views/real_estate_property_views.xml',
+        'views/real_estate_property_type_views.xml',
+        'views/real_estate_tag_views.xml',
+        'views/real_estate_property_maintenance_request_views.xml',
+        'views/res_users_views.xml',
+        'views/real_estate_menus.xml',
+    ],
+    'application': True,
+    'installable': True,
+    'license': 'LGPL-3',
```
mash-odoo, line 28: What is a license and why did you specifically write `LGPL-3`?
Are there any other options available?

real_estate/views/real_estate_property_views.xml
```diff
@@ -0,0 +1,167 @@
...
+                />
+                <filter
+                        name="In_active_record"
+                        string="IN Active"
+                        domain="[('active', '=', False)]"
+                />
+                <filter
+                        string="Garden orientation"
+                        name="name"
+                        domain="['|',('garden_orientation', '=' , 'north'),
```
mash-odoo, line 189: What if you do not add an operator in the domain?

real_estate/views/real_estate_property_views.xml
```diff
@@ -0,0 +1,167 @@
...
+                </group>
+            </search>
+        </field>
+    </record>
+
+    <record id="real_estate_action_view" model="ir.actions.act_window">
+        <field name="name">Real Estate</field>
+        <field name="res_model">real.estate</field>
+        <field name="view_mode">list,form</field>
+        <field name="context">
```
mash-odoo, line 259: How to give a default group_by?

real_estate/views/real_estate_property_offer_views.xml
```diff
@@ -0,0 +1,65 @@
...
+        <field name="arch" type="xml">
+            <list decoration-danger="status == 'refused'"
+                  decoration-success="status == 'accepted'">
+                <field name="price"/>
+                <field name="partner_id"/>
+                <field name="validity"/>
+                <field name="date_deadline"/>
+                <button
+                        name="action_accept"
+                        type="object"
```
mash-odoo, line 16: When do we use `type="object"`?
Also what are the other types that we can use?

real_estate/models/real_estate_property_offer.py
```diff
@@ -0,0 +1,149 @@
...
+    #         if offer.property_id and offer.property_id.stage == 'new':
+    #             offer.property_id.stage = 'offer_received'
+    #     return offers
+
+    @api.model
+    def create(self, vals):
+        for val in vals:
+            price = val.get('price')
+            property_id = val.get('property_id')
+            property = self.env['real.estate'].browse(property_id)
```
mash-odoo, line 42: when would you use `.browse()` and `.search()`

real_estate/models/real_estate_property.py
```diff
@@ -0,0 +1,150 @@
...
+            self.garden_orientation = 'north'
+        else:
+            self.garden_area = 0
+            self.garden_orientation = False
+
+    @api.ondelete(at_uninstall=False)
+    def _check_property_delete(self):
+        for record in self:
+            if record.stage not in ('new', 'cancelled'):
+                raise UserError(
```
mash-odoo: When do we use UserError and ValidationError?

mash-odoo (changes requested):
Hello!!
Thanks for your work!
Here are some questions and suggestions..

real_estate/models/real_estate_property_offer.py
rugot-odoo: create_date is auto-set by Odoo after the record is created, but it can be False during creation, so the fallback to fields.Date.today() is needed.
Also, the if/else can be removed —  using offer.create_date or fields.Date.today() is cleaner

real_estate/views/real_estate_property_views.xml
rugot-odoo: These options control record creation from the Many2one dropdown
no_create: hides the Create option.
no_create_edit: hides the Create and Edit, option (so users can’t open the form to create a new record from here)

real_estate/__manifest__.py
rugot-odoo: A license defines how others are allowed to use, modify, and redistribute the code.

I chose LGPL-3 because it’s the standard license for Odoo Community modules. It allows the code to be used in both open-source and private projects, while ensuring that improvements to the LGPL-covered code itself remain open.

Yes, there are other options (as listed in the field):
GPL/ AGPL:  stronger copyleft, everything must stay open
LGPL:  more flexible, widely used in the Odoo ecosystem
OEEL / OPL / Proprietary: restricted, not open source

real_estate/views/real_estate_property_views.xml
rugot-odoo: In Odoo domains, if I don’t specify any logical operator, it defaults to an AND operation between conditions.

real_estate/views/real_estate_property_views.xml
rugot-odoo: <field name="context"> {'search_default_group_by_active_record': 1} </field>

real_estate/views/real_estate_property_offer_views.xml
rugot-odoo: We use type="object" when the button needs to call a Python method on the model.
It triggers a server-side function defined on the record

Other  type :
action: triggers an ir.actions(like opening another view, wizard, report, etc.) (client side)
edit: switches the form into edit mode
base64 / file: used with binary fields for file upload/download button

real_estate/models/real_estate_property_offer.py
rugot-odoo: .browse() when already have the record ID and just want a recordset for them. It does not query the database, it simply wraps the IDs in a recordset.

.search() when  don’t know the IDs and need to find records based on a domain. This performs an actual database query.

real_estate/models/real_estate_property.py
rugot-odoo: ValidationError when we want to block saving a record because the data is invalid (usually in @api.constrains) 

UserError when we want to stop an action and show a message to the user during a workflow (e.g. button click)

real_estate/models/real_estate_property.py
```diff
@@ -0,0 +1,155 @@
...
+from odoo import models, fields, api
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools import float_is_zero, float_compare, _
+
+
+class RealEstate(models.Model):
+    _name = 'real.estate'
+    _description = 'Real Estate Property'
+    _order = "id desc"
+    _inherit = ['mail.thread', 'mail.activity.mixin']
```
mash-odoo, line 12: What's the purpose of inheriting `mail.thread` and `mail.activity.mixin`

real_estate/models/res_users.py
```diff
@@ -0,0 +1,27 @@
...
+    )
+    total_unsold = fields.Float(
+        string="Total Unsold Value",
+        compute="_compute_total_unsold_value",
+        store=True
+    )
+
+    @api.depends('property_ids.stage', 'property_ids.expected_price')
+    def _compute_total_unsold_value(self):
+        sum_unsold = dict(self.env['real.estate']._read_group(
```
mash-odoo, line 20: Why is using _read_group here a better choice?

real_estate/views/real_estate_property_views.xml
```diff
@@ -0,0 +1,216 @@
...
+                                <group string="Interior Details">
+                                    <field name="bedrooms"/>
+                                    <field name="bathrooms"/>
+                                    <field name="facades"/>
+                                </group>
+                            </group>
+                        </page>
+                        <page string="Offers">
+                            <group>
+                                <field name="offer_ids" readonly="stage in ('offer_accepted', 'sold', 'cancelled')"/>
```
mash-odoo: Make the list view editable instead of opening a form view. Also i don't want property field to be shown in this view... can you handle that as well?

real_estate/views/real_estate_property_views.xml
```diff
@@ -0,0 +1,216 @@
...
+                        name="name"
+                        domain="['|',('garden_orientation', '=' , 'north'),
+                                ('garden_orientation', '=' , 'south')]"
+                />
+                <group>
+                    <filter
+                            string="Orientation"
+                            name="garden_orientation"
+                            context="{'group_by': 'garden_orientation'}"
+                    />
```
mash-odoo, line 200: Indentation issue

real_estate/models/real_estate_property.py
```diff
@@ -0,0 +1,155 @@
...
+            aggregates=['price:max'],
+        ))
+        for record in self:
+            record.best_price = dataa.get(record, 0.0)
+
+    @api.depends('maintenance_request_ids.cost')
+    def _compute_total_maintenance_cost(self):
+        for record in self:
+            costs = record.maintenance_request_ids.mapped('cost')
+            record.total_maintenance_cost = sum(costs) if costs else 0.0
```
mash-odoo, line 166: Explain why this is the best way to compute the maintenance cost? 
If not, is there any other way?

real_estate/models/real_estate_property.py
```diff
@@ -0,0 +1,155 @@
...
+            self.garden_area = 10
+            self.garden_orientation = 'north'
+        else:
+            self.garden_area = 0
+            self.garden_orientation = False
+
+    @api.ondelete(at_uninstall=False)
+    def _check_property_delete(self):
+        for record in self:
+            if record.stage not in ('new', 'cancelled'):
```
mash-odoo: Try to execute this without using a loop

real_estate/models/real_estate_property.py
```diff
@@ -0,0 +1,155 @@
...
+        ('south', 'South'),
+        ('east', 'East'),
+        ('west', 'West')
+    ])
+    active = fields.Boolean(default=True)
+    tag_ids = fields.Many2many(
+        "real.estate.tag", string="Tags", ondelete='cascade')
+    offer_ids = fields.One2many(
+        "real.estate.property.offer", "property_id", string="Offers", tracking=True)
+    total_area = fields.Float(compute="_compute_total", store=True)
```
mash-odoo, line 48: Can you perform searching/filtering on `total_area` without using `store=True`?

real_estate/models/real_estate_property.py
```diff
@@ -0,0 +1,155 @@
...
+    ], default='new', tracking=True)
+    buyer_id = fields.Many2one(
+        'res.partner',
+        string='Buyer',
+        copy=False, tracking=True)
+    selling_price = fields.Float()
+    maintenance_request_ids = fields.One2many(
+        "real.estate.property.maintenance.request", "property_id", string="Maintenance Requests")
+    total_maintenance_cost = fields.Float(compute="_compute_total_maintenance_cost", store=True, string="Total Cost")
+    salesperson_id = fields.Many2one(
```
mash-odoo, line 72: Can you add the code for getting default value of salesperson?

real_estate/views/real_estate_property_views.xml
```diff
@@ -0,0 +1,216 @@
...
+                                <field name="selling_price"/>
+                            </t>
+                        </div>
+                        <footer>
+                            <field name="activity_ids" widget="kanban_activity"/>
+                            <field name="salesperson_id" widget="many2one_avatar_user" class="ms-auto"/>
+                        </footer>
+                    </t>
+                    <t t-name="menu">
+                        <a t-if="widget.deletable" type="delete" class="dropdown-item">Delete</a>
```
mash-odoo, line 247: What does this `widget.deletable` do?

mash-odoo (changes requested):
Hello, 
Thank you for your work..
Here are some comments..

real_estate/views/real_estate_property_views.xml
rugot-odoo: Implemented the requested changes by overriding the list view

real_estate/models/real_estate_property.py
rugot-odoo: We inherit from mail.thread and mail.activity.mixin to enable the chatter and activity features on the model.
mail.thread adds message tracking, followers, and logging in the chatter.
mail.activity.mixin enables scheduling and managing activities

real_estate/models/res_users.py
rugot-odoo: _read_group performs the aggregation directly in SQL instead of Python loops
This is faster, more scalable, and avoids unnecessary queries

real_estate/models/real_estate_property.py
rugot-odoo: no, This is not the best approach for large datasets because it loops in Python and uses mapped, which increases time complexity as records grow
A better and more scalable solution is to use _read_group, which performs the aggregation at the SQL level and avoids Python-side loops.

real_estate/models/real_estate_property.py
rugot-odoo: Yes, it’s possible without store=True by defining a custom search method on the computed field

real_estate/models/real_estate_property.py
rugot-odoo: default=lambda self: self.env.user

real_estate/views/real_estate_property_views.xml
rugot-odoo: widget.deletable is a flag provided by the Odoo JS widget that indicates whether the current record can be deleted, It’s typically set based on user permissions and record state

awesome_owl/controllers/controllers.py
```diff
@@ -2,9 +2,6 @@
 from odoo.http import request, route
 
 class OwlPlayground(http.Controller):
-    @http.route(['/awesome_owl'], type='http', auth='public')
+    @http.route(['/awesome_owl'], type='http', auth='public', website=True)
```
mash-odoo: What is `website=True` used for?

awesome_owl/static/src/card/card.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates xml:space="preserve">
```
mash-odoo: what is `xml:space="preserve"` used for?

awesome_owl/static/src/playground.xml
```diff
@@ -1,10 +1,23 @@
...
+    <div class="o_playground p-4">
 
-    <t t-name="awesome_owl.playground">
-        <div class="p-3">
-            hello world
-        </div>
-    </t>
+      <div class="d-flex gap-3 flex-wrap">
+        <Card title="'First Card'">
+          <Counter onChange.bind="onCounter1Change"/>
```
mash-odoo: when do we use `.bind`?

PR #1103, opened 2026-01-01
[ADD] Estate Property Module

kukor-odoo:
- Implemented the complete Estate Property tutorial by following the official Odoo documentation from start to finish.
- Designed and implemented custom ORM models, relational fields, computed fields, and constraints to enforce                                 business rules.
- Created form, list, kanban and search views, menus, and actions to provide a usable and structured UI.
- Configured security access rights and record rules to control data visibility and operations.


estate/models/estate_property.py
```diff
@@ -0,0 +1,53 @@
+from datetime import date
+from dateutil.relativedelta import relativedelta
+from odoo import fields, models
```
mash-odoo: For ordering of imports, refer this: https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#imports

estate/models/estate_property.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class EstateProperty(models.Model):
+    _name = "estate_model"
```
mash-odoo: Please follow the proper guidelines for naming

estate/models/estate_property.py
```diff
@@ -0,0 +1,53 @@
...
+    _description = "This is to say that this is the description of the Estate Model"
+
+    name = fields.Char("Name", default="Unknown", required=True)
```
mash-odoo: You don't need to explicitly write the string if it is same as the field name.

estate/models/estate_property.py
```diff
@@ -0,0 +1,53 @@
...
+        ],
+        help="This is the direction of the proprety, which side the preperty is facing.",
+    )
+    state = fields.Selection(
+        default="new",
+        string="State",
+        selection=[
+            ("new", "New"),
+            ("offer received", "Offer Received"),
+            ("offer accepted", "Offer Accepted"),
```
```suggestion
            ("offer_received", "Offer Received"),
            ("offer_accepted", "Offer Accepted"),
```
Use underscores while defining keys for selection field

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,18 @@
+from odoo import fields, models
+
+
+class EstatePropertyOffer(models.Model):
+    _name = "estate_offer"
```
mash-odoo: Follow the proper naming format, also try to keep it same as the class name

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_model,access_estate_model,model_estate_model,base.group_user,1,1,1,1
+access_estate_type,access_estate_type,model_estate_type,base.group_user,1,1,1,1
+access_estate_tags,access_estate_tags,model_estate_tags,base.group_user,1,1,1,1
+access_estate_offer,access_estate_offer,model_estate_offer,base.group_user,1,1,1,1
```
mash-odoo: Always add a line at the end of file

estate/views/estate_property_offers.xml
```diff
@@ -0,0 +1,37 @@
+<odoo>
+    <!-- listview -->
+    <record id="estate_offer_list" model="ir.ui.view">
+        <field name="name">estate_offer.list</field>
```
mash-odoo: For naming a view: `<model_name>_view_<view_type>`, where view_type is `kanban`, `form`, `list`, `search`, …

estate/views/estate_property_offers.xml
```diff
@@ -0,0 +1,37 @@
...
+                <field name="status" />
+            </list>
+        </field>
+    </record>
+
+    <record id="estate_offer_view_form" model="ir.ui.view">
+        <field name="name">estate_offer.view.form</field>
+        <field name="model">estate_offer</field>
+        <field name="arch" type="xml">
+            <form string="">
```
mash-odoo: Please explain me the reason for adding `string=""`

estate/views/estate_property_offers.xml
```diff
@@ -0,0 +1,37 @@
...
+                <sheet>
+                    <group>
+                        <group>
+                            <field name="price" />
```
mash-odoo:
```suggestion
                            <field name="price"/>
```

estate/views/estate_property_offers.xml
```diff
@@ -0,0 +1,37 @@
...
+                        </group>
+                        <group>
+                            <field name="partner_id" />
```
mash-odoo:
```suggestion
                            <field name="partner_id"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,103 @@
+<odoo>
+    <record id="estate_model_action" model="ir.actions.act_window">
+        <field name="name">Property</field>
+        <field name="res_model">estate_model</field>
+    </record>
+
+    <!-- listview -->
+    <record id="estate_model_list" model="ir.ui.view">
+        <field name="name">estate_model.list</field>
```
mash-odoo: Please apply changes as suggested above about naming the view properly

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,103 @@
...
+        </field>
+    </record>
+
+    <!-- formview -->
+    <record id="estate_model_form" model="ir.ui.view">
+        <field name="name">estate_model.form</field>
+        <field name="model">estate_model</field>
+        <field name="arch" type="xml">
+            <form>
+
```
```suggestion
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,103 @@
...
+                                <group>
+                                    <field name="garage" />
+                                    <field name="garden" />
+                                    <field name="garden_orientation" />
+                                </group>
+                            </group>
+                        </page>
+                        <page string="Offers">
+                            <group>
+                                <field name="offers_id" string=""/>
```
mash-odoo: > string=""

Is this really needed?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,103 @@
...
+                            </group>
+                        </page>
+                        <page string="More Info">
+                            <group>
+                                <field name="partner_id" string="Salesperson"/>
+                                <field name="buyer_id" string="Buyer"/>
+                            </group>
+
```
mash-odoo: Do not leave unnecessary lines

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,103 @@
...
+        <field name="arch" type="xml">
+            <search>
+                <field name="name" />
+                <field name="postcode" />
+                <field name="last_seen" />
+                <field name="selling_price" />
+                <field name="last_seen" />
+                <field name="garden" />
+                <field name="garage" />
+                <filter string="State" name="state" domain="['|', ('state', '=', 'new') ,('state', '=', 'offer received')]" />
```
```suggestion
                <filter string="State" name="state" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer received')]" />
```
Also you can give better name for the filter for example “Available / Active” or “New & Offer Received”

mash-odoo (changes requested):
Hello @kukor-odoo,
Nice start for the module.
Please apply the suggested changes wherever necessary.

estate/views/estate_property_views.xml
kukor-odoo: Actually ma’am, I wanted to hide the “Offer” label. Even without defining the string attribute in estate_property_views.xml or in estate_property.py, the label was still appearing.

To verify this behavior, I initially passed an empty string, and the label stopped showing.

Later, I used nolabel="1", which allowed me to hide the label properly without having to pass an empty string.

estate/views/estate_property_views.xml
kukor-odoo: Applied, ma’am. Every issue related to styling and naming has been implemented in my code.

estate/views/estate_property_views.xml
kukor-odoo: Yes, ma’am. I was considering improving the filter name to something more descriptive, like “New & Offer Received”, so it clearly represents the filtered states.

estate/views/estate_property_offers.xml
```diff
@@ -0,0 +1,46 @@
...
+                        </group>
+                        <group>
+                            <field name="validity"/>
+                        </group>
+                        <group>
+                            <field name="date_deadline"/>
+                        </group>
+                        <group>
+                            <field name="status"/>
+                        </group>
```
mash-odoo, line 41: Any specific reason for wrapping fields in different groups?

estate/models/estate_property.py
```diff
@@ -0,0 +1,73 @@
...
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    last_seen = fields.Datetime(default=fields.Datetime.now)
+    garden_area = fields.Integer("Garden Area (sqm)")
+    active = fields.Boolean(default=True)
+    partner_id = fields.Many2one('res.partner', string="Salesperson")
+    buyer_id = fields.Many2one('res.users')
+    property_type_id = fields.Many2one('estate.property.type')
+    property_tag_id = fields.Many2many('estate.property.tags')
+    offers_id = fields.One2many('estate.property.offer', 'property_id')
```
mash-odoo: Please follow the naming conventions of these fields.
For example Many2Many should always have `_ids` suffix.
You can refer from here: https://www.odoo.com/documentation/19.0/developer/tutorials/server_framework_101/07_relations.html?

estate/models/estate_property.py
```diff
@@ -0,0 +1,73 @@
...
+            ('cancelled', "Cancelled"),
+        ],
+        help="This field tells us the state of the property."
+    )
+    total_area = fields.Integer(compute='_compute_total_area')
+    best_price = fields.Integer(compute='_compute_best_price')
+
+    @api.depends('living_area', 'garden_area')
+    def _compute_total_area(self):
+        self.total_area = (self.living_area or 0) + (self.garden_area or 0)
```
mash-odoo, line 87: Addin  `or 0` here is a good practice. So can you tell me the cases where writing this is important?

estate/models/estate_property.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @api.depends('offers_id')
```
mash-odoo: Since best_price depends only on the offer price, is there a reason we want it to recompute on changes to other `offers_id` fields as well?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,107 @@
+<odoo>
+    <record id="estate_property_action" model="ir.actions.act_window">
+        <field name="name">Properties</field>
```
mash-odoo: For naming a view: `<model_name>_view_<view_type>`, where view_type is `kanban`, `form`, `list`, `search`, …

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,107 @@
...
+                <field name="garage"/>
+                <filter string="Garage" name="garage" domain="[('garage', '=', True)]"/>
+                <filter string="Archived" name="active" domain="[('active', '=', False)]"/>
+                <group>
+                    <filter string="Postcode" name="postcode" context="{'group_by': 'postcode'}"/>
+                </group>
+            </search>
+        </field>
+    </record>
+</odoo>
```
mash-odoo, line 195: Add a line at the end of file

mash-odoo (changes requested):
Hello,
I am getting [this](https://app.screencastify.com/watch/fhckFVBmzshYuMl0J4uK?checkOrg=f6824e66-4e21-4104-b0a4-df127d8351de) traceback, can you please resolve it?
Also, please update your PR title and description.

estate/views/estate_property_offers.xml
kukor-odoo: You are right. Wrapping each field individually is unnecessary - a single <group> is sufficient here.
Just out of curiosity, want to show all fields in the same line

estate/models/estate_property.py
kukor-odoo: Without `or 0` : Works in most standard Odoo ORM scenarios, but can be fragile with external inputs
With `or 0` `: Explicit, defensive, and safe across all data sources

It's a small safeguard that costs nothing but prevents potential issues down the line.

estate/models/estate_property.py
kukor-odoo: There's no reason to recompute when other `offers_id` fields change. By using `offers_id.price`, we ensure recomputation only happens when the price field changes, not when unrelated fields like status or partner are modified.

estate/models/estate_property.py
```diff
@@ -0,0 +1,109 @@
...
+
+from odoo import api, fields, models
+from odoo.exceptions import UserError
+
+
+class EstateProperty(models.Model):
+    _name = 'estate.property'
+    _description = "Estate Property"
+
+    _order = "id desc"
```
```suggestion
    _order = "id desc"

```

estate/models/estate_property.py
```diff
@@ -0,0 +1,109 @@
...
+    state = fields.Selection(
+        default='new',
+        selection=[
+            ('new', "New"),
+            ('offer_received', "Offer Received"),
+            ('offer_accepted', "Offer Accepted"),
+            ('sold', "Sold"),
+            ('cancelled', "Cancelled"),
+        ],
+
```
```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,109 @@
...
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    last_seen = fields.Datetime(default=fields.Datetime.now)
+    garden_area = fields.Integer("Garden Area (sqm)")
+    active = fields.Boolean(default=True)
+    partner_id = fields.Many2one('res.users', string="Salesperson")
+    buyer_id = fields.Many2one('res.partner')
+    property_type_id = fields.Many2one('estate.property.type')
+    property_tag_ids = fields.Many2many('estate.property.tags')
+    offers_id = fields.One2many('estate.property.offer', 'property_id')
```
```suggestion
    offer_ids = fields.One2many('estate.property.offer', 'property_id')
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,84 @@
...
+
+    price = fields.Float()
+    status = fields.Selection(
+        selection=[('accepted', "Accepted"), ('refused', "Refused")], copy=False
+    )
+    partner_id = fields.Many2one('res.partner', required=True)
+    property_id = fields.Many2one('estate.property', required=True)
+    validity = fields.Integer(default=7)
+    date_deadline = fields.Date(
+        "Deadline", compute='_compute_deadline', inverse='_inverse_deadline'
```
```suggestion
        "Deadline", compute='_compute_date_deadline', inverse='_inverse_date_deadline'
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+    _offer_price_check = models.Constraint(
+        'CHECK(price >= 0)', "Offer price should be strictly positive"
+    )
+    property_type_id = fields.Many2one(
+        related='property_id.property_type_id', store=True)
+    check_button = fields.Char(store=False)
```
mash-odoo: Is there a need to add `store=False` here, also why is it a Char field?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,84 @@
...
+    def action_accept(self):
+        for offer in self:
+            if offer.property_id.buyer_id:
+                raise UserError(
+                    "An offer has already been accepted for this property.")
+
+            offer.status = 'accepted'
+            offer.property_id.buyer_id = offer.partner_id
+            offer.property_id.selling_price = offer.price
+            offer.property_id.state = 'offer_accepted'
```
mash-odoo: Can you optimize this and write it in any other way?

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,27 @@
...
+
+class EstatePropertyType(models.Model):
+    _name = 'estate.property.type'
+    _description = "Estate Property Type"
+    _order = "name"
+
+    sequence = fields.Integer('Sequence', default=1,
+                              help="Used to order Property Type")
+    name = fields.Char("Property Type", required=True)
+    ofr_ids = fields.One2many('estate.property.offer', 'property_type_id')
```
mash-odoo: Please give some meaningful name

estate/views/estate_property_offers.xml
```diff
@@ -0,0 +1,47 @@
...
+        <field name="res_model">estate.property.offer</field>
+        <field name="domain">[('property_type_id', '=', active_id)]</field>
+    </record>
+
+    <!-- listview -->
+    <record id="estate_property_offer_view_list" model="ir.ui.view">
+        <field name="name">estate.property.offer.view.list</field>
+        <field name="model">estate.property.offer</field>
+        <field name="arch" type="xml">
+            <list string="Channel" editable="bottom" decoration-success="status == 'accepted'"
```
mash-odoo: Can we use decorations in form view?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,119 @@
...
+                    </h1>
+                    <group>
+                        <field name="property_tag_ids" widget="many2many_tags" string="Property Tag"
+                        options="{'color_field' : 'color'}"/>
+                    </group>
+                    <separator/>
+                    <group>
+                        <group>
+                            <field name="state" string="Status"/>
+                            <field name="property_type_id" options="{'no_create': True, 'no_edit': True}"/>
```
mash-odoo: What's the purpose of adding these options?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,119 @@
...
+            <search>
+                <field name="name"/>
+                <field name="postcode"/>
+                <field name="last_seen"/>
+                <field name="selling_price"/>
+                <field name="last_seen"/>
+                <field name="garden"/>
+                <field name="garage"/>
+                <field name="living_area" string="Living Area" filter_domain="[('living_area', '>=', self)]"/>
+                <filter string="Archived" name="unactive" domain="[('active', '=', False)]"/>
```
mash-odoo: When property is accepted -> status is sold -> make the property archived -> make a filter to view archived properties

estate/views/estate_type_views.xml
```diff
@@ -0,0 +1,63 @@
...
+        <field name="res_model">estate.property.type</field>
+    </record>
+
+    <!-- listview -->
+    <record id="estate_property_type_view_list" model="ir.ui.view">
+        <field name="name">estate.property.type.view.list</field>
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <list string="Channel" editable="bottom">
+                <field name="sequence" widget="handle"/>
```
mash-odoo, line 13: What does this handle widget do?

mash-odoo (changes requested):
Hello,
Good going on this task..
Please go through my questions and suggestions

estate/models/estate_property_offer.py
kukor-odoo: I used `store=False` because I didn’t want the field to be stored in the database—it was only needed temporarily for some logic that would change later. Initially, the field was a Boolean, but I later changed it to a Char field.

The reason `check_button` is a Char field is that I needed to pass a string value from it to the XML file. I switched from Boolean to Char so I could pass a string instead and observe the change, but I didn’t notice any difference.

estate/views/estate_property_offers.xml
kukor-odoo: Yes, we can use a decorator in the form view, for example:

`<field name="facades" decoration-success="garden == True"/>`

When the garden field is checked, the value of facades will be displayed in green.

estate/views/estate_property_views.xml
kukor-odoo: This option allows the user to select only from the already created property types. Normally, Odoo provides the ability to create or edit a value directly from the selection field. However, option `'no_create': True` prevents the user from creating new records, and option `'no_edit': True` removes the option to edit existing records while selecting a property type in the form view.

image


image

As shown in the screenshots above, only the two existing property types are available for selection. The user must choose one of these and cannot create a new type or modify an existing one from the dropdown.

estate/views/estate_property_views.xml
kukor-odoo: Here is what needs to be changed:

```
property = offer.property_id
property.state = 'sold'
property.active = False
```
 
 This updates the property’s state from offer_accepted to sold and sets the active field to False.

 To handle archived records, we also need to add a filter with a domain in the search view:
 
 ` <filter string="Archived" name="a" domain="[('active', '=', False)]"/> `

estate/views/estate_type_views.xml
kukor-odoo: The widget="handle" is used to enable drag-and-drop reordering in a list view.
Here, 

`<field name="sequence" widget="handle"/>`

It will use the sequence field to control the order of these records, and allow the user to change that order by dragging rows up and down.

estate/models/estate_property_offer.py
kukor-odoo: Yes this can be the alternative of the about code and the optimal way to write it:
```
def action_accept(self):
        if self.property_id.state == 'offer_accepted':
            raise UserError(
                "An offer has already been accepted for this property."
            )
        self.write({'status': 'accepted'})
        for offer in self:
            offer.property_id.write({
                'selling_price': offer.price,
                'state': 'sold',
                'active': False
            })
```

estate/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+from . import estate_property
+from . import estate_property_investor
+from . import estate_property_maintenance
+from . import estate_property_offer
+from . import estate_property_tags
+from . import estate_property_type
+from . import ResUsers
```
mash-odoo:
```suggestion
from . import res_users
```
Follow the naming convention

estate_account/views/estate_account.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    
+
+</odoo>
```
mash-odoo: What's the need of this file?

estate/views/res_users_view.xml
```diff
@@ -0,0 +1,17 @@
+<odoo>
+    <record id="res_user_view_form" model="ir.ui.view">
+        <field name="name">res.users.view.form.inherit.estate.property</field>
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" position="inside">
```
mash-odoo:
```suggestion
            <notebook position="inside">
```
You can directly give paths like this

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,183 @@
...
+                    <t t-name="card">
+                        <strong>
+                            <field name="name"/>
+                        </strong>
+                        <field name="property_tag_ids" widget="many2many_tags" options="{'color_field' : 'color'}"/>
+                        <div>
+                            Expected Price:
+                            <field name="expected_price"/>
+                        </div>
+                        <t t-if="record.property_type_id.raw_value">
```
mash-odoo, line 155: Why do we need to write here raw_value?

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+            ('new', "New"),
+            ('offer_received', "Offer Received"),
+            ('offer_accepted', "Offer Accepted"),
+            ('sold', "Sold"),
+            ('cancelled', "Cancelled"),
+        ],
+        required=True,
+        help="This field tells us the state of the property.",
+    )
+    total_area = fields.Integer(compute='_compute_total_area')
```
mash-odoo: Can we perform search on the `total_area` without using `store=True`?

estate/models/estate_property_maintenance.py
```diff
@@ -0,0 +1,29 @@
...
+        ]
+    )
+
+    @api.constrains('cost', 'status')
+    def _check_cost(self):
+        for record in self:
+            if record.status == 'approved' and record.cost <= 0:
+                raise UserError(
+                    'Approved cost must be greater then 0'
+                )
```
mash-odoo: Can you optimize this and try without using for loop?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,183 @@
...
+        </field>
+    </record>
+
+    <!-- formview -->
+    <record id="estate_property_view_form" model="ir.ui.view">
+        <field name="name">estate.property.view.form</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form>
+                <header>
```
mash-odoo, line 35: Can you add one more button in the header for accepting the best offer directly by a click?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,183 @@
...
+                            </div>
+                        </t>
+                        <footer class="pt-1 d-flex align-items-center">
+                            <div class="d-flex">
+                                <h4>
+                                    <field name="priority" widget="priority"/>
+                                </h4>
+                            </div>
+                            <div class="ms-auto d-flex">
+                                <field name="partner_id" widget="many2one_avatar_user" class="ms-2"/>
```
mash-odoo, line 187: What is this widget used for?

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,28 @@
+from odoo import models
+from odoo.orm.commands import Command
+
+
+class EstateProperty(models.Model):
+    _inherit = 'estate.property'
+
+    def action_sold(self):
+        res = super().action_sold()
```
mash-odoo, line 9: What is the use of calling super()?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,183 @@
...
+    </record>
+
+    <!-- kanban view -->
+    <record id="estate_property_view_kanban" model="ir.ui.view">
+        <field name="name">estate.property.view.kanban</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <kanban default_group_by="property_type_id" highlight_color="color" options="{'color_field' : 'color'}">
+                <field name="state"/>
+                <progressbar field="state" colors='{"offer_accepted": "success", "offer_received": "warning", "cancelled": "danger"}' sum_field="expected_price"/>
```
mash-odoo, line 141: Do you know where are all these attributes of xml file defined?

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,27 @@
+from odoo import api, fields, models
+
+
+class EstatePropertyType(models.Model):
```
mash-odoo, line 4: Why did you use here `models.Model`?
What are the other types of models that you can use and what are their specifications?

mash-odoo (changes requested):
Hello!
Good going on the task..
Please look at the comments..

estate_account/views/estate_account.xml
kukor-odoo: There is no longer any need for this file in the current state of the Estate module. However, it was required earlier when I was creating and displaying fields in the form view. At that time, I was working with data from the `estate.property` table and needed to show those records, so I inherited the fields from that module and used this file for that purpose.

After successfully completing the task, the file was no longer needed, but I forgot to remove it.

estate/views/res_users_view.xml
kukor-odoo: Got it, thanks for pointing that out. I’ve removed the `<xpath>`  and directly used the  `<notebook>` to add the content inside the existing notebook.

```
<notebook position="inside">
                <page string="Properties">
                        <group>
                            <field name="property_ids"/>
                            <field name="unsold_cost"/>
                        </group>
                </page>
</notebook>
```

estate/views/estate_property_views.xml
kukor-odoo: We use raw_value because `record.property_type_id` is always an object in the UI, even when the field is empty, so `t-if="record.property_type_id"` would always be true.

raw_value holds the actual database value (for a many2one, the ID or false).
So `t-if="record.property_type_id.raw_value"` correctly checks whether the field is really set.

With `raw_value`
image


Without `raw_value`
image

Here, both the property_type are empty, so if the value will be empty then also the condition will be satisfy which should not be valid, so we use raw_value.

estate/models/estate_property_maintenance.py
kukor-odoo: Yes, with the help of the filter we can use the same logic.
```
@api.constrains('cost', 'status')
    def _check_cost(self):
        invalid_records = self.filtered(
            lambda r: r.status == 'approved' and r.cost <= 0
        )
        if invalid_records:
            raise UserError('Approved cost must be greater than 0')
```

estate/views/estate_property_views.xml
kukor-odoo: For accepting the best offer directly here is the xml button and python logic.

XML Button
`<button name="action_offer_accept" type="object" string="Accept Offer" invisible="state in ['sold', 'cancelled']"/>`

Python Logic 
```
 def action_offer_accept(self):
        result = self.env['estate.property.offer'].search(
            [('property_id', '=', self.ids),
             ('price', '=', self.best_price)]
        )
        result.action_accept()
```

Here `action_accept()` and `action_offer_accept()` hold the same logic so it would be better to reuse the code 
so here is the `action_accept()`

```
def action_accept(self):
        if self.property_id.state == 'offer_accepted':
            raise UserError(
                _("An offer has already been accepted for this property.")
            )

        other_offer = self.property_id.offer_ids - self
        other_offer.write({'status': 'refused'})
        self.write({'status': 'accepted'})
        for offer in self:
            offer.property_id.write({
                'selling_price': offer.price,
                'state': 'sold',
                'active': False
            })
```

estate/views/estate_property_views.xml
kukor-odoo: `many2one_avatar_user` is used to display a many2one user field with the user’s avatar, making it easy to visually identify and select users (typically for res.users fields).

image

estate/views/estate_property_views.xml
kukor-odoo: These attributes are defined in ir_ui_view.py, where Odoo specifies the allowed XML tags and attributes for views.

estate/models/estate_property_type.py
kukor-odoo: `models.Model` is the standard Odoo model class for creating regular database-backed models. It is used when you want to store records in the database, with full support for fields, ORM methods, and business logic.

Other model types include:

`models.TransientModel`
Used for temporary models (like wizards). Records are automatically cleared from the database periodically. If you need some data to be saved, you can use ir.config_parameter to store it.

`models.AbstractModel`
An abstract base model, not stored in the database. Used to share methods or fields across multiple models without creating a table.

estate/models/estate_property.py
kukor-odoo: Yes, to make a computed field with store=False searchable, we need to define the `search` attribute on the field, for example on total_area.

This can be implemented in two ways.

Python Level Filtering
```
def _search_total_area(self, operator, value):
    if operator not in ('=', '!=', '>', '>=', '<', '<='):
        raise UserError(_("Operator %s is not supported") % operator)

    records = self.search([])

    def match(record):
        if operator == '>':
            return record.total_area > value
        if operator == '<':
            return record.total_area < value

    matched_records = records.filtered(match)
    return [('id', 'in', matched_records.ids)]
```

Using SQL
```
def _search_total_area(self, operator, value):
    if operator not in ('=', '!=', '>', '>=', '<', '<='):
        raise UserError(_("Operator %s is not supported") % operator)

    query = f"""
        SELECT id
        FROM estate_property
        WHERE (COALESCE(living_area, 0) + COALESCE(garden_area, 0)) {operator} %s
    """

    self.env.cr.execute(query, (value,))
    ids = [row[0] for row in self.env.cr.fetchall()]

    return [('id', 'in', ids)]
```

Here `COALESCE` is a SQL function.
It returns the first non-NULL value from the list you give it.

estate_account/models/estate_property.py
kukor-odoo: Calling `super()` is necessary because we want the original method to be executed before our overridden method.

The reason is that `action_sold()` of estate.property contains the logic and validations, which must be executed first.

```
def action_sold(self):
    if self.state == 'cancelled':
        raise UserError("The property is cancelled!")
    if not self.offer_ids:
        raise UserError("You cannot sell a property without any offers!")
    else:
        for record in self.maintenance_id:
            if record.status != 'done':
                raise UserError("Property is not under maintenance!")
        self.state = 'sold'

```

All of this logic should be executed first, and only then should the `action_sold()` method of estate_account be executed.

If we do not call `super()`, none of this logic will run, which would break the expected behavior of the system.

PR #1157, opened 2026-02-10
venpr: estate: completed chapters 1–14 of Real Estate tutorial  and owl 1st chapter

venpr-odoo:
Chapters 1–7: Estate module implementation

This PR completes Chapters 1 to 7 of the Odoo tutorial for the estate module.

Covered topics:
- Module scaffolding and manifest configuration
- Basic models and fields
- Menu items, actions, and views
- Custom list, form, and search views
- Property types and property tags
- Relationships between models (many2one, one2many, many2many)
- Property offers linked to properties and integrated in property views

This PR brings the estate module to a stable state up to Chapter 7.


estate/models/estate_property.py
```diff
@@ -0,0 +1,64 @@
+from datetime import timedelta
+from odoo import fields, models
```
mash-odoo: For ordering of imports, refer this: https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#imports

estate/models/estate_property.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate Property"
```
mash-odoo, line 10:
```suggestion
    _description = "Real Estate Property"

```

estate/models/estate_property.py
```diff
@@ -0,0 +1,64 @@
...
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        [
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
```
```suggestion
            ('north', "North"),
            ('south', "South"),
            ('east', "East"),
            ('west', "West"),
```
Try keeping key in lower case

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,26 @@
+from odoo import models, fields
+
+
+class EstatePropertyOffer(models.Model):
+    _name = "estate.property.offer"
+    _description = "Estate Property Offer"
+
+    price = fields.Float(string="Offer Price")
+
```
```suggestion
```

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+    <menuitem
+        id="estate_menu_root"
+        name="Real Estate"
+        sequence="10"
```
mash-odoo: What is this sequence used for?

estate/security/estate_security.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+    <!-- Estate User group -->
+    <record id="group_estate_user" model="res.groups">
+        <field name="name">Estate User</field>
+    </record>
```
mash-odoo: What's the purpose of making groups?

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,33 @@
+<odoo>
+
```
mash-odoo:
```suggestion
```

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,33 @@
...
+        <field name="model">estate.property.offer</field>
+        <field name="arch" type="xml">
+            <list>
+                <field name="price"/>
+                <field name="partner_id"/>
+                <field name="status"/>
+            </list>
+        </field>
+    </record>
+
```
mash-odoo:
```suggestion
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,137 @@
...
+<odoo>
+
+    
+    <record id="estate_property_action" model="ir.actions.act_window">
+        <field name="name">Properties</field>
+        <field name="res_model">estate.property</field>
+        <field name="view_mode">list,form</field>
+    </record>
+
+
```
```suggestion
```
Do not leave unnecessary lines in the file. Go through all the files and update the changes wherever necessary.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,137 @@
...
+                            </group>
+                        </page>
+                    </notebook>
+
+                
+
+                    
+
+
+
```
```suggestion
```
Do not leave unnecessary lines in the file. Go through all the files and update the changes wherever necessary.

mash-odoo (changes requested):
Hello!
Thank you for your work!
Here are some suggestions, mostly linter as of now!
Also, please update your PR title
Furthermore, please update description in your commits while pushing, regarding the chapters and topics covered.

estate/views/estate_menus.xml
venpr-odoo: At first, I did not see the Real Estate app in the Odoo dashboard.
So I added sequence="10" to make sure the menu appears in proper order.

Later I understood that the real issue was not sequence —
it was because the Admin user did not belong to the Estate group.

So the problem was related to group access, not sequence ordering.

The sequence is only used to control menu position, not visibility.

estate/security/estate_security.xml
venpr-odoo: Initially, I used `base.group_user` for access control.

Later, I created custom groups named `group_estate_user` and `group_estate_manager`. I added a security XML file and linked the access rights to these new groups.

After making these changes, I could not see the Real Estate app when logged in as Admin.

Then I realized that the Admin user was not assigned to the new Estate groups. Because of that, the menu was hidden due to missing group permissions.

So the issue was not related to the menu configuration or the sequence. It was purely because the Admin user did not belong to the Estate group.

When I switched back to using `base.group_user`, it worked because the Admin user already belongs to that group. That is why the app became visible again.

estate/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+from . import estate_property
+from . import estate_property_type
+from . import estate_property_tag
+from . import estate_property_offer
```
mash-odoo, line 4: Order this alphabetically

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
+from datetime import timedelta
+
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+
+class EstateProperty(models.Model):
```
mash-odoo, line 8: What are the different types of models?

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+    _name = "estate.property"
+    _description = "Real Estate Property"
+    _order = "id desc"
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
```
mash-odoo, line 16: Why do we need `Char` field type for postcode?

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+    expected_price = fields.Float(required=True, default=0.0)
+    selling_price = fields.Float(readonly=True, copy=False)
+    best_price = fields.Float(compute="_compute_best_price")
+    bedrooms = fields.Integer(default=2, copy=False)
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    total_area = fields.Float(compute="_compute_total_area", store=True)
```
mash-odoo, line 30: When do you need to use `store=True`?

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+            ("west", "West"),
+        ]
+    )
+    active = fields.Boolean(default=True)
+    state = fields.Selection([
+        ('new', 'New'),
+        ('offer_received', 'Offer Received'),
+        ('offer_accepted', 'Offer Accepted'),
+        ('sold', 'Sold'),
+        ('cancelled', 'Cancelled'),
```
```suggestion
        ('new', "New"),
        ('offer_received', "Offer Received"),
        ('offer_accepted', "Offer Accepted"),
        ('sold', "Sold"),
        ('cancelled', "Cancelled"),
```
The strings which are to be displayed in the UI should be in double quotes

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+    buyer_id = fields.Many2one(
+        "res.partner",
+        string="Buyer",
+        copy=False,
+        readonly=True,
+    )
+    user_id = fields.Many2one(
+        "res.users",
+        string="Salesperson",
+        default=lambda self: self.env.user,
```
mash-odoo, line 62: Why do we need a `lambda` function here?

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+    tag_ids = fields.Many2many("estate.property.tag", string="Tags")
+    offer_ids = fields.One2many(
+        "estate.property.offer",
+        "property_id",
+        string="Offers",
+    )
+    _expected_price_check = models.Constraint(
```
mash-odoo, line 81: What's the difference between SQL constraints and Python constraints?

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for record in self:
+            prices = record.offer_ids.mapped("price")
```
mash-odoo, line 116: What's the use of `mapped` function?

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = False
+
+    @api.constrains('selling_price', 'expected_price')
```
mash-odoo, line 142: What are decorators?

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+            if record.state == "sold":
+                raise UserError("Sold property cannot be cancelled.")
+            record.state = "cancelled"
+
+    def action_sold(self):
+        for record in self:
+            if record.state == 'cancelled':
+                raise UserError("Cancelled property cannot be sold.")
+            accepted_offer = record.offer_ids.filtered(
+                lambda o: o.status == 'accepted'
```
mash-odoo: You need to be more explicit than just `'o'`, it doesn't help while reading the code

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,113 @@
...
+        for record in self:
+            create_date = record.create_date.date() if record.create_date else fields.Date.today()
+            record.date_deadline = create_date + timedelta(days=record.validity)
+
+    def _inverse_date_deadline(self):
+        for record in self:
+            create_date = record.create_date.date() if record.create_date else fields.Date.today()
+            record.validity = (record.date_deadline - create_date).days
+
+    @api.model_create_multi
```
mash-odoo, line 58: When do we use `model_create_multi`?

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,14 @@
...
+
+class EstatePropertyTag(models.Model):
+    _name = "estate.property.tag"
+    _description = "Estate Property Tag"
+    _order = "name"
+
+    name = fields.Char(required=True)
+    color = fields.Integer(string="Color")
+    _name_unique = models.Constraint(
+        'UNIQUE(name)',
```
mash-odoo, line 13: Is this case sensitive?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,147 @@
...
+    </record>
+    <record id="estate_property_form_view" model="ir.ui.view">
+        <field name="name">estate.property.form</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form string="Property">
+                <sheet>
+                    <header>
+                        <button name="action_sold"
+                                type="object"
```
mash-odoo, line 114: When do we use `type="object"`

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,47 @@
...
+                    </div>
+                    <group>
+                        <field name="name"/>
+                        <field name="property_ids">
+                            <list>
+                                <field name="name"/>
+                                <field name="expected_price"/>
+                                <field name="state"/>
+                                <field name="date_availability" optional="hide"/>
+                            </list>
```
mash-odoo, line 40: When do we need to explicitly mention a list?

mash-odoo (changes requested):
Hello!
Thanks you for your work!!
Here are some questions and suggestions..

estate/models/estate_property.py
venpr-odoo: Postcodes are identifiers, not numeric values. They may contain leading zeros or letters depending on country, so using Char ensures correctness and avoids data loss.

estate/models/estate_property.py
venpr-odoo: We use store=True when the computed field needs to be searchable, used in filters or group by, for performance optimization, and especially when the field is a dependency of another computed field to ensure proper recomputation chaining.

estate/models/estate_property_offer.py
venpr-odoo: We use @api.model_create_multi when overriding the create() method to support batch record creation efficiently, since Odoo may create multiple records in a single call.

estate/models/estate_property_tag.py
venpr-odoo: Yes, it is case-sensitive by default. PostgreSQL treats 'Garden' and 'garden' as different values under a standard UNIQUE constraint. If we need case-insensitive uniqueness, we would have to explicitly use lower(name) or a case-insensitive column type.

estate/views/estate_property_views.xml
venpr-odoo: We use type="object" when the button should call a Python method defined in the model to execute business logic on the current record(s).

estate/views/estate_property_type_views.xml
venpr-odoo: We explicitly define a <list> when we want to control the columns and layout of a One2many/Many2many field inside a form view, or when defining a standalone list view. If no customization is needed, Odoo can use the default list view automatically.

estate/models/estate_property.py
venpr-odoo: We use lambda because the default value needs to be calculated when the record is created.
It allows us to assign the current logged-in user dynamically.

estate/models/estate_property.py
venpr-odoo: SQL constraint works at the database level and is faster.
Python constraint works in Odoo and allows more complex business logic.

estate/models/estate_property.py
venpr-odoo: mapped() is used to get a specific field value from multiple records.
It helps us extract values easily from a recordset.

estate/models/estate_property.py
venpr-odoo: Decorators tell Odoo when and how a method should run.
They connect our method to Odoo’s ORM lifecycle like create, write, or compute.

estate_account/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "Real Estate - Accounting",
+    "version": "1.0",
+    "category": "Tutorials",
+    "summary": "Real Estate Accounting Link Module",
+    "author": "Prasad Venkat",
+    "license": "LGPL-3",
+    "depends": ["estate", "account"],
+    "data": [],
+    "application": False,
```
mash-odoo: This is by default False only, so you can avoid writing this.

estate_account/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "Real Estate - Accounting",
+    "version": "1.0",
+    "category": "Tutorials",
+    "summary": "Real Estate Accounting Link Module",
+    "author": "Prasad Venkat",
+    "license": "LGPL-3",
+    "depends": ["estate", "account"],
```
mash-odoo, line 8: Currently, if we install this, `estate` and `account` will be installed to, but what if I want to have this `estate_account` installed automatically when I have estate and account installed in my system?

estate/views/res_users_views.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <record id="res_users_view_form_inherit_estate" model="ir.ui.view">
+        <field name="name">res.users.form.inherit.estate</field>
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" position="inside">
```
mash-odoo, line 8:
```suggestion
            <notebook position="inside">
```
If the path is direct, you can try writing it directly without using `xpath` and `expr`

estate/views/estate_request_views.xml
```diff
@@ -0,0 +1,63 @@
...
+                            invisible="state not in ['Assigned']"/>
+                    <button name="action_done" type="object" string="Stop"
+                            class="oe_highlight"
+                            invisible="state != 'Progress'"/>
+                    <button name="action_cancel" type="object" string="Cancel"
+                            class="oe_link" 
+                            invisible="state in ['Done', 'cancelled']"/>
+                    <field name="state"
+                        widget="statusbar"
+                        statusbar_visible="new,Assigned,Progress,Done,cancelled"/>
```
mash-odoo: What's the purpose of `statusbar_visible`?
What's the point of using this if you are mentioning all the values?

estate/models/estate_request.py
```diff
@@ -0,0 +1,68 @@
...
+        ('medium', 'Medium'),
+        ('high', 'High'),
+    ], string="Priority", default='medium')
+    state = fields.Selection([
+        ('new', "New"),
+        ('Assigned', "Assigned"),
+        ('Progress', "Progress"),
+        ('Done', "Done"),
+        ('cancelled', "Cancelled"),
+    ], default='new', copy=False)
```
mash-odoo, line 35:
```suggestion
    state = fields.Selection([
        ('new', "New"),
        ('assigned', "Assigned"),
        ('progress', "Progress"),
        ('done', "Done"),
        ('cancelled', "Cancelled"),
    ], default='new', copy=False)
```
Keep the keys in lower case

estate/views/estate_request_views.xml
```diff
@@ -0,0 +1,63 @@
...
+    </record>
+    <record id="estate_property_request_form" model="ir.ui.view">
+        <field name="name">estate.property.request.form</field>
+        <field name="model">estate.request</field>
+        <field name="arch" type="xml">
+            <form>
+                <header>
+                    <button name="action_progress" type="object" string="Start"
+                            class="oe_highlight" 
+                            invisible="state not in ['Assigned']"/>
```
mash-odoo:
```suggestion
                            invisible="state != 'assigned'"/>
```

estate/views/estate_request_views.xml
```diff
@@ -0,0 +1,63 @@
+<odoo> 
+    <record id="estate_property_request_action_property" model="ir.actions.act_window">
+        <field name="name">Property Requests</field>
+        <field name="res_model">estate.request</field>
+        <field name="view_mode">list,form</field>
+        <field name="domain">[('request_id', '=', active_id)]</field>
+        <field name="context">{'default_request_id': active_id}</field>
```
mash-odoo, line 7: Can you explain the difference between domain and context?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,219 @@
...
+                            <p>
+                                <strong>Expected Price:</strong>
+                                <field name="expected_price"/>
+                            </p>
+                            <p t-if="record.state.raw_value == 'offer_received'">
+                                <strong>Best Price:</strong>
+                                <field name="best_price"/>
+                            </p>
+
+                            <p t-if="record.state.raw_value == 'offer_accepted'">
```
mash-odoo, line 78: What's the purpose of using `raw_value`?
Why can't I directly use `record.state`?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,219 @@
...
+                                    <field name="validity"/>
+                                    <field name="date_deadline"/>   
+                                    <field name="status" 
+                                        column_invisible="parent.state != 'offer_accepted'"/>
+                                    <button name="action_accept"
+                                            type="object"
+                                            icon="fa-check"
+                                            class="btn-success"
+                                            title="Accept Offer"
+                                            invisible="status in ['accepted','refused'] or parent.state == 'offer_accepted'"/>
```
mash-odoo, line 193:
```suggestion
                                            invisible="status or parent.state == 'offer_accepted'"/>
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,177 @@
...
+
+    request_ids = fields.One2many(
+        "estate.request",
+        "request_id",
+        string="Requests",
+    )
+    request_count = fields.Integer(
+        string="Request Count",
+        compute="_compute_request_count",
+        store=True,
```
mash-odoo, line 78: Why do we need to store this value here?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,127 @@
...
+        required=True,
+    )
+    property_id = fields.Many2one(
+        "estate.property",
+        string="Property",
+        required=True,
+    )
+    property_type_id = fields.Many2one(
+        "estate.property.type",
+        related="property_id.property_type_id",
```
mash-odoo, line 33: Why do we need to use here `related`?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,127 @@
...
+            other_pending_offers = offer.property_id.offer_ids.filtered(
+                lambda o: o.status != 'refused' and o != offer
+            )
+            other_pending_offers.write({'status': 'refused'})
+
+    def action_refuse(self):
+        for offer in self:
+            property_rec = offer.property_id
+            if property_rec.state in ['sold', 'cancelled']:
+                raise UserError("You cannot refuse an offer on a sold or cancelled property.")
```
mash-odoo, line 125: I don't think we would need this UserError as we are making the button invisible when the status is set.
So this can be a dead code.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,219 @@
...
+                                    <button name="action_refuse"
+                                            type="object"
+                                            icon="fa-times"
+                                            class="btn-danger"
+                                            title="Refuse Offer"
+                                            invisible="status in ['accepted', 'refused'] or parent.state == 'offer_accepted'"/>
```
mash-odoo, line 199: Where is the function written for this button?
Also, I am not sure if this button should exist here as it is created to refuse offer.

estate/models/estate_property_visit.py
```diff
@@ -0,0 +1,53 @@
...
+        string="Agent",
+        required=True
+    )
+    visit_date = fields.Datetime(string="Visit Date")
+
+    state = fields.Selection([
+        ('schedule', 'Scheduled'),
+        ('done', 'Done'),
+        ('cancel', 'Cancelled')
+    ], default="schedule")
```
mash-odoo, line 23:
```suggestion
    state = fields.Selection([
        ('schedule', "Scheduled"),
        ('done', "Done"),
        ('cancel', "Cancelled")
    ], default='schedule')
```

estate/models/estate_property_visit.py
```diff
@@ -0,0 +1,53 @@
...
+        for rec in self:
+            rec.state = 'done'
+
+    def action_cancel(self):
+        for rec in self:
+            rec.state = 'cancel'
+
+    def action_reset(self):
+        for rec in self:
+            rec.state = 'schedule'
```
mash-odoo: Make the state editable instead.
Then there won't be a need for this function!

estate/views/estate_property_visit_views.xml
```diff
@@ -0,0 +1,69 @@
...
+                            class="btn-primary"
+                            invisible="state != 'schedule'"/>
+                    <button name="action_cancel"
+                            string="Cancel"
+                            type="object"
+                            invisible="state != 'schedule'"/>
+                    <button name="action_reset"
+                            string="Reset"
+                            type="object"
+                            invisible="state == 'schedule'"/>
```
mash-odoo: This can be removed

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,219 @@
...
+        <field name="search_view_id" ref="estate_property_search_view"/>
+        <field name="context">{'search_default_available': 1}</field>
+    </record>
+    <record id="estate_property_list_view" model="ir.ui.view">
+        <field name="name">estate.property.list</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <list string="Properties"
+                decoration-success="state == 'offer_received'"
+                decoration-bf="state == 'offer_accepted'"
```
mash-odoo, line 36: Can I apply decoration in the form view?

mash-odoo (changes requested):
Hello!
Thank you for your work..
Here are some comments and suggestions..
Also please update your PR title and description

estate_account/__manifest__.py
venpr-odoo: To install estate_account automatically when both estate and account are installed, we should add "auto_install": True in the manifest. This makes Odoo automatically install the module when all dependencies are present.

estate/views/estate_request_views.xml
venpr-odoo: If all states are listed, then statusbar_visible does not change anything. It becomes unnecessary.

Normally it is used when we want to show only some states in the status bar and hide other states.

estate/views/estate_property_views.xml
venpr-odoo: The function for this button is written in the estate.property.offer model as the method action_refuse.

This button is placed inside the offers list (offer_ids) in the property form. The intention here is to allow the user to refuse a specific offer directly from the list of offers related to that property.

estate/views/estate_property_views.xml
venpr-odoo: No, decorations like decoration-success, decoration-bf, decoration-danger can only be used in list or kanban views. They are not supported in form views.

estate/views/estate_request_views.xml
venpr-odoo: domain is used to filter records that should be shown in the view.
In this case it shows only the records where request_id is equal to the current record (active_id).

context is used to pass additional information to the view or model.
Here it sets a default value, so when creating a new record the field request_id will automatically be filled with the current record (active_id)

estate/views/estate_property_views.xml
venpr-odoo: raw_value gives the actual value stored in the field.

In kanban templates, record.state is not the direct value. It is an object used for display. Because of that, comparing record.state may not work correctly.

record.state.raw_value returns the real value stored in the database (like offer_accepted), so it is used in conditions like t-if.

estate/models/estate_property.py
venpr-odoo: We store this value so it is saved in the database and does not need to be recomputed every time the record is opened. This improves performance, especially when the value is used in list views, search, or grouping.

estate/models/estate_property_offer.py
venpr-odoo: We use related to get the property type from the related property record.

So the offer automatically shows the same property type as the property. This avoids storing the same data again in the offer model.

estate/models/estate_property_offer.py
venpr-odoo: This logic was written earlier before the invisible condition was added to the button. After adding the invisible attribute, the user can no longer click the button in those states. Because of that, this UserError may not be necessary now, and it was not removed after adding the UI condition.

PR #1159, opened 2026-02-10
19.0 tutorial-sasri

sasri-odoo:
Implements Chapter 7 of the estate tutorial by introducing the property model, UI views, and relational fields using many2one, one2many, and many2many


estate/models/estate_property.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate Property"
+
+    name = fields.Char()
+    description = fields.Text(required=True)
+    postcode = fields.Char()
+
```
```suggestion
```
Do not leave lines between fields

estate/models/estate_property.py
```diff
@@ -0,0 +1,78 @@
...
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+
+    garden_orientation = fields.Selection(
+        [
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
```
```suggestion
            ('north', "North"),
            ('south', "South"),
            ('east', "East"),
            ('west', "West"),
```
try keeping the key in lower case

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
+access_estate_property_type,access_estate_property_type,model_estate_property_type,base.group_user,1,1,1,1
+access_estate_property_tag,access_estate_property_tag,model_estate_property_tag,base.group_user,1,1,1,1
+access_estate_property_offer,access_estate_property_offer,model_estate_property_offer,base.group_user,1,1,1,1
```
mash-odoo: Always add an extra line at the end of the file

estate/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "Real Estate",
+    "version": "1.0",
+    "category": "Tutorial",
+    "summary": "Manage real estate properties",
+    "author": "sasri-odoo",
+    "license": "LGPL-3",
+    "depends": [
+        "base",
+    ],
```
mash-odoo:
```suggestion
    "depends": [ "base"],
```

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+    <record id="estate_property_offer_list_view" model="ir.ui.view">
+        <field name="name">estate.property.offer.list</field>
```
mash-odoo:
```suggestion
    <record id="estate_property_offer_view_list" model="ir.ui.view">
        <field name="name">estate.property.offer.view.list</field>
```
For naming a view: `<model_name>_view_<view_type>`, where view_type is `kanban`, `form`, `list`, `search`, …

mash-odoo (changes requested):
Hello!
Thank you for your work!
Please revert the changes for `awesome_owl` , `awesome_dashboard` and `website_airproof` as of now
Here are some suggestions, mostly linter as of now!
Also, please update your PR title and description.
Furthermore, please update description in your commits while pushing, regarding the chapters and topics covered.

estate/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+from . import estate_property
+from . import estate_property_type
+from . import estate_property_tag
+from . import estate_property_offer
```
mash-odoo, line 2: Arrange the imports alphabetically

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+from odoo import api, fields, models
+
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate Property"
+    _order = "id desc"
```
mash-odoo, line 12: What does this do?

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+
+    name = fields.Char()
+    description = fields.Text(required=True)
+    postcode = fields.Char()
```
mash-odoo, line 17: Why do we need `Char` field type for postcode?

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+    state = fields.Selection(
+        [
+            ('new', "New"),
+            ('offer_received', "Offer Received"),
+            ('offer_accepted', "Offer Accepted"),
+            ('sold', "Sold"),
+            ('cancelled', "Cancelled"),
+        ],
+        readonly=True,
+        default="new",
```
```suggestion
        default='new',
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = (record.living_area or 0.0) + (record.garden_area or 0.0)
+
+    @api.depends("offer_ids.price")
+    def _computer_best_price(self):
+        for record in self:
+            if record.offer_ids:
+                record.best_price = max(record.offer_ids.mapped("price"))
```
mash-odoo, line 128: What is the function `mapped` used for?

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+            else:
+                record.best_price = 0.0
+
+    @api.onchange("garden")
+    def _onchange_garden(self):
```
mash-odoo, line 162: What are the parameters by which you decide what to use: compute or onchange?

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+        for record in self:
+            if record.state == "sold":
+                raise UserError("A sold property cannot be cancelled")
+            record.state = "cancelled"
+
+    def action_sold(self):
+        for record in self:
+            if record.state == "cancelled":
+                raise UserError("A cancelled property cannot be sold")
+            accepted = record.offer_ids.filtered(
```
mash-odoo: What does `filtered` do?
Do you think it's the best solution to use filtered here or do you have any other way?

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+
+    _check_expected_price = models.Constraint(
+        "CHECK(expected_price > 0)",
+        "The expected price must be strictly positive.",
+    )
+
+    _check_selling_price = models.Constraint(
+        "CHECK(selling_price > 0)",
+        "The selling price must be strictly positive",
+    )
```
mash-odoo: Please check the order of writing the functions and constraints from [here](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions) and reorder it

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+
+    @api.constrains("selling_price", "expected_price")
+    def _check_selling_price(self):
+        for record in self:
+            if float_is_zero(record.selling_price, precision_rounding=0.01):
+                continue
+
+            minimum_price = record.expected_price * 0.9
+
+            if float_compare(
```
mash-odoo: What does `float_compare` do?

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+                record.selling_price,
+                minimum_price,
+                precision_rounding=0.01,
+            ) < 0:
+                raise ValidationError(
```
mash-odoo: What are the different types of errors?
Why did you think to add here `ValidationError` and not `UserError`?

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,14 @@
...
+
+class EstatePropertyTag(models.Model):
+    _name = "estate.property.tag"
+    _description = "Property Tag"
+    _order = "name"
+
+    name = fields.Char(required=True)
+
+    _unique_name = models.Constraint(
+        "UNIQUE(name)",
```
mash-odoo, line 13: Does this `UNIQUE` work for all cases or it has a drawback?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,139 @@
...
+                        <button name="action_cancel"
+                                type="object"
+                                string="Cancel"
+                                class="btn-secondary"/>
+                    <group>
+                        <group>
+                            <field name="name"/>
+                            <field name="state"
+                                   widget="statusbar"
+                                   statusbar_visible="new,offer_received,offer_accepted,sold,cancelled"
```
mash-odoo, line 90: What's the purpose of `statusbar_visible`?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,139 @@
...
+                <separator/>
+
+                <filter
+                    name="available"
+                    string="Available"
+                    domain="[
+                        '|',
+                        ('state', '=', 'new'),
+                        ('state', '=', 'offer_received')
+                    ]"
```
mash-odoo, line 221: Can this be written in any other way?
Also what if I don't write any operator here?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,80 @@
...
+        for record in self:
+            if record.property_id.state in ["cancelled", "sold"]:
+                raise UserError("Cannot modify offers for a sold or cancelled property")
+            if record.status == "accepted":
+                record.property_id.selling_price = 0.0
+                record.property_id.buyer_id = False
+                record.property_id.state = "new"
+            record.status = "refused"
+
+    _check_offer_price = models.Constraint(
```
mash-odoo, line 144: When do you use database level constraint and python level constraint?

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,41 @@
...
+
+    <record id="estate_property_type_view_list" model="ir.ui.view">
+        <field name="name">estate.property.type.view.list</field>
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <list>
+                <field name="sequence" widget="handle"/>
+                <field name="name"/>
+            </list>
+        </field>
```
mash-odoo, line 18: Can you make this list editable and eliminate the usage of form view?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,80 @@
...
+            if record.create_date and record.date_deadline:
+                delta = record.date_deadline - record.create_date.date()
+                record.validity = delta.days
+
+    def action_accept(self):
+        for record in self:
+            if record.property_id.state in ["cancelled", "sold"]:
+                raise UserError("Cannot accept an offer for a cancelled or sold property")
+            accepted_offers = record.property_id.offer_ids.filtered(
+                lambda o: o.status == "accepted"
```
mash-odoo, line 122: You need to be more explicit than just `'o'`, it doesn't help while reading the code

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+    )
+    buyer_id = fields.Many2one(
+        "res.partner",
+        string="Buyer",
+        copy=False,
+    )
+    user_id = fields.Many2one(
+        "res.users",
+        string="Salesperson",
+        default=lambda self: self.env.user,
```
mash-odoo, line 65: What's the purpose of using `lambda` function here?

mash-odoo (changes requested):
Hello!
Thank you for your work!
Here are some few questions and suggestions...

estate/models/estate_property.py
sasri-odoo: The class EstateProperty defines an Odoo model named "estate.property" with a description. If we are specifically talking about the '_order = "id desc"' line, it sets the default record sorting so that newer properties appear first.

estate/models/estate_property.py
sasri-odoo: That is because postcode strictly isn't a number. Depending upon the country, it may even be a letter, number,etc.

estate/models/estate_property.py
sasri-odoo: mapped is used to get values of a specific field from a set of records. Here we require the price values from the record set. Hence we use mapped.

estate/models/estate_property.py
sasri-odoo: compute is generally decided when the fields value is automatically calculated based on other fields, and als o if we want the values always to be up to date.
onchange is more of a UI modification which is used to visually update the changed field to the user.

estate/models/estate_property.py
sasri-odoo: filtered() is used to select records from a record set that meet a certain condition. Certainly, there are other ways to do this, for example, using a simple for loop to check each record,but while that works, .filtered() is more efficient, readable.Efficient because it is a Odoo builtin method used to directly operate on the record sets.

estate/models/estate_property.py
sasri-odoo: float_compare is used to safely compare decimal numbers because direct float comparisons can give precision errors. moreover, float_compare allows specifying a precision, this ensures more accuracy.

estate/models/estate_property.py
sasri-odoo: The types of errors are UserError, ValidationError, RedirectWarning, ConcurrencyError.(found in exceptions.py file in the codebase).
In here, i have added validationerror instead of usererror because validationerror(as the name suggests) prevents the record from saving into the database because it fails the business rule. Even if I replace validationerror with usererror, the user still sees the error but it is less robust.

estate/models/estate_property_tag.py
sasri-odoo: UNIQUE doesnt work well for ALL cases correctly. It is case sensitive, meaning we can have two tags named 'TAG' AND 'tag' which is practically the same name.

estate/views/estate_property_views.xml
sasri-odoo: It is used to control which states are shown in the visual status bar. without 'statusbar_visible' all states in the fields selection are displayed, which may not be optimal for a user who wants to customize his/her view.

estate/views/estate_property_views.xml
sasri-odoo: Yes, it can be written in other ways. The immediate and simple one would be using 'in' instead of 'or'.
If we do not mention any operator, AND operation would be performed automatically.

estate/models/estate_property_offer.py
sasri-odoo: we use database level constraints for simple and universal rules like UNIQUE and NOT NULL and also if we want to enforce it directly into the database.Whereas, python level constraints are used when the rules are complex, or conditional(depends or multuple records).

estate/views/estate_property_type_views.xml
sasri-odoo: Yes, we can do that using 'editable = "top"'  which eliminates the form view and new records can be directly added inline.

estate/models/estate_property.py
sasri-odoo: lambda function is used here to set the default value for the field when creating a new record.Coming to why only 'lambda', because it ensures the self.env.user is evaluated only when a new record is created, without that the value would be set while defining the class and get stored which is incorrect.

estate/models/estate_property.py
mash-odoo: So, when we have a direct domain available, we can directly use `filtered_domain`.
As per your case you can directly write `record.offer_ids.filtered_domain([('status', '=', 'accepted')])`.
The can avoid the use of lambda function.

estate/models/estate_property_maintenance.py
```diff
@@ -0,0 +1,54 @@
...
+        "res.users",
+        string="technician"
+    )
+    state = fields.Selection(
+        [
+            ("new", "New"),
+            ("assigned", "assigned"),
+            ("in_progress", "In progress"),
+            ("done", "done"),
+            ("cancel", "cancel"),
```
mash-odoo:
```suggestion
            ('new', "New"),
            ('assigned', "assigned"),
            ('in_progress', "In progress"),
            ('done', "done"),
            ('cancel', "cancel"),
```

estate_account/__manifest__.py
```diff
@@ -0,0 +1,11 @@
+{
+    "name": 'Estate Account',
+    "version": "1.0",
+    "category": "Tutorial",
+    "author": "sasri-odoo",
+    "license": "LGPL-3",
+    "depends": ["estate", "account"],
+    "data": [],
+    "application": True,
```
mash-odoo, line 10: Currently, if we install this estate_account, `estate` and `account` will be installed to, but what if I want to have this `estate_account` installed automatically when I have estate and account installed in my system?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,207 @@
...
+                </sheet>
+            </form>
+        </field>
+    </record>
+
+    <record id="estate_property_view_kanban" model="ir.ui.view">
+        <field name="name">estate.property.view.kanban</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <kanban default_group_by="property_type_id" draggable="false">
```
mash-odoo, line 169: What are `default_group_by` and `records_draggable` used for?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,207 @@
...
+                                >
+                                    <field name="price"/>
+                                    <field name="partner_id"/>
+                                    <field name="status" readonly="1"/>
+                                    <button name="action_accept"
+                                            type="object"
+                                            string="Accept"
+                                            icon="fa-check"
+                                            class="btn-primary"
+                                            invisible="status in ['accepted','refused']"/>
```
```suggestion
                                            invisible="status"/>
```
Here you need to have this invisible when any of the selection is set and here we have just two options, so instead of writing both, you can just give the field name. What will happen here is that, if the status field is set, your button will become invisible.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,207 @@
...
+                                    <button name="action_refuse"
+                                            type="object"
+                                            string="Refuse"
+                                            icon="fa-times"
+                                            class="btn-secondary"
+                                            invisible="status in ['accepted','refused']"/>
```
```suggestion
                                            invisible="status"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,207 @@
...
+                            <field name="name"/>
+                            <field name="state"
+                                   widget="statusbar"
+                                   statusbar_visible="new,offer_received,offer_accepted,sold,cancelled"
+                            />
+                            <field name="postcode"/>
+                            <field name="expected_price"/>
+                            <field name="selling_price"/>
+                            <field name="property_type_id"
+                                   options="{'no_create': True, 'no_edit': True}"
```
mash-odoo, line 96: What's the purpose of `no_create` and `no_open`?

estate/views/estate_property_views.xml
mash-odoo: So here, I don't think you will require this attribute as you are displaying all the values.

estate/views/estate_property_maintenance.xml
```diff
@@ -0,0 +1,69 @@
...
+                                statusbar_visible="new,assigned,in_progress,done,cancel"
+                                />
+                            <field name="name"/>
+                            <field name="description"/>
+                        </group>
+                        <group>
+                            <field name="property_id"/>
+                            <field name="technician_id"/>
+                            <field name="estimated_cost"/>
+                            <field name="actual_cost" readonly = "state not in 'done'"/>
```
mash-odoo:
```suggestion
                            <field name="actual_cost" readonly = "state != 'done'"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,207 @@
...
+                    ]"
+                />
+
+                <separator/>
+
+                <filter
+                    name="group_by_postcode"
+                    string="group by postcode"
+                    context="{'group_by': 'postcode'}"
+                />
```
mash-odoo, line 230: Can you make a default group by?

estate/views/estate_res_user_views.xml
```diff
@@ -0,0 +1,25 @@
...
+<odoo>
+
+
+    <record id="views_users_form_inherit_estate" model="ir.ui.view">
+        <field name="name">res.users.form.inherit.estate</field>
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form"/>
+        <field name="arch" type="xml">
+
+            <xpath expr="//notebook" position="inside">
```
mash-odoo:
```suggestion
            <notebook position="inside">
```
If the path is direct, you can try writing it directly

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,107 @@
...
+        string="Deadline",
+        compute="_compute_date_deadline",
+        inverse="_inverse_date_deadline"
+    )
+    property_type_id = fields.Many2one(
+        "estate.property.type",
+        string="Property Type",
+        related="property_id.property_type_id",
+        store=True,
+    )
```
mash-odoo, line 45: What is the logic behind introducing this field here?

mash-odoo (changes requested):
Hello!
Thank you for your work..
Here are a few suggestions and questions..
Also, can you please update the PR title and description?

estate_account/__manifest__.py
sasri-odoo: In that case, we use auto_install: True, which checks if the dependencies are installed first, if they are, it proceeds to install the module automatically.

estate/views/estate_property_views.xml
sasri-odoo: default_group_by is used to group records by a particular field (in this case, property type) when the Kanban view is opened.
The draggable=False that i have written prevents changing the property within the group in the kanban view. To answer your question, records_draggable is used to restrict moving records from one group to another.

estate/views/estate_property_views.xml
sasri-odoo: inside the form view of a property as we are using property_type which is a many2one field, By default, the view allows creating a new property type. no_create and no_edit here are used to prevent that .

estate/views/estate_property_views.xml
sasri-odoo: if you mean making the groupby postcode a default filter, yes, we can make it by declaring in the action.
after implementing this is what the code looks like 
<field name="context">
            {
                'search_default_available': 1,
                'group_by': 'postcode'
            }
        </field>

estate/models/estate_property_offer.py
sasri-odoo: the field property_type_id is used here because, without this the offer only knows its property, to find the property type we would have to go through property model which is not efficient. by adding this field the offer knows its property type directly so it becomes easy to operate on.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,107 @@
...
+    @api.model
+    def create(self, vals_list):
+        for vals in vals_list:
+            property_record = self.env['estate.property'].browse(vals.get('property_id'))
+
+            if property_record.state in ('offer_accepted', 'sold'):
+                raise UserError("Cannot create new offer. An offer is already accepted.")
+            if property_record.offer_ids:
+                highest_price = max(property_record.offer_ids.mapped('price'))
+                if vals.get('price') <= highest_price:
```
mash-odoo, line 102:
```suggestion
                if vals.get('price', 0) <= highest_price:
```
If `vals.get('price')` returns None (price not passed), it can throw an error.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,107 @@
...
+                    raise UserError("New offer must not cost less than the previous offers")
+
+        offers = super().create(vals_list)
+        for offer in offers:
+            property_record.state = 'offer_received'
```
mash-odoo: `property_record` is set inside the `for vals in vals_list` loop but used outside it which can be buggy!
Instead you can try: `offer.property_id.state = 'offer_received'`

estate/models/estate_property.py
```diff
@@ -0,0 +1,197 @@
...
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for record in self:
+            if record.offer_ids:
+                record.best_price = max(record.offer_ids.mapped("price"))
+            else:
+                record.best_price = 0.0
+
+    @api.constrains("selling_price", "expected_price")
+    def _check_selling_price(self):
```
mash-odoo, line 133: There's a `models.Constraint` named **_check_selling_price** and an `@api.constrains `method also named **_check_selling_price**. The Python method will silently shadow the SQL constraint in your case. So you can rename either of the two constraints!

estate/models/estate_property_maintenance.py
```diff
@@ -0,0 +1,54 @@
...
+        for record in self:
+            if record.state == "in_progress":
+                record.state = "done"
+
+    def action_start(self):
+        for record in self:
+            if record.state == "assigned":
+                record.state = "in_progress"
+                if record.estimated_cost == 0:
+                    raise ValidationError("Estimate price is a required field")
```
mash-odoo: The state is set to `"in_progress"` before the validation check. If the error is raised, the state would have already been changed which can cause an issue. The validation should come first, before changing state.

estate/models/estate_property.py
```diff
@@ -0,0 +1,197 @@
...
+                raise UserError("A sold property cannot be cancelled")
+            record.state = "cancelled"
+
+    def action_sold(self):
+        for record in self:
+            if record.state == "cancelled":
+                raise UserError("A cancelled property cannot be sold")
+            accepted = record.offer_ids.filtered_domain([('status', '=', 'accepted')])
+            if not accepted:
+                raise UserError("You must accept an offer before selling.")
```
mash-odoo, line 188: You validate… but never change the record state to sold

estate/models/estate_property_maintenance.py
mash-odoo: Field is already a required one, so maybe you can skip this validation?

estate/views/estate_res_user_views.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+
+    <record id="views_users_form_inherit_estate" model="ir.ui.view">
```
mash-odoo, line 5: you can directly keep the file name as **res_users_views.xml** and py file as **res_users.py.**

estate/views/estate_property_visit_views.xml
```diff
@@ -0,0 +1,57 @@
...
+        <field name="arch" type="xml">
+            <form>
+                <sheet>
+                    <group>
+                        <field name="property_id"/>
+                        <field name="partner_id"/>
+                        <field name="visit_date"/>
+                        <field name="state"
+                               widget="statusbar"
+                                />
```
mash-odoo, line 30:
```suggestion
                        <field name="state" widget="statusbar"/>
```
Can be written in a single line

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,216 @@
...
+                                <field name="buyer_id" readonly="state=='sold'"/>
+                                <field name="user_id"/>
+                            </group>
+                        </page>
+                        <page string="Offers">
+                            <field name="offer_ids" readonly="state in ['sold','cancelled']">
+                                <list editable="top"
+                                      decoration-danger="status == 'refused'"
+                                      decoration-success="status == 'accepted'" 
+                                >
```
```suggestion
                                <list
                                    editable="top"
                                    decoration-danger="status == 'refused'"
                                    decoration-success="status == 'accepted'" 
                                >
```

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,52 @@
...
+        <field name="res_model">estate.property.type</field>
+        <field name="view_mode">list,form</field>
+    </record>
+
+    <record id="estate_property_type_view_list" model="ir.ui.view">
+        <field name="name">estate.property.type.view.list</field>
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <list>
+                <field name="sequence" widget="handle"/>
```
mash-odoo, line 15: What if you do not give the widget over here?

mash-odoo (commented):
Hello!
Good going on the task..
Here are a few comments!
Also please change the PR title and description.

awesome_owl/static/src/card/card.js
```diff
@@ -0,0 +1,21 @@
...
+    };
+
+    setup() {
+        this.state = useState({ 
+                    open: true
+        });
+    };
+
+    toggle() {
+        this.state.open != this.state.open;
```
mash-odoo:
```suggestion
        this.state.open = !this.state.open;
```
Your current logic won't work and the card won't toggle as you have used a comparison operator...instead you should use an assignment operator.

awesome_owl/static/src/todo/todo_list.xml
```diff
@@ -0,0 +1,21 @@
...
+        <t t-foreach="todos" t-as="todo" t-key="todo.id">
+            <TodoItem
+                todo="todo"
+                toggleState.bind="toggleState"
+                removeTodo.bind = "removeTodo"
+                />
+        </t>
+    </div>
+</t>
+</templates>
```
mash-odoo:
```suggestion
<templates xml:space="preserve">
    <t t-name="awesome_owl.todolist">
        <div>
            <input
                t-ref="input"
                type="text"
                placeholder="Enter a new task"
                t-on-keyup="addTodo"
            />
            <t t-foreach="todos" t-as="todo" t-key="todo.id">
                <TodoItem
                    todo="todo"
                    toggleState.bind="toggleState"
                    removeTodo.bind = "removeTodo"
                />
            </t>
        </div>
    </t>
</templates>
```
Please maintain a proper indentation across all files

awesome_owl/static/src/todo/todo_list.js
```diff
@@ -0,0 +1,52 @@
...
+
+    this.inputRef = useRef("input");
+    onMounted(() => {
+        this.inputRef.el.focus();
+            }     
+        )
+    }
+
+    addTodo(ev) {
+        if (ev.keyCode === 13) {
```
mash-odoo:
```suggestion
        if (ev.key === "Enter")
```
`keyCode` is deprecated. You can use this instead.

awesome_owl/static/src/todo/todo_item.js
```diff
@@ -0,0 +1,18 @@
+import { Component } from "@odoo/owl";
+
+export class TodoItem extends Component {
+
+    static template = "awesome_owl.todoitem";
+
+    static props = {
+        todo: Object,
+        toggleState: Function,
+        removeTodo: Function
```
mash-odoo, line 10:
```suggestion
        onToggle: Function,
        onRemove: Function
```
Just a suggestion..upto you to implement or not

awesome_owl/static/src/todo/todo_list.js
```diff
@@ -0,0 +1,52 @@
...
+
+    static components = {TodoItem};
+
+    setup() {
+    this.todos = useState([]);  
+    this.nextId = 0;
+
+    this.inputRef = useRef("input");
+    onMounted(() => {
+        this.inputRef.el.focus();
```
mash-odoo:
```suggestion
        this.inputRef.el?.focus();
```
This can prevent some edge case like for eg. if ref is not ready.

awesome_owl/static/src/todo/todo_list.js
```diff
@@ -0,0 +1,52 @@
...
+        )
+    }
+
+    addTodo(ev) {
+        if (ev.keyCode === 13) {
+            const text = ev.target.value;
+
+            if (!text) {
+                return;
+            }
```
mash-odoo, line 25:
```suggestion
            const text = ev.target.value.trim();
            if (!text) return;
```
You can try preventing empty spaces in todos by this

awesome_owl/static/src/card/card.js
```diff
@@ -0,0 +1,21 @@
...
+    static template = "awesome_owl.card";
+    static props = {
+        title: String,
+        slots: { type: Object, optional: true },
+    };
+
+    setup() {
+        this.state = useState({ 
+                    open: true
+        });
```
mash-odoo:
```suggestion
        this.state = useState({ open: true });
```
Inconsistent formatting

awesome_owl/static/src/card/card.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
```
mash-odoo, line 2: What is `xml:space="preserve"` used for?
Also..please format this entire file and remove extra spaces and blank lines

awesome_owl/static/src/counter/counter.js
```diff
@@ -0,0 +1,22 @@
+import { Component, useState } from "@odoo/owl";
+
+export class Counter extends Component {
+    static template = "awesome_owl.counter";
+
+    static props = {
+        onchange: {type: Function, optional: true}
```
mash-odoo, line 7: What is `optional` used for?

awesome_owl/static/src/counter/counter.js
```diff
@@ -0,0 +1,22 @@
...
+
+    setup() {
+        this.state = useState({ value: 0 });
+    }
+
+    increment() {
+        this.state.value++;
+        if(this.props.onchange) {
+            this.props.onchange();
+        }
```
mash-odoo, line 18:
```suggestion
        this.props.onchange?.(this.state.value);
```
I think this would be better. WDYT?

mash-odoo (changes requested):
Hello!
Thank you for your work!
Here are a few questions..
Also please update your PR title and description

awesome_owl/static/src/card/card.xml
sasri-odoo: it ensures that whitespace are kept as written in the code instead of being normalized by the xml parser. 
And yes mam, ill surely format all the files and push them again.

awesome_owl/static/src/counter/counter.js
sasri-odoo: it means, the prop is not required when the component is used, it allows component to be reused without always passing a callback

awesome_owl/static/src/counter/counter.js
sasri-odoo: Yes, this is better. It simplifies the code and also passes the updated value.i'll update it.

PR #1160, opened 2026-02-13
[IMP] purchase: add global discount support on RFQ lines

mekot-odoo:
This PR introduces a global discount mechanism that allows users to
apply a percentage or fixed amount discount across all RFQ lines in a
single action, improving usability and efficiency.


purchase_discount/wizard/purchase_order_discount.py
```diff
@@ -0,0 +1,36 @@
...
+        if self.discount_type == "percent":
+            order.order_line.write({"discount": self.discount})
+        elif order.amount_untaxed != 0:
+            order.order_line.write({"discount": 0})
+            self.discount = (self.discount * 100) / order.amount_untaxed
+            order.order_line.write({"discount": self.discount})
+        else:
+            raise UserError(
+                "Cannot apply fixed discount because Total Amount is already zero."
+            )
```
kcv-odoo: This seems so wrong and giving me very weird behaviors, Could you try to understand it more and make it work properly? You can take inspiration from Discount on SO and Loyalty module or even try to use it's logic but it should not give any weird behavior 😉

purchase_discount/wizard/purchase_order_discount.xml
```diff
@@ -0,0 +1,25 @@
...
+    (
+                    <field name="discount_percent" nolabel="1"/>
+%
+    )
+                </span>
+                <footer>
+                    <button name="action_apply_discount" string="Apply" type="object" class="btn btn-primary"/>
+                    <button special="cancel" string="Discard" class="btn btn-secondary" data-hotkey="x"/>
+                </footer>
+            </form>
```
kcv-odoo, line 40: Make it more user friendly currently is seems very ugly 🙁

purchase_discount/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "purchase discount",
+    "version": "1.0.0",
+    "depends": ["base", "purchase"],
+    "author": "Mehul Kotak",
+    "category": "Task-1",
+    "description": "This perfrom global discount in purchase",
+    "license": "LGPL-3",
+    "application": True,
+    "installable": True,
```
kcv-odoo: It should not be application and auto_install should be true

purchase_discount/models/purchase_order.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def action_purchase_global_discount(self):
+        self.ensure_one()
+        return {
+            "name": "Discount",
+            "type": "ir.actions.act_window",
+            "res_model": "purchase.order.discount",
+            "view_mode": "form",
+            "target": "new",
+        }
```
kcv-odoo: Make button `type="action` instead object

purchase_discount/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_purchase_order_discount,access_purchase_order_discount,model_purchase_order_discount,base.group_user,1,1,1,1
```
kcv-odoo: Is it really useful for non purchase user?

kcv-odoo (commented):
Thank you for the PR!

It seems not working properly and giving weird behavior in different cases, Can you explain bit more about what you did? why you did?

From functional POV:
- In one case, the amount becomes 0, and in another case it shows 10 after applying the global discount.

- Issue with Global Discount Update: If I change the global discount from 10 to 1, the amount becomes 0, which should not happen.
The current logic (especially in the wizard) does not seem correct and needs to be reviewed.

could you try to make it work Properly according to Specs and add remaining things missing from the specs

@bit-odoo Did you reviewed this task? as it seems wrong and not working properly

Have a nice day!

purchase_discount/security/ir.model.access.csv
mekot-odoo: Thank you for pointing it out.

You’re right — granting access to base.group_user would allow all internal users to manage purchase order discount records, which is not necessary. This feature is intended to be used only by purchase users.

I will update the access rights to restrict it to group_purchase_user instead of all internal users.

purchase_discount/__manifest__.py
mekot-odoo: Thank you for the suggestion.

I’ve updated the manifest accordingly — the module is no longer marked as an application, and auto_install has been set to True.

purchase_discount/models/purchase_order.py
mekot-odoo: I’ve replaced the object method call with a type="action" button and linked it directly to the corresponding window action.

purchase_discount/wizard/purchase_order_discount.py
mekot-odoo: "Hello! Thank you for the feedback. After further investigation, I identified why the behavior was inconsistent: my previous logic was calculating the discount against the current amount_untaxed, which already included previous discounts. This caused the total to drift incorrectly when updating the value (e.g., changing from 10% to 1%).

To fix this and align with the standard logic used in SO/Loyalty:

    Consistent Base: I am refactoring the calculation to use the Base Untaxed Amount (Sum of price_unit * product_qty) so the discount is always calculated against the original 100% value.

    Standard Write: I’m removing the manual reset to zero and instead performing a single, clean write() to the lines based on this consistent base total.

    Validation: I'm adding constraints to ensure the discount stays within a logical 0-100% range.

purchase_discount/wizard/purchase_order_discount.py
```diff
@@ -0,0 +1,59 @@
...
+            raise UserError("There are no lines on this order to discount.")
+
+        base_amount = self._get_base_untaxed_amount(order)
+
+        if self.discount_type == "percent":
+            target_discount = self.discount
+        else:
+            if base_amount <= 0:
+                raise UserError("Cannot apply an amount discount to a $0 order.")
+            target_discount = (self.discount * 100) / base_amount
```
kcv-odoo, line 60: I don't think this is right? can you explain what is done and why it should be like this?

purchase_discount/wizard/purchase_order_discount.py
```diff
@@ -0,0 +1,59 @@
...
+    @api.constrains("discount", "discount_type")
+    def _check_discount_limit(self):
+        if self.discount < 0 or self.discount > 100:
+            raise ValidationError("Discount value is invalid")
+
+    def _get_base_untaxed_amount(self, order):
+        total = 0.0
+        for line in order.order_line:
+            total += line.price_unit * line.product_qty
+            return total
```
kcv-odoo: Don't we have total untaxed amount on the order? why we are doing this?

purchase_discount/wizard/purchase_order_discount.py
```diff
@@ -0,0 +1,59 @@
...
+
+class PurchaseOrderDiscount(models.TransientModel):
+    _name = "purchase.order.discount"
+    _description = "Apply Global Discount on Purchase Order"
+
+    discount = fields.Float(string="Discount", required=True)
+    discount_type = fields.Selection(
+        [("amount", "$"), ("percent", "%")], default="percent", string="Discount Type"
+    )
+    discount_percent = fields.Float(
```
kcv-odoo, line 16: shouldn't discount_type also be required?

purchase_discount/wizard/purchase_order_discount.py
```diff
@@ -0,0 +1,59 @@
...
+        if self.discount_type == "percent":
+            self.discount_percent = self.discount
+        elif base_amount > 0:
+            self.discount_percent = (self.discount * 100) / base_amount
+        else:
+            self.discount_percent = 0
+
+    def action_apply_discount(self):
+        self.ensure_one()
+        order = self.env["purchase.order"].browse(self.env.context.get("active_id"))
```
kcv-odoo: Can you create order_id field instead this? so we don't need to browse from context everytime we need order record

purchase_discount/wizard/purchase_order_discount.py
mekot-odoo: Yes, we do have the untaxed amount on the order, but that untaxed amount already have the discounted total. In case where we need to calculate the percentage for a discount of type 'amount', and there is already a discount applied, we need the original untaxed base amount to compute the correct percentage. That is why calculating the base amount separately is necessary.

purchase_discount/wizard/purchase_order_discount.py
mekot-odoo: Yes, you are right it should be required. i will do it.

purchase_discount/wizard/purchase_order_discount.py
mekot-odoo: So basically, if discount_type is 'amount', it first checks whether the base price is greater than zero to avoid a division-by-zero error. If the base amount is valid, then it calculates the  percentage from the given discount amount.

purchase_discount/wizard/purchase_order_discount.py
mekot-odoo: Yes, that's good idea , i will change it

purchase_discount/wizard/purchase_order_discount.xml
mekot-odoo: Ok, Now i made exactly as they are asking

purchase_discount/wizard/purchase_order_discount.py
kcv-odoo: Would it give desire result?

IMO If I say I am giving 100 rupee fixed discount on the order then it should deduct 100 rupees from total order amount regardless of number of order line, regardless of total amount

so I didn't understand how you trying to achieve it and is it working properly?

(PS. I would be better to add UI efferct instead raising User error on every step of the flow like here we could hide `Discount` button if order does not have amount by computing base_amount somewhere)

purchase_discount/wizard/purchase_order_discount.py
kcv-odoo: return is in For loop, Are you sure it is giving desire result? I don't think this idea to get this amount differently and keep other computation separate

purchase_discount/wizard/purchase_order_discount.xml
```diff
@@ -0,0 +1,47 @@
...
+    <record id="purchase_order_line_wizard_form" model="ir.ui.view">
+        <field name="name">purchase.order.discount.form</field>
+        <field name="model">purchase.order.discount</field>
+        <field name="arch" type="xml">
+            <form>
+                <sheet>
+                    <group>
+                        <div class="d-flex align-items-center gap-2">
+                            <label for="discount" string="Discount"/>
+                            <div style="width:50px;">
```
kcv-odoo: Try to use bootstrap here (same for other places)

purchase_discount/wizard/purchase_order_discount.py
mekot-odoo: > Would it give desire result?
> 
> IMO If I say I am giving 100 rupee fixed discount on the order then it should deduct 100 rupees from total order amount regardless of number of order line, regardless of total amount
> 
> so I didn't understand how you trying to achieve it and is it working properly?
> 
> (PS. I would be better to add UI efferct instead raising User error on every step of the flow like here we could hide `Discount` button if order does not have amount by computing base_amount somewhere)

Yes sir you are right but in this task https://www.odoo.com/odoo/project/18468/tasks/5926008 they are mention to perform global discount and global discount can be (% or amount) and what you tell me is fixed amount discount which is different type of discount which we can see in SO ,but you can also check mock-up of the task in which they clearly give us that if we global discount provide as amount then we have to first convert it into percentage and then have to apply to each products in order line. This is the thing which i understood from the task mock-up .

purchase_discount/wizard/purchase_order_discount.py
mekot-odoo: Yes, sir you are right by mistakenly i put return in loop , my logic is simple to calculate total price of all given product without  discount but mistakenly put it in the loop , i will fixed it.

purchase_discount/wizard/purchase_order_discount.py
kcv-odoo: Yess, I did see that but does it mean I get total amount convert it to % and then apply same percentage on each line, will it give desire result?

I am not able to test it because of constrain (which should only apply on % type)
image


and percentage is also not computing properly, I don't think 419.75 should be 52% of 36500 :thinking: 
image

kcv-odoo: > what you tell me is fixed amount discount which is different type of discount which we can see in SO

It should kind of work same as fix discount AFAIK using discount field on order line, like if I apply 100 discount then it should deduct 100 from order total

PR #1168, opened 2026-02-16
19.0 tutorials viwar

VIWAR-ODOO:
# completed till Chapter - 4
* created a custom module
* created the folder structure
* created the model
* added access rights in the ir.model.access.csv in the security folder


VIWAR-ODOO:
# Chapter - 5 - Done Actions and Menus
* created action
* created the menu
* created the three levels of menu 
* and created the first entry in the property - estate_model

**<h4>Learned using `<data>` tag in the XML if we define `<data noupdate="1">` then records within it will be loaded when the module is installed, and without the noupdate="1" the records will get reloaded at module install and update.</h4>**

VIWAR-ODOO:
# Chapter 5 done 
* added `active` and `state` field,
* added the `default values` for some fields
* added `copy=false` property to fields to not copy while duplicating
* added the the 3months + from now as default date in the `date availability` field

estate/models/estate_property.py
```diff
@@ -0,0 +1,42 @@
+from odoo import fields, models
+
+
+class TestModel(models.Model):
+    _name = "estate.property"
+    _description = "Test Model for real estate"
+    name = fields.Char(default="Unknown")
```
```suggestion
    _description = "Test Model for real estate"

    name = fields.Char(default="Unknown")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,42 @@
...
+    last_seen = fields.Datetime("Last Seen", default=fields.Datetime.now)
+    description = fields.Text()
+    postcode = fields.Char()
```
mash-odoo, line 21: Why do we need `Char` field type for postcode?

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
```
mash-odoo: Always add an extra line at the end of the file.

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version='1.0'?>
+<odoo>
+    <data>
```
mash-odoo, line 3: What's the purpose of using `data` tag here?

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    'category': 'real estate',
+    'description': "real estate App.",
+    'data': [
+        'security/ir.model.access.csv',
+        'views/estate_property_views.xml',
+        'views/estate_menus.xml'
+    ],
+    'application': True,
+    'installable': True,
+    'license': 'LGPL-3',
```
mash-odoo, line 26: What are the different types of licenses?

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    'author': 'viwar-odoo',
+    'category': 'real estate',
+    'description': "real estate App.",
+    'data': [
+        'security/ir.model.access.csv',
+        'views/estate_property_views.xml',
+        'views/estate_menus.xml'
+    ],
+    'application': True,
+    'installable': True,
```
mash-odoo, line 25: What does this `'installable': True` do?

estate/models/estate_property.py
```diff
@@ -0,0 +1,42 @@
+from odoo import fields, models
+
+
+class TestModel(models.Model):
```
```suggestion
class EstateProperty(models.Model):
```
Give a meaningful name of the class. Try keeping it same as the file name.

mash-odoo: Why did you use here `models.Model`?
What are the different types of models?

mash-odoo (changes requested):
Hello!
Thank you for your work..
Here are some suggestions and questions..

estate/models/estate_property.py
VIWAR-ODOO: thank you for suggestion, i'll keep in mind to leave a extra line

estate/models/estate_property.py
VIWAR-ODOO: the main reason to use the `char` was to include the `Zeros` in the starting of the number whereas in the Integer the starting `Zeros` is skipped, and another reason is to save on storage , because `char` is of `1 byte` and integer is of `4 bytes` , and another reason is the postcode might also required a alphabet to be inserted in that case `char` is suitable

estate/security/ir.model.access.csv
VIWAR-ODOO: point noted

estate/views/estate_menus.xml
VIWAR-ODOO: the purpose of data tag is to use the attribute of noupdate=1 ,in older versions the data tag used to bind multiple components within, but in odoo19 without the <data> tag the <odoo> tag does that all for us, and i saw this in an example on the github from which i learned, so now thank you for pointing out, i can remove that and remove the unnecessary thing

estate/__manifest__.py
VIWAR-ODOO: * Odoo 18 Community Edition is licensed under - `LGPL-3`
* Odoo 18 Enterprise Edition is licensed under the Odoo Enterprise Edition License v1.0 - `OEEL-1`
* Odoo Apps by Odoo SA (including the website themes) are published under the Odoo Proprietary License v1.0 - `OPL`

estate/__manifest__.py
VIWAR-ODOO: * installable defines if the module is installable or not in the odoo19 ecosystem
* and if we remove it by default the value is `True`
* and if we set the value to `False` then the module wont show a button of activate

estate/models/estate_property.py
VIWAR-ODOO: point noted

VIWAR-ODOO: * models is the module that we are importing from the odoo core and model is a specific class in that module
* the model is used to define the definitions and odoo directly maps the data to the database

* and once we create a model it auto generates some fields in the table 

>id
create_date
create_uid
write_date
write_uid


**there are three types of model**
* Model
* abstract model
* transient model

VIWAR-ODOO:
## Chapter 8 COMPLETED

estate/data/estate_property_demo.xml
```diff
@@ -0,0 +1,31 @@
...
+
+        <record id="estate_property_offer1" model="estate.property.offer">
+                    <field name="id">1</field>
+                    <field name="price">100000</field>
+                    <field name="property_id">1</field>
+        </record>
+
+
+    </data>
+</odoo>
```
mash-odoo, line 49:
```suggestion
<odoo>
    <record id="estate_property_tag_1" model="estate.property.tag">
        <field name="id">1</field>
        <field name="name">tag1</field>
    </record>
    <record id="estate_property_type_1" model="estate.property.type">
        <field name="id">1</field>
        <field name="name">type1</field>
    </record>
    <record id="estate_property_1" model="estate.property">
          <field name="id">1</field>
          <field name="name">property</field>
          <field name="tag_ids" eval="[Command.set([ref('estate_property_tag_1')])]" />
          <field name="property_type_id">1</field>
    </record>
    <record id="estate_property_offer1" model="estate.property.offer">
          <field name="id">1</field>
          <field name="price">100000</field>
          <field name="property_id">1</field>
    </record>
</odoo>
```
I don't think there would be any use of this data tag unless you want to use it with `noupdate`

estate/views/estate_property_visit_kanban_view.xml
```diff
@@ -0,0 +1,45 @@
...
+                            </div>
+                            <div>
+                               
+                                <field name="tag_ids" 
+                                 widget="many2many_tags"  options="{'color_field': 'color'}"/>
+                            </div>
+                        </t>
+                        <t t-name="new">
+
+                        </t>
```
mash-odoo: What is this used for?

mash-odoo (changes requested):
Hello!
Thank you for your work.
Please keep in mind the coding guidelines. Do not leave unnecessary lines and spaces in the file. Keep a check on a proper indentation.
It's better to review a code if the readability is good otherwise it becomes difficult to understand the code blocks!
Leave one extra line at the end of the files.

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,68 @@
...
+            sequence="10"
+            parent="settings_menu_root"
+            action="estate_property_visit_kanban_view"
+            />
+
+
+
+
+    </data>
+</odoo>
```
mash-odoo:
```suggestion
            action="estate_property_visit_kanban_view"
            />
</odoo>

```

estate/views/estate_property_tag_view.xml
```diff
@@ -0,0 +1,42 @@
...
+        </form>
+    </field>
+</record>
+
+
+
+
+
+
+</odoo>
```
mash-odoo:
```suggestion
    </record>
</odoo>

```

estate/views/estate_property_type_view.xml
```diff
@@ -0,0 +1,63 @@
...
+                                        
+                                        
+                                 
+                                
+                            </page>
+                    </notebook>
+                           
+                   
+                    
+                </sheet></group>
```
mash-odoo: Can you please clean this up?

estate/views/estate_property_view.xml
```diff
@@ -0,0 +1,140 @@
...
+                                        <field name="technician_id"/>
+                                        <field name="state"/>
+                                        <field name="estimate_cost"/>
+                                        <field name="actual_cost"/>
+                                        <field name="current_stage"/>
+                            </list>
+
+                            </field>
+
+                            </page>
```
mash-odoo: Clean up the code and follow proper indentation everywhere

mash-odoo (changes requested):
Please clean up the code.
I will resume the review after this.

estate/data/estate_property_demo.xml
VIWAR-ODOO: yes that makes sense, there is no use of data tag unless and until we need to include the noupdate

estate/views/estate_property_visit_kanban_view.xml
VIWAR-ODOO: The <t t-name=""> syntax is a template definition directive used within the Odoo QWeb template engine,
as you told to push the latest commit in the meeting there was some code i haven't cleaned up so this code is of no use and i have removed it already in my latest commit

estate/views/estate_menus.xml
VIWAR-ODOO: thank you for the suggestion

estate/views/estate_property_tag_view.xml
VIWAR-ODOO: thank you for the suggestion

estate/views/estate_property_type_view.xml
VIWAR-ODOO: i have cleaned up the code in my latest commit already

estate/views/estate_property_view.xml
VIWAR-ODOO: yes mam i have cleaned up and formatted the code already in my latest commit

PR #1170, opened 2026-02-16
renol - Technical Training

renol-odoo:
Hi, first commit here!


chapter2_estate/__manifest__.py
blse-odoo, line 1: The tutorials will keep using the same module in each chapter. If you want to, you can simply call your module `estate` instead of `chapter2_estate`

chapter1_firstmodule/__manifest__.py
blse-odoo, line 1: You don't really need to create a module for the first chapter. It was more of an overview of the module structure ;)

blse-odoo (commented):
It is a good start, congrats on your first PR!
I noticed you used very short commit messages, did you have a look at the git guidelines? Just to get familiar with the process, I think would be better to try to follow them a bit

chapter2_estate/models/estate_property.py
```diff
@@ -7,17 +9,37 @@ class EstateProperty(models.Model):
...
-        help="One of the four cardinal points (North, South, East, West).",
+        selection=[
+            ('north', 'North'), 
+            ('south', 'South'), 
+            ('east', 'East'), 
+            ('west', 'West')
+        ],
+        help="Select one of the four cardinal points (North, South, East, West).",
+    ),
+    active = fields.Boolean(default=True),
```
blse-odoo: I found the mistake in this diff that makes the `garden_orientation` field not created. The `active` field is also not created. I'm telling you how to fix it if you want, otherwise I leave you a chance to find out yourself ;)

renol-odoo: Thanks for the hint! I found it, f***ing comma

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,17 @@
...
+    price = fields.Float(required=True)
+    status = fields.Selection(
+        string="Status",
+        selection=[
+            ('accepted', 'Accepted'), 
+            ('refused', 'Refused')
+        ], 
+        copy=False,
+    )
+    partener_id = fields.Many2one("res.partner", string="Buyer", required=True)
```
blse-odoo: Just a typo here: `partener_id` should be `partner_id` ;)

estate/models/estate_property.py
blse-odoo: Could you check if this default value is 3 month after the record creation, or 3 month after the model is loaded on startup?

estate/models/estate_property_offer.py
renol-odoo: Corrected in 39015359bbe1a94f6c9db8833c6d98c74b93f15c

estate/models/estate_property.py
renol-odoo: Just checked with `datetime.now()` and indeed it keeps showing sever startup time instead of record creation time. I'll look for a fix tomorrow ;)

estate/models/estate_property_tag.py
```diff
@@ -4,5 +4,7 @@ class EstatePropertyTag(models.Model):
     _name = "estate.property.tag"
     _description = "Tag describing a particular property characteristic"
     _check_name = models.Constraint('UNIQUE(name)', 'A tag with this name already exists')
+    _order = "name asc"
 
     name = fields.Char(required=True)
+    color = fields.Integer()
```
blse-odoo: The color is not shown on the tags in the list of properties 😢

estate/views/estate_property_tag_views.xml
```diff
@@ -7,6 +7,16 @@
         <field name="view_mode">list,form</field>
     </record>
 
+    <record id="estate_property_tag_list_view" model="ir.ui.view">
+        <field name="name">estate_property_tag_list</field>
+        <field name="model">estate.property.tag</field>
+        <field name="arch" type="xml">
+            <list string="Tags" editable="bottom">
+                <field name="name"/>
+            </list>
```
blse-odoo, line 17: Could you add the color in this list?

blse-odoo (commented):
It is looking good!
There is one bug: if user tries to refuse an offer, there is a crash 😮 
(and like we said, it fails to start from a fresh db)

estate/models/estate_property_tag.py
renol-odoo: Solved in ce251438f5592b5a28187986070761459f5cf499 ;)

estate/views/estate_property_tag_views.xml
renol-odoo: Also solved in ce251438f5592b5a28187986070761459f5cf499

renol-odoo:
I failed to reproduce the bug mentioned above (crash on refused offer). Let me know if it is still there or in which conditions it is happening... 🧐

blse-odoo:
I think it is fixed by https://github.com/odoo/tutorials/pull/1170/changes/121e523457ec5fcd7963021f6851b8234b1bcdeb#diff-8000c23a2907f34f0cc387be9dc4178411fda81b674724854bd4d048f9e11b16L50

estate/models/estate_property.py
```diff
@@ -83,7 +83,14 @@ def _check_selling_above_90(self):
             compare = tools.float_utils.float_compare(record.selling_price, record.expected_price*0.9, precision_digits=2)
             if offer_accepted and compare != 1:
                 raise exceptions.ValidationError("Cannot sell bellow 90% of expected price!")
-            
+    
+    @api.ondelete(at_uninstall=False)
+    def _unlink_if_new_or_canceled(self):
+        for record in self:
+            if record.state not in ['new', 'canceled']:
+                raise exceptions.UserError('Cannot delete a property with offers or that is sold!')
```
blse-odoo, line 94: I'm not sure what is causing the following error with these steps:
- Create a property and save it
- Add an offer and save
- Cancel the property
- Delete the property (confirm delete) → This causes an error

blse-odoo (commented):
This looks good 👍

estate/models/res_users.py
```diff
@@ -0,0 +1,11 @@
+from odoo import models, fields
+
+class EstateUser(models.Model):
+    _inherit = "res.users"
```
blse-odoo, line 5: You can have a look at the other `_inherit = "res.users"` in odoo. You may want to adapt the code here to follow the naming convention for the file and the class ;)

estate/models/estate_property.py
renol-odoo: Once an offer has been linked to a property, the property cannot be deleted (even is state=canceled) because of the forgein key constraint in the DB. I've added an error message in 72992f8e7773c8e9d69962c0903dfa403948083e to avoid the crash. Another option is to delete the property AND the offer(s) linked to it

estate/models/res_users.py
renol-odoo: Yep, updated and squashed into 52efca0b9fc28485cfbd466a9aece3a2a6cb2603

awesome_owl/static/src/playground.js
```diff
@@ -1,5 +1,23 @@
...
+import { TodoList } from "./todo_list";
+
 
 export class Playground extends Component {
     static template = "awesome_owl.playground";
+    static props = []
+    static components = {Counter, Card, TodoList}
+
+    html1 = "<div class='text-primary'>some content</div>";
+    html2 = markup("<div class='text-primary'>some content</div>");
```
blse-odoo, line 13: I did not find where these `html1` and `html2` are used, is it just the remains of some experiments around markup and html injection?

Using the `markup` function like that, in a PR for odoo, should result in the security checks to raise a warning, as it allows to pass any string as the first argument, not just a literal string. So calls to this function is a potential source of injection (in case attackers control the first argument, they can add arbitrary html, for example `<script> malicious_code() </script>`). In your case this is safe, as you pass a literal string (you control the string as you wrote it).
The preferred way to call `markup` with a literal string, is to use it like this:
```js
html2 = markup`<div class='text-primary'>some content</div>`;
```

renol-odoo: Yes, that was part of [this piont](https://www.odoo.com/documentation/19.0/developer/tutorials/discover_js_framework/01_owl_components.html#using-markup-to-display-html) in chapter 1 but then was replace by other content

renol-odoo:
Thank you for the reviews and feedback! I'll now close the PR

PR #1171, opened 2026-02-16
ahham - Technical Training Estate module 

ahmedhamila:


estate/__manifest__.py
```diff
@@ -0,0 +1,14 @@
...
+    'name': 'Estate',
+    'version': '0.0',
+    'summary': 'Real Estate Management',
+    'depends': [
+        'base',
+    ],
+    'data': [
+        'security/ir.model.access.csv',
+    ],
+    'installable': True,
```
ltinel: FYI, I believe this is the default, so you don't strictly need to specify it.

estate/__manifest__.py
```diff
@@ -0,0 +1,14 @@
...
+    'application': True,
+    'author': 'ahham',
```
ltinel: In general, we put `Odoo S.A.` here, but it doesn't matter much in this context :)

ltinel (commented):
Good job, almost nothing to say 🙂

ltinel:
@ahmedhamila Regarding your commits:
- Could you create one commit per chapter (no need for a `FIX` commit, you can just squash the fix with the "main" commit).
- `FIX` commits are for bug fixes, not linting and formatting (See [1]). You can use `CLN` or `LINT` for those (but squashing the commits is better in this case).
- Could you use imperative style for the first line of your commit message? I.e. "add" instead of "added".
- The module name (following `[ADD]`) can be lower case.

[1] https://www.odoo.com/documentation/master/contributing/development/git_guidelines.html

estate/views/estate_property_views.xml
```diff
@@ -6,4 +6,77 @@
...
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <search>
+                <field name="name"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="facades"/>
+                <filter name="Available Properties" domain="['|',('state','=','new'),('state','=','offer_received')]"/>
```
ltinel: This can be a bit simpler:
```suggestion
                <filter string="Available" name="available" domain="[('state', 'in', ('new', 'offer_received'))]"/>
```

estate/models/estate_property.py
```diff
@@ -9,18 +9,18 @@ class EstateProperty(models.Model):
     _name = "estate.property"
     _description = "Real Estate Property"
 
-    name = fields.Char(string="Name", required=True)
+    name = fields.Char(string="Title", required=True)
     description = fields.Text(string="Description")
```
ltinel, line 24: Just FYI (no need to change anything): if you don't specify a `string` parameter to a field, it will use the field name as string (and I believe it correctly converts underscores to spaces and capitalizes the first letter of each word). So if the `string` parameter and the field name are the same, you don't strictly need to specify it.
```suggestion
    description = fields.Text()
```

estate/models/estate_property.py
```diff
@@ -25,3 +25,9 @@ class EstateProperty(models.Model):
                                           selection=GARDEN_ORIENTATION)
     active = fields.Boolean(string="Active", default=True)
     state = fields.Selection(string="Status", selection=PROPERTY_STATE, required=True, copy=False, default="new")
+
+    property_type_id = fields.Many2one("estate.property.type", string="Property Type")
+    buyer_id = fields.Many2one("res.partner", string="Buyer")
+    salesman_id = fields.Many2one("res.users", string="Salesman", default=lambda self: self.env.user)
+    property_tag_ids = fields.Many2many("estate.property.tag")
```
ltinel:
```suggestion
    property_tag_ids = fields.Many2many("estate.property.tag", string="Tags")
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,12 @@
+from odoo import models, fields
+
+PROPERTY_OFFER_STATE = [("accepted", "Accepted"), ("refused", "Refused")]
```
ltinel:
```suggestion
PROPERTY_OFFER_STATE = [("accepted", "Accepted"), ("refused", "Refused")]


```

estate/views/estate_property_views.xml
```diff
@@ -33,8 +33,10 @@
                             <field name="name"/>
                         </h1>
                     </div>
+                    <field name="property_tag_ids" widget="many2many_tags"/>
```
ltinel: Could you move this into the `oe_title` div?

estate/models/estate_property.py
```diff
@@ -31,3 +31,25 @@ class EstateProperty(models.Model):
     salesman_id = fields.Many2one("res.users", string="Salesman", default=lambda self: self.env.user)
     property_tag_ids = fields.Many2many("estate.property.tag")
     offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers")
+
+    total_area = fields.Integer(string="Total Area (sqm)", compute="_compute_total_area")
+    best_offer = fields.Float(string="Best Offer", compute="_compute_best_offer")
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
```
ltinel: It's slightly nicer to use a more descriptive name :) (same below)
```suggestion
        for property in self:
```

estate/models/estate_property_offer.py
```diff
@@ -10,3 +13,16 @@ class PropertyOffer(models.Model):
 
     partner_id = fields.Many2one("res.partner", required=True)
     property_id = fields.Many2one("estate.property", required=True)
+
+    validity = fields.Integer(string="Validity (days)", default=7)
+    date_deadline = fields.Date(compute="_compute_date_deadline", inverse="_inverse_date_deadline")
+
+    @api.depends("validity")
+    def _compute_date_deadline(self):
+        for record in self:
```
ltinel: Same here (and below)
```suggestion
        for offer in self:
```

estate/models/estate_property_offer.py
```diff
@@ -10,3 +13,16 @@ class PropertyOffer(models.Model):
...
+            if record.create_date:
```
ltinel: This will prevent the deadline from being set when the record is being created. Could you use a fallback (e.g. today) if `create_date` isn't set yet? Same for the inverse method.

estate/models/estate_property_offer.py
```diff
@@ -10,3 +13,16 @@ class PropertyOffer(models.Model):
 
     partner_id = fields.Many2one("res.partner", required=True)
     property_id = fields.Many2one("estate.property", required=True)
+
+    validity = fields.Integer(string="Validity (days)", default=7)
+    date_deadline = fields.Date(compute="_compute_date_deadline", inverse="_inverse_date_deadline")
+
+    @api.depends("validity")
```
ltinel: This also depends on `create_date` :)

ltinel (commented):
Great work 😄

ltinel:
FYI, if you scroll to the bottom of your PR, you'll find some Runbot links (i.e. the `ci/style` and `ci/tutorial` checks), which are failing. If you click on the links, you'll find some suggestions. Could you implement those to make the checks pass?

estate/models/estate_property.py
```diff
@@ -53,3 +54,18 @@ def _onchange_garden(self):
         else:
             self.garden_area = 0
             self.garden_orientation = False
+
+    def action_cancel_property(self):
+        for record in self:
+            if record.state == "sold":
+                raise UserError("Sold properties cannot be cancelled")
+            else:
+                record.state = "cancelled"
```
ltinel: No need for the `else` condition since there's a `raise` statement in the `if` (which shortcuts the method anyway):
```suggestion
            record.state = "cancelled"
```

estate/models/estate_property.py
```diff
@@ -53,3 +54,18 @@ def _onchange_garden(self):
...
+        return True
+    def action_sold_property(self):
+        for record in self:
+            if record.state == "cancelled":
+                raise UserError("Cancelled properties cannot be sold")
+            else:
+                record.state = "sold"
```
ltinel: Same here:
```suggestion
            record.state = "sold"
```

estate/models/estate_property_offer.py
```diff
@@ -26,3 +26,17 @@ def _inverse_date_deadline(self):
...
                 record.validity = (record.date_deadline - record.create_date.date()).days
+
+    def action_accept_offer(self):
+        for record in self:
+            record.status = "accepted"
+            record.property_id.buyer_id = record.partner_id
+            record.property_id.selling_price = record.price
+            for offer in record.property_id.offer_ids:
+                if offer != record:
+                    offer.status = "refused"
```
ltinel: Instead of this, you could refuse all offers (including the accepted one), before accepting the offer in question (as you do above):
```py
record.property_id.offer_ids.status = "refused"
```
But this line should be moved before the `record.status = "accepted"` statement.

This works because the ORM has special handling for recordsets: this line will set the status of all offers in the recordset to `refused`.

estate/models/estate_property_offer.py
```diff
@@ -26,3 +26,17 @@ def _inverse_date_deadline(self):
         for record in self:
             if record.create_date and record.date_deadline:
                 record.validity = (record.date_deadline - record.create_date.date()).days
+
+    def action_accept_offer(self):
```
ltinel, line 79: I guess you should also mark the property as having an accepted offer?
```py
record.property_id.state = "offer_accepted"
```

estate/models/estate_property_offer.py
```diff
@@ -26,3 +26,17 @@ def _inverse_date_deadline(self):
...
+            record.status = "accepted"
+            record.property_id.buyer_id = record.partner_id
+            record.property_id.selling_price = record.price
+            for offer in record.property_id.offer_ids:
+                if offer != record:
+                    offer.status = "refused"
+        return True
+    def action_refuse_offer(self):
+        for record in self:
+            record.status = "refused"
```
ltinel: Maybe we should also update the related property, in case the offer was accepted and then refused? (it's a bit of an edge case though 😅 )

estate/models/estate_property.py
```diff
@@ -56,16 +66,25 @@ def _onchange_garden(self):
...
+                property.state = "sold"
         return True
+
+    @api.constrains("selling_price")
+    def _check_selling_price_minimum_ratio(self):
+        for property in self:
+            if not property.selling_price:
+                continue
+            if float_compare(property.selling_price, property.expected_price * 0.9, precision_digits=2) < 0:
+                raise ValidationError("The selling price must be at least 90% of the expected price.")
```
ltinel: Could you move these before the action methods? See also https://www.odoo.com/documentation/master/contributing/development/coding_guidelines.html#symbols-and-conventions

estate/models/estate_property.py
```diff
@@ -56,16 +66,25 @@ def _onchange_garden(self):
...
                 raise UserError("Cancelled properties cannot be sold")
             else:
-                record.state = "sold"
+                property.state = "sold"
         return True
+
+    @api.constrains("selling_price")
+    def _check_selling_price_minimum_ratio(self):
+        for property in self:
+            if not property.selling_price:
```
ltinel: This is a bit risky, it would be safer to use `float_is_zero` (which allows for precision errors).

ltinel (commented):
Thanks, great work again :)

estate/models/estate_property_type.py
```diff
@@ -1,13 +1,37 @@
...
     _name = "estate.property.type"
     _description = "Real Estate Property Type"
+    _order = "name"
 
+    # -------------------------------------------------------------------------
+    # Field declarations
+    # -------------------------------------------------------------------------
     name = fields.Char(string="Property Type", required=True)
+    property_ids = fields.One2many("estate.property", "property_type_id", string="Properties")
+    sequence = fields.Integer('Sequence', default=1)
```
ltinel: Since you're adding a sequence, we should probably order the records according to that sequence (and use the name as fallback ordering).
FYI, in Odoo, we usually use a default sequence of 10 :)

estate/views/estate_property_type_views.xml
```diff
@@ -6,4 +6,48 @@
...
+        <field name="name">estate.property.type.form</field>
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <form>
+                <sheet>
+                    <div name="button_box" class="oe_button_box">
+                        <button class="oe_stat_button" type="action" name="%(estate.estate_property_offer_action)d" icon="fa-book">
+                            <div class="o_stat_info">
+                                <span class="o_stat_text">Offers</span>
+                            </div>
```
ltinel: I don't think your stat button shows the offer count:
```suggestion
                            <field name="offer_count" widget="statinfo" string="Offers"/>
```

estate/views/res_users_views.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo>
+
+    <record id="res_users_view_form" model="ir.ui.view">
+        <field name="name">res.users.form.inherit.estate</field>
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" position="inside">
```
ltinel: Alternative (and shorter) syntax:
```suggestion
            <notebook position="inside">
```

estate/models/estate_property_offer.py
```diff
@@ -61,10 +62,17 @@ def _inverse_date_deadline(self):
     # -------------------------------------------------------------------------
     @api.model
     def create(self, vals):
```
ltinel: Could you update the annotation to support the creation of multiple records at once?
```suggestion
    @api.model_create_multi
    def create(self, vals_list):
```

estate/models/estate_property_offer.py
```diff
@@ -67,13 +67,14 @@ def create(self, vals):
...
-        offer = super().create(vals)
-        if property_record.state == "new":
-            property_record.state = "offer_received"
-        return offer
+
+        offers = super().create(vals)
+        for offer in offers:
+            if offer.property_id.state == "new":
+                offer.property_id.state = "offer_received"
+        return offers
```
ltinel:
```suggestion
            if property_record.state == "new":
                property_record.state = "offer_received"
        return super().create(vals)
```

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
...
+    # -------------------------------------------------------------------------
+    # Private attributes
+    # -------------------------------------------------------------------------
+    _inherit = "estate.property"
+
+    # -------------------------------------------------------------------------
+    # Action methods
+    # -------------------------------------------------------------------------
+    def action_sold_property(self):
+        self.env["account.move"].create({
```
ltinel, line 16: I think you also need a journal id to create an account move

estate_account/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+{
+    'name': 'Estate Account',
+    'version': '0.0',
+    'summary': 'Real Estate Management with Account',
+    'depends': [
+        'base',
```
ltinel: By depending on `estate`, you automatically depend on `base`, so you don't need to specify it explicitly :)

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
...
+            "partner_id": self.buyer_id.id,
+            "move_type": "out_invoice",
+            "invoice_line_ids": [
+                Command.create({
+                    "name": self.name,
+                    "quantity": 0.06,
+                    "price_unit": self.selling_price,
```
ltinel: It's a bit weird to invoice fees like this 😅 Could you instead invoice a single quantity, but reduce the price?
```suggestion
                    "quantity": 1,
                    "price_unit": self.selling_price * 0.06,
```

estate/views/estate_property_views.xml
```diff
@@ -99,4 +99,30 @@
...
+        <field name="arch" type="xml">
+            <kanban default_group_by="property_type_id" records_draggable="False">
+                <field name="state"/>
+                <templates>
+                    <t t-name="card">
+                        <strong class="o_kanban_record_title">
+                            <field name="name"/>
+                        </strong>
+                        <div>
+                            Expected Price:<field name="expected_price"/>
```
ltinel: Nit: (same below)
```suggestion
                            Expected Price: <field name="expected_price"/>
```

ltinel (commented):
Good job :)

estate/models/estate_property_offer.py
ahmedhamila: Isn't that done by default when attached to a create method as shown in code below : 
```
def model(method: C) -> C:
    """ Decorate a record-style method where ``self`` is a recordset, but its
        contents is not relevant, only the model is. Such a method::

            @api.model
            def method(self, args):
                ...

    """
    if method.__name__ == 'create':
        return model_create_multi(method)  # type: ignore
    method._api_model = True  # type: ignore
    return method
```

ltinel: Indeed :) But I find it a bit cleaner to make it explicit

awesome_owl/static/src/components/card/card.js
```diff
@@ -0,0 +1,15 @@
...
+        title: { type: String },
+        slots: Object,
+    };
+    setup() {
+        this.state = useState({ isOpen: true });
+    }
+    toggleOpen() {
+        this.state.isOpen = !this.state.isOpen;
+    }
+    static template = "awesome_owl.card";
```
ltinel: Could you move this above `static props = ...`? Same for the other components

awesome_owl/static/src/components/card/card.js
```diff
@@ -0,0 +1,15 @@
+import { Component, useState } from "@odoo/owl";
+
+export class Card extends Component {
+    static props = {
+        title: { type: String },
```
ltinel: Nit:
```suggestion
        title: String,
```

awesome_owl/static/src/components/card/card.js
```diff
@@ -0,0 +1,15 @@
...
+        slots: Object,
```
ltinel, line 7: Just FYI, you can give this a `shape` if you know what it will look like. See also https://github.com/odoo/owl/blob/master/doc/reference/props.md#props-validation

awesome_owl/static/src/components/card/card.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+    <t t-name="awesome_owl.card">
```
ltinel: Nit: (see https://www.odoo.com/documentation/master/developer/reference/frontend/owl_components.html#using-owl-components)
```suggestion
    <t t-name="awesome_owl.Card">
```

awesome_owl/static/src/components/counter/counter.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+
+    <t t-name="awesome_owl.counter">
```
ltinel: Nit:
```suggestion
    <t t-name="awesome_owl.Counter">
```

awesome_owl/static/src/components/todoo/todo_item/todo_item.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+    <t t-name="awesome_owl.todo_item">
```
ltinel: Nit:
```suggestion
    <t t-name="awesome_owl.TodoItem">
```

awesome_owl/static/src/components/todoo/todo_item/todo_item.xml
```diff
@@ -0,0 +1,15 @@
...
+        <div>
```
ltinel: Is this `div` really needed?

awesome_owl/static/src/components/counter/counter.xml
```diff
@@ -0,0 +1,12 @@
...
+        <div class="d-flex gap-2 border border-dark align-items-center p-2">
+            <span>Counter: <t t-esc="state.value"/>
+            </span>
+            <button t-on-click="increment" class="btn btn-primary" type="button">Increment</button>
```
ltinel: `type` shouldn't be needed :)

awesome_owl/static/src/components/todoo/todo_list/todo_list.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+    <t t-name="awesome_owl.todo_list">
```
ltinel: Nit:
```suggestion
    <t t-name="awesome_owl.TodoList">
```

awesome_owl/static/src/components/todoo/todo_list/todo_list.js
```diff
@@ -0,0 +1,37 @@
...
+                description: event.target.value,
+                isCompleted: false,
+            });
+            event.target.value = "";
+        }
+    }
+    toggleState(event, id) {
+        console.log(event);
+        this.todos.find((todo) => todo.id === id).isCompleted =
+            event.target.checked;
```
ltinel: Instead of relying on the event, you could also simply toggle `isCompleted`, i.e. `todo.isCompleted = !todo.isCompleted`

awesome_owl/static/src/components/todoo/todo_list/todo_list.js
```diff
@@ -0,0 +1,37 @@
...
+import { TodoItem } from "../todo_item/todo_item";
+import { useAutofocus } from "../../../utils";
+export class TodoList extends Component {
+    static props = {};
+    static template = "awesome_owl.todo_list";
+    setup() {
+        this.todos = useState([]);
+        useAutofocus("todo_input");
+        this.toggleState = this.toggleState.bind(this);
+        this.removeTodo = this.removeTodo.bind(this);
```
ltinel: I believe you can bind directly in the view, which is slightly more elegant IMO. E.g. `toggleState.bind="toggleState`

awesome_owl/static/src/components/todoo/todo_list/todo_list.js
```diff
@@ -0,0 +1,37 @@
...
+    }
+    addTodo(event) {
+        if (event.keyCode === 13 && event.target.value.length) {
+            this.todos.push({
+                id: this.todos.length
+                    ? this.todos[this.todos.length - 1].id + 1
+                    : 1,
```
ltinel, line 16: This should be fine, but is slightly risky IMO. What if we want to allow reordering the TODOs? I'd opt for a simple counter in this case (which also has limitations, but should suffice for a simple TODO list 🙂 )

awesome_owl/static/src/components/todoo/todo_list/todo_list.js
```diff
@@ -0,0 +1,37 @@
...
+                description: event.target.value,
+                isCompleted: false,
+            });
+            event.target.value = "";
+        }
+    }
+    toggleState(event, id) {
+        console.log(event);
+        this.todos.find((todo) => todo.id === id).isCompleted =
```
ltinel: This should work but we might want to make sure that the todo was actually found before accessing it.

awesome_owl/static/src/components/counter/counter.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+
+    <t t-name="awesome_owl.counter">
+        <div class="d-flex gap-2 border border-dark align-items-center p-2">
+            <span>Counter: <t t-esc="state.value"/>
```
ltinel: `t-esc` is deprecated. Could you use `t-out` instead?

awesome_owl/static/src/playground.js
```diff
@@ -1,5 +1,24 @@
...
+    static props = {};
     static template = "awesome_owl.playground";
+
+    setup() {
+        this.card1 = {
+            title: "Card 1",
+        };
+        this.card2 = {
+            title: "Card 2",
+        };
```
ltinel: Feel free to inline those in the view :)

awesome_owl/__manifest__.py
```diff
@@ -34,6 +34,7 @@
             'web/static/lib/bootstrap/scss/_variables.scss',
             'web/static/lib/bootstrap/scss/_maps.scss',
             ('include', 'web._assets_bootstrap'),
+            ('include', 'web._assets_bootstrap_backend'),
```
ltinel, line 37: Shouldn't be needed normally, but I saw that the `primary-button` style was missing from the assets, so I guess it makes sense.

ltinel (commented):
Thanks, great work again :)

awesome_owl/static/src/components/todoo/todo_list/todo_list.js
```diff
@@ -0,0 +1,37 @@
...
+        }
+    }
+
+    removeTodo(id) {
+        const index = this.todos.findIndex((todo) => todo.id === id);
+        if (index !== -1) {
+            this.todos.splice(index, 1);
+        }
+    }
+    static components = { TodoItem };
```
ltinel: Could you move this with the other static properties? (same in `playground.js`)

awesome_owl/static/src/playground.js
```diff
@@ -1,5 +1,17 @@
-import { Component } from "@odoo/owl";
-
+import { Component, useState } from "@odoo/owl";
+import { Counter } from "./components/counter/counter";
+import { Card } from "./components/card/card";
+import { TodoList } from "./components/todoo/todo_list/todo_list";
 export class Playground extends Component {
-    static template = "awesome_owl.playground";
+    static template = "awesome_owl.Playground";
+    static props = {};
```
ltinel: No need for this if there are no props :)
```suggestion
```

ltinel:
Feel free to mark my comments as "resolved" once you've addressed them :)

PR #1174, opened 2026-02-16
Tutorials

coder-odoo:


estate/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from . import estate_property
```
AlessandroLupo: Remember to add an empty line at the end of each file ;)

estate/models/estate_property.py
```diff
@@ -0,0 +1,26 @@
...
+    bedrooms = fields.Integer()
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        string='Type',
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')],
+        help="Type is used to choose the orientation")
```
AlessandroLupo, line 28: I suppose that you followed [this example](https://github.com/odoo/odoo/blob/b0e0035b585f976e912e97e7f95f66b525bc8e43/addons/crm/report/crm_activity_report.py#L31-L34). Just to be clear: 'Type' in `string='Type'` is not a special keyword. You can use a more specific string, for example "Garden Orientation". 👍

estate/__manifest__.py
```diff
@@ -0,0 +1,9 @@
+{
+    'name': "Real Estate",
+    'depends': ['base'],
+    'author': "Odoo",
+    'category': 'Category',
+    'description': """
+    A app for real estate
+    """
+}
```
AlessandroLupo: You could add a couple of extra fields to the manifest, try for example to:
- add a `license` field, to avoid the warning "Missing 'license' key in manifest for 'estate', defaulting to LGPL-3"
- add a `application` field, to make your "Estate" app appear among the other Odoo apps

AlessandroLupo (changes requested):
Nice work! :D I added a couple of comments. Please make sure that your commit messages follow the [guidelines](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html). 👍

AlessandroLupo (commented):
Thanks for the changes, it looks fine now! 👍

estate/views/estate_list_views.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="estate_property_view_list" model="ir.ui.view">
+        <field name="name">estate.property.list</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <list string="Properties">
+                <field name="name"/>
+                <field name="tag_ids"/>
```
AlessandroLupo: You could use `widget="many2many_tags"` to have the actual tag names displayed on each line ;)

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,1
+estate.access_property_type,access_property_type,estate.model_estate_property_type,base.group_user,1,1,1,1
+estate.access_property_tag,access_property_tag,estate.model_estate_property_tag,base.group_user,1,1,1,1
+estate.access_property_offer,access_property_offer,estate.model_estate_property_offer,base.group_user,1,1,1,1
```
AlessandroLupo: No empty line at the end of file 😄

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,10 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import fields, models
+
+
+class EstatePropertyType(models.Model):
+    _name = "estate.property.type"
+    _description = "A table for estate properties types"
```
AlessandroLupo: Nitpick: briefly explore the Odoo codebase searching for `_descrption`. You will see that we tend to use a more concise style. I would use something like "Estate Property Type".

AlessandroLupo (commented):
Good job, it's nice that you got a green runbot. 👍 I added a couple of minor comments.
As a git exercise, you could try to do an interactive rebase and leave only one commit per chapter (you could also add the chapter number to the commit messages). To achieve this, you could just squash the commit fixing the runbot with the previous commit. Let me know if you need support with git!

estate/views/estate_list_views.xml
```diff
@@ -4,15 +4,15 @@
         <field name="name">estate.property.list</field>
         <field name="model">estate.property</field>
         <field name="arch" type="xml">
-            <list string="Properties">
+            <list string="Properties" decoration-success="state == 'offer_received' or state == 'offer_accepted'" decoration-bf="state == 'offer_accepted'" decoration-muted="state == 'sold'">
```
AlessandroLupo: You could also use  `decoration-success="state in ('offer_received', 'offer_accepted')"`

estate/models/res_users.py
AlessandroLupo, line 1: Try to search the Odoo codebase for the line `_inherit = 'res.users'`. You will see what is the common style for inherited classes. The class should be named "ResUsers" and the file should be named "res_users". ;)

AlessandroLupo (commented):
Good job with the git exercise! 👍

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import models
+
+
+class EstatePropertyInherited(models.Model):
```
AlessandroLupo: Nitpick: the conventional name here would be `EstateProperty` (no need to add "Inherited") ;)

AlessandroLupo (commented):
Good job! 👍

awesome_owl/static/src/card/card.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+
+    <t t-name="awesome_owl.card">
+        <div class="card d-inline-block m-2" style="width: 18rem;">
+            <div class="card-body">
+                <h5 class="card-title"><t t-esc="props.title"/></h5>
+                <p class="card-text">
+                    <t t-out="props.content"/>
```
AlessandroLupo: You could also do:
```
<h5 class="card-title" t-out="props.title"/>
                <p class="card-text" t-out="props.content"/>
```

awesome_owl/static/src/counter/counter.js
```diff
@@ -0,0 +1,19 @@
...
+
+    setup() {
+        this.state = useState({ value: 0 });
+    }
+
+    increment() {
+        this.state.value++;
+        if (this.props.onChange) {
+            this.props.onChange();
+        }
```
AlessandroLupo, line 17: Here you could use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining): `this.props.onChange?.()`.

awesome_owl/static/src/todo/todo_list.js
```diff
@@ -0,0 +1,24 @@
...
+export class TodoList extends Component {
+    static template = "awesome_owl.todo_list";
+
+    setup() {
+        this.todos = useState([]);
+        this.counter = 0;
+    }
+
+    addTodo(ev) {
+        if (ev.keyCode === 13 && ev.target.value != "") {
```
AlessandroLupo: You could also use `ev.key === "Enter"` (see [here](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key)). `keyCode` is deprecated

awesome_owl/static/src/todo/todo_list.js
```diff
@@ -0,0 +1,24 @@
+import { Component, useState } from "@odoo/owl";
+import { TodoItem } from "./todo_item";
+
+export class TodoList extends Component {
+    static template = "awesome_owl.todo_list";
+
+    setup() {
+        this.todos = useState([]);
```
AlessandroLupo: Here I would do `this.state = useState({todos: []})`, and later access the todos array using `this.state.todos`, so it is clear that todos is in the component state.

AlessandroLupo (commented):
Nice work 👍

awesome_owl/static/src/card/card.js
```diff
@@ -3,7 +3,8 @@ import { Component } from "@odoo/owl";
 export class Card extends Component {
     static template = "awesome_owl.card";
     static props = {
-        title: String,
-        content: String,
+        title: { type: String },
```
AlessandroLupo, line 6: Just to be clear, `title: String` was fine too (it is assumed implicitly that you are defining the type), but feel free to use the style you prefer 👍

awesome_owl/static/src/todo/todo_list.js
```diff
@@ -20,5 +24,23 @@ export class TodoList extends Component {
...
+            }
+        }
+    }
+
+    removeTodo = (todoId) => {
+        const index = this.todos.findIndex((t) => t.id === todoId);
+        if (index > 0) {
+            this.todos.splice(index, 1);
+        }
+    }
```
AlessandroLupo, line 42: Will this work if we try to remove the first element in the todo list (index 0)?

AlessandroLupo (commented):
You're doing well! Just two comments

coder-odoo:
Thanks for your comments. I'm on it

awesome_owl/static/src/counter/counter.js
```diff
@@ -0,0 +1,17 @@
...
+        onChange: { type: Function, optional: true },
+    };
+
+    setup() {
+        this.state = useState({ value: 0 });
+    }
+
+    increment() {
+        this.state.value++;
+        this.props.onChange?.(this.props.onChange());
```
AlessandroLupo: Probably what you want to do here is `this.props.onChange?.()`, which means "if `onChange` exists, call it".
The way you coded it, `this.props.onChange()` gets evaluated once, and then passed as an argument to a second call to `this.props.onChange()`  (that's why the counter increases by two times instead of one). Moreover the purpose of `?.()` is undermined, because if `this.props.onChange` does not exist, you would still get an error. ;)

awesome_owl/static/src/todo/todo_list.js
```diff
@@ -0,0 +1,44 @@
...
+export class TodoList extends Component {
+    static template = "awesome_owl.todo_list";
+
+    setup() {
+        this.state = useState({todos: []});
+        this.counter = 0;
+    }
+
+    focusInput() {
+        this.myRef.el.focus();
```
AlessandroLupo: Something seems incomplete here, right? `myRef` is not defined and `focusInput` is never called?
See how it is done [here](https://github.com/odoo/odoo/blob/5143840911e1f05be3f5f5c7fa786f05e4966312/addons/barcodes/static/src/components/manual_barcode.js#L21-L23), for example ;)

AlessandroLupo (commented):
You're doing good, I left couple of comments. 👍  Don't hesitate to ask if something is not clear!

awesome_dashboard/static/src/services/statistics_service.js
```diff
@@ -0,0 +1,17 @@
+import { registry } from "@web/core/registry";
+import { rpc } from "@web/core/network/rpc";
+import { memoize } from "@web/core/utils/functions";
+
+const loadStatistics = memoize(async () => {
+    return rpc("/awesome_dashboard/statistics", {});
+});
```
AlessandroLupo: Just a small remark. The function `return rpc("/awesome_dashboard/statistics", {});` does nothing asyncronous. It actually just returns the promise from `rpc` without awaiting for it. So tecnically the keyword `async` is not needed. Anyway since it does no harm and makes the code clearer, someone may prefer to add it. So it is just a matter of style, up to you 👍 

Also, `return` can be made implicit:
```
const loadStatistics = memoize(async () => rpc("/awesome_dashboard/statistics", {}));
```

awesome_dashboard/static/src/dashboard.scss
AlessandroLupo, line 1: You could have `.o_dashboard_row`, `.o_header_dashboard_item`, and `.o_average_quantity` nested in `.o_dashboard`

AlessandroLupo (commented):
The dashboard is looking nice! 👍 Two minor comments

awesome_dashboard/static/src/dashboard/Dashboard/dashboard.js
```diff
@@ -0,0 +1,69 @@
...
+class AwesomeDashboard extends Component {
+    static template = "awesome_dashboard.AwesomeDashboard";
+    static components = { Layout, DashboardItem };
+
+    setup() {
+        this.allItems = registry.category("awesome_dashboard").getAll();
+        this.actionService = useService("action");
+        this.dialogService = useService("dialog");
+        this.statisticsService = useService("awesome_dashboard.statistics");
+        this.statistics = useState(this.statisticsService.statistics);
```
AlessandroLupo: Why do you need `useState` here?

awesome_dashboard/static/src/dashboard/Dashboard/dashboard.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    loadRemovedItemIds() {
+        try {
+            const value = localStorage.getItem(STORAGE_KEY);
+            const parsed = value ? JSON.parse(value) : [];
+            return Array.isArray(parsed) ? parsed : [];
+        } catch {
+            return [];
+        }
```
AlessandroLupo, line 34: What kind of error do you expect to catch here? I see that the case where the key does not exists is already correctly covered in the `try` block. In general there may be errors that we may want to display to the user, instead of just silently failing. For example, I see from the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage#exceptions) that accessing `localStorage` could throw a `SecurityError` if the user configured the browser to discard persisting data. In this case, we may want to show a notification to inform the user. (This is just to let you know, you don't have to do it 😄 )

awesome_dashboard/static/src/dashboard/Dashboard/dashboard.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    saveRemovedItemIds(removedItemIds) {
+        localStorage.setItem(STORAGE_KEY, JSON.stringify(removedItemIds));
```
AlessandroLupo: If the idea is to catch every error coming from `localStorage`, you may want a try/catch also here

awesome_dashboard/static/src/dashboard/Dashboard/dashboard.js
```diff
@@ -0,0 +1,69 @@
...
+        this.state.removedItemIds = removedItemIds;
+    }
+
+    get items() {
+        const removedIds = new Set(this.state.removedItemIds);
```
AlessandroLupo, line 47: Good idea to use a `Set` to have a faster search. Two minor comments:
1. Here we have a maximum of 6 elements, so no big difference between O(1) or O(n) searches
2. You could keep `this.state.removedItemIds` always as a set, so we don't need to create a set every time we access `items`

awesome_dashboard/static/src/dashboard/Dashboard/dashboard.js
```diff
@@ -0,0 +1,69 @@
...
+            res_model: "crm.lead",
+            views: [[false, "list"], [false, "form"]],
+        });
+    }
+
+    openSettings() {
+        this.dialogService.add(DashboardSettings, {
+            title: "Dashboard settings",
+            removedItemIds: this.state.removedItemIds,
+            applyConfiguration: (removedItemIds) => this.saveRemovedItemIds(removedItemIds),
```
AlessandroLupo, line 68: Or `applyConfiguration: this.staveRemovedItemIds.bind(this)`

awesome_dashboard/static/src/dashboard/services/statistics_service.js
```diff
@@ -0,0 +1,50 @@
...
+    },
+});
+
+let refreshToken = 0;
+
+const fetchStatistics = memoize(async () => rpc("/awesome_dashboard/statistics", {}));
+
+async function loadStatistics() {
+    const result = await fetchStatistics(refreshToken);
+    statistics.average_quantity = result.average_quantity;
```
AlessandroLupo: You could also use something like [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)(statistics, result)

awesome_dashboard/static/src/dashboard/services/statistics_service.js
```diff
@@ -0,0 +1,50 @@
...
+    nb_new_orders: 0,
+    total_amount: 0,
+    orders_by_size: {
+        s: 0,
+        m: 0,
+        xl: 0,
+    },
+});
+
+let refreshToken = 0;
```
AlessandroLupo, line 22: Why do we need the refreshToken?

AlessandroLupo (commented):
Good job! I left some comments but they all are minor observations / questions

estate/models/estate_property.py
```diff
@@ -63,11 +63,13 @@ def _onchange_partner_id(self):
...
         for record in self:
-            if record.state != 'canceled':
-                record.state = 'sold'
-            else:
+            if record.state != 'offer_accepted':
+                error_msg = "You cannot sell a property that has not an accepted offer."
+                raise UserError(error_msg)
+            if record.state == 'canceled':
                 error_msg = "You cannot sell a canceled property."
                 raise UserError(error_msg)
```
AlessandroLupo, line 68: Note that this section of code `if record.state == 'cancelled'` will never be reached, because if state is cancelled the previous `if` condition is true and an error is raised there

AlessandroLupo (commented):
Good job with the tests 👍 Have you tested them on your local Odoo instance? Don't hesitate to ask if you need help with that.

PR #1176, opened 2026-02-16
luleg - Technical Training - Creation of the Real Estate Module

Astremy:
Here is my progress on the [Technical training](https://www.odoo.com/documentation/19.0/developer/tutorials/server_framework_101.html)

Don't hesitate to give me feedbacks about what I could improve.


AlessandroLupo (commented):
Nice work! 👍 

Ideally, we should have a single commit per tutorial chapter. But I see that you introduced some code in advance (for example, `EstateTagCategory` is introduced in chapter 7, right?). Is there a specific reason for this? As a git exercise, you could try to move some of this content to the right commit when you reach the correct chapter. 

It is very good that you fixed the styling error. You could try to do it in the already existing commits instead of adding a new one.

Depending on your confidence with git, this may be boring or may be very interesting. Feel free to ask for help if you need :)

Astremy:
Thank you for the feedbacks!
The thing is, I just tried to make the structure of the model as close as possible to [first image of chapter two](https://www.odoo.com/documentation/19.0/developer/tutorials/server_framework_101/02_newapp.html) (I didn't evn notice the other images before now, so... I could put even more fields into my model). I didn't look at the tutorial for that, I just went and dug in odoo's code. In the same way, it seems like I also did the chapter 10 part before finishing chapter 2...

I kinda just go and find what I need for what I want to do and do it ^^'

Well, about moving parts of the code... it would probably be pretty annoying, I know a bit of git but in no way in hell do I already know how to do that in particular (moving code around commits) (especially now that it's pushed and assuming I don't force push, which I would love not to do).

AlessandroLupo:
Good job :+1: 

I see your runbot is red, please try to fix the problems and make it green. It's just a styling error and a problem about an invisible element, feel free to ask for help :)

estate/models/estate_property_offer.py
```diff
@@ -9,8 +9,9 @@ class PropertyOffer(models.Model):
 
     property_id = fields.Many2one("estate.property", string="Property", required=True)
     partner_id = fields.Many2one("res.partner", string="Partner", index=True, required=True)
+    property_type_id = fields.Many2one("estate.property.type", string="Property Type", related="property_id.property_type_id")
 
-    # Beginning of the deadline part
+    # Deadline Part
```
AlessandroLupo, line 14: It is a good practice to leave comments on your code. But consider cleaning them a bit. In this case, you have "deadline part", then a bit after you have "beginning of the deadline part" and then "end of the deadline part". In most of the cases, commenting only the beginning of a block is enough :)

estate/security/ir.model.access.csv
```diff
@@ -2,5 +2,4 @@ id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
 access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
 access_estate_property_type,access_estate_property_type,model_estate_property_type,base.group_user,1,1,1,1
 access_estate_property_offer,access_estate_property_offer,model_estate_property_offer,base.group_user,1,1,1,1
-access_estate_tag,access_estate_tag,model_estate_tag,base.group_user,1,1,1,1
-access_estate_tag_category,access_estate_tag_category,model_estate_tag_category,base.group_user,1,1,1,1
\ No newline at end of file
+access_estate_tag,access_estate_tag,model_estate_tag,base.group_user,1,1,1,1
```
AlessandroLupo: Empty line is missing at the end of file ;)

estate/views/estate_property_views.xml
```diff
@@ -63,7 +64,18 @@
                             </group>
                         </page>
                         <page string="Offers">
-                            <field name="offer_ids"/>
+                            <field name="offer_ids" readonly="stage in ['sold','cancelled']">
+                                <list string="Offers" editable="bottom" default_order="translated_price desc" decoration-success="status=='accepted'" decoration-danger="status=='refused'" sample="1">
+                                    <field name="translated_price" options="{'currency_field': 'property_currency_id'}" />
+                                    <field name="price" options="{'currency_field': 'currency_id'}" />
```
AlessandroLupo, line 70: I found a bit confusing that both "Price" and "Original price" are shown here. Also, "price" is shown without currency symbol. Maybe consider showing only the original price?

AlessandroLupo (commented):
Nice work with the currency conversion part. Just a few minor comments 👍

estate/views/estate_property_views.xml
Astremy: What's shown on top is the best offer in the same currency as the expected sell price, so, for commercials to be able to know which offer is the best one without having to do the conversion themselves, I need to show the translated price next to the offered price.

You could say that you don't need it since the offers are ordered related to the translated price, but then how do you know how much less would be the second best offer?

AlessandroLupo: Oh okay now I see your point. You could use something like "price in the local currency" just to make it clearer, but that's up to you :+1:

estate/models/estate_property.py
```diff
@@ -107,6 +107,12 @@ def action_set_as_sold(self):
             property.stage = "sold"
         return True
 
+    @api.ondelete(at_uninstall=False)
+    def _unlink_except_if_advanced_stage(self):
+        for property in self:
+            if not property.stage in ["new", "cancelled"]:
```
AlessandroLupo: or `if property.state not in [....]`

estate/models/res_users.py
```diff
@@ -0,0 +1,8 @@
+from odoo import fields, models
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    # Domain doesn't work
+    property_ids = fields.One2many("estate.property", "seller_id", string="Properties List",
+    domain="[('available_from', '<=', 'today'), ('stage', '!=', 'cancelled'), ('stage', '!=', 'sold')]")
```
AlessandroLupo: Domain doesn't work because it is supposed to be a list, not a string :D
Also consider doing `('stage', 'not in', ('cancelled', 'sold'))`

estate/views/res_user_views.xml
```diff
@@ -0,0 +1,16 @@
...
+<odoo>
+    <record id="res_users_view_form" model="ir.ui.view">
+        <field name="name">res.users.view.form.inherit.property</field>
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" position="inside">
+                <page string="Real Estate Properties">
+                    <!-- The filter doesn't work, I'm too sick for this. To fix later -->
+                    <field name="property_ids" readonly="1" domain="[('available_from', '&gt;=', 'today'), ('stage', '!=', 'cancelled'), ('stage', '!=', 'sold')]"/>
```
AlessandroLupo: I think you won't need the `domain` here anymore after having fixed the one in `ResUsers`

AlessandroLupo (commented):
Good job. You're not the first one having a problem with the one2many domain in ResUsers, see my comment :D

estate/models/res_users.py
Astremy: Fixed! Thank you!

estate/models/res_users.py
```diff
@@ -0,0 +1,9 @@
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    # Domain doesn't work
```
AlessandroLupo, line 7: It does work now, right?

estate/views/estate_property_views.xml
```diff
@@ -106,9 +106,35 @@
...
+                    <t t-name="card">
+                        <div class="row">
+                            <div class="col-8">
+                                <field class="fw-bolder" name="name"/>
+                            </div>
+                            <div>
+                                Expected Price: 
+                                <field name="expecting_price"/>
+                            </div>
+                            <div t-if="best_offer == 0">
```
AlessandroLupo, line 124: Why `t-if="best_offer == 0"` ?

AlessandroLupo (commented):
Nice work! I had just two small questions

estate/models/res_users.py
Astremy: Oh yeah

awesome_dashboard/static/src/dashboard/core/statistics_service.js
```diff
@@ -0,0 +1,22 @@
...
+
+        setInterval(async () => {
+            const getStatistics = await rpc("/awesome_dashboard/statistics");
+            data.statistics = getStatistics;
+            console.log(data)
+        }, 1_000*10)
+
+        return {
+            data: data
+        }
```
AlessandroLupo, line 16: This is equivalent to `return { data }`

AlessandroLupo, line 16: It could be simplified a bit, for example
```javascript
async start() {
    const statistics = reactive({});
    const updateStatistics = async () => {
        const newStatistics = await rpc("/awesome_dashboard/statistics");
        Object.assign(statistics, newStatistics);
    }
    setInterval(updateStatistics, 1_000*10);
    await updateStatistics();
    return statistics
}
```

awesome_dashboard/static/src/dashboard/pie_chart/pieChart.js
```diff
@@ -0,0 +1,52 @@
...
+        onWillStart(() => loadJS(["/web/static/lib/Chart/Chart.js"]));
+
+        useEffect(() => this.renderChart());
+        onWillUnmount(this.onWillUnmount);
+    }
+
+    onWillUnmount() {
+        if (this.chart) {
+            this.chart.destroy();
+        }
```
AlessandroLupo, line 28: Also possible:
```javascript
this.chart?.destroy();
```

awesome_owl/static/src/counter/counter.js
```diff
@@ -0,0 +1,18 @@
...
+        onChange: { type: Function, optional: true }
+    }
+
+    setup() {
+        this.state = useState({ value: 0 });
+    }
+
+    increment() {
+        this.state.value++;
+        if (this.props.onChange) this.props.onChange();
```
AlessandroLupo, line 16: Also:
```javascript
this.props.onchange?.();
```

awesome_owl/static/src/playground.js
```diff
@@ -1,5 +1,21 @@
...
+import { Card } from "./card/card";
+import { TodoList } from "./todoList/todoList";
 
 export class Playground extends Component {
     static template = "awesome_owl.playground";
+
+    value2 = markup("<a href='http://odoo.com'>Test</a>");
+
+    setup() {
+        this.total = useState({ value: 0 });
```
AlessandroLupo, line 12: Just a minor comment about naming: maybe here I would have used 
```javascript
this.state = useState({ total: 0 });
```

awesome_owl/static/src/playground.js
```diff
@@ -1,5 +1,21 @@
...
+    }
+
+    incrementSum() {
+        console.log("aaa")
+        this.total.value++;
+    }
+
+    static components = { Counter, Card, TodoList };
```
AlessandroLupo, line 20: It gives the same result, but the `components` declaration usually are put on top

AlessandroLupo (commented):
Nice work, I left some minor comments. Make sure to remove your `console.log` and to add empty lines at the end of every file! 👍

awesome_dashboard/static/src/dashboard/core/statistics_service.js
Astremy: We tried returning a reactive object, but it didn't seems to work when importing. You have to go through another layer, so sadly you can't just return statistics. Also I made a few changes not to make requests if you don't need to (will come in a future commit).
Are you sure using `Object.assign(statistics, newstatistics)` would work? It may help removing one layer.

awesome_owl/static/src/counter/counter.js
Astremy: That is cursed code, that's some very specific syntax and I feel like it would be better to keep it like that as it is more readable? idk, both solutions seems okay.

awesome_dashboard/static/src/dashboard/pie_chart/pieChart.js
Astremy: (also I stole the code from [here](https://github.com/odoo/odoo/blob/16.0/addons/web/static/src/views/graph/graph_renderer.js#L63) so that code review is on odoo's code :p

awesome_dashboard/static/src/dashboard/core/statistics_service.js
AlessandroLupo: I believe that when we tried we were dealing with a complex object and I didn't quite figure out the naming. That's why I came up with this comment, to propose a simpler object. I tried it and it works fine, provided that you adapt also the AwesomeDashboard component. I confirm it works fine with this setup method:
```javascript
setup() {
        this.action = useService("action");
        this.statisticsService = useService("statistics");
        this.statisticsData = useState(this.statisticsService);
        const updateData = () => {
            this.result = this.statisticsData
            this.data = Object.entries(this.statisticsData.orders_by_size).map(([key, value]) => ({
                label: key, 
                value: value
            }));
        }
        onWillStart(updateData)
        useEffect(updateData, () => [this.statisticsData.orders_by_size])
    }
```

AlessandroLupo (commented):
Good work 👍  I answered your comment on the statistics service

PR #1177, opened 2026-02-16
[ADD] estate: property model with fields, views, and menu access

karimgamaleldin:
- Add estate_property model with fields
- Add active field with default True
- Set date_availability default to 3 months from today
- Create list, form, and search views for estate.property
- Add search filters and group by postcode
- Configure access rights in ir.model.access.csv


estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
...
+    return three_months_later
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate Property Model"
+ 
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(copy=False, default=get_date_in_3_months())
```
blse-odoo: The `default` argument will be the result of `get_date_in_3_months` when this model is loaded, not when records are created: If you start your database on the 01/01, `get_date_in_3_months()` will return the 01/04. But this value will always be used, if a user create a new record a few days later for example on the 17/01, the default will still be `01/04`. You can use a lamdba (like for `salesperson` in the next commit) to compute it when the record is created

estate/models/__init__.py
```diff
@@ -1,2 +1,4 @@
 from . import estate_property
-# from . import estate_property_type
\ No newline at end of file
+from . import estate_property_type
+from . import estate_property_tag
+from . import estate_property_offer
```
blse-odoo: Some of the files do not finish by a newline character. `git` is a bit limited about that, and this may cause diffs to be slightly larger than they need to be.
Could you ensure that all files ends with a newline?

blse-odoo (commented):
This looks good. Could you make one commit per chapter, to make it easier to review?

estate/models/estate_property_offer.py
```diff
@@ -22,4 +22,17 @@ def _inverse_date_deadline(self):
             start = fields.Date.to_date(offer.create_date) or fields.Date.today()
             if offer.date_deadline and start:
                 delta = offer.date_deadline - start
-                offer.validity = delta.days
\ No newline at end of file
+                offer.validity = delta.days
+
+    def accept_offer(self):
```
blse-odoo, line 41: Playing around a bit, it looks like we can accept many offers. We should only be able to accept one

estate/models/estate_property.py
```diff
@@ -78,6 +87,12 @@ def _onchange_garden(self):
             self.garden_orientation = False
 
 
+    @api.constrains("selling_price", "expected_price")
+    def _check_selling_price(self):
```
blse-odoo: The name `_check_selling_price` is used here to define the constraint between the 2 prices, but it is also used to the sql constraint above. That is a bit confusing, could you change one of the name? :)

blse-odoo (commented):
👍 
I noticed you always used the `[ADD]` tag in the commit, you can check out the git guidelines to see the other ones (`ADD` should be used for commits that new modules)

estate/models/estate_property.py
karimgamaleldin: Done. Let me know if this looks better

estate/models/__init__.py
karimgamaleldin: Cleaned! All files now end with a newline.

estate/models/estate_property.py
karimgamaleldin: I missed that conflict! I've renamed the Python constraint method so it's distinct from the SQL constraint name. Thanks for spotting that.

estate/models/estate_property_offer.py
karimgamaleldin: Thank you for testing that out! You are absolutely right. I have updated the accept_offer method to raise a UserError if the property already has an accepted offer. Additionally, as part of the current tutorial chapter, I have modified the views to hide the 'Accept' and 'Refuse' buttons once an offer is accepted, which should prevent this situation from occurring in the UI.

karimgamaleldin:
> 👍 I noticed you always used the `[ADD]` tag in the commit, you can check out the git guidelines to see the other ones (`ADD` should be used for commits that new modules)

Thanks for pointing that out. I’ve updated the commit tags accordingly. I initially thought [ADD] was equivalent to feat: in Conventional Commits, but I understand now that [ADD] should only be used when introducing new modules. I’ll follow the git guidelines more closely going forward.

blse-odoo:
I can't try the current version, there is some error when I start the server with an empty database. Could you have a look?

karimgamaleldin:
> I can't try the current version, there is some error when I start the server with an empty database. Could you have a look?

I created a completely new database from scratch and started the server, and I’m not seeing any errors on my side. Everything starts correctly.

Could you please send me the exact error message you're getting when starting with an empty database? That will help me investigate further.

blse-odoo:
The runbot encounters the same error, you can look at the logs for the "ci/tutorial" (or we can have a look together tomorrow)

estate/views/estate_property_views.xml
```diff
@@ -80,15 +82,19 @@
         <field name="name">estate.property.list</field>
         <field name="model">estate.property</field>
         <field name="arch" type="xml">
-            <list string="Properties">
+            <list string="Properties"
+                decoration-success="state in ('offer_received', 'offer_accepted')"
```
blse-odoo, line 86: I noticed that we do not get in the state `offer_received` after we have offers, but only after we refused one. Do you think you could improve that? ;)

karimgamaleldin: Thanks for pointing that out. I was already working on this, and I’ve just pushed the update.

estate/models/estate_sales_person.py
```diff
@@ -0,0 +1,7 @@
+from odoo import fields, models
+
+
+class EstateSalesPerson(models.Model):
```
blse-odoo: We usually re-use the same name for the class and the file (you can search for other `_inherit = "res.users"` in the code)

karimgamaleldin: I hadn’t noticed that convention before. After reviewing the codebase and confirming that the existing extensions use res_users.py and res_users_views.xml, I’ve renamed the class and related files to align with that pattern.

Thank you for highlighting this.

estate/security/estate_property_rules.xml
blse-odoo, line 1: From a fresh database, I can't access the estate app with the admin account. What should I do to be able to access it? 🙂

karimgamaleldin: Solved! Because the admin wasn't assigned to one of the real estate roles, he wasn't able to access the app. I've now set it up so he is assigned automatically as soon as the application is installed. (During development, he had been assigned manually as a manager before the access rights were added.)

PR #1180, opened 2026-02-17
Functional Training - Estate module

antonio-pimentel:


estate/models/estate_property.py
```diff
@@ -4,18 +4,33 @@ class EstateProperty(models.Model):
...
     
-    name = fields.Char(required=True)
+    name = fields.Char('Title', required=True)
     description = fields.Text()
     postcode = fields.Char()
-    date_availability = fields.Date()
+    date_availability = fields.Date(
+        string='Available From',
+        copy=False,
+        default=fields.Date.add(fields.Date.today(), months=+3))
```
blse-odoo: The declaration of the fields is only read once when the server starts. Here the default value will be computed just once the day the server is started, not every time a record is created.
If you starts the server on the 01/01, but create a new record on the 17/02, the default will be the value computed at start up, thus 01/04.
You can pass a lambda to `default`, and it will be computed on each record creation

blse-odoo (commented):
This is a good start!
Regarding the commit's messages, did you have a look at the git guidelines? (It's a bit of a detail, but it's good to get familiar with the process)

estate/views/estate_property_views.xml
blse-odoo, line 1: This looks all good!
(and good idea to rename `estate.properties` to match convention)
Your commit message mentions chapter 5, but it looks like this is chapter 6 (can you maybe amend your commit message to reflect that?)

antonio-pimentel: Thanks! Done

blse-odoo:
Just a small detail about the message of your last commit: the `[ADD]` tag is for new modules, I believe the `[IMP]` tag is more appropriate in this case ([link to the guidelines](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html#tag-and-module-name))

estate/views/estate_property_tag_views.xml
```diff
@@ -5,6 +5,16 @@
         <field name="view_mode">list,form</field>
     </record>
 
+    <record id="estate_property_tag_view_list" model="ir.ui.view">
+        <field name="name">estate.property.tag.list</field>
+        <field name="model">estate.property.tag</field>
+        <field name="arch" type="xml">
+            <list string="Tags" editable="top">
+                <field name="name"/>
```
blse-odoo, line 13: Could we add the possibility to edit the color of the tag from this list view?

blse-odoo (commented):
Very good :)
(just for the exercise) Could you make a commit message with a description, in addition to the title?

estate/views/estate_property_tag_views.xml
antonio-pimentel: Yes. Added the color field with the `color_picker` widget to the list view :)

awesome_dashboard/static/src/dashboard/dashboard_item/dashboard_item.js
```diff
@@ -0,0 +1,16 @@
...
+    static template = "awesome_dashboard.DashboardItem"
+    static props = {
+        slots: {
+            type: Object,
+            shape: { default: Object },
+        },
+        size: {
+            type: Number,
+            optional: true,
+            default: 1,
```
blse-odoo, line 13: This line does not set a default value for the `size` props. You can confirm that by removing the fallback `|| 1` (where `DashboardItem` is used) and see that the layout changes.
Look for `static defaultProps` to see how default values should be provided for props 😉

awesome_dashboard/static/src/dashboard/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,39 @@
+import { loadJS } from "@web/core/assets";
+import { getColor } from "@web/core/colors/colors";
+import { Component, onWillStart, useRef, onMounted, onWillUnmount } from "@odoo/owl";
+
+export class PieChart extends Component {
+    static template = "awesome_dashboard.PieChart";
+    static props = {
+        data: Object,
```
blse-odoo, line 8: It looks like the chart is not updated when the statistic service updates the values.
You can have a look at https://github.com/odoo/owl/blob/v2.8.2/doc/reference/hooks.md#lifecycle-hooks for a bt more information about the life-cycle hooks that you can use :)

awesome_dashboard/static/src/dashboard/dashboard.xml
```diff
@@ -0,0 +1,42 @@
...
+                </t>
+            </div>
+        </Layout>
+    </t>
+
+    <t t-name="awesome_dashboard.ConfigurationDialog">
+        <Dialog title="'Dashboard items configuration'">
+            Which cards do you whish to see ?
+            <t t-foreach="items" t-as="item" t-key="item.id">
+                <CheckBox value="item.enabled" onChange="(ev) => this.toggleCheckbox(ev, item)">
```
blse-odoo, line 30: A small detail: we usually reserve `ev` name for variables that contain events. Here it just contains the new value

awesome_dashboard/static/src/dashboard/statistics_service.js
```diff
@@ -0,0 +1,22 @@
+import { reactive } from "@odoo/owl";
+import { rpc } from "@web/core/network/rpc";
+import { registry } from "@web/core/registry";
+
+const statisticsService = {
+
+    start() {
+        const statistics = reactive({ isReady: false });
```
blse-odoo, line 8: It looks like the `isReady` field is not used anywhere else. If that is the case, I think you can remove it completely

blse-odoo (commented):
It looks good 👍 
I made a few comments, don't hesitate to ask questions if they are not clear :)

PR #1185, opened 2026-02-26
[ADD] unpublish_product: auto unpublish products when free_qty is zero

mekot-odoo:
-Implement automatic unpublishing of products when their free quantity
 reaches zero.

### DIFF
--- a/unpublish_product/__manifest__.py
+++ b/unpublish_product/__manifest__.py
@@ -1,0 +1,12 @@
+{
+    "name": "auto unpublish product",
+    "version": "1.0.0",
+    "depends": ["base", "product", "website_sale", "stock"],
+    "author": "Mehul Kotak",
+    "category": "Task-3",
+    "description": "eCommerce",
+    "license": "LGPL-3",
+    "auto_install": True,
+    "installable": True,
+    "data": ["views/res_config_settings_view.xml"],
+}
--- a/unpublish_product/models/res_config_settings.py
+++ b/unpublish_product/models/res_config_settings.py
@@ -1,0 +1,10 @@
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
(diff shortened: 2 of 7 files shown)


unpublish_product/models/res_config_settings.py
```diff
@@ -0,0 +1,10 @@
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    auto_unpublish_out_of_stock = fields.Boolean(
+        string="Auto Unpublish Out-of-Stock Products",
+        config_parameter="unpublish_product.auto_unpublish_out_of_stock",
+    )
```
kcv-odoo, line 10: make this setting company dependent, or website dependent 🤔

unpublish_product/models/stock_quant.py
```diff
@@ -0,0 +1,13 @@
...
+class StockQuant(models.Model):
+    _inherit = "stock.quant"
+
+    def write(self, vals):
+        res = super().write(vals)
+
+        products = self.mapped("product_id.product_tmpl_id")
+        products.check_and_unpublish()
+
+        return res
```
kcv-odoo, line 13: Don't do it on write of quant, do it in product template directly

unpublish_product/__manifest__.py
```diff
@@ -0,0 +1,12 @@
...
+    "name": "auto unpublish product",
+    "version": "1.0.0",
+    "depends": ["base", "product", "website_sale", "stock"],
+    "author": "Mehul Kotak",
+    "category": "Task-3",
+    "description": "eCommerce",
+    "license": "LGPL-3",
+    "auto_install": True,
+    "installable": True,
+    "data": ["views/res_config_settings_view.xml"],
```
kcv-odoo, line 11: Don't create new module directly update in community, make hook or something like it, it should check direct quantity if stock is not installed then override in website_sale_stock to adept depending on stock

PR #1188, opened 2026-03-10
Developing a real estate advertisement module from scratch

pasaw-odoo:
Setup and install real estate module
Declaring model class and creating estate property table in database via odoo orm
implementing access rights


estate/demo/demo.xml
```diff
@@ -0,0 +1,198 @@
+<odoo>
+  <data>
```
mash-odoo: What's the purpose of this data tag?
Will there be any impact if we remove it?

estate/demo/demo.xml
```diff
@@ -0,0 +1,198 @@
...
+      <field name="living_area">95</field>
+      <field name="facades">2</field>
+      <field name="garage">False</field>
+      <field name="garden">False</field>
+      <field name="garden_area">0</field>
+      <field name="garden_orientation">north</field>
+    </record>
+
+  </data>
+</odoo>
```
mash-odoo: Always leave an extra line at the end of the file.

estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
+from dateutil.relativedelta import relativedelta
+from odoo import fields, models
+
+
+class EstateProperty(models.Model):
```
mash-odoo, line 7: Follow [these](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html?highlight=import#imports) guidelines to import in a correct order and proper format.

estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
...
+    _name = 'estate.property'
+    _description = 'Real Estate Property'
+
+    name = fields.Char(string='Property Name', required=True, help='Enter the name of the property')
+    description = fields.Text(string='Property Description', help='Enter a description of the property')
```
```suggestion
    name = fields.Char(string="Property Name", required=True, help='Enter the name of the property')
    description = fields.Text(string="Property Description", help='Enter a description of the property')
```
Functional strings (the strings which are shown to user) should be in double quotes.

estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
...
+    postcode = fields.Char(string='Postcode', help='Enter the postcode of the property')
```
```suggestion
    postcode = fields.Char(help='Enter the postcode of the property')
```
Even if you don't mention string here, it will directly take the field name.

estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
...
+    date_availability = fields.Date(string='Availability Date', help='Enter the date when the property becomes available', copy=False, default=lambda self: fields.Date.today() + relativedelta(months=3))
```
mash-odoo: Nit: try to make it <100 for a better readability.
```suggestion
    date_availability = fields.Date(
        string='Availability Date',
        help='Enter the date when the property becomes available',
        copy=False,
        default=lambda self: fields.Date.today() + relativedelta(months=3)
    )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
...
+    garage = fields.Boolean(string='Garage', help='Check if the property has a garage')
+    garden = fields.Boolean(string='Garden', help='Check if the property has a garden')
+    garden_area = fields.Integer(string='Garden Area', help='Enter the area of the garden in square meters')
+    garden_orientation = fields.Selection(
+        selection=[
+            ('north', 'North'),
+            ('south', 'South'),
+            ('east', 'East'),
+            ('west', 'West'),
+        ],
```
mash-odoo, line 46:
```suggestion
        selection=[
            ('north', "North"),
            ('south', "South"),
            ('east', "East"),
            ('west', "West"),
        ],
```
keep the key in single quotes and value in double quotes

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,9 @@
+
+<odoo>
+    <menuitem id="estate_menu_root" name="Estate">
+        <menuitem id="estate_first_level_menu" name="Advertisements">
+            <menuitem id="estate_property_menu_action" action="estate_property_action"/>
+        </menuitem>
+    
+    </menuitem>
```
mash-odoo, line 11:
```suggestion
        </menuitem>
    </menuitem>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,84 @@
...
+            <search>
+                <field name="name"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="selling_price"/>
+                <field name="date_availability"/>
+                <filter
+                    name="available"
+                    string="Available"
+                    domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"
```
mash-odoo: Can you optimize this domain?

estate/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    'data': [
+        'security/ir.model.access.csv',
+        'views/estate_property_views.xml',
+        'views/estate_menus.xml',
+    ],
+    'demo': [
+        'demo/demo.xml',
+    ],
+    'application': True,
+    'installable': True,
```
mash-odoo: What's the meaning of `installable` true?

mash-odoo (changes requested):
Hello!
Great start to the task! 🐣 
Here are some comments and suggestions, please have a look.
Also change the PR title and description!

estate/__manifest__.py
pasaw-odoo: installable true means it becomes visible in all filter and we can install that specific module

pasaw-odoo: we dont need to write this as it is already true in default manifest which is in module.py

estate/demo/demo.xml
pasaw-odoo: The <data> tag was used in older Odoo versions and is now completely optional. But if we are using noupdate="1", the logic is — when a user renames a field and saves it in the DB, but the .xml file on disk doesn't change, then on upgrade the old data comes back from the XML. But with noupdate="1" it doesn't update and keeps what the user had renamed. And if we are writing inside the <odoo> tag and want to apply noupdate="1" to all records we can put it directly on <odoo>, but if we want to apply it to only a few records then we can add a separate <data noupdate="1"> block for those specific ones.

estate/demo/demo_tag.xml
```diff
@@ -0,0 +1,43 @@
...
+
+    <record id="estate_tag_pet_friendly" model="estate.property.tag">
+        <field name="name">Pet Friendly</field>
+    </record>
+
+    <record id="estate_tag_gated" model="estate.property.tag">
+        <field name="name">Gated Community</field>
+    </record>
+
+</odoo>
```
mash-odoo, line 43: Always add an extra line at the end of the file!!

estate/models/estate_property.py
```diff
@@ -0,0 +1,82 @@
...
+        selection=[
+            ('north', "North"),
+            ('south', "South"),
+            ('east', "East"),
+            ('west', "West"),
+        ],
+        string="Garden Orientation",
+        help='Select the orientation of the garden'
+    )
+    active = fields.Boolean(string="Active", default=True, help='Set to False to archive the property')
```
mash-odoo, line 50:
```suggestion
    active = fields.Boolean(default=True, help='Set to False to archive the property')
```
No need to add the string. By default it will take the string from the field name

estate/models/estate_property.py
```diff
@@ -0,0 +1,82 @@
...
+        ],
+        string="Status",
+        required=True,
+        copy=False,
+        default='new',
+        help='Current status of the property'
+    )
+    property_type_id = fields.Many2one('estate.property.type', string="Property Type")
+    buyer_id = fields.Many2one('res.partner', string="Buyer", copy=False)
+    seller_id = fields.Many2one('res.users', string="Salesperson", default=lambda self:self.env.user)
```
mash-odoo: What is a lambda function? And why do we need to use this function here?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,37 @@
+from odoo import api,fields, models
+from datetime import timedelta
+
+class EstatePropertyOffer(models.Model):
+    _name = 'estate.property.offer'
+    _description = 'Estate Property Offer'
```
```suggestion
    _description = "Estate Property Offer"
```
This will be displayed in UI so you can keep it in double quotes

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,37 @@
...
+
+    price = fields.Float(string="Price")
+
+    status = fields.Selection(
+        selection=[
+            ('accepted', "Accepted"),
+            ('refused', "Refused"),
+        ],
+        string="Status",
```
mash-odoo, line 23:
```suggestion
```
You don't need to explicitly mention the string unless you want a different display name than the field name.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,130 @@
+<odoo>
+   <record id="estate_property_view_list" model="ir.ui.view">
+        <field name="name">estate.property.list</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <list limit="50" decoration-danger="state == 'canceled'" decoration-success="state == 'sold'">
+                <field name="name" optional="show"/>
+                <field name="property_type_id" optional="show"/>
+                <field name="postcode" optional="show"/>
+                <field name="tag_ids" widget="many2many_tags" optional="show"/>
```
mash-odoo, line 21: What's the purpose of this widget?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,130 @@
+<odoo>
+   <record id="estate_property_view_list" model="ir.ui.view">
+        <field name="name">estate.property.list</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <list limit="50" decoration-danger="state == 'canceled'" decoration-success="state == 'sold'">
+                <field name="name" optional="show"/>
+                <field name="property_type_id" optional="show"/>
+                <field name="postcode" optional="show"/>
```
mash-odoo, line 20: What is the purpose of this `optional` attribute?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,130 @@
...
+            <field name="name"/>
+            <field name="postcode"/>
+            <field name="expected_price"/>
+            <field name="bedrooms"/>
+            <field name="living_area"/>
+            <field name="facades"/>
+            <filter 
+                string="Available" 
+                name="available"
+                domain="[('state','in',['new','offer_received'])]"/>
```
```suggestion
                domain="[('state', 'in', ['new', 'offer_received'])]"/>
```
Leave an extra space after adding a comma.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,130 @@
...
+            <filter 
+                string="Archived"
+                name="archived"
+                domain="[('active','=',False)]"/>
+            <separator/>
```
mash-odoo, line 151: Why do we use this separator?
Also do you know where is it coming from?

estate/__manifest__.py
```diff
@@ -0,0 +1,27 @@
...
+    'category': 'Real Estate',
+    'data': [
+        'security/ir.model.access.csv',
+        'views/estate_property_views.xml',
+        'views/estate_property_type_views.xml',
+        'views/estate_property_tags_views.xml',
+        'views/estate_property_offer_views.xml',
+        'views/estate_menus.xml',
+
+    ],
```
mash-odoo, line 21:
```suggestion
        'views/estate_menus.xml',
    ],
```

mash-odoo (changes requested):
Hello!
Thank you for your work..
Here are a few questions and suggestions.

estate/views/estate_property_views.xml
pasaw-odoo: I used <separator/> because it separates filters into groups. Filters within the same group use OR logic, and if we want to combine filters from two different groups, AND logic is applied between them. it came from search_arch_parser.js

estate/views/estate_property_views.xml
pasaw-odoo: it is just use for appearance

estate/views/estate_property_views.xml
pasaw-odoo: The optional attribute is a field attribute used in list views. If we don't want to see a specific column, we can hide or unhide it using this attribute

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,52 @@
...
+            record.date_deadline = start + timedelta(days=record.validity)
+
+    def _inverse_date_deadline(self):
+        for record in self:
+            start = record.create_date.date() if record.create_date else fields.Date.today()
+            record.validity = (record.date_deadline - start).days
+
+    def action_accept(self):
+        if self.property_id.state == 'sold':
+            raise UserError(_("you cant accept a sold property"))
```
mash-odoo, line 58: What is this  `_` used for?

estate/models/estate_property.py
```diff
@@ -0,0 +1,94 @@
+from dateutil.relativedelta import relativedelta
+from odoo import api, fields, models, _
+from odoo.exceptions import UserError
+
+
+class EstateProperty(models.Model):
+    _name = 'estate.property'
+    _description = "Real Estate Property"
+
+    name = fields.Char(string="Property Name", required=True, help='Enter the name of the property')
```
mash-odoo, line 22:
```suggestion
    name = fields.Char(string="Property Name", required=True, help="Enter the name of the property")
```
The help string is going to be displayed, so it should be wrapped in double quotes.
Also, how can we create a new attribute like help over here... say `info`?

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
...
+            ('sold', "Sold"),
+            ('canceled', "Canceled")
+        ],
+        string="Status",
+        required=True,
+        copy=False,
+        default='new',
+        compute="_compute_state",
+        store=True,
+        readonly=False,
```
mash-odoo: I don't think this is the best way to achieve what you are doing. Making a compute field is heavy and then you are also storing it. If you just want to change the state, probably you can override the write method instead.

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
...
+            else:
+                record.state = 'canceled'
+
+    def action_reset(self):
+        for record in self:
+            record.state = 'new'
+            record.offer_ids.unlink()
+
+    @api.constrains("selling_price", "expected_price")
+    def set_price(self):
```
mash-odoo, line 134:
```suggestion
    def _check_selling_price(self):
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,151 @@
...
+                            string = "Mark as sold"
+                            type = "object"
+                            confirm =" Are you sure you want to sold this property ?"
+                            invisible = "state == 'sold'"/>
+                    <button name = "action_cancel"
+                            string = "Mark as Cancel"
+                            type = "object"
+                            confirm = "Are you sure you want to cancel this property ?"
+                            invisible = "state == 'canceled'"/>
+                    <field name="state" widget="statusbar" statusbar_visible="new,offer_received,offer_accepted,sold" options="{'clickable': '1'}"/>
```
mash-odoo: Let the state be changed based on the condition. We won't need this option clickable.

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
...
+    _order = "id desc"
+
+    _check_expected_price = models.Constraint(
+        'CHECK(expected_price > 0)',
+        'The property expected price must be strictly positive',
+    )
+    _check_selling_price = models.Constraint(
+        'CHECK(selling_price >= 0)',
+        'The property selling price must be positive',
+    )
```
mash-odoo, line 20: Follow these guidelines while ordering fields, constraints, actions etc

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
+from dateutil.relativedelta import relativedelta
+from odoo import api, fields, models, _
+from odoo.exceptions import UserError, ValidationError
```
mash-odoo, line 3: Refer [this](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#imports) for proper ordering of imports

estate/models/estate_property.py
```diff
@@ -0,0 +1,200 @@
...
+    property_type_id = fields.Many2one('estate.property.type', string="Property Type")
+    buyer_id = fields.Many2one('res.partner', string="Buyer", copy=False)
+    seller_id = fields.Many2one('res.users', string="Salesperson", default=lambda self: self.env.user)
+    tag_ids = fields.Many2many('estate.property.tag', string="Property Tags", compute="_compute_tags", readonly=False ,store = True)
+    
+    offer_ids = fields.One2many('estate.property.offer', 'property_id', string="Offers" )
+
+    total_area = fields.Integer(string="Total Area", compute="_compute_total_area", help="Total area of the property including living area and garden area", store=True)
+    best_price = fields.Float(string="Best Offer", compute="_compute_best_price", help="Best offer received for the property", store=True)
+    offer = fields.Float(string="Total Offer", compute="_compute_total_offer", help="This shows the total offers this property has", store=True)
```
mash-odoo, line 77: Where is this field used exactly?
Like we are already showing the offers in a separate page. So what are we trying to do with this field?

estate/models/estate_property.py
```diff
@@ -0,0 +1,200 @@
...
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for record in self:
+            record.best_price = max(record.offer_ids.mapped('price')) if record.offer_ids else 0
+
+    @api.depends("offer_ids")
+    def _compute_total_offer(self):
+        for record in self:
+            record.offer = len(record.mapped('offer_ids'))
```
mash-odoo, line 93: Not sure if this is really needed

estate/models/estate_property.py
```diff
@@ -0,0 +1,200 @@
...
+        low_value = tag_value.filtered(lambda r : r.name == 'Low Interest')
+        if not low_value:
+            low_value = self.env['estate.property.tag'].create({'name': 'Low Interest'})
+        quick_sale = tag_value.filtered(lambda r : r.name == 'Quick sale')
+        if not quick_sale:
+            quick_sale = self.env['estate.property.tag'].create({'name': 'Quick sale'})
+        for record in self:
+            if record.expected_price > 50000:
+                record.tag_ids |= high_value
+            if 0 < len(record.mapped('offer_ids')) < 2 :
```
```suggestion
            if 0 < len(record.offer_ids) <= 2 :
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,200 @@
...
+                if offer.price == record.best_price:
+                    offer.status = 'accepted'
+                    other = record.offer_ids - offer
+                    other.status = 'refused'
+                    record.state = 'offer_accepted'
+                
+    @api.depends("expected_price","state","offer_ids","create_date")
+    def _compute_tags(self):
+        tag_value = self.env['estate.property.tag'].search([])
+        high_value = tag_value.filtered(lambda r : r.name == 'High Value')
```
```suggestion
        tags = self.env['estate.property.tag'].search([
            ('name', 'in', ['High Value', 'Low Interest', 'Quick Sale'])
        ])
        tags_map = {tag.name: tag for tag in tags}
```
You can also try doing in this way and use `.get` to check its presence. WDYT?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,157 @@
...
+                            string = "Mark as sold"
+                            type = "object"
+                            confirm =" Are you sure you want to sold this property ?"
+                            invisible = "state == 'sold'"/>
+                    <button name = "action_cancel"
+                            string = "Mark as Cancel"
+                            type = "object"
+                            confirm = "Are you sure you want to cancel this property ?"
+                            invisible = "state == 'canceled'"/>
+                    <button name ="action_best"
```
mash-odoo, line 51:
```suggestion
                    <button name ="action_accept_best_offer"
```|Try giving a name which is self explanatory of what it needs to perform.

File: estate/models/estate_property.py [hunk 33]
```diff
@@ -0,0 +1,200 @@
...
+    def set_price(self):
+        for record in self:
+            if record.selling_price > 0 and record.selling_price < record.expected_price * 0.9:
+                raise ValidationError("The selling price cannot be lower than 90 percent of the expected price")
+
+    def action_best(self):
+        for record in self:
+            for offer in record.offer_ids:
+                if offer.price == record.best_price:
+                    offer.status = 'accepted'
```
mash-odoo, line 148:
```suggestion
            best_offer = record.offer_ids.filtered(
                lambda offer: offer.price == record.best_price
            )[:1]
```
Your code might create issue if multiple offers have same price. Can you try this?

estate/models/estate_property.py
```diff
@@ -0,0 +1,200 @@
...
+        for record in self:
+            if record.state == 'sold':
+                raise UserError(_("Sold properties cannot be canceled"))
+            else:
+                record.state = 'canceled'
+
+    def action_reset(self):
+        for record in self:
+            record.state = 'new'
+            record.offer_ids.unlink()
```
mash-odoo, line 131: If we see the business perspective, i feel that writing this would lose the offer data.
I don't think there should be a reset button here. If you are doing this, maybe you can try archiving the offers instead. WDYT?

estate/models/estate_property.py
```diff
@@ -0,0 +1,200 @@
...
+                raise ValidationError("The selling price cannot be lower than 90 percent of the expected price")
+
+    def action_best(self):
+        for record in self:
+            for offer in record.offer_ids:
+                if offer.price == record.best_price:
+                    offer.status = 'accepted'
+                    other = record.offer_ids - offer
+                    other.status = 'refused'
+                    record.state = 'offer_accepted'
```
mash-odoo: Also I think that you are already doing the same operations in the accept_offer in the offer model. So you can maybe remove the code duplicacy.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,69 @@
...
+        self.status = 'accepted'
+        other = self.property_id.offer_ids - self
+        other.write({'status': 'refused'})
+        self.property_id.state = 'offer_accepted'
+        self.property_id.buyer_id = self.partner_id
+        self.property_id.selling_price = self.price
+
+    def action_reject(self):
+        if self.property_id.state == 'sold':
+            raise UserError(_("You cant reject a sold property"))
```
mash-odoo, line 68: Instead of raising an error, you can hide the buttons once the offer is sold

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,157 @@
...
+                    <notebook>
+                        <page string="Description">
+                            <group>
+                                <field name="description"/>
+                                <field name="bedrooms"/>
+                                <field name="living_area"/>
+                                <field name="facades"/>
+                                <field name="garage"/>
+                                <field name="garden"/>
+                                <field name="garden_area" invisible="garden == False"/>
```
mash-odoo, line 87:
```suggestion
                                <field name="garden_area" invisible="not garden"/>
```

mash-odoo (changes requested):
Hello!
Thank you for your work..
I have added some questions and suggestions.
Have a look at it...
Please update your PR title and description.

PR #1199, opened 2026-03-16
Up to chapter 3

abotaha23:


estate/models/estate_property.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
```
leclerc-leo: This isn't needed anymore and can be removed.

estate/__manifest__.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+
```
leclerc-leo: same here, this isn't needed anymore

estate/__manifest__.py
leclerc-leo, line 1: You need to add more details in the manifest,

For example, you should have `base` as a dependency.
The author should be set to `Odoo S.A.` with a license `LGPL-3`.
You can also add the category `Tutorials`

estate/__manifest__.py
```diff
@@ -0,0 +1,10 @@
...
+{
+    'name': 'Estate',
+    'version': '1.9',
```
```suggestion
    'version': '1.0',
```

When creating a module, it should be on `1.0`.

estate/models/estate_property.py
```diff
@@ -0,0 +1,23 @@
...
+from odoo import fields, models
+
+
+class RecurringPlan(models.Model):
```
leclerc-leo: A better name for this class would be `EstateProperty` to match the _name you've set

estate/models/__init__.py
leclerc-leo, line 1: You can check the [Odoo Git Guidelines](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html) which will tell you more about commit names, and commit content.

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_test_model,access_test_model,model_estate_property,base.group_user,1,1,1,1
```
leclerc-leo: you should probably use use a better name and id for your access rule to make it easier when debugging.
For example, access_estate_property_user, as this rule is for the model estate_property and for the group user.

estate/models/estate_property.py
```diff
@@ -11,13 +12,19 @@ class RecurringPlan(models.Model):
     name = fields.Char(required=True)
     description = fields.Text()
     postcode = fields.Char()
-    date_availability = fields.Date()
+    date_availability = fields.Date(default=fields.Date.today()+relativedelta(months=3), copy=False)
```
leclerc-leo: You should use a lambda instead of directly giving it a value.
As the value used by default here would be computed when the server starts so if your server ran for 2 months, the date_availability would be only offset by a month.

estate/models/estate_property.py
```diff
@@ -11,13 +12,19 @@ class RecurringPlan(models.Model):
...
+    bedrooms = fields.Integer(default=2)
     living_area = fields.Integer()
     facades = fields.Integer()
     garage = fields.Boolean()
     garden = fields.Boolean()
     garden_area = fields.Integer()
-    garden_orientation = fields.Selection(selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')])
\ No newline at end of file
+    garden_orientation = fields.Selection(selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')])
+    active = fields.Boolean(default=True) # testing the reserved fields thing
```
leclerc-leo:
```suggestion
    active = fields.Boolean(default=True)
```

estate/models/estate_property.py
```diff
@@ -11,13 +12,19 @@ class RecurringPlan(models.Model):
...
+    state = fields.Selection(selection=[('new', 'New'), ('offer received', 'Offer Received'), ('offer accepted', 'Offer Accepted'), ('sold', 'Sold'), ('cancelled', 'Cancelled')],
+                            default='new',
+                            required=True,
+                            copy=False
+                        )
```
leclerc-leo: Here's how we would format this, Also you should use double quotes for any string that's going to be displayed to the user.
```suggestion
    state = fields.Selection(
        selection=[
            ('new', "New"),
            ('offer received', "Offer Received"),
            ('offer accepted', "Offer Accepted"),
            ('sold', "Sold"),
            ('cancelled', "Cancelled"),
        ],
        default='new',
        required=True,
        copy=False,
    )
```

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,19 @@
+<odoo>
+    <menuitem id="test_menu_root" name="Real Estate">
+        <menuitem id="test_first_level_menu" name="First Level">
+            <menuitem id="test_model_menu_action" action="estate_property_action"/>
```
leclerc-leo: Try to use id that represent what is inside.
This will help when you are debugging to identify which one has an issue more easily and will help others that read your code understand it better.
```suggestion
    <menuitem id="menu_real_estate" name="Real Estate">
        <menuitem id="menu_real_estate_first_level" name="First Level">
            <menuitem id="menu_real_estate_properties" action="estate_property_action"/>
```

estate/views/estate_property_views.xml
```diff
@@ -1,4 +1,81 @@
 <odoo>
+    <record id="estate_property_view_tree" model="ir.ui.view">
+        <field name="name">estate.property.list</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <list string="Tests">
+                <field name="name" string="Title"/>
```
leclerc-leo: Instead of settings string="X" in the view, It's better to set the string="X" on the field directly.
That way, if you display the field in different view, you wont have to add string= in both of those views.

leclerc-leo (changes requested):
Great job 👍 

Try to use more meaningful names and ids which will help other understand your code.

.vscode/settings.json
leclerc-leo, line 1: Try not to commit your .vscode (or changes to .gitignore for example)

A way to prevent this, is to have a src folder containing the community, enterprise and tutorials folder that way your .vscode is create in that folder.

estate/models/estate_property.py
```diff
@@ -0,0 +1,134 @@
+from odoo import api, fields, models, exceptions
+from odoo.tools.float_utils import float_compare, float_is_zero
+from dateutil.relativedelta import relativedelta
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
```
leclerc-leo: We always use single quotes in python, and only use double quotes for string shown to the users (such as for _description, string=, the string in selection field, SQL constraints, exceptions, ...)
```suggestion
    _name = 'estate.property'
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,134 @@
...
+                record.best_offer = 0
+
+    @api.constrains('selling_price', 'expected_price')
+    def _check_price(self):
+        for record in self:
+            if float_is_zero(record.selling_price, 3):
+                continue
+            if float_compare(record.selling_price, 0.9 * record.expected_price, 3) == -1:
+                raise exceptions.ValidationError(
+                    "The selling price cannot be less than 90% of the expected price")
```
leclerc-leo: 2 smalls nitpicks, when importing something from a submodule from odoo, we always try to import the submodule directly.
Also exceptions aren't directly translated, for this you need to use self.env._("string").
```suggestion
                raise ValidationError(self.env._("The selling price cannot be less than 90% of the expected price"))
```

estate/security/ir.model.access.csv
leclerc-leo: still relevant.

estate/views/estate_property_offers_views.xml
```diff
@@ -0,0 +1,56 @@
...
+                <field name="date_deadline"/>
+                <field name="property_id" nolabel="1"/> <!-- just to use it in the offer creation constraints-->
+                <button name="action_accept_offer" string=" " type="object" icon="fa-check" invisible="status"/>
+                <button name="action_refuse_offer" string=" " type="object" icon="fa-times" invisible="status"/>
+            </list>
+        </field>
+    </record>
+
+    <record id="estate_property_offers_view_form" model="ir.ui.view">
+        <field name="name">estate.property.offerform</field>
```
leclerc-leo:
```suggestion
        <field name="name">estate.property.offer.form</field>
```

estate/views/estate_property_types_views.xml
```diff
@@ -0,0 +1,49 @@
+<odoo>  
+    <record id="estate_property_type_view_form" model="ir.ui.view">
+        <field name="name">estate.property.type.form</field>
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <form>
+                <field name="name" nolable="1"/>
+                    <div class="oe_button_box" name="button_box">
+                        <button name="%(estate_property_offer_action)d" type="action"
```
leclerc-leo, line 9: I dont think `%()d` is needed but I might be wrong (I just never seen it used but I was able to find example in the codebase of it)
```suggestion
                        <button name="estate_property_offer_action" type="action"
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,146 @@
...
+            <list decoration-success="state == 'offer received' or state == 'offer accepted'" decoration-bf="state == 'offer accepted'" decoration-muted="state == 'sold'">
+                <field name="name"/>
+                <field name="property_type_id"/>
+                <field name="tag_ids" options="{'color_field': 'color'}" widget="many2many_tags"/>
+                <field name="postcode"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="expected_price"/>
+                <field name="selling_price"/>
+                <field optional="hide" name="date_availability"/>
```
leclerc-leo: Try to keep the same ordering of parameters between your xml tags
```suggestion
                <field name="date_availability" optional="hide"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,146 @@
...
+                        </group>
+                        <group>
+                            <field name="best_offer"/>
+                        </group>
+                        <group>
+                            <field name="date_availability"/>
+                        </group>
+                        <group>
+                            <field name="selling_price"/>
+                        </group>
```
leclerc-leo, line 47: Easier to understand since group takes 50% of the width of the parent when a parent <group> is present
```suggestion
                        <group>
                            <field name="property_type_id" options="{'no_create': true}"/>
                            <field name="expected_price"/>
                            <field name="date_availability"/>
                        </group>
                        <group>
                            <field name="postcode"/>
                            <field name="best_offer"/>
                            <field name="selling_price"/>
                        </group>
```

estate_account/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+{
+    'name': 'Estate Accounting',
+    'version': '1.0',
+    'author': 'Odoo S.A.',
+    'license': 'AGPL-3',
+    'application': True,
```
leclerc-leo: We call this kind of modules "bridge" modules since they are used as bridge between different apps.
And usually they aren't they own kind of apps, but more like additional features when both apps are installed.
```suggestion
```

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,39 @@
...
+        )
+        if not journal:
+            raise exceptions.UserError(
+                "Please define a 'Sale' journal in Accounting settings."
+            )
+
+        self.env['account.move'].create({
+            'partner_id': self.buyer_id.id,
+            'move_type': 'out_invoice',
+            'journal_id': journal.id,
```
leclerc-leo: There's no need to specify a journal here.
Searching on a random one of type sale isn't great as the result you get can get pretty random so you dont always have the same journal.

```suggestion
```

leclerc-leo (changes requested):
🚀

estate/views/estate_property_types_views.xml
abotaha23: it gives an error without it, I think because it is not a functional action in a python file, but rather a view action, don't know if this is correct or not I'm just guessing :)

awesome_owl/static/src/TodoList/TodoList.js
```diff
@@ -0,0 +1,42 @@
+import { Component, useState } from "@odoo/owl";
+import { TodoItem } from "../TodoItem/TodoItem";
+import { useAutofocus } from "../utils";
+
+export class TodoList extends Component {
+    static template = "awesome_owl.TodoList";
+
+    static components = { TodoItem };
+
+    static props = {};
```
leclerc-leo: Nitpick / How I would have done it: I think it looks a bit better when they are grouped together (but just my opinion)
```suggestion
    static template = "awesome_owl.TodoList";
    static components = { TodoItem };
    static props = {};
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,138 @@
...
+            record.state = 'sold'
+
+    def action_mark_as_canceled(self):
+        for record in self:
+            if record.state != 'sold':
+                record.state = 'canceled'
+            else:
+                raise UserError(self.env._(
+                    "Sold properties cannot be canceled!")
+                )
```
leclerc-leo:
```suggestion
            if record.state == 'sold':
                raise UserError(self.env._("Sold properties cannot be canceled!"))
            record.state = 'canceled'
```

PR #1204, opened 2026-03-16
19.0-adkha-Technical Training

Adham-K-Fahmy:


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
leclerc-leo: Try not to push .vscode and similar files as those are usually reserved for the basic configuration used by everyone.

estate/models/estate_property.py
```diff
@@ -0,0 +1,28 @@
+from odoo import fields, models
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
```
leclerc-leo: You should always use single quotes except for string that will be displayed to the user such as the _description, string="X" for the fields, ...
```suggestion
    _name = 'estate.property'
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,28 @@
...
+    _description = "Estate properties"
+
+    name = fields.Char('Name', required=True)
+    description = fields.Char("Description", translate=True)
```
leclerc-leo: Char fields are translatable by default, you need to add this argument

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
```
leclerc-leo: Nitpick, you can name it `access_estate_property_user`, as this rule applies to user.
This can help when you set one for user, admin, system, ...

estate/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+{
```
leclerc-leo, line 1: your module is missing the author (`Odoo S.A.`) and a license.

leclerc-leo (changes requested):
Great job 👍

estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_property, estate_property_type, estate_property_tag, estate_property_offer, res_users
```
leclerc-leo: Nitpick: We try to have every imports in an __init__.py file to start by `from . import `
```suggestion
from . import estate_property
from . import estate_property_type
from . import estate_property_tag
from . import estate_property_offer
from . import res_users
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,89 @@
+from odoo import api, fields, models, exceptions, tools
+
+
+class EstateProperty(models.Model):
+    _name = 'estate.property'
+    _description = 'Estate properties'
```
leclerc-leo: We use single quotes everywhere except for strings displayed to the user (in _description, string=, UserErrors, ...) which uses double quotes
```suggestion
    _description = "Estate properties"
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,89 @@
...
+        string='State',
+        selection=[('new', 'New'), ('offer_received', 'Offer Received'), ('offer_accepted', 'Offer Accepted'), ('sold', 'Sold'), ('cancelled', 'Cancelled')],
+        required=True,
+        copy=False,
+        default='new')
+    property_type_id = fields.Many2one('estate.property.type', string='Type')
+    buyer_id = fields.Many2one('res.partner', string='Buyer')
+    salesman_id = fields.Many2one('res.users', string='Salesman', default=lambda self: self.env.user)
+    tag_ids = fields.Many2many('estate.property.tag', string='Tags')
+    offer_ids = fields.One2many('estate.property.offer', 'property_id', string='Offers')
```
leclerc-leo: It's better to specify the comodel_name and inverse_name in One2many and Many2many fields in case the order of arguments changes and to make it easier for someone reviewing/reading your code to know which is which.
```suggestion
    offer_ids = fields.One2many(comodel_name='estate.property.offer', inverse_name='property_id', string='Offers')
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,89 @@
...
+            if not tools.float_utils.float_is_zero(record.selling_price, 2) and record.selling_price < 0.9 * record.expected_price:
+                raise exceptions.ValidationError('Selling price cannot be lower than 90% of the expected price')
+
+    @api.ondelete(at_uninstall=False)
+    def _property_delete(self):
+        if self.state != 'new' and self.state != 'cancelled':
+            raise exceptions.UserError('Cannot delete property with a state other than \'New\' or \'Cancelled\'')
+        self.offer_ids.unlink()
+
+    def change_state_to_sold(self):
```
leclerc-leo: When a function is called by the user we usually try to have it prefixed by `action_` to help know it's going to be called by the end user.

estate/models/estate_property.py
```diff
@@ -0,0 +1,89 @@
...
+    def _onchange_garden(self):
+        for record in self:
+            record.garden_area = (10 if record.garden else 0)
+            record.garden_orientation = ('north' if record.garden else None)
+
+    @api.constrains('selling_price')
+    def _check_selling_price(self):
+        for record in self:
+            if not tools.float_utils.float_is_zero(record.selling_price, 2) and record.selling_price < 0.9 * record.expected_price:
+                raise exceptions.ValidationError('Selling price cannot be lower than 90% of the expected price')
```
leclerc-leo: For odoo submodules such as exceptions, we prefer to import them directly rather than using submodule.<something>


```suggestion
                raise ValidationError('Selling price cannot be lower than 90% of the expected price')
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,89 @@
...
+        for record in self:
+            record.best_price = max(record.offer_ids.mapped('price'), default=0.0)
+
+    @api.onchange('garden')
+    def _onchange_garden(self):
+        for record in self:
+            record.garden_area = (10 if record.garden else 0)
+            record.garden_orientation = ('north' if record.garden else None)
+
+    @api.constrains('selling_price')
```
leclerc-leo: We try to have elements inside a class be ordered in the following:
- _name, _description, _order
- fields
- models.Constraint
- @api.constains
- _compute + _inverse (if you have a field with a compute and inverse, it's better to have them after each other)
- and the rest of the functions

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,66 @@
...
+            if record.create_date:
+                record.date_deadline = fields.Date.add(record.create_date, days=record.validity)
+
+    def _compute_validity(self):
+        for record in self:
+            if record.create_date and record.date_deadline:
+                record.validity = (record.date_deadline - fields.Date.to_date(record.create_date)).days
+
+    @api.model
+    def create(self, vals):
```
leclerc-leo: It's preferable to use `model_create_multi` to be able to handle imports of csv files, ...
```suggestion
    @api.model_create_multi
    def create(self, vals_list):
```

estate/views/estate_menus.xml
leclerc-leo, line 1: We always use double quotes in .xml files.
*I know it's weird when we use single quotes in py*

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,29 @@
...
+                    'name': 'Administrative Fees',
+                    'quantity': 1,
+                    'price_unit': 100.00
+                }),
+                Command.create({
+                    'name': 'Additional Costs',
+                    'quantity': 1,
+                    'price_unit': 0.06 * self.selling_price
+                }),
+            ]})
```
leclerc-leo: To make it easier to know that you're closing the line_ids and then the account.move.create
```suggestion
            ],
        })
```

estate_account/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+{
+    'name': 'Real Estate Account',
+    'category': 'Tutorials',
+    'depends': [
+        'base',
```
leclerc-leo: you dont need `base` here as you already depends on `estate` which depends on `base`

.gitignore
leclerc-leo, line 1: You shouldn't be editing .gitignore files

odoos.toml
leclerc-leo, line 1: Try to keep those file outside of your working directory.

For example use:
src/odoo
src/enterprise
src/upgrade
src/tutorial

and put your files into src which is not in a git repository and you shouldn't this sort of issues anymore.

leclerc-leo (changes requested):
Great work 👍

PR #1205, opened 2026-03-16
19.0 tutorial dacat

Yetso:


estate/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import fields, models
+from dateutil.relativedelta import relativedelta
+
+
+class RecurringPlan(models.Model):
+    _name = "estate.property"
```
lost-odoo: We typically use single quotes for strings, and reserve double quotes for user-facing text.
```suggestion
    _name = 'estate.property'
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
...
+    _description = "estate property"
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(default=fields.Date.today() + relativedelta(months=3), copy=False)
```
lost-odoo: You should use a lambda here; otherwise, this value will be evaluated at server start and remain static.

estate/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
```
lost-odoo: You can remove this it is not used anymore ^^

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,1
```
lost-odoo: Just a suggestion but cleaner IMO
```suggestion
access_estate_property_user,access_estate_property_user,model_estate_property,base.group_user,1,1,1,1
```

estate/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
lost-odoo: Here also

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<odoo>
+
```
lost-odoo, line 108: nitpicking but you could remove those spaces here and below same for other xml files 😄 .

lost-odoo (changes requested):
Hello, Good Job 😄 , I already made a small review.

estate/models/estate_property.py
lost-odoo: Do not resolve the comment while the change hasn't been done yet ^^

estate/models/estate_property.py
```diff
@@ -0,0 +1,97 @@
...
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "estate property"
+
+    name = fields.Char(required=True, string="Title")
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        default=lambda self: fields.Date.add(fields.Date.today(), months=3), copy=False
+    )
```
lost-odoo: Could be on one line

estate/models/estate_property.py
```diff
@@ -0,0 +1,97 @@
...
+    selling_price = fields.Float(readonly=True, copy=False)
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer(string="Living Area (sqm)")
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        [('north', "North"), ('south', "South"), ('east', "East"), ('west', "West")]
+    )
```
lost-odoo: This too

estate/models/estate_property.py
```diff
@@ -0,0 +1,97 @@
...
+            ('cancelled', "Cancelled"),
+        ],
+        required=True,
+        default='new',
+        copy=False,
+    )
+
+    _check_expected_price = models.Constraint(
+        'CHECK(expected_price > 0)', "The expected price must be stricly positive"
+    )
```
lost-odoo: Same, could also be on one line IMO

estate/models/estate_property.py
```diff
@@ -0,0 +1,97 @@
...
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        [('north', "North"), ('south', "South"), ('east', "East"), ('west', "West")]
+    )
+    total_area = fields.Integer(compute='_compute_total_area')
+    property_type_id = fields.Many2one('estate.property.type', string="House Type")
+    buyer_id = fields.Many2one('res.partner', string="Buyer", copy=False)
+    seller_id = fields.Many2one('res.users', string="Seller", default=lambda self: self.env.user)
+    tag_ids = fields.Many2many('estate.property.tag', string="Tags")
+    offer_ids = fields.One2many('estate.property.offer', 'property_id', string="")
```
lost-odoo: You could name your parameters IMO. It would be cleaner than having some named and some not.

estate/models/estate_property.py
```diff
@@ -0,0 +1,97 @@
...
+
+    _check_expected_price = models.Constraint(
+        'CHECK(expected_price > 0)', "The expected price must be stricly positive"
+    )
+
+    _check_selling_price = models.Constraint('CHECK(selling_price > 0)', "The selling price must be stricly positive")
+
+    @api.depends('living_area', 'garden_area')
+    def _compute_total_area(self):
+        for record in self:
```
```suggestion
        for property in self:
```

Same for other occurences. It is not wrong and is just a suggestion ^^

estate/models/estate_property.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @api.depends('offer_ids.price')
+    def _compute_best_offer(self):
+        for record in self:
+            record.best_offer = max(record.offer_ids.mapped('price'), default=0)
+
+    @api.onchange('garden')
+    def _onchange_garden(self):
+        self.garden_area = 10 if self.garden else 0
+        self.garden_orientation = 'north' if self.garden else ''
```
lost-odoo: It would be better to do the if only once than having it two times 😄

estate/models/estate_property.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @api.constrains('selling_price', 'expected_price')
+    def _check_selling_expected_price(self):
+        for record in self:
+            if not float_is_zero(record.selling_price, precision_digits=2) and (
+                float_compare( record.expected_price * 0.9, record.selling_price, precision_digits=2,) > 0):
```
lost-odoo: Be careful here, you have a final ',' and some useless spacing ^^. You should check you runbot and make sure it is green 😄

estate/models/estate_property.py
```diff
@@ -0,0 +1,97 @@
...
+    def action_sold(self):
+        for record in self:
+            if record.state == 'cancelled':
+                raise UserError(self.env_("Canceled properties cannot be sold"))
+            record.state = 'sold'
+        return True
+
+    @api.ondelete(at_uninstall=False)
+    def _ondelete(self):
+        if not any((property_id.state in ('new', 'cancelled')) for property_id in self):
```
lost-odoo: I think this condition is wrong. As if there are invalid states of properties in self but at least one is good it won't trigger the UserError 🤔

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,59 @@
...
+    @api.depends('create_date', 'validity')
+    def _inverse_date_deadline(self):
+        for record in self:
+            record.validity = (record.date_deadline - record.create_date.date()).days
+
+    def action_accept_offer(self):
+        for record in self:
+            for offer in record.property_id.offer_ids:
+                if offer.status == 'accepted':
+                    raise UserError(self.env_("Another offer has already been accepted."))
```
lost-odoo:
```suggestion
            if accepted_offers := record.property_id.offer_ids.filtered(lambda offer: offer.status == 'accepted'):
                raise UserError(self.env_("Another offer has already been accepted."))
```

I think this should be cleaner IMO

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def action_refuse_offer(self):
+        for record in self:
+            record.status = 'refused'
+        return True
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            PropertyModel = self.env['estate.property'].browse(vals['property_id'])
```
lost-odoo: This is not a "PropertyModel" as you made a browse but a recordset so here it would be  as simple as: 'property'

PropertyModel would be used if you do:

`PropertyModel = self.env['estate.property']`

estate/models/res_users.py
```diff
@@ -0,0 +1,15 @@
...
+
+class ResUsers(models.Model):
+    _inherit = ['res.users']
+    _name = 'res.users'
+
+    property_ids = fields.One2many(
+        'estate.property',
+        'seller_id',
+        string="Available Properties",
+        domain=[('state', 'in', {'new', 'offer_received'})],
```
lost-odoo: Would be better to use a tuple or a list 😄

estate_account/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    'depends': [
+        'base',
+        'estate',
+        'account',
+    ],
+    'application': True,
+    'installable': True,
+    'author': 'Odoo S.A.',
+    'license': 'LGPL-3',
+
```
lost-odoo, line 17: missing `auto_install` key. This key is usefull because it auto install the module if all the dependencies are installed.

lost-odoo (changes requested):
Good Job !

PR #1206, opened 2026-03-16
jucop - Technical Training 19.0

jucop-odoo:


estate/models/estate_property.py
```diff
@@ -0,0 +1,106 @@
+from dateutil.relativedelta import relativedelta
+
+from odoo import fields, models, api
+from odoo.exceptions import UserError
+from odoo.tools import float_compare
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
```
leclerc-leo: You should always use single quotes, except for string displayed to the users such as _description, string of fields, userErrors, ...
```suggestion
    _name = 'estate.property'
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,106 @@
...
+    _description = "Estate property"
+
+    _rec_name = 'name'
```
leclerc-leo: You dont need to specify the _rec_name here as name is used by default.

estate/models/estate_property.py
```diff
@@ -0,0 +1,106 @@
...
+
+    _check_expected_price = models.Constraint(
+        'CHECK (expected_price > 0)',
+        'The expected price must be greater than 0.'
+    )
+    _check_selling_price = models.Constraint(
+        'CHECK (selling_price >= 0)',
+        'The selling price must be greater or equal than 0.'
+    )
```
leclerc-leo, line 52: Nitpick: The constraints should be after the declaration of your fields

estate/models/estate_property.py
```diff
@@ -0,0 +1,106 @@
...
+
+    name = fields.Char(required=True, string='Title')
+    description = fields.Text(string='Description')
+    postcode = fields.Char(required=True, string='Postcode')
+    date_availability = fields.Date(required=True, default=fields.Date.today() + relativedelta(months=+3), copy=False, string='Available From')
```
leclerc-leo: You should use a lambda instead of directly giving it a value.
As the value used by default here would be computed when the server starts so if your server ran for 2 months, the date_availability would be only offset by a month.

estate/models/estate_property.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @api.depends('offer_ids.price')
+    def _compute_best_price(self):
+        for record in self:
+            record.best_price = max(record.offer_ids.mapped('price')) if len(record.offer_ids) > 0 else 0
+
+    def action_mark_as_sold(self):
+        for record in self:
+            if record.state == 'canceled':
+                raise UserError("Cannot mark a canceled property as sold")
```
leclerc-leo: This wont be translate, you should use self.env._ to make it translatable
```suggestion
                raise UserError(self.env._("Cannot mark a canceled property as sold"))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,106 @@
...
+            if not record.available_for_offers:
+                raise UserError("Cannot add or update an offer on a property that is not available for offers")
+
+            record.selling_price = offer.price
+            record.buyer_id = offer.partner_id
+            record.state = 'offer_accepted'
+
+    def _compute_available_for_offers(self):
+        for record in self:
+            record.available_for_offers = record.state in ('new', 'offer_received')
```
leclerc-leo: Try to use sets rather than tuples when using in, as it mean you are going to look for the element in O(1).
*Small note: this usually has no impact for 2 elements tuple but it's a great habit to have so when you declare 3 elements or more you dont have to switch it to a set*

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,1
```
leclerc-leo: small Nitpick, you can dont need to specify `estate` here and you can specify the group in the id and name of the access rules to make it easier to understand at what level it applies (to users, admins, system)
```suggestion
access_estate_property_user,access_estate_property_user,model_estate_property,base.group_user,1,1,1,1
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,124 @@
...
+                <field name="living_area"/>
+                <field name="facades"/>
+                <field name="state"/>
+                <separator/>
+                <filter string="Available" name="available" domain="[('state', 'in', ('new', 'offer_received'))]"/>
+                <separator/>
+                <group>
+                    <filter string="Postcode" name="postcode" context="{'group_by':'postcode', 'residual_visible':True}"/>
+                    <filter string="Buyer" name="buyer" context="{'group_by':'buyer_id', 'residual_visible':True}"/>
+                    <filter string="Salesman" name="salesman" context="{'group_by':'seller_id', 'residual_visible':True}"/>
```
leclerc-leo: What's `residual_visible` here for ?
I doesn't seems to be anywhere else in the PR.

leclerc-leo (changes requested):
Very good work 👍

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property,access_estate_property_user,estate.model_estate_property,base.group_user,1,1,1,1
```
leclerc-leo: You dont need to specify estate here.
```suggestion
access_estate_property,access_estate_property_user,model_estate_property,base.group_user,1,1,1,1
```

estate/models/res_users.py
```diff
@@ -0,0 +1,10 @@
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = ['res.users']
```
leclerc-leo: When just doing an override of a class it's a bit better to do the inherit without a list
```suggestion
    _inherit = 'res.users'
```

leclerc-leo (commented):
🚀

PR #1208, opened 2026-03-19
qugeo - Technical Training

qugeo-odoo:


estate/models/__init__.py
```diff
@@ -0,0 +1,2 @@
+
```
leclerc-leo: Nitpick: you can remove this empty line
```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,31 @@
+from odoo import fields, models
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
```
leclerc-leo: We always use single quotes and only use double quotes for string shown to the users (in _description, string=X, selection fields, exceptions, ...)
```suggestion
    _name = 'estate.property'
```

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,1
+estate.access_estate_property_type,access_estate_property_type,estate.model_estate_property_type,base.group_user,1,1,1,1
+estate.access_estate_property_tag,access_estate_property_tag,estate.model_estate_property_tag,base.group_user,1,1,1,1
+estate.access_estate_property_offer,access_estate_property_offer,estate.model_estate_property_offer,base.group_user,1,1,1,1
```
leclerc-leo: estate is not needed here since you already are inside the estate module.
Small tip: when multiple access rule exist for different groups (user, admin, system, ...) when add `_<group>` at the end (like `_user`)
```suggestion
access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
access_estate_property_type,access_estate_property_type,model_estate_property_type,base.group_user,1,1,1,1
access_estate_property_tag,access_estate_property_tag,model_estate_property_tag,base.group_user,1,1,1,1
access_estate_property_offer,access_estate_property_offer,model_estate_property_offer,base.group_user,1,1,1,1
```

estate/views/estate_property_view_form.xml
```diff
@@ -0,0 +1,58 @@
...
+                            <field name="expected_price"/>
+                            <field name="selling_price"/>
+                        </group>
+                    </group>
+                    <notebook>
+                        <page string="Details">
+                            <group>
+                                <field name="description"/>
+                                <field name="bedrooms"/>
+                                <field name="living_area" string="Living Area (sqm)"/>
```
leclerc-leo: It's usually better to set this string directly on the field.
Like you did, you would need to set this string every time you use it in a view.

estate/__manifest__.py
```diff
@@ -0,0 +1,29 @@
+{
+    'name': "estate",
+    'author': 'qugeo',
```
leclerc-leo: The author of modules you write should always be `Odoo S.A.`.
```suggestion
    'author': 'Odoo S.A.',
```

leclerc-leo (changes requested):
Great work 👍

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class EstateProperty(models.Model):
+    _name = 'estate.property'
+    _description = "Estate properties"
+    _order = 'id desc'
+
+    active = fields.Boolean(default=True)
+    state = fields.Selection([
+        ("new", "New"),
```
leclerc-leo: Only the 2nd argument of the tuple is shown to the user
```suggestion
        ('new', "New"),
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+    _check_selling_price = models.Constraint(
+        'CHECK(selling_price >= 0)',
+        "The selling price of a property must be positive.",
+    )
+
+    @api.constrains('selling_price', 'expected_price')
+    def _check_date_end(self):
+        for record in self:
+            if (len(record.property_offer_ids) != 0) and tools.float_utils.float_compare(record.expected_price * 0.9, record.selling_price, precision_digits=2) == 1:
+                raise exceptions.ValidationError("The selling price cannot be inferior to 90% of the expected price")
```
leclerc-leo: For ordering we try to have the following
- fields (you can order them a bit like you want and have separator lines when needed to group some together or not)
- sql contrains
- @api.contrains
- computed functions for fields (followed by their inverse if one exists)
- remaining inverse functions
- rest of the functions (but try to group together actions, orm methods (create, write, unlink, ...), ...)

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = ""
+
+    def estate_property_action_sold(self):
+        for record in self:
+            if (record.state == "cancelled"):
+                raise exceptions.UserError("Cancelled properties cannot be sold.")
```
leclerc-leo: userErrors need to be translated

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+        if (self.garden):
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = ""
+
+    def estate_property_action_sold(self):
+        for record in self:
+            if (record.state == "cancelled"):
```
leclerc-leo: in python, () are not needed for a if
```suggestion
            if record.state == "cancelled":
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+            if (record.state == "sold"):
+                raise exceptions.UserError("Sold properties cannot be cancelled.")
+            else:
+                record.state = "cancelled"
+        return True
+
+    @api.ondelete(at_uninstall=False)
+    def _check_state_before_unlink(self):
+        for record in self:
+            if record.state not in ['new', 'cancelled']:
```
leclerc-leo: While for 2 elements set and array have about the same performance, it's always great to build an habit of using sets instead of array (to always use them when 3 and more elements are present)
```suggestion
            if record.state not in {'new', 'cancelled'}:
```

This only apply to Python and not XML, as list are better there since the template can be compiled using Python or JS and array work in both

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,59 @@
...
+    validity = fields.Integer(default=7)
+    date_deadline = fields.Date(compute='_compute_deadline', inverse='_inverse_deadline', string="Deadline")
+
+    @api.depends('create_date', 'validity')
+    def _compute_deadline(self):
+        for record in self:
+            if (record.create_date):
+                record.date_deadline = fields.Date.add(record.create_date, days=record.validity)
+            else:
+                record.date_deadline = fields.Date.add(fields.Date.today(), days=record.validity)
```
leclerc-leo: something like this should work and duplicate less code
```suggestion
            create_date = record.create_date or fields.Date.today()
            record.date_deadline = fields.Date.add(create_date, days=record.validity)
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,59 @@
...
+            record.status = "refused"
+        return True
+
+    @api.model
+    def create(self, vals_list):
+        for vals in vals_list:
+            propertyId = self.env['estate.property'].browse(vals['property_id'])
+            curPrice = vals['price']
+            if propertyId.best_price > curPrice:
+                raise exceptions.UserError(f"Cannot create an offer with a lower price than the best offer:{propertyId.best_price}")
```
leclerc-leo:
```suggestion
                raise UserError(self.env._("Cannot create an offer with a lower price than the best offer: %s", propertyId.best_price))
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,59 @@
...
+        for record in self:
+            record.status = "refused"
+        return True
+
+    @api.model
+    def create(self, vals_list):
+        for vals in vals_list:
+            propertyId = self.env['estate.property'].browse(vals['property_id'])
+            curPrice = vals['price']
+            if propertyId.best_price > curPrice:
```
leclerc-leo: vals['price'] is clear enough to not declare another variable for it
```suggestion
            if propertyId.best_price > vals['price']:
```

estate/views/estate_menus.xml
leclerc-leo, line 1: We always used double quotes in xml file (only use single quote for rare case when a JSON object is given as an argument and it need to be parsed using JSON.parse in JS).
I know it doesn't make any sense when we are using single quotes in Python.

estate/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+        'security/ir.model.access.csv',
+
+        'views/estate_property_views.xml',
+        'views/estate_property_type_views.xml',
+        'views/estate_property_tag_views.xml',
+        'views/estate_property_offer_views.xml',
+
+        'views/estate_menus.xml',
+
+        'views/res_users_view_form.xml'
```
leclerc-leo: Try to always have trailing comas, so when someone is going to edit the file to add a line, he wont have to add diff on your line and make it harder to find who added the original line.
```suggestion
        'views/res_users_view_form.xml',
```

estate_account/__manifest__.py
```diff
@@ -0,0 +1,10 @@
+{
+    'name': 'Estate Accounting',
+    'author': 'Odoo S.A.',
+    'depends': ['base', 'estate', 'account'],
```
leclerc-leo: estate already depends on base so there's no need to add it here.
```suggestion
    'depends': ['estate', 'account'],
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+            else:
+                record.state = "sold"
+        return True
+
+    def estate_property_action_cancelled(self):
+        for record in self:
+            if (record.state == "sold"):
+                raise exceptions.UserError("Sold properties cannot be cancelled.")
+            else:
+                record.state = "cancelled"
```
leclerc-leo:
```suggestion
            record.state = "cancelled"
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+
+    best_price = fields.Float(compute='_compute_best_price', string="Best Offer")
+
+    @api.depends('property_offer_ids.price')
+    def _compute_best_price(self):
+        for record in self:
+            if len(record.property_offer_ids) > 0:
+                record.best_price = max(record.property_offer_ids.mapped('price'))
+            else:
+                record.best_price = 0.0
```
leclerc-leo:
```suggestion
            record.best_price = max(0, *record.property_offer_ids.mapped('price'))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
+from odoo import fields, models, api, exceptions, tools
```
leclerc-leo: We always try to import directly the functions/classes when it's from an odoo submodule.
(there was some exceptions like models.Model, the best usually is to look in the codebase what's done for each import before adding them and after a bit you'll usually know the exceptions)
```suggestion
from odoo import fields, models, api, tools
from odoo.exceptions imports UserError, ...
```

leclerc-leo (changes requested):
Good job 👍

PR #1234, opened 2026-04-21
First module created

covey-odoo:
Creation of Estate Module:
- Initial model created
- Initial menu created


estate/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
```
delcourtfl: This line is not needed anymore

estate/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
...
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        string='Type',
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')],
+        help="this is used to indicated the garden orientation"
+    )
```
delcourtfl: It's recommended to have one empty line at the end of each file

estate/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    """,
+
+    'description': """
+        test
+    """,
+
+    'author': "covey",
+    'website': "",
+    'category': 'Tutorials',
+    'version': '0.1',
```
```suggestion
    'version': '19.0.0.1.0',
```

estate/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    'application': True,
+    'installable': True,
+    'depends': ['base'],
```
```suggestion
    'depends': [
        'base',
    ],
```

estate/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    'summary': """
+        test"
+    """,
+
+    'description': """
+        test
+    """,
+
+    'author': "covey",
+    'website': "",
```
```suggestion
    'author': 'covey',
    'website': '',
```
Try to avoid mixing single and double quote when it's not needed

delcourtfl (commented):

Also it seems your first commits have the wrong author set (most probably because your config was not properly set before pushing, might want to check into this as a general git exercise)

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+                float_compare(
+                    record.selling_price,
+                    record.expected_price * 0.9,
+                    precision_digits=2,
+                )
+                < 0
+            ):
+                raise ValidationError(
+                    "Selling price cannot be lower than 90%of expected price!"
+                )
```
delcourtfl: This function is not that readable, I suppose it's what a linter would suggest to avoid too long lines (+80 char), but I think it can be improved in a few ways (avoid inconsistent line breaks, use temp variables, ...), for example:
```suggestion
            if any(
                float_is_zero(value, precision_digits=2)
                for value in [record.selling_price, record.expected_price]
            ):
                raise ValidationError(
                    "Selling price and expected price must be greater than zero."
                )
            if float_compare(
                record.selling_price,
                record.expected_price * 0.9,
                precision_digits=2,
            ) < 0:
                raise ValidationError(
                    "Selling price cannot be lower than 90% of expected price!"
                )
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,59 @@
...
+from odoo import fields, models, api
+
+
+class EstatePropertyOffer(models.Model):
+    _name = "estate.property.offer"
+    _description = "Property Offer"
+
+    _check_offer_price_positive = models.Constraint(
+        "CHECK(price > 0)", "The offer price must be positive."
+    )
```
delcourtfl: Always try to keep the constraints after the fields (see https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions for the model attribute order)

estate/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    "description": """
+        test
+    """,
+    "author": "covey",
+    "website": "",
+    "category": "Tutorials",
+    "version": "19.0.0.1.0",
+    "application": True,
+    "installable": True,
+    "depends": ["base"],
```
```suggestion
    "depends": [
        "base",
    ],
```

delcourtfl (changes requested):

estate/models/estate_property.py
```diff
@@ -0,0 +1,104 @@
...
+    def _onchange_garden(self):
+        for record in self:
+            if record.garden:
+                record.garden_area = 10
+                record.garden_orientation = "north"
+            else:
+                record.garden_area = 0
+                record.garden_orientation = None
+
+    def cancel_sale(self):
```
delcourtfl: Better to explicitly keep action in the name in this case (so that we know it's coming from the ui): `action_cancel_sale`

estate/models/estate_property.py
```diff
@@ -0,0 +1,104 @@
...
+        for record in self:
+            record.state = "cancelled"
+
+    def mark_as_sold(self):
```
delcourtfl: Same as above: `action_mark_as_sold`

estate/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import test_estate_property
```
delcourtfl: :cry:

delcourtfl (changes requested):
Good work, just a few formatting nitpicks.

- one for estate module

(And don't forget to check the runbot style CI 👀)

PR #1235, opened 2026-04-21
sever - Technical Training

SebVrc:
Pull request for the tutorials

### DIFF
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -1,0 +1,18 @@
+{
+    "name": "Estate",
+    "author": "sever",
+    "depends": [
+        "base",
+    ],
+    "application": True,
+    "license": "LGPL-3",
+    "data": [
+        "views/estate_property_tag_views.xml",
+        "views/estate_property_type_views.xml",
+        "views/estate_property_offer_views.xml",
+        "views/estate_property_views.xml",
+        "views/estate_property_users_view.xml",
+        "views/estate_menus.xml",
+        "security/ir.model.access.csv",
+    ],
+}
--- a/estate/models/estate_property.py
+++ b/estate/models/estate_property.py
@@ -1,0 +1,110 @@
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate Property Module"
+    _order = "id desc"
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        default=fields.Date.add(fields.Date.today(), months=3),
+        copy=False,
+    )
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float(readonly=True, copy=False)
+    bedrooms = fields.Integer(default=2)
+    living_areas = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        [("north", "North"), ("south", "South"), ("east", "East"), ("west", "West")],
+    )
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
+        [
+            ("new", "New"),
+            ("offer_received", "Offer Received"),
+            ("offer_accepted", "Offer Accepted"),
+            ("sold", "Sold"),
+            ("cancelled", "Cancelled"),
+        ],
+        default="new",
+        required=True,
+        copy=False,
+    )
+    property_type_id = fields.Many2one("estate.property.type")
+    buyer_id = fields.Many2one("res.partner", copy=False)
+    seller_id = fields.Many2one("res.users", default=lambda self: self.env.user)
+    property_tag_ids = fields.Many2many("estate.property.tag")
+    offer_ids = fields.One2many("estate.property.offer", "property_id")
+    total_area = fields.Integer(compute="_compute_total_area")
+    best_price = fields.Float(compute="_compute_best_price")
+
+    _check_expected_price = models.Constraint(
+        "CHECK(expected_price > 0)",
+        "The expected price must be strictly positive.",
+    )
+    _check_selling_price = models.Constraint(
+        "CHECK(selling_price >= 0)",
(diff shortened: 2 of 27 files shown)


estate/__manifest__.py
```diff
@@ -0,0 +1,5 @@
+{
+    'name': "Estate",
+    'depends': ['base'],
```
```suggestion
    "depends": [
        "base",
    ],
```

estate/__manifest__.py
```diff
@@ -0,0 +1,5 @@
...
+    'application': True
+}
```
delcourtfl: It's recommended to have one empty line at the end of each file

delcourtfl (changes requested):
Hello there ! Just added a few general comments.

estate/models/estate_property.py
```diff
@@ -0,0 +1,84 @@
+from odoo import api, exceptions, fields, models
```
delcourtfl: It's cleaner to only import exceptions which are used in this file:
```suggestion
from odoo import api, fields, models
from odoo.exceptions import UserError
```

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,11 @@
...
+    <menuitem id="estate_menu_root" name="Real Estate">
+        <menuitem id="estate_advertisements_menu" name="Advertisements">
+            <menuitem id="estate_property_menu_action" action="estate_property_action"/>
+        </menuitem>  
+        <menuitem id="estate_settings_menu" name="Settings">
+            <menuitem id="estate_property_type_menu_action" action="estate_property_type_action"/>
+            <menuitem id="estate_property_tag_menu_action" action="estate_property_tag_action"/>
+        </menuitem>  
+    </menuitem>
+</odoo>
```
delcourtfl: Don't forget to add the empty line at the end (even for xml files)

.gitignore
delcourtfl, line 1: You should remove the `.gitignore` from your modifications, a clean way to do it would be to amend the commit which added it in the first place

README.md
```diff
@@ -9,3 +9,4 @@ tutorial's solutions, and one for the
 [Master the Odoo web framework](https://www.odoo.com/documentation/latest/developer/tutorials/master_odoo_web_framework.html)
 tutorial's solutions. For example, `17.0`, `17.0-discover-js-framework-solutions` and
 `17.0-master-odoo-web-framework-solutions`.
+
```
delcourtfl: This can be removed as well

delcourtfl (changes requested):

estate/models/estate_property.py
```diff
@@ -0,0 +1,102 @@
...
+                if len(property.offer_ids) > 0
+                else 0
+            )
+
+    @api.constrains("expected_price", "selling_price")
+    def _check_selling_price(self):
+        for property in self:
+            if (not float_is_zero(property.selling_price, 3) and float_compare(property.selling_price, 0.9 * property.expected_price, 3) == -1):
+                msg = "The selling price cannot be lower than 90 percents of the expected price."
+                raise ValidationError(msg)
```
delcourtfl: Might be a bit too long (linter would not like it ^^). Try to reformat a bit:
```suggestion
            too_low = float_compare(property.selling_price, 0.9 * property.expected_price, 3) == -1
            if not float_is_zero(property.selling_price, 3) and too_low:
                msg = "The selling price cannot be lower than 90 percents of the expected price."
                raise ValidationError(msg)
```

delcourtfl (changes requested):
Hello there ! Good work already, just one nitpick.

estate/models/estate_property.py
SebVrc: Is it because the condition is too long to stand on 1 line or because of the number of tokens ?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,79 @@
...
+        for offer in self:
+            if offer.create_date:
+                offer.validity = int((offer.date_deadline - offer.create_date.date()).days)
+            else:
+                offer.validity = int((offer.date_deadline - fields.Date.today()).days)
+
+    @api.model_create_multi
+    def create(self, vals):
+        for offer in vals:
+            linked_property = self.env["estate.property"].browse(offer["property_id"])
```
delcourtfl, line 53: Avoid making search and browse calls on each iteration of the loop (this is quite important as it could cause performance issues). Batching the search and filtering inside the loop for the relevant results could be a better approach for that.

estate/views/estate_property_users_view.xml
```diff
@@ -0,0 +1,14 @@
...
+        <field name="inherit_id" ref="base.view_users_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//page[@name='preferences']" position="after">
+                <page name="real_estate_properties" string="Real Estate Properties">
+                    <field name="property_ids"/>
+                </page>
+            </xpath>
+        </field>
+    </record>
+</odoo>
```
delcourtfl, line 14: :cry:

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,36 @@
...
+                "quantity": 1,
+                "price_unit": 100,
+            })
+            invoice_lines_vals.append(administrative_fee_line)
+
+            invoice_vals["invoice_line_ids"] = invoice_lines_vals
+            invoice_vals_list.append(invoice_vals)
+
+        # Creating invoice
+        self.env["account.move"].sudo().with_context(default_move_type="out_invoice").create(invoice_vals_list)
```
delcourtfl, line 35: Good job on batching the values in a single create call ! (better for the performances)
But always try to avoid `sudo` if possible, it bypasses the user rights and I don't think it's needed in this case.

estate/models/estate_property.py
```diff
@@ -0,0 +1,110 @@
...
+            self.garden_area = 0
+            self.garden_orientation = None
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_if_new_or_cancelled_property(self):
+        for property in self:
+            if property.state != "new" and property.state != "cancelled":
+                msg = "Only new or cancelled properties can be deleted"
+                raise UserError(msg)
+        return super().unlink()
```
delcourtfl, line 94: You don't need the `super().unlink()` with the `ondelete` decorator

delcourtfl (changes requested):
Hello there ! Good work, not much to say except a few nitpicks.

PR #1236, opened 2026-04-21
19.0 belam technical training

belam-odoo:
# First pull request in my technical training

This PR includes a hello message added and then removed.
And the first chapter of the training : the minimalistic creation of the estate module.


estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
+from odoo import fields, models
+
+
+class EstateProperty(models.Model):
+    _name = "estate_property"
+    _description = "Lorem Ipsum technical description of a Real Estate object"
```
delcourtfl: Try to avoid mixing single and double quote in the same file when it's not needed

estate/security/__init__.py
delcourtfl, line 1: Not sure this should have been added there

estate/__manifest__.py
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "Real Estate",
+    "version": "0.1",
+    "category": "Custom",
+    "summary": "My first application",
+    "installable": True,
+    "application": True,
+    "author": "belam",
+    "depends": {"base"},
```
```suggestion
    "depends": [
        "base",
    ],
```
Keep the depends as a list there. Also most of the time it's better to keep arrays formatted in a way that will limit further diff, and doesn't break the git blame when we need to add some additional dependencies

estate/__manifest__.py
```diff
@@ -8,4 +8,5 @@
     "author": "belam",
     "depends": {"base"},
     "license": "LGPL-3",
+    "data": {"security/ir.model.access.csv"},
```
delcourtfl: Same comment as for the `depends`:
```suggestion
    "data": [
        "security/ir.model.access.csv",
    ],
```

delcourtfl (changes requested):
Hello there ! Good work already, just added a few comments.

estate/models/estate_property.py
delcourtfl: Also the `_name` should be `"estate.property"` to follow the notation guidelines.

Note that the `_description` is used as a technical field as well not as the general description of the model, so for example here it would be better to just set it to `"Estate Property"`

estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import customer, estate_property, estate_property_type, offer, tag
```
delcourtfl: We prefer using one line per import in module `__init__.py`, it's easier to track
```suggestion
from . import customer
from . import estate_property
from . import estate_property_type
from . import offer
from . import tag
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,158 @@
...
+    total_area = fields.Integer(compute="_compute_total_area", string="Total Area")
+    garden_orientation = fields.Selection(
+        string="Garden Orientation",
+        selection=[
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
+        ],
+        help="If you don't know where West is, wait for the sun to go to sleep. Its bedroom lies West.",
```
delcourtfl, line 70: Well that's a good helper ^^

estate/models/estate_property.py
```diff
@@ -0,0 +1,224 @@
...
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for ep in self:
+            ep.best_price = max(ep.offer_ids.mapped("price")) if ep.offer_ids else None
+
+    @api.onchange("has_garden")
+    def _onchange_has_garden(self):
+        if self.has_garden:
+            self.garden_area = 10
+            self.garden_orientation = "north"
```
delcourtfl, line 141: Might be better to have an else condition here, so that if the property `has_garden` is False, the area and orientation values are set back to their default values

estate/models/estate_property.py
```diff
@@ -0,0 +1,224 @@
...
+                float_compare(
+                    ep.selling_price,
+                    lowest_selling_price,
+                    precision_digits=2,
+                )
+                == -1
+            ):
+                raise ValidationError(
+                    "Selling price must be at least 90% of expected price",
+                )
```
delcourtfl, line 122: This function is not that readable, I suppose it's what a linter would suggest to avoid too long lines (+80 char), but I think it can be improved in a few ways (avoid inconsistent line breaks, use temp variables, ...), 
```suggestion
            if float_is_zero(ep.selling_price, precision_digits=2):
                continue
            lowest_selling_price = ep.expected_price * 0.9
            if float_compare(
                record.selling_price,
                lowest_selling_price
                precision_digits=2,
            ) == -1:
                raise ValidationError(
                    "Selling price cannot be lower than 90% of expected price"
                )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,224 @@
...
+
+    @api.constrains("offer_ids.status")
```
delcourtfl: Don't think this would work, might want to override the `write()` call instead or move the logic to the offer itself. From the doc:
`@constrains only supports simple field names, dotted names (fields of relational fields e.g. partner_id.customer) are not supported and will be ignored.`

estate/models/estate_property.py
```diff
@@ -0,0 +1,224 @@
...
+    def _check_state(self):
+        # todo see how this needs to be done
+        for ep in self:
+            match ep.state:
+                case "accepted":
+                    pass
+                case "sold":
+                    pass
+                case "cancelled":
+                    pass
```
delcourtfl: Not sure what you were trying to do here, avoid pushing unfinished logic

estate/models/estate_property.py
```diff
@@ -0,0 +1,224 @@
...
+            ep.selling_price = accepted_offers[0].price
+
+        return True
+
+    def action_set_cancelled(self):
+        # todo: split action and validation (validation goes into python constraint)
+        for ep in self:
+            # todo later : add warning - dont know how to do this from the model
+            if ep.state == "sold":
+                raise UserError(_("A cancelled property cannot be sold"))
```
delcourtfl, line 177:
```suggestion
                raise UserError(_("A sold property cannot be cancelled"))
```

estate/models/offer.py
```diff
@@ -0,0 +1,99 @@
...
+        comodel_name="estate.property",
+        string="Estate Property",
+        required=True,
+    )
+    partner_id = fields.Many2one("res.partner", string="Customer", required=True)
+    price = fields.Float(string="Price", required=True)
+    _check_expected_price_positive = models.Constraint(
+        "CHECK(price > 0)",
+        "Selling price must be positive",
+    )
```
delcourtfl: Always try to keep the fields declaration before the constraints (see https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions for the model attribute order)

estate/models/offer.py
```diff
@@ -0,0 +1,99 @@
...
+                raise UserError(_("This offer has already been %s", o.status))
+            o.status = "refused"
+        return True
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_by_user(self):
+        for o in self:
+            if o.status == "accepted":
+                raise UserError(_("An accepted offer cannot be deleted."))
+            super().unlink()
```
delcourtfl: You don't need the `super().unlink()` with the `ondelete` decorator

estate/views/__init__.py
delcourtfl, line 1: Should be removed, this is not needed in the `views` folder

estate/views/customer_views.xml
```diff
@@ -0,0 +1,52 @@
...
+                            <field name="phone"/>
+                            <field name="email"/>
+                        </group>
+                    </group>
+                </sheet>
+            </form>
+        </field>
+    </record>
+
+</odoo>
```
delcourtfl: Always keep an empty line at the end, even for xml files

estate/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "Real Estate",
+    "version": "0.1",
```
```suggestion
    "version": "19.0.0.1.0",
```

README.md
```diff
@@ -9,3 +9,4 @@ tutorial's solutions, and one for the
 [Master the Odoo web framework](https://www.odoo.com/documentation/latest/developer/tutorials/master_odoo_web_framework.html)
 tutorial's solutions. For example, `17.0`, `17.0-discover-js-framework-solutions` and
 `17.0-master-odoo-web-framework-solutions`.
+
```
delcourtfl: You can remove this line now, try to amend the commit which added it

estate/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    "category": "Custom",
+    "summary": "My first application",
+    "installable": True,
+    "application": True,
+    "author": "belam",
+    "depends": ["base"],
```
delcourtfl: Most of the time it's better to keep arrays formatted in a way that will limit further diff, and doesn't break the git blame when we need to add some additional dependencies
```suggestion
    "depends": [
        "base",
     ],
```

delcourtfl (changes requested):
Hello there ! Good work already, just a few comments/nitpicks.

(Also check the runbot style CI :eyes:)

estate/models/estate_property.py
```diff
@@ -0,0 +1,177 @@
...
+            if not accepted_offer:
+                raise UserError(_("You must accept an offer before selling the property."))
+
+            ep.write(
+                {
+                    "state": "sold",
+                    "customer_id": accepted_offer[0].partner_id.id,
+                    "selling_price": accepted_offer[0].price,
+                },
+            )
```
delcourtfl: Cleaner formatting:
```suggestion
            ep.write({
                "state": "sold",
                "customer_id": accepted_offer[0].partner_id.id,
                "selling_price": accepted_offer[0].price,
            })
```

estate/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import test_estate_property
```
delcourtfl: 😢

delcourtfl (changes requested):
Good work, small review here for the formatting.

- one for estate module

(And don't forget to check the runbot style CI 👀)

PR #1237, opened 2026-04-21
sevan - Technical Training

SebVde:
Initialized the Sevan Estate module as part of the Odoo technical training.
The module is set as an installable application.


estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_property
```
vandroogenbd: It is good practice to add a trailing empty line to your files

https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file

estate/models/estate_property.py
```diff
@@ -0,0 +1,26 @@
+from odoo import fields, models
+
+class Property(models.Model):
```
vandroogenbd: By convention, we leave 2 empty lines before Python class declarations

estate/models/estate_property.py
```diff
@@ -0,0 +1,26 @@
...
+
+    # Model definition
+    _name = "estate.property"
+    _description = "Estate Property"
+
+    # Fields
+    name = fields.Char(required=True)
```
vandroogenbd: Nice one on separating private attributes from fields 👍

estate/models/estate_property.py
```diff
@@ -0,0 +1,26 @@
...
+    bedrooms = fields.Integer()
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        string='Orientation',
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
+    )
```
vandroogenbd, line 47: This is correct, but you could also write it down as

```
  string='Orientation',
  selection=[
    ('north', 'North'),
    ('south', 'South'),
    ('east', 'East'),
    ('west', 'West'),
  ],
)
```

Although this approach takes more lines, it has one advantage. Can you guess what? 😉

estate/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
vandroogenbd: Same empty line comment

estate/__manifest__.py
```diff
@@ -0,0 +1,8 @@
+{
+    'name': "Sevan Estate",
+    'depends': ['base'],
+    'summary': "Just a simple estate app for Sevan.",
+    'application': True,
+    'installable': True,
+    'author': "Sevan Corp.",
+}
```
vandroogenbd: Same empty line comment

vandroogenbd (commented):
Good start, here's a few comments 🙂

estate/models/estate_property.py
SebVde: Readability I suppose?

estate/models/estate_property.py
```diff
@@ -0,0 +1,110 @@
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
```
vandroogenbd: Order is off (see [doc](https://www.odoo.com/documentation/18.0/contributing/development/coding_guidelines.html#imports))

```suggestion
from datetime import datetime
from dateutil.relativedelta import relativedelta

from odoo import api, fields, models
from odoo.exceptions import UserError, ValidationError
from odoo.tools.float_utils import float_compare, float_is_zero
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,110 @@
...
+            ('cancelled', 'Cancelled'),
+        ],
+        required=True,
+        copy=False,
+        default='new',
+        store=True,
+        compute='_compute_if_offer',
+    )
+
+    _check_expected_price = models.Constraint(
```
vandroogenbd: Constraints should be after computes (see [doc](https://www.odoo.com/documentation/18.0/contributing/development/coding_guidelines.html#imports:~:text=In%20a%20Model%20attribute%20order%20should%20be))

estate/models/estate_property.py
```diff
@@ -0,0 +1,110 @@
...
+        'CHECK(expected_price > 0)',
+        'The expected price should be strictly positive.',
+    )
+
+    @api.constrains('expected_price', 'selling_price')
+    def _check_selling_price_above_90(self):
```
vandroogenbd: This makes it sound like the selling price should be above 90 as a flat value

```suggestion
    def _check_selling_price_to_expected_price_ratio(self):
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids")
+    def _compute_best_price(self):
+        for record in self:
+            record.best_price = max(record.offer_ids.mapped("price")) if record.offer_ids else 0.0
```
vandroogenbd, line 79: Clear and concise 👌

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+    @api.onchange("garden")
+    def _onchange_garden_values(self):
+        self.garden_area = 10 if self.garden else 0
+        self.garden_orientation = 'north' if self.garden else None
+
+    def action_cancel_property(self):
+        if self.state == 'sold':
+            raise UserError("You can't cancel a property that is already sold.")
+        self.state = 'cancelled'
+        return True
```
vandroogenbd: In actions, we usually do a call to write. This way it is easily expandable.

```suggestion
        return self.write({
            'state': 'cancelled',
        })
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class Property(models.Model):
+    # Model definition
+    _name = "estate.property"
+    _description = "Estate Property"
+    _order = "id desc"
+
+    # Fields
+    name = fields.Char('Title', required=True)
```
vandroogenbd: Try to stay consistent with `'` and `"` (for the whole file)

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids")
+    def _compute_best_price(self):
+        for record in self:
+            record.best_price = max(record.offer_ids.mapped("price")) if record.offer_ids else 0.0
+
+    @api.depends("offer_ids")
+    def _compute_if_offer(self):
```
vandroogenbd: This name doesn't really make sense, you're computing the state.

```suggestion
    def _compute_state(self):
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+        for record in self:
+            if (record.state == 'new' and len(record.offer_ids) > 0):
+                record.state = 'offer_received'
+
+    @api.onchange("garden")
+    def _onchange_garden_values(self):
+        self.garden_area = 10 if self.garden else 0
+        self.garden_orientation = 'north' if self.garden else None
```
vandroogenbd: Tricky (?) question time!

At first glance, this method could also be an `inverse`. Can you tell why it makes more sense for it to be an `onchange`?

<details>
  <summary>Answer</summary>
  
`inverse` are triggered whenever the field is written. `onchange` are [called by form views](https://www.odoo.com/documentation/18.0/developer/reference/backend/orm.html#odoo.api.onchange).

It doesn't make much sense to set arbitrary default values when programmatically updating records, so `onchange` is the correct approach here.
  
</details>

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+        if self.state == 'sold':
+            raise UserError("You can't cancel a property that is already sold.")
+        self.state = 'cancelled'
+        return True
+
+    def action_sold_property(self):
+        if self.state == 'cancelled':
+            raise UserError("You can't sell a property that is cancelled.")
+        self.state = 'sold'
+        return True
```
vandroogenbd: Same as above

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+                record.state = 'offer_received'
+
+    @api.onchange("garden")
+    def _onchange_garden_values(self):
+        self.garden_area = 10 if self.garden else 0
+        self.garden_orientation = 'north' if self.garden else None
+
+    def action_cancel_property(self):
+        if self.state == 'sold':
+            raise UserError("You can't cancel a property that is already sold.")
```
vandroogenbd: Try to make the comments more "generic" (use passive tense when possible eg `you can't do x` => `x is not allowed`)

```suggestion
            raise UserError("Sold properties cannot be cancelled.")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+        return True
+
+    def action_sold_property(self):
+        if self.state == 'cancelled':
+            raise UserError("You can't sell a property that is cancelled.")
+        self.state = 'sold'
+        return True
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_except_new_or_cancelled(self):
```
vandroogenbd: `_unlink` is a generic function of models. Since you are overriding it (you call `super().unlink()` eventually), you should keep the name.

```suggestion
    def _unlink(self):
```

The fact that there is an override of `_unlink` is enough to tell the user that there is some additional condition.

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+        if self.state in ('new', 'cancelled'):
+            raise UserError("Can't delete a new or cancelled property.")
+        return super().unlink()
```
vandroogenbd: What about unlinking multiple records using the list view? The function needs to be adapted to match such a common use case 😉

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,80 @@
...
+    property_id = fields.Many2one(
+        "estate.property",
+        string="Property",
+        required=True,
+    )
+    validity = fields.Integer("Validity (days)", default=7)
+    date_deadline = fields.Date("Deadline", compute="_compute_deadline", inverse="_inverse_deadline")
+    property_type_id = fields.Many2one("estate.property.type", store=True, related="property_id.property_type_id")
+
+    _check_price = models.Constraint(
```
vandroogenbd: Same comment about ordering

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,80 @@
...
+        'CHECK(price > 0.00)',
+        "The offer's amount should be strictly positive.",
+    )
+
+    @api.depends("validity")
+    def _compute_deadline(self):
+        for record in self:
+            if not record.create_date:
+                record.create_date = datetime.today()
+            record.date_deadline = record.create_date + relativedelta(days=record.validity)
```
vandroogenbd: I get why you want to do this, but it's not a good idea. Don't change the value of another field in a compute, this is opening the door for inconsistent behaviour.

```suggestion
        for record in self:
            date = datetime.today() if not record.create_date else record.create_date.date()
            record.date_deadline = date + relativedelta(days=record.validity)
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def _inverse_deadline(self):
+        for record in self:
+            if not record.create_date:
```
vandroogenbd: Same comment here

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,80 @@
...
+                record.create_date = datetime.today()
+            record.validity = (record.date_deadline - record.create_date.date()).days
+
+    def action_accept_offer(self):
+        if any(record != self and record.status == 'accepted' for record in self.property_id.offer_ids):
+            raise UserError("Only one offer can be accepted")
+        if (self.property_id.garden_orientation == 'south' and
+        float_compare(self.property_id.expected_price, self.price, 2) == 1):
+            raise ValidationError("For properties South oriented gardens, only offers having a price higher than the expected value of the property can be accepted")
```
vandroogenbd: 😉

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,80 @@
...
+            record.date_deadline = record.create_date + relativedelta(days=record.validity)
+
+    def _inverse_deadline(self):
+        for record in self:
+            if not record.create_date:
+                record.create_date = datetime.today()
+            record.validity = (record.date_deadline - record.create_date.date()).days
+
+    def action_accept_offer(self):
+        if any(record != self and record.status == 'accepted' for record in self.property_id.offer_ids):
```
vandroogenbd: The `record != self` part is useless, because there is not point in accepting an offer twice anyways.

Here's another way to do the same for reference.

```suggestion
        if "accepted" in self.mapped("property_id.offer_ids.state"):
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,80 @@
...
+                record.create_date = datetime.today()
+            record.date_deadline = record.create_date + relativedelta(days=record.validity)
+
+    def _inverse_deadline(self):
+        for record in self:
+            if not record.create_date:
+                record.create_date = datetime.today()
+            record.validity = (record.date_deadline - record.create_date.date()).days
+
+    def action_accept_offer(self):
```
vandroogenbd: You should iterate over self, what if the action is called on multiple records from the list view?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,80 @@
...
+        if any(record != self and record.status == 'accepted' for record in self.property_id.offer_ids):
+            raise UserError("Only one offer can be accepted")
+        if (self.property_id.garden_orientation == 'south' and
+        float_compare(self.property_id.expected_price, self.price, 2) == 1):
+            raise ValidationError("For properties South oriented gardens, only offers having a price higher than the expected value of the property can be accepted")
+        self.status = 'accepted'
+        self.property_id.buyer = self.partner_id
+        self.property_id.selling_price = self.price
+        self.property_id.state = 'offer_accepted'
+        return True
```
vandroogenbd: Again, try to use `write` in such cases, to commit all of the changes at once.
```suggestion
        return self.property_id.write({
            "state": "offer_accepted",
            "selling_price": self.price,
            "buyer": self.partner_id,
        })
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+        selection=[
+            ('north', 'North'),
+            ('south', 'South'),
+            ('east', 'East'),
+            ('west', 'West'),
+        ],
+    )
+    total_area = fields.Integer("Total Area (sqm)", compute="_compute_total_area")
+    property_type_id = fields.Many2one("estate.property.type", string="Property Type")
+    buyer = fields.Many2one("res.partner", copy=False)
```
vandroogenbd: Many2one fields are suffixed with `_id` to identify them quickly.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def action_refuse_offer(self):
+        for record in self:
+            if record.status == 'accepted':
+                record.property_id.buyer = None
```
vandroogenbd: Same as above, write the changes

estate/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    'license': 'LGPL-3',
+    'depends': ['base'],
+    'summary': "Just a simple estate app for Sevan.",
+    'application': True,
+    'installable': True,
+
+    'data': [
+        'security/ir.model.access.csv',
+
+        # Be careful with the order!
```
vandroogenbd, line 12: 👀

estate_account/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+{
+    'name': "Estate Account",
+    'license': 'LGPL-3',
+    'depends': ['estate', 'account'],
+    'summary': "Link module to be able to create invoices for sold estate properties.",
+    'application': True,
+    'installable': True,
+
+    # 'data': [
```
vandroogenbd: 👀

estate_account/models/estate_account.py
```diff
@@ -0,0 +1,34 @@
...
+    _inherit = "estate.property"
+
+    def action_sold_property(self):
+        invoice_vals_list = []
+        for sold_property in self:
+            sold_to_partner_id = None
+            for offer in sold_property.offer_ids:
+                if offer.status == 'accepted':
+                    sold_to_partner_id = offer.partner_id
+                    break
```
vandroogenbd: Can't you just use `sold_property`'s `buyer_id` here?

PR #1238, opened 2026-04-21
[ADD] module: add new module real estate (Onboarding)

yogietama:
Create Real Estate Module for onboarding program


estate/__manifest__.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
```
delcourtfl: This line is not needed anymore

estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_property
```
delcourtfl: It's recommended to have one empty line at the end of each file

estate/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+
+{
+    'name': 'Real Estate',
+    'version': '1.9',
```
```suggestion
    'version': '19.0.0.1.0',
```

estate/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+    'category': 'Real Estate/Properties',
+    'sequence': 15,
+    'summary': 'Track leads and close opportunities',
+    'website': 'https://www.odoo.com/app/estate',
+    'depends': ['base'],
```
```suggestion
    'depends': [
        'base',
    ],
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,55 @@
...
+            name="Real Estate"
+            action="estate_property_action"
+            parent="menu_real_estate"
+            sequence="2"/>
+        
+        <menuitem id="property_menu"
+            name="Real Estate"
+            action="estate_property_action"
+            parent="menu_real_estate"
+            sequence="2"/> -->
```
delcourtfl: Avoid pushing commented code

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,55 @@
...
+            </field>
+        </record>
+
+        <menuitem name="Real Estate"
+            id="menu_real_estate"
+            sequence="20"
+            web_icon="estate,static/description/icon.png"
+            groups="base.group_user,base.group_partner_manager"/>
+
+        <!--  -->
```
delcourtfl: Might not be needed as well

estate/models/estate_property.py
```diff
@@ -0,0 +1,29 @@
+from odoo import fields, models
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
```
delcourtfl, line 7: Try to avoid mixing single and double quote in the same file when it's not needed

estate/models/estate_property.py
```diff
@@ -0,0 +1,29 @@
...
+    _description = "Real Estate Property"
+
+    name = fields.Char(required = True, default = "Unknown")
```
delcourtfl:
```suggestion
    name = fields.Char(required=True, default="Unknown")
```
Extra spaces are not needed here

estate/models/estate_property.py
```diff
@@ -0,0 +1,29 @@
...
+    description = fields.Text()
+    postcode = fields.Char()
+    date_avaibility = fields.Date( 
+        string = 'Date Avaibility'
+    )
```
delcourtfl: Same here for the extra spaces, also the line break might be a bit too much. Also note that for non relational fields (and non selection fields) the change can be reduced to:
```suggestion
    date_availability = fields.Date('Date Availability')
```
But for simple field such as these, the string might not be needed at all as they are automatically generated from the variable name

estate/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+        'views/estate_property_views.xml'
+    ],
+    'demo': [],
+    'installable': True,
+    'application': True,
+    'assets': {
+        'web.assets_backend': [],
+        'web.assets_backend_lazy': [],
+        'web.assets_tests': [],
+        'web.assets_unit_tests': [],
```
delcourtfl: Try to reduce the diff if possible, these are not needed in this case (as you don't add anything for now)

delcourtfl (changes requested):

delcourtfl:
Hello there ! Good work already, just added a few comments.

Also it seems your first commit has the wrong author set (most probably because your config was not properly set before pushing, might want to check into this as a general git exercise)

estate/models/estate_property.py
```diff
@@ -0,0 +1,92 @@
...
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate Property"
+
+    # define constrain
+    _check_expected_price = models.Constraint(
+        'CHECK(expected_price > 0)',
+        'The expected price must be strictly positive',
+    )
```
delcourtfl: Always try to keep the fields declaration before the constraints (see https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions for the model attribute order)

estate/models/estate_property.py
```diff
@@ -0,0 +1,92 @@
...
+    def _check_price_offer(self):
+        m1 = "The selling price cannot be lower than 90% of the expected price"
+        m2 = "You must reduce the expected price if you want to accept this offer"
+        for record in self:
+            if record.selling_price > 0:
+                offer_percentage = record.selling_price / record.expected_price
+                # if offer percentage lower then 90%
+                if float_compare(offer_percentage, 
+                                 0.9, precision_digits=2) == -1:
+                    raise ValidationError(f"{m1}. {m2}")
```
delcourtfl: I believe it's better to keep the error message in the error call (or just beside it, unless it repeats) to avoid having to look back and forth in the code to understand the issue.
```suggestion
                    raise ValidationError(
                           "The selling price cannot be lower than 90% of the expected price\n"
                           "You must reduce the expected price if you want to accept this offer"
                    )
```

delcourtfl: I believe in practice it would be better to keep the multiplication instead of the division to avoid x / 0 issue (`expected_price` is not checked here). Also try using the `float_is_zero` to ensure there is no issue with floats (and you can format function calls in a cleaner way):
```suggestion
            if float_is_zero(record.selling_price, precision_digits=2):
                continue
            min_price = 0.9 * property.expected_price
            if float_compare(record.selling_price, min_price, precision_digits=2) == -1:
                raise ValidationError("...")
```

estate/models/estate_propery_offer.py
```diff
@@ -0,0 +1,65 @@
...
+    def _inverse_deadline(self):
+        for record in self:
+            if record.deadline:
+                diff = record.deadline - fields.Date.today()
+                record.validity_days = diff.days
+
+    def action_accepted(self):
+        for record in self:
+            if record.property_id.selling_price == 0:
+                record.status = "accepted"
```
delcourtfl, line 62: Check if it doesn't already have an accepted offer ?

estate/models/estate_propery_offer.py
```diff
@@ -0,0 +1,65 @@
...
+                record.property_id.selling_price = record.price
+                record.property_id.partner_id = record.partner_id
+            # else mean, property already sold
+            else:
+                return
```
delcourtfl: Some cleanup might be needed here

estate/models/estate_propery_offer.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def action_refused(self):
+        for record in self:
+            if record.status == "accepted":
+                raise ValidationError(
+                    "You're not eligible to refused an accepted offer")
```
delcourtfl: I prefer this notation but I think both are okay in the codebase
```suggestion
                raise ValidationError(
                    "You're not eligible to refused an accepted offer"
                )
```

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,1
+estate.access_estate_property_type,access_estate_property_type,estate.model_estate_property_type,base.group_user,1,1,1,1
+estate.access_estate_property_tag,access_estate_property_tag,estate.model_estate_property_tag,base.group_user,1,1,1,1
+estate.access_estate_property_offer,access_estate_property_offer,estate.model_estate_property_offer,base.group_user,1,1,1,1
```
delcourtfl: New line needed here 👀 (it's the case in all odoo files)

estate/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+
+{
+    'name': 'Real Estate',
+    'version': '19.0.0.1.0',
+    'category': 'Real Estate/Properties',
+    'sequence': 15,
+    'summary': 'Track leads and close opportunities',
+    'website': 'https://www.odoo.com/app/estate',
+    'depends': [
+        'base'
```
delcourtfl: Nitpicking, add what is needed to easily add new elements later on:
```suggestion
        'base',
```

delcourtfl (changes requested):

(Might also need to check runbot style CI for some additional issues)

estate/models/estate_propery_offer.py
yogietama: done updated

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+                # if offer percentage lower then 90%
+                if float_compare(record.selling_price, min_price, precision_digits=2) == -1:
+                    raise ValidationError(
+                           "The selling price cannot be lower than 90% of the expected price\n"
+                           "You must reduce the expected price if you want to accept this offer"
+                    )
+
+    def action_sold(self):
+        for record in self:
+            if record.state == "sancelled":
```
delcourtfl: Small typo there
```suggestion
            if record.state == "cancelled":
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def action_cancel(self):
+        for record in self:
+            if record.state == "sold":
+                raise UserError("Sold properties cannot be cancelled")
+            record.state = "cancelled"
+
+    @api.model
+    def create(self, vals_list):
+        return super().create(vals_list)
```
delcourtfl: Better to remove if unused

estate/models/estate_propery_offer.py
```diff
@@ -0,0 +1,91 @@
...
+            if record.property_id.garden_orientation == "south" and record.property_id.expected_price > record.price:
+                raise ValidationError("The south-facing garden can only be accepted if above expected price")
+
+            if record.property_id.selling_price == 0:
+                record.status = "accepted"
+                record.property_id.selling_price = record.price
+                record.property_id.partner_id = record.partner_id
+                record.property_id.state = "offer accepted"
+
+    def refused_offer(self):
```
delcourtfl: Better to explicitly keep action in the name in this case: `action_refuse_offer`

estate/models/estate_propery_offer.py
```diff
@@ -0,0 +1,91 @@
...
+                )
+            record.status = "refused"
+
+    @api.model
+    def create(self, val_list):
+        for vals in val_list:
+            property_id = vals.get('property_id')
+
+            if property_id:
+                property_record = self.env['estate.property'].browse(property_id)
```
delcourtfl: Avoid making search and browse calls on each iteration of the loop (this is quite important as it could cause performance issues). Batching the search and filtering inside the loop for the relevant results could be a better approach for that.

estate/views/estate_property_menu.xml
```diff
@@ -0,0 +1,48 @@
...
+        />
+        <menuitem 
+            id = "estate_property_tags_menu"
+            name = "Property Tags"
+            parent = "estate_menu_settings"
+            action = "estate_property_tag_action"
+            sequence = "2"
+        />
+    </data>
+</odoo>
```
delcourtfl: 😢

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,27 @@
...
+                        'quantity': 1,
+                        'price_unit': record.selling_price * 0.06,
+                    }),
+                    Command.create({
+                        'name': 'Administrative Fees',
+                        'quantity': 1,
+                        'price_unit': 100.00,
+                    }),
+                ],
+            })
```
delcourtfl: For performance you could also avoid creating in the loop, you can make a list of values to create and call the `create` on all of them at once at the end.

delcourtfl (changes requested):
Hello there ! Really good work, not much to say, you got the general idea for module development.

- one for estate module

(And don't forget to check the runbot style CI :eyes:)

yogietama:
Hi @delcourtfl ,
I've already updated and improved the code to address potential performance issues. I also squashed the commits into the estate and estate_account modules. The Runbot style checks passed!  :)

Thank you so much for checking my PR. I really appreciate you taking the time to review it and giving detailed feedback on the parts I missed.

image

PR #1240, opened 2026-04-21
lugva - Technical Training

lugva-odoo:


estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,1
+estate.access_estate_property_type,access_estate_property_type,estate.model_estate_property_type,base.group_user,1,1,1,1
+estate.access_estate_property_tag,access_estate_property_tag,estate.model_estate_property_tag,base.group_user,1,1,1,1
+estate.access_estate_property_offer,access_estate_property_offer,estate.model_estate_property_offer,base.group_user,1,1,1,1
```
delcourtfl: It's recommended to have one empty line at the end of each file

estate/models/estate_property.py
```diff
@@ -0,0 +1,65 @@
...
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer("Living Area (sqm)")
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer("Garden Area (sqm)")
+
+    garden_orientation = fields.Selection(
+        selection=[
+            ('north', 'North'),
```
delcourtfl: Try to avoid mixing single and double quote in the same file when it's not needed

estate/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+{
+    'name': 'Real Estate',
+    'depends': [
+        'base',
+    ],
```
delcourtfl, line 5: Nice you did the proper list formatting, this will limit further diff, and doesn't break the git blame when we need to add some additional dependencies

estate/models/estate_property.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+
+    property_type_id = fields.Many2one(
+        "estate.property.type", string="Property Type"
+    )
+    buyer_id = fields.Many2one(
+        "res.partner", string="Buyer", copy=False
+    )
+    salesperson_id = fields.Many2one(
+        "res.users", string="Salesperson", default=lambda self: self._uid
```
delcourtfl: Don't think it would work as expected, use the environment for this:
```suggestion
        "res.users", string="Salesperson", default=lambda self: self.env.user
```

delcourtfl (changes requested):

estate/models/estate_property.py
delcourtfl: You can check the runbot for this, it raises a `DeprecationWarning`

lugva-odoo: thanks! I actually tried both ways, and using `_uid` by default I had Administrator, and by using `env.user` I got System. On the top right of Odoo I am visible as administrator, so I thought `_uid` was correct. I will fix it to `env.user`

delcourtfl: Strange, it should give you the administrator in both ways, but maybe you activated the superuser option in the debug tools at some point ?

lugva-odoo: yes I did indeed, so this is the reason
Thank you!

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
+from dateutil.relativedelta import relativedelta
+
+
```
delcourtfl: Don't need to add two empty lines between imports, it's mostly done before classes

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+
+    total_area = fields.Integer("Total area", compute="_compute_total_area")
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    best_price = fields.Float("Best offer", compute="_compute_best_price")
```
delcourtfl: Always try to keep the fields declaration before the computation logic, same for the constraints (see https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions for the model attribute order)

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,14 @@
...
+
+    <!-- First Item in the top bar -->
+    <menuitem id="estate_first_level_menu" name="Advertisements" parent="estate_menu_root" />
+    <menuitem id="estate_property_menu_action" action="estate_property_action" parent="estate_first_level_menu"/>
+
+    <!-- Second Item in the top bar -->
+    <menuitem id="estate_second_level_menu" name="Settings" parent="estate_menu_root" />
+    <menuitem id="estate_property_type_menu_action" action="estate_property_type_action" parent="estate_second_level_menu"/>
+    <menuitem id="estate_property_tag_menu_action" action="estate_property_tag_action" parent="estate_second_level_menu"/>
+</odoo>
```
delcourtfl: Having an empty line at the end is also true for xml files

estate/views/estate_property_type_view_form.xml
```diff
@@ -0,0 +1,41 @@
...
+                    <button
+                        type="object"
+                        name="action_view_offers"
+                        string="Offers"
+                        class="oe_stat_button"
+                        icon="fa-money"
+                    >
+                        <field
+                            name="offer_count"
+                        />
```
delcourtfl: The break might be a bit overkill here (even if it's true than if we add some attributes to the field it would make the diff cleaner):
```suggestion
                        <field name="offer_count"/>
```

.gitignore
delcourtfl, line 1: You should remove the `.gitignore` from your modifications, a clean way to do it would be to amend the commit which added it in the first place

delcourtfl (changes requested):

(Also you should update the title of your first commit to make the runbot CI style happy)

.gitignore
lugva-odoo: Understand, thank you! What should I do now? Should I remove the changes I did and make a new commit, and then amend?

delcourtfl: You don't really need to create a new commit, a better way would be to amend the original by doing an interactive rebase.
For example you could do:
-  `git rebase -i HEAD~n` to try to rebase the last `n` commits and set 'edit' to the one you want in the editor
- Modify everything you need in the commit
- `git add .` to add the modifications
- `git commit --amend` to apply the changes
- `git rebase --continue` to finish the rebase
- `git push dev --force-with-lease` to update the remote

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,108 @@
...
+    # Make sure that price is positive
+    _check_price = models.Constraint(
+        "CHECK(price > 0)",
+        "Expected price must be positive"
+    )
+
+    property_type_id = fields.Many2one(
+        related="property_id.property_type_id",
+        store=True,
+    )
```
delcourtfl: Follow the recommended ordering, the constraints would be better under the fields

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,108 @@
...
+            )
+            if exists_higher_offer:
+                raise UserError("Can't create an offer with a lower price")
+
+            # change status to received
+            current_property = self.env["estate.property"].browse(
+                vals["property_id"])
+            current_property.state = "received"
+
+        return super().create(vals_list)
```
delcourtfl, line 114: This would work, but it should be possible to improve this (performance-wise) by avoiding the `search-count` and `browse` calls on each iteration of the loop.

delcourtfl (commented):
Hello there ! Really good work, not much to say, you got the general idea for module development.

A test was added to your branch, if you have some extra time you can try to run/adapt it to ensure it works properly on your implementation.

- one for estate module

estate/models/estate_property_offer.py
lugva-odoo: yeah I thought about it as well, but I thought that "maybe" there could be a situation where different elements in `vals_list` would have a different `property_id`, but indeed it should be optimised

delcourtfl: Yes it should be taken into account as well, batching the search and filtering inside the loop for the relevant results could be a good approach for that (but maybe it was a bit too much for the tutorial).

PR #1241, opened 2026-04-21
tutorial server framework joemo

djoewie:


estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_property
```
cgun-odoo: Don't forget to have an empty line at the end of every file. This will be automatic if you use a formatter like Ruff

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,1
```
```suggestion
access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
```
You don't need to start with the module name here as that will be done automatically.

estate/models/estate_property.py
```diff
@@ -0,0 +1,22 @@
+from odoo import fields, models
+
+
+class EstateProperty(models.Model):
+    _name = 'estate_property'
```
cgun-odoo: For technical names we separate words with .
```suggestion
    _name = 'estate.property'
```
https://www.odoo.com/documentation/18.0/contributing/development/coding_guidelines.html#symbols-and-conventions

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
cgun-odoo: No need to have this line for xml files
```suggestion
```

estate/models/__init__.py
cgun-odoo: You can check which files are missing the line on the PR

estate/models/estate_property.py
```diff
@@ -0,0 +1,22 @@
...
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float()
+    bedrooms = fields.Integer()
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
```
```suggestion
        [('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
```
The first positional argument is already 'selection' so it's redundant to have it.

estate/security/ir.model.access.csv
cgun-odoo: Important thing to notice: You need to keep 'base' in `base.group_user` because it's a different module. But if you just write `model_estate_property` odoo knows to look for that in the `estate` module.

estate/models/estate_property.py
```diff
@@ -1,22 +1,51 @@
...
+from odoo import api, fields, models
 
 
 class EstateProperty(models.Model):
-    _name = 'estate_property'
+    _name = 'estate.property'
     _description = 'Estate Property'
 
-    name = fields.Char(required=True)
+    name = fields.Char('Title', required=True)
```
cgun-odoo: For strings, as far as I know in R&D they put user facing strings in double quotes and technical strings in single quotes. In PS-Tech we put everything in double quotes.
Again something that will easily be fixed with a formatter configuration

estate/models/estate_property.py
```diff
@@ -1,22 +1,51 @@
...
     description = fields.Text()
     postcode = fields.Char()
-    date_availability = fields.Date()
+    date_availability = fields.Date('Available From', default=fields.Date.today() + relativedelta(months=3), copy=False)
```
cgun-odoo: Here you're passing a value as default. This value will be calculated once when you run the server. 
However, what you want to do is to pass a function, so that the value is calculated every time the default value is needed.

estate/models/estate_property.py
```diff
@@ -1,22 +1,51 @@
...
     garden_orientation = fields.Selection(
-        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
+        [('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
+    )
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
+        [('new', 'New'), ('offer', 'Offer Received'), ('accepted', 'Offer Accepted'), ('sold', 'Sold'), ('canceled', 'Cancelled')],
+        default='new'
     )
+    property_type_id = fields.Many2one('estate.property.type', 'Property Type')
```
cgun-odoo: No need to add the string here, the default string for `property_type_id` is already Property Type

estate/models/estate_property.py
```diff
@@ -0,0 +1,51 @@
...
+    garden_orientation = fields.Selection(
+        [('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
+    )
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
+        [('new', 'New'), ('offer', 'Offer Received'), ('accepted', 'Offer Accepted'), ('sold', 'Sold'), ('canceled', 'Cancelled')],
+        default='new'
+    )
+    property_type_id = fields.Many2one('estate.property.type', 'Property Type')
+    sales_person_id = fields.Many2one('res.users', "Salesman", copy=False, default=lambda self: self.env.user)
```
```suggestion
    salesperson_id = fields.Many2one('res.users',  copy=False, default=lambda self: self.env.user)
```
salesperson is one word i guess, not super sure

estate/models/estate_property.py
```diff
@@ -0,0 +1,51 @@
...
+    buyer_id = fields.Many2one('res.partner', 'Buyer', copy=False)
```
```suggestion
    buyer_id = fields.Many2one('res.partner, copy=False)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,51 @@
...
+    tag_ids = fields.Many2many('estate.property.tag')
+    offer_ids = fields.One2many('estate.property.offer', 'property_id', 'offers')
```
```suggestion
    offer_ids = fields.One2many('estate.property.offer', 'property_id')
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,51 @@
...
+            estate_property.total_area = estate_property.living_area + estate_property.garden_area
+
+    @api.depends('offer_ids.price')
+    def _compute_best_offer(self):
+        for estate_property in self:
+            estate_property.best_offer = max(estate_property.offer_ids.mapped('price'), default=None)
+
+    @api.onchange('garden')
+    def _onchange_garden(self):
+        self.garden_area = 10 if self.garden else 0
```
cgun-odoo, line 64: Do you realize why you don't need to iterate over the records in `self` in the onchange method like you did in the compute?

estate/models/estate_property.py
```diff
@@ -0,0 +1,51 @@
...
+        self.garden_orientation = 'north' if self.garden else None
```
```suggestion
        if self.garden:
           self.garden_area = 10
           self.garden_orientation = "north"
       else:
           self.garden_area = 0
           self.garden_orientation = None
```
I'm sure you already thought about it. Why did you choose the inline if's instead of the blocks?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,24 @@
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models
+
+
+class EstatePropertyOffer(models.Model):
+    _name = 'estate.property.offer'
+    _description = 'Any offer that is put towards an property'
```
cgun-odoo: The model description in Odoo doesn't really work as a description. It's basically the user facing name of the model. So we usually write the method name again. If you think a description is needed for your model you can add it like a docstring

```suggestion
    _name = 'estate.property.offer'
    _description = 'Estate Property Offer'
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,24 @@
...
+
+    price = fields.Float()
+    status = fields.Selection([('accepted', 'Accepted'), ('refused', 'Refused')], copy=False)
+    partner_id = fields.Many2one('res.partner', required=True)
+    property_id = fields.Many2one('estate.property', required=True)
+    validity = fields.Integer('Validity (days)', default=7)
+    date_deadline = fields.Date('Deadline', compute='_calculate_deadline', inverse='_inverse_deadline')
```
cgun-odoo: By default you should name the compute/inverse methods as `_compute_{field_name}`, `_inverse_{field_name}`. Of course, there can be exceptions (a compute method for multiple fields)
```suggestion
    date_deadline = fields.Date('Deadline', compute='_compute_date_deadline', inverse='_inverse_date_deadline')
```

cgun-odoo: https://www.odoo.com/documentation/18.0/contributing/development/coding_guidelines.html#symbols-and-conventions

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,8 @@
+from odoo import fields, models
+
+
+class EstatePropertyType(models.Model):
+    _name = 'estate.property.type'
+    _description = 'Type of Estate Property'
```
cgun-odoo:
```suggestion
    _description = 'Estate Property Type'
```

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
+access_estate_property_type,access_estate_property_type,model_estate_property_type,base.group_user,1,1,1,1
+access_estate_property_tag,access_estate_property_tag,model_estate_property_tag,base.group_user,1,1,1,1
+access_estate_property_offer,access_estate_property_offer,model_estate_property_offer,base.group_user,1,1,1,0
```
cgun-odoo: Blank line missing at the end of this file

cgun-odoo: Shouldn't records of this model be deleted by anyone? Maybe someone creates a record by accident. Does it make sense for an admin to delete it maybe

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,32 @@
...
+                <field name="price"/>
+                <field name="partner_id"/>
+                <field name="validity"/>
+                <field name="date_deadline"/>
+            </list>
+        </field>
+    </record>
+
+    <record id="estate_property_offer_view_form" model="ir.ui.view">
+        <field name="name">estate.property.offer.list</field>
```
cgun-odoo:
```suggestion
        <field name="name">estate.property.offer.view.form</field>
```

cgun-odoo: https://www.odoo.com/documentation/18.0/contributing/development/coding_guidelines.html#security-view-and-action

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,32 @@
...
+                </sheet>
+            </form>
+        </field>
+    </record>
+
+
+    <record id="estate_property_tag_action" model="ir.actions.act_window">
+        <field name="name">Property Tags</field>
+        <field name="res_model">estate.property.tag</field>
+        <field name="view_mode">list,form</field>
```
cgun-odoo: Normally you don't need the form view for such a small model. You can just create/update it from the list view. I don't know if the tutorial asked for it so no need to remove.

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,32 @@
...
+                        <field name="partner_id"/>
+                        <field name="validity"/>
+                        <field name="date_deadline"/>
+                        <field name="status"/>
+                    </group>
+                </sheet>
+            </form>
+        </field>
+    </record>
+</odoo>
```
cgun-odoo: Line at the end of the file

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,32 @@
...
+    </record>
+</odoo>
```
cgun-odoo: Line at the end of the file

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,32 @@
...
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <list string="Types">
+                <field name="name"/>
+            </list>
+        </field>
+    </record>
+
+    <record id="estate_property_type_view_form" model="ir.ui.view">
+        <field name="name">estate.property.type.list</field>
```
cgun-odoo:
```suggestion
        <field name="name">estate.property.type.view.form</field>
```

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,32 @@
...
+                </sheet>
+            </form>
+        </field>
+    </record>
+
+
+    <record id="estate_property_type_action" model="ir.actions.act_window">
+        <field name="name">Property Types</field>
+        <field name="res_model">estate.property.type</field>
+        <field name="view_mode">list,form</field>
```
cgun-odoo, line 40: Same for the form view of a small model

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,32 @@
...
+    </record>
+</odoo>
```
cgun-odoo: Line at the end of the file

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,93 @@
+<odoo>
+
+    <record id="estate_property_view_search_name" model="ir.ui.view">
```
cgun-odoo:
```suggestion
    <record id="estate_property_view_search" model="ir.ui.view">
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,93 @@
...
+        <field name="name">estate.property.list</field>
```
cgun-odoo:
```suggestion
        <field name="name">estate.property.view.search</field>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,93 @@
...
+        <field name="arch" type="xml">
+            <search string="Properties">
+                <field name="name"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="facades"/>
+                <field name="property_type_id"/>
+                <filter name="available" domain="['|',('state','=','new'),('state','=','offer')]"/>
```
cgun-odoo:
```suggestion
                <filter name="available" domain="[('state', 'in', ['new', 'offer'])]"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,93 @@
...
+                <field name="living_area"/>
+                <field name="expected_price"/>
+                <field name="selling_price"/>
+                <field name="date_availability"/>
+            </list>
+        </field>
+    </record>
+
+    <record id="estate_property_view_form" model="ir.ui.view">
+        <field name="name">estate.property.list</field>
```
cgun-odoo:
```suggestion
        <field name="name">estate.property.view.form</field>
```

estate/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+        'base'
+    ],
+    'application': True,
+    'data': [
+        "views/estate_property_type_views.xml",
+        "views/estate_property_tag_views.xml",
+        "views/estate_property_views.xml",
+        "views/estate_property_offer_views.xml",
+        "views/estate_menus.xml",
+        'security/ir.model.access.csv',
```
cgun-odoo: We usually put the security files on the top, since what they define can be used in the views but not the opposite

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,8 @@
+from odoo import fields, models
+
+
+class EstatePropertyTag(models.Model):
+    _name = 'estate.property.tag'
+    _description = 'Tags for properties'
```
cgun-odoo:
```suggestion
    _description = 'Estate Property Tag'
```

cgun-odoo (commented):
Good job 👍 You can check the runbot, you're getting a warning

Image

estate/models/estate_property.py
djoewie: Do you have a formatter configuration I can start from, because now I use the formatter of pycharm by default.
I forgot to enable it for the first couple of commits.

estate/models/estate_property.py
djoewie: indeed this is one word, is there a way to not loose data when renaming a field?

cgun-odoo: Yes there is, you can check:
https://www.odoo.com/documentation/18.0/developer/reference/upgrades/upgrade_scripts.html

For this specific case:
https://www.odoo.com/documentation/18.0/developer/reference/upgrades/upgrade_utils.html#odoo.upgrade.util.fields.rename_field

No need for now, but do it if you find it interesting

estate/models/estate_property.py
djoewie: because when calling this method it is done with a specific record that is changed, while with the compute it  needs to be done on all the records that need to be shown which is a list of records

estate/models/estate_property.py
djoewie: Indeed thought about it, and I prefer to read what fields will change over how it will change
combined with the fact that in my experience the if blocks have higher chance of missing a field if more then 4 fields are being changed based on the if statement

but in the end this certainly is a matter of convention

estate/security/ir.model.access.csv
djoewie: true, that makes sense, I was thinking in terms of you could refuse an offer if it was made by accident
I just did not wanted everyone to be able to delete that record as this would create a big hole for data inconsistance.

estate/models/estate_property.py
cgun-odoo: That's a good reason, there is no convention for this I guess. It's just perefrence, what you did is completely fine

estate/models/estate_property.py
cgun-odoo: One thing to add:
onchange methods only run in the form view. Therefore, you're always sure that it will run for only one record. (If the method is not called by us in a unit test)

estate/models/estate_property.py
```diff
@@ -1,51 +1,94 @@
...
+    property_type_id = fields.Many2one("estate.property.type")
+    salesman_id = fields.Many2one("res.users", copy=False, default=lambda self: self.env.user)
+    buyer_id = fields.Many2one("res.partner", copy=False, readonly=True)
+    tag_ids = fields.Many2many("estate.property.tag")
+    offer_ids = fields.One2many("estate.property.offer", "property_id")
+    total_area = fields.Integer("Total Area (sqm)", compute="_compute_total_area")
+    best_offer = fields.Float(compute="_compute_best_offer")
+
+    _check_expected_price = models.Constraint("CHECK(expected_price>0)", "The expected price needs to be bigger then 0")
+    _check_selling_price = models.Constraint("CHECK(selling_price>=0)", "The selling price needs to be positive")
```
cgun-odoo, line 50: sql Constraints should be above fields with other private properties.
Refer to this to check the order in a model:
https://www.odoo.com/documentation/18.0/contributing/development/coding_guidelines.html#symbols-and-conventions

estate/models/estate_property.py
```diff
@@ -1,51 +1,94 @@
...
+    def action_state_to_sold(self):
+        self.ensure_one()
+        if self.state == "canceled":
+            raise UserError("Sold properties can not be canceled")
+        self.state = "sold"
+
+    def action_state_to_canceled(self):
+        self.ensure_one()
+        if self.state == "sold":
+            raise UserError("Canceled properties can not be sold")
```
cgun-odoo: The Error messages for `action_state_to_sold` and `action_state_to_canceled` felt like the opposite of what it should be

estate/models/estate_property.py
```diff
@@ -1,51 +1,94 @@
...
+        self.garden_orientation = "north" if self.garden else None
+
+    @api.constrains("selling_price", "expected_price")
+    def _check_selling_price(self):
+        for estate_property in self:
+            if (
+                not float_is_zero(estate_property.selling_price, 2)
+                and float_compare(estate_property.selling_price, estate_property.expected_price * 0.9, 2) < 0
+            ):
+                raise UserError("The property can not be sold for a price lower than 90% of the expected price")
```
cgun-odoo: This one should be a ValidationError as it's thrown from an api.constrains
Check this for exceptions in Odoo
https://www.odoo.com/documentation/18.0/developer/reference/backend/orm.html?highlight=usererror#odoo.exceptions.UserError

estate/models/estate_property.py
```diff
@@ -1,51 +1,94 @@
...
+    def action_state_to_canceled(self):
+        self.ensure_one()
+        if self.state == "sold":
+            raise UserError("Canceled properties can not be sold")
+        self.state = "canceled"
+
+    def accept_offer(self, price, buyer):
+        self.ensure_one()
+        if self.state not in ["new", "offer"]:
+            raise UserError("The offer can not be accepted")
```
cgun-odoo: You can explain more like with the message like you did for the other ones.
Why can't they accept it?

estate/models/estate_property_offer.py
```diff
@@ -1,24 +1,41 @@
...
-
-    @api.depends('validity', 'create_date')
-    def _calculate_deadline(self):
+    status = fields.Selection([("accepted", "Accepted"), ("refused", "Refused")], copy=False)
+    partner_id = fields.Many2one("res.partner", required=True)
+    property_id = fields.Many2one("estate.property", required=True)
+    validity = fields.Integer("Validity (days)", default=7)
+    date_deadline = fields.Date("Deadline", compute="_calculate_date_deadline", inverse="_inverse_date_deadline")
+
+    _check_price = models.Constraint("CHECK(price>0)", "The selling price needs to be bigger then 0")
```
cgun-odoo, line 20: Should go up

estate/views/estate_property_tag_views.xml
```diff
@@ -1,32 +1,17 @@
...
-        <field name="arch" type="xml">
-            <form string="Tag">
-                <sheet>
-                    <group>
-                        <field name="name"/>
-                    </group>
-                </sheet>
-            </form>
-        </field>
-    </record>
```
cgun-odoo: If you don't want them to access the form view you should remove it from the action as well

estate/views/estate_property_views.xml
```diff
@@ -65,17 +75,17 @@
...
-                                <field name="garden_area"/>
-                                <field name="garden_orientation"/>
+                                <field name="garden_area" invisible="not garden"/>
+                                <field name="garden_orientation" invisible="not garden"/>
                                 <field name="total_area"/>
                             </group>
                         </page>
                         <page string="Offers">
-                            <field name="offer_ids"/>
+                            <field name="offer_ids" readonly="state in ['accepted','sold','canceled']"/>
```
cgun-odoo, line 112: You can define the list view of offers here so that you don't need to have a file for it (i guess you don't access that view anywhere else)

Kind of like
```
<field name="one2many_field">
      <list>
           <field/>
           <field/>
      </list>
 </field>
```

cgun-odoo (commented):
Added some comments to align more with the conventions

estate/models/estate_property.py
djoewie: I was following the 19 guidelines where it was indicating that this is below the field declaration

estate/models/estate_property.py
djoewie: you are absolutely right

estate/models/estate_property.py
djoewie: would you go for 1 generic message, or for individual messages based on the different states?

estate/views/estate_property_tag_views.xml
djoewie: you are going faster then the tutorial, nice, just came across this.

estate/models/estate_property.py
cgun-odoo: I would go for multiple messages to be more explanatory. If there are too many different states that require too many messages, I would think about creating a helper function that generates an error message.

estate/models/estate_property.py
```diff
@@ -71,24 +71,41 @@ def _check_selling_price(self):
...
+            )
         self.selling_price = price
         self.buyer_id = buyer
         self.state = "accepted"
+
+    def offer_made(self, price):
+        self.ensure_one()
+        if self.state not in ["new", "offer"]:
+            raise UserError(
+                "An offer can can only be accepted when the building is still for sale and no other offer is already accepted"
```
cgun-odoo, line 109: "An offer can only be *made*" probably

estate/models/estate_property.py
```diff
@@ -71,24 +71,41 @@ def _check_selling_price(self):
...
+            )
+        if float_compare(price, self.best_offer, 2) <= 0:
+            raise UserError("Only offers that are over the current best offer can be accepted")
```
cgun-odoo, line 112: Shouldn't this check be in `accept_offer`?

estate/models/estate_property_offer.py
```diff
@@ -27,6 +28,12 @@ def _inverse_date_deadline(self):
         for offer in self:
             offer.validity = (offer.date_deadline - fields.Date.today()).days
 
+    @api.model
+    def create(self, vals_list):
+        for offer in vals_list:
```
cgun-odoo: Here we usually do 
`for vals in vals_list:`
As they aren't offers yet

estate/models/estate_property_type.py
```diff
@@ -8,8 +8,15 @@ class EstatePropertyType(models.Model):
...
 
     _unique_name = models.UniqueIndex(
         "(name)",
         "Another entry with the same name already exists.",
     )
+
+    @api.depends("offer_ids")
+    def _compute_offer_count(self):
+        for offer in self:
+            offer.offer_count = len(offer.offer_ids)
```
cgun-odoo: This should match the model
`for property_type in self:`

estate/views/res_user_views.xml
```diff
@@ -0,0 +1,15 @@
+<odoo>
+    <!-- Update User form !-->
+    <record id="res_users_form_view_properties" model="ir.ui.view">
```
cgun-odoo: https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#inheriting-xml

ID of the inherited record should stay the same:

`<record id="view_users_form" model="ir.ui.view">`

I know that some people in the department actually don't like this and they try to give more explanatory id and names to view records.
IMO, naming them the same helps check where the record has been updated easily so I follow the convention.

estate/views/res_user_views.xml
```diff
@@ -0,0 +1,15 @@
...
+        <field name="name">res.users.form.properties</field>
```
cgun-odoo: According to the same guideline this should be named:
`res.users.view.form.inherit.estate`

estate/models/estate_property_offer.py
```diff
@@ -15,6 +15,7 @@ class EstatePropertyOffer(models.Model):
     property_id = fields.Many2one("estate.property", required=True)
     validity = fields.Integer("Validity (days)", default=7)
     date_deadline = fields.Date("Deadline", compute="_calculate_date_deadline", inverse="_inverse_date_deadline")
+    property_type_id = fields.Many2one(related="property_id.property_type_id", stored=True)
```
cgun-odoo: Why did you need to make this field stored?

estate/models/estate_property.py
djoewie: question was to only allow offers that are higher than previous made offers

estate/views/res_user_views.xml
djoewie: makes sense, did not get that from the tutorial

estate/models/estate_property_offer.py
djoewie: done to make it possible to have the inverse field in property_type as one2many, if not stored then these items could not be found from a property_type trough SQL queries

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,21 @@
+from odoo import Command, models
+from odoo.exceptions import ValidationError
+
+
+class EstateProperty(models.Model):
+    _inherit = "estate.property"
+
+    def action_state_to_sold(self):
+        if not self.buyer_id:
+            raise ValidationError("Properties can only be sold if the buyer is filled in")
```
cgun-odoo: This should be a `UserError` since it's not a constraint https://www.odoo.com/documentation/18.0/developer/reference/backend/orm.html?highlight=usererror#odoo.exceptions.UserError

estate/models/estate_property_offer.py
cgun-odoo: There is one problem though: It should be `store=True` without the 'd'. Right now, the field is probably not stored.
It works because related fields are special, you can search on them even if they aren't stored.

cgun-odoo (commented):
You can try to make the runbot green now

PR #1242, opened 2026-04-21
rohar - Technical Training

rohar-odoo:


estate/models/estate_property.py
```diff
@@ -0,0 +1,28 @@
+from odoo import fields, models
+
+
+class EstateProperty(models.Model):
+    _name = "estate_property"
```
Mathilde411:
```suggestion
    _name = "estate.property"
```
Model names in odoo use dots, usually with the module name inside:

estate/models/estate_property.py
```diff
@@ -0,0 +1,28 @@
...
+    _description = "Real estate properties"
```
Mathilde411:
```suggestion
    _description = "Real Estate Property"
```
Little nitpicky but this should be singular

estate/models/estate_property.py
```diff
@@ -0,0 +1,28 @@
...
+    garden_area = fields.Integer("Garden area")
+    garden_orientation = fields.Selection(
+        string="Garden orientation",
+        selection=[
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
+        ],
+    )
```
Mathilde411, line 37:
```suggestion
    name = fields.Char("Property Name", required=True)
    description = fields.Text()
    postcode = fields.Char()
    date_availability = fields.Date()
    expected_price = fields.Float(required=True)
    selling_price = fields.Float()
    bedrooms = fields.Integer("Number of Bedrooms")
    living_area = fields.Integer()
    facades = fields.Integer()
    garage = fields.Boolean()
    garden = fields.Boolean()
    garden_area = fields.Integer()
    garden_orientation = fields.Selection(
        selection=[
            ("north", "North"),
            ("south", "South"),
            ("east", "East"),
            ("west", "West"),
        ],
    )
```
string can be inferred from the variable name

estate/models/estate_property.py
```diff
@@ -0,0 +1,28 @@
...
+    description = fields.Text("Description")
+    postcode = fields.Char("Postcode")
+    date_availability = fields.Date("Date availability")
+    expected_price = fields.Float("Expected price", required=True)
+    selling_price = fields.Float("Selling price")
+    bedrooms = fields.Integer("Number of bedrooms")
+    living_area = fields.Integer("Living area")
+    facades = fields.Integer("Facades")
+    garage = fields.Boolean("Garage")
+    garden = fields.Boolean("Garden")
```
Mathilde411: Here it's the tutorial telling you to use the names, but I prefer using is_... or has_... for boolean fields, it makes the code more readable :)

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,1
```
Mathilde411:
```suggestion
estate_property_access_user,estate.property.user,model_estate_property,base.group_user,1,1,1,1
```
Comments for each column:
1st column is the XMLid of the access right, it should be `<model_name>_access_<concerned_group>` (similarly to record rules in https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#xml-ids-and-naming)
second column is the name
2nd column is the name of the rule, the one that will be displayed in the ui, should say the model name and the targeted group (there is not really a fixed way, but a common one is with dots, like suggested)
3rd one is the xmlid of the model targeted, you don't need to add the module name in front if the xmlid is from the same module as you, same for 4th column that represents the targetted group

estate/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from . import models
```
Mathilde411:
```suggestion
from . import models
```
Not needed here

estate/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    Real Estate Advertisement
+    """,
+    # data files always loaded at installation
+    "data": [
+        "security/ir.model.access.csv",
+    ],
+    # data files containing optionally loaded demonstration data
+    "demo": [],
+    "application": True,
+    "installable": True,
```
Mathilde411: installable's default is True here, not needed

estate/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "auto_install": True,
```
Mathilde411: Can you tell me what this does ? Why did you add it ?

Mathilde411: Can you tell me what auto_install does ? Why did you add it ?

estate/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "Estate",
+    "version": "1.0",
+    "depends": ["base"],
```
Mathilde411:
```suggestion
    "depends": [
        "base",
    ],
```
Putting oit like this allows anyone to see when a certain dependency was added using a git blame

estate/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "Estate",
+    "version": "1.0",
```
Mathilde411:
```suggestion
    "version": "19.0.1.0.0",
```
When making modules outside of standard we use this synthax so that it'll only install on the correct version of Odoo.
For example here the module will only be installable on odoo 19.0

Mathilde411 (changes requested):
Pretty good so far !
Runbot is green so good job !
Also would you be able to squash your commits so that you only have one per chapter ?

estate/__manifest__.py
rohar-odoo: So it's rather Odoo's version than the module's version ?

Mathilde411: > So it's rather Odoo's version than the module's version ?

The version of a module that is not part of odoo standard will be:

`<odoo_major>.<odoo_minor>.<module_major>.<module_minor>.<module_patch>`

If odoo detects this patern it will only allow you to install the module in odoo version <odoo_major>.<odoo_minor>

So for example a module with version 17.0.x.y.z will only be installable in Odoo 17.0

rohar-odoo: Also kind of stupid question but I know that I'm on Odoo 19.0, but "19.0.1.0.0" ?

estate/__manifest__.py
rohar-odoo: To have it already installed since it's dependency (base) is installed ? I put it there thinking I'll skip the install step later in Odoo

estate/__manifest__.py
rohar-odoo: Right ! Side question on the same topic: the license field has a default value, but not setting it result in a warning anyway ?

Mathilde411: It has a default value, but you should still specify it. Licenses should be a deliberate choice you make with your model. So it'll run but it'll warn you that you should provide it because it's an info that odoo wants to have

estate/__manifest__.py
Mathilde411: This is used for modules that should be installed when their deps are met.
For example if you have stock and sales installed you'd want to be able to create pickings from sales orders.
So it auto installs the sale_stock model. There you don't need it, so you should remove it.

Btw you can use -i in the command line to install a module on startup :)

estate/models/estate_property.py
rohar-odoo: I misunderstood it as a description of the table, which contains multiples properties (and the linked example was for the recurring plans, [plural](https://github.com/odoo/odoo/blob/e80911aaead031e7523173789e946ac1fd27c7dc/addons/crm/models/crm_recurring_plan.py#L1-L9)).

Mathilde411: This should not be the case. Most are singular
image

estate/__manifest__.py
rohar-odoo: Ok it make sense

estate/models/estate_property.py
rohar-odoo: I agree and I changed it, then the tutorial introduced the `active`  reserved field, which is boolean and now I have two different conventions in the same file 😵‍💫

Mathilde411: Yeah, active is the exception

estate/models/estate_property.py
```diff
@@ -0,0 +1,75 @@
...
+            ("new", "New"),
+            ("offer_received", "Offer Received"),
+            ("offer_accepted", "Offer Accepted"),
+            ("sold", "Sold"),
+            ("cancelled", "Cancelled"),
+        ],
+    )
+    buyer_id = fields.Many2one("res.partner", string="Buyer", copy=False)
+    salesperson_id = fields.Many2one(
+        "res.users", string="Salesperson", default=lambda self: self.env.user
```
Mathilde411: You don't need the string here either :)
In a many2one the field xyz_id has automatically the string Xyz
In one2many and many2many it's the same but with xyz_ids

estate/models/estate_property.py
```diff
@@ -0,0 +1,92 @@
...
+    @api.depends("garden_area", "living_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.garden_area + record.living_area
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for record in self:
+            record.best_price = (
+                max(record.mapped("offer_ids.price")) if record.offer_ids else 0.0
```
Mathilde411:
```suggestion
            record.best_price = max(record.offer_ids.mapped('price'), default=0)
```
if xyz is a relational field of rec, than rec.xyz will always be a recordset you can use.
Also nobody knows this but max and min have a default parameter for empty iterables :D

estate/models/estate_property.py
```diff
@@ -0,0 +1,92 @@
...
+    def estate_property_action_sold(self):
+        for record in self:
+            if record.state == "cancelled":
+                raise UserError("A cancelled property cannot be sold !")
+
+            record.state = "sold"
+            return True
+
+    def estate_property_action_cancel(self):
+        for record in self:
```
Mathilde411:
```suggestion
    def action_sold(self):
        for record in self:
            if record.state == "cancelled":
                raise UserError("A cancelled property cannot be sold !")

            record.state = "sold"
            return True

    def action_cancel(self):
        for record in self:
```
The naming convention you used is for xmlids only :)
For method we usually use "action_..." but you'll see other stuff sometimes too

estate/models/estate_property.py
```diff
@@ -0,0 +1,92 @@
...
+            self.garden_orientation = None
+            self.garden_area = 0
+
+    def estate_property_action_sold(self):
+        for record in self:
+            if record.state == "cancelled":
+                raise UserError("A cancelled property cannot be sold !")
+
+            record.state = "sold"
+            return True
```
Mathilde411:
```suggestion
    def estate_property_action_sold(self):
        for record in self:
            if record.state == "cancelled":
                raise UserError("A cancelled property cannot be sold !")

            record.state = "sold"
        return True
```
Here you return inside the loop, so if the action is made on multiple records, only the first would be treated

estate/models/estate_property.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def estate_property_action_cancel(self):
+        for record in self:
+            if record.state == "sold":
+                raise UserError("A sold property cannot be cancelled !")
+
+            record.state = "cancelled"
+            return True
```
Mathilde411: Same as above

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class EstatePropertyOffer(models.Model):
+    _name = "estate.property.offer"
+    _description = "Property offer"
+
+    price = fields.Float()
+    status = fields.Selection(
+        copy=False, selection=[("accepted", "Accepted"), ("refused", "Refused")]
+    )
```
Mathilde411, line 15:
```suggestion
status = fields.Selection([
            ("accepted", "Accepted"), 
            ("refused", "Refused"),
    ], copy=False)
```
You don't need the kwarg for selection

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,48 @@
...
+    partner_id = fields.Many2one("res.partner", string="Partner", required=True)
+    property_id = fields.Many2one("estate.property", string="Property", required=True)
```
Mathilde411: Same as above, strings not required here :)

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,48 @@
...
+            record.validity = days.days
+
+    def property_offer_action_accept(self):
+        for record in self:
+            property_id = record.property_id
+            if property_id.state == "sold":
+                raise UserError("Property already sold !")
+
+            if property_id.state == "cancelled":
+                raise UserError("Cancelled property cannot be sold !")
```
Mathilde411:
```suggestion
            if property_id.state in ("sold", "cancelled"):
                raise UserError("A sold or cancelled property can't be sold.")
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,48 @@
...
+        copy=False, selection=[("accepted", "Accepted"), ("refused", "Refused")]
+    )
+    partner_id = fields.Many2one("res.partner", string="Partner", required=True)
+    property_id = fields.Many2one("estate.property", string="Property", required=True)
+    validity = fields.Integer("Validity (days)", default=7)
+    date_deadline = fields.Date(
+        "Deadline", compute="_compute_date_deadline", inverse="_inverse_date_deadline"
+    )
+
+    @api.depends("validity")
```
Mathilde411:
```suggestion
    @api.depends("validity", "create_date")
```
You have to have a dependency on every field you use

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,48 @@
...
+                raise UserError("Cancelled property cannot be sold !")
+
+            record.status = "accepted"
+            property_id.state = "sold"
+            property_id.buyer_id = record.partner_id
+            property_id.selling_price = record.price
+
+    def property_offer_action_refuse(self):
+        for record in self:
+            record.status = "refused"
```
Mathilde411:
```suggestion
    def action_refuse(self):
        self.status = "refused"
```
You can assign fields in a recordset

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,12 @@
...
+    <menuitem id="real_estate_root" name="Real Estate">
+        <menuitem id="advertisement" name="Advertisement">
+            <menuitem id="estate_menu_action" action="estate_property_new"/>
+        </menuitem>
+        <menuitem id="settings" name="Settings">
+            <menuitem id="estate_property_type_action" action="estate_property_type_new"/>
+            <menuitem id="estate_property_tag_action" action="estate_property_tag_new"/>
+        </menuitem>
+    </menuitem>
+</odoo>
```
Mathilde411: Missing newline

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,12 @@
...
+<odoo>
+    <menuitem id="real_estate_root" name="Real Estate">
+        <menuitem id="advertisement" name="Advertisement">
+            <menuitem id="estate_menu_action" action="estate_property_new"/>
+        </menuitem>
+        <menuitem id="settings" name="Settings">
+            <menuitem id="estate_property_type_action" action="estate_property_type_new"/>
+            <menuitem id="estate_property_tag_action" action="estate_property_tag_new"/>
+        </menuitem>
+    </menuitem>
```
Mathilde411: For submenus the xmlid should be `<module_name>_menu_<precision>` or if linked to a model `<model_name>_menu`
So for example `id="estate_property_type_action"` should be `id="estate_property_menu"` and `id="advertisement"` should be `id="estate_menu_advertisement"`

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="estate_property_offer_views" model="ir.actions.act_window">
```
Mathilde411:
```suggestion
    <record id="estate_property_offer_action" model="ir.actions.act_window">
```
XMLid for actions

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="estate_property_tag_new" model="ir.actions.act_window">
```
Mathilde411: Same, xmlid is not correct for the action

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,107 @@
...
+                <field name="living_area"/>
+                <field name="facades"/>
+                <field name="tag_ids" widget="many2many_tags"/>
+                <filter name="available" string="Available" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"/>
+                <filter name="postcode" string="Postcode" context="{'group_by':'postcode'}"/>
+            </search>
+        </field>
+    </record>
+
+    <record id="estate_property_new" model="ir.actions.act_window">
```
Mathilde411: Same again

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,107 @@
...
+            <form string="Properties">
+                <header>
+                    <button name="estate_property_action_sold" type="object" string="Sold"/>
+                    <button name="estate_property_action_cancel" type="object" string="Cancel"/>
+                </header>
+                <sheet>
+                    <h1>
+                        <field name="name"/>
+                    </h1>
+                    <field name="tag_ids" widget="many2many_tags"/>
```
Mathilde411: Generally we put it inside a <div class="oe_title"> so:
```suggestion
                    <div class="oe_title">
                        <h1>
                            <field name="name"/>
                        </h1>
                        <field name="tag_ids" widget="many2many_tags"/>
                    </div>
```

Mathilde411 (changes requested):
Good job here !
Here are a few things that I'd have done another way :)

estate/models/estate_property.py
rohar-odoo: 🤯

estate/models/estate_property.py
```diff
@@ -0,0 +1,113 @@
...
+        for record in self:
+            if float_is_zero(record.selling_price, precision_digits=2):
+                return
+            if float_compare(record.selling_price, 0.9 * record.expected_price, precision_digits=2) == -1:
+                raise ValidationError("The selling price cannot be lower than 90% of the expected price.")
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_if_new_or_cancelled(self):
+        if any(not record.state in ("new", "cancelled") for record in self):
+            raise UserError("Can only delete an new or cancelled property.")
```
Mathilde411, line 113: Be careful about attribute ordering

estate/models/estate_property.py
```diff
@@ -0,0 +1,113 @@
...
+        "A property expected price must be stricly positive",
+    )
+
+    _check_selling_price = models.Constraint(
+        "CHECK(selling_price >= 0)",
+        "A property selling price must be positive",
+    )
+
+    @api.constrains("selling_price", "expected_price")
+    def check_selling_price(self):
```
Mathilde411, line 103: should be named '_check_selling_price'

estate/models/estate_property.py
```diff
@@ -0,0 +1,113 @@
...
+        for record in self:
+            if float_is_zero(record.selling_price, precision_digits=2):
+                return
```
Mathilde411, line 106: Why an exception for 0 ?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @api.model
+    def create(self, vals_list):
+        property = self.env["estate.property"].browse(vals_list[0]["property_id"])
+        if max(property.offer_ids.mapped("price"), default=0.0) > vals_list[0]["price"]:
+            raise UserError("New offer must be higher than existing ones.")
+
+        property.state = "offer_received"
+
+        return super().create(vals_list)
```
Mathilde411, line 43: What if I create multiple records ? (multiple values in vals_list) then only the first one would be checked

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,65 @@
...
+            if (property_id.has_garden and property_id.garden_orientation == "south" and float_compare(record.price, property_id.expected_price, precision_digits=2) == -1):
+                raise ValidationError("The selling price cannot be lower than 90% of the expected price.")
+
+            record.status = "accepted"
+            property_id.state = "sold"
+            property_id.buyer_id = record.partner_id
+            property_id.selling_price = record.price
+
+    def action_refuse(self):
+        self.status = "refused"
```
Mathilde411, line 65: Attribute ordering :)

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,65 @@
...
+            record.validity = days.days
+
+    def action_accept(self):
+        for record in self:
+            property_id = record.property_id
+            if property_id.state in ("sold", "accepted"):
+                raise UserError("A sold or cancelled property can't be sold.")
+
+            if (property_id.has_garden and property_id.garden_orientation == "south" and float_compare(record.price, property_id.expected_price, precision_digits=2) == -1):
+                raise ValidationError("The selling price cannot be lower than 90% of the expected price.")
```
Mathilde411, line 57: This should be in a check

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+estate_property_access_user,estate.property.user,model_estate_property,base.group_user,1,1,1,1
+estate_property_type_access_user,estate.property.type.user,model_estate_property_type,base.group_user,1,1,1,1
+estate_property_tag_access_user,estate.property.tag.user,model_estate_property_tag,base.group_user,1,1,1,1
+estate_property_offer_acces_user,estate.property.offer.user,model_estate_property_offer,base.group_user,1,1,1,1
```
Mathilde411, line 5: typo

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,24 @@
...
+        <field name="model">estate.property.offer</field>
+        <field name="arch" type="xml">
+            <list string="Offers">
+                <field name="price"/>
+                <field name="status"/>
+                <field name="partner_id"/>
+                <field name="property_id"/>
+                <field name="validity" optional="true"/>
+                <field name="date_deadline" optional="true"/>
+                <field name="property_type_id" optional="true"/>
```
Mathilde411, line 14: Optional values are hide or show

estate_account/models/estate_account.py
```diff
@@ -0,0 +1,6 @@
+from odoo import models
+
+
+class EstateAccount(models.Model):
+    _name = "estate.account"
+    _description = "Real estate accounting"
```
Mathilde411, line 6: You don't need a model here ? Why did you add it ?

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,28 @@
+from odoo import models, Command
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _inherit = ["estate.property"]
```
Mathilde411, line 6: Can just be:
```suggestion
    _inherit = "estate.property"
```

estate_account/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+        ],
+    "author": "Roman (rohar)",
+    "description": """
+    Real Estate Accounting
+    """,
+    "data": [
+        "security/ir.model.access.csv",
+    ],
+    "license": "LGPL-3",
+    "auto_install": True,
```
Mathilde411, line 16: This makes sense here ! Do you know why ?

Mathilde411 (changes requested):
Oops forgot to submit it !
Pretty good overall !
Fix the last few points and you'll be good to go !

PR #1243, opened 2026-04-21
Technical Training - stoss

stoss-odoo:


estate/data/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
```
Mathilde411: Most common way would be:
```suggestion
id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
```
But this works too, it's just not as common :)

estate/models/estate_property.py
```diff
@@ -0,0 +1,27 @@
...
+    living_area = fields.Integer("Living Area")
+    facades = fields.Integer("Facades")
+    garage = fields.Boolean("Garage")
+
+    garden = fields.Boolean("Garden")
+    garden_area = fields.Integer("Garden Area")
+    garden_orientation = fields.Selection(string="Orientation", selection=[("north", "North"), ("south", "South"), ("east", "East"), ("west", "West")])
+
+    active = fields.Boolean("Active", default=True)
+    state = fields.Selection(string="State", selection=[("new", "New"), ("offer_received", "Offer Received"), ("offer_accepted", "Offer Accepted"), ("sold", "Sold"), ("cancelled", "Cancelled")])
```
Mathilde411:
```suggestion
    name = fields.Char(required=True)
    description = fields.Text()
    postcode = fields.Char()
    date_availability = fields.Date(copy=False, default=date_utils.add(fields.Date.today() + date_utils.relativedelta(months=3)))
    expected_price = fields.Float(required=True)
    selling_price = fields.Float(readonly=True, copy=False)

    bedrooms = fields.Integer(default=2)
    living_area = fields.Integer()
    facades = fields.Integer()
    garage = fields.Boolean()

    garden = fields.Boolean()
    garden_area = fields.Integer()
    garden_orientation = fields.Selection(selection=[("north", "North"), ("south", "South"), ("east", "East"), ("west", "West")])

    active = fields.Boolean(default=True)
    state = fields.Selection(selection=[("new", "New"), ("offer_received", "Offer Received"), ("offer_accepted", "Offer Accepted"), ("sold", "Sold"), ("cancelled", "Cancelled")])
```
No need for the string arguments there, it is extrapolated from the variable name, so the best practice is to only add it when necessary :)

Mathilde411: Also you can format this (e.g. by jumping lines in the lists) as it will make the whole thing more readable. On my screen it goes out of the reading area.

estate/models/estate_property.py
```diff
@@ -0,0 +1,27 @@
+import datetime
+
+from odoo import models, fields
+import odoo.tools.date_utils as date_utils
+
+class Property(models.Model):
+    _name = 'estate.property'
+    _description = 'Test Model for the Estate App'
```
Mathilde411:
```suggestion
    _description = 'Estate Property'
```
_description is the name the model is going to have when seen in the UI. Not an actual description

estate/models/estate_property.py
```diff
@@ -0,0 +1,27 @@
...
+
+    name = fields.Char("Name", required=True)
+    description = fields.Text("Description")
+    postcode = fields.Char("Postcode")
+    date_availability = fields.Date("Date Availability", copy=False, default=date_utils.add(fields.Date.today() + date_utils.relativedelta(months=3)))
```
Mathilde411: The default here needs to be a function. It works here because you are starting the server on the same day you are using it, but if you try 3 months from now it'll still be the same default date  (3 months after starting the server)

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" ?>
+<odoo>
+    <menuitem id="estate_menu_root" name="Estate">
+        <menuitem id="estate_first_level_menu" name="Advertisements">
+            <menuitem id="estate_property_menu_action" action="estate_property_action"/>
+        </menuitem>
+    </menuitem>
+</odoo>
```
Mathilde411: Missing newline here

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" ?>
+<odoo>
+    <menuitem id="estate_menu_root" name="Estate">
+        <menuitem id="estate_first_level_menu" name="Advertisements">
+            <menuitem id="estate_property_menu_action" action="estate_property_action"/>
```
Mathilde411: The xmlids in menus should be <model_name>_menu, or <model_name>_menu_do_stuff for submenu.
So here it'd have to be something like: 
```suggestion
        <menuitem id="estate_menu_advertisements" name="Advertisements">
            <menuitem id="estate_property_menu" action="estate_property_action"/>
```

estate/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+{
+    'name': "Estate",
+    'version': '69.0',
```
Mathilde411:
```suggestion
    'version': '19.0.1.0.0',
```

estate/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    'depends': ['base'],
```
Mathilde411:
```suggestion
    'depends': [
        'base',
    ],
```
That way if you add a new dependency you will only have one line in the diff and it'll be easy to see what was added

Mathilde411 (commented):
Quite good !
Only small comments. Also try to be purposeful about the type of quote you use. In standard ' is preferred.

estate/models/estate_property.py
```diff
@@ -0,0 +1,133 @@
...
+
+class Property(models.Model):
+    _name = "estate.property"
+    _description = "Estate Property"
+
+    # General information
+    name = fields.Char(string="Title", required=True)
+    description = fields.Text()
+    tag_ids = fields.Many2many(comodel_name="estate.property.tag", string="Tags")
+    type_id = fields.Many2one(comodel_name="estate.property.type", string="Type")
```
Mathilde411:
```suggestion
    tag_ids = fields.Many2many("estate.property.tag", string="Tags")
    type_id = fields.Many2one("estate.property.type", string="Type")
```
Most people use positional arguments for the comodel/inverse_field. But this also works

estate/models/estate_property.py
```diff
@@ -0,0 +1,133 @@
...
+            ("east", "East"),
+            ("west", "West"),
+        ]
+    )
+    # Sales info
+    date_availability = fields.Date(
+        copy=False,
+        default=lambda x: date_utils.add(
+            fields.Date.today() + date_utils.relativedelta(months=3)
+        ),
```
Mathilde411:
```suggestion
        default=lambda self: fields.Date.today() + date_utils.relativedelta(months=3),
```
No need for the add here, + already works for relativedelats

estate/models/estate_property.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+    offer_ids = fields.One2many(
+        comodel_name="estate.property.offer",
+        inverse_name="property_id",
```
Mathilde411: Same as above no need to pass them as kwargs, use positional

estate/models/estate_property.py
```diff
@@ -0,0 +1,133 @@
...
+        inverse="_compute_state",
+    )
+    expected_price = fields.Float(required=True)
+    best_offer = fields.Integer(compute="_compute_best_offer")
+    selling_price = fields.Float(readonly=True, copy=False)
+    buyer_id = fields.Many2one(
+        comodel_name="res.partner", string="Buyer", copy=False, readonly=True
+    )
```
Mathilde411:
```suggestion
buyer_id = fields.Many2one(
        "res.partner", copy=False, readonly=True
    )
```
on a many2one buyer_id automatically has Buyer as it's string

Mathilde411: Actually missed it, but same for type_id

estate/models/estate_property.py
```diff
@@ -0,0 +1,133 @@
...
+    salesman_id = fields.Many2one(
+        comodel_name="res.users",
+        string="Salesman",
+        default=lambda self: self.env.user.id,
```
Mathilde411:
```suggestion
        "res.users",
        default=lambda self: self.env.user,
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,133 @@
...
+            ("sold", "Sold"),
+            ("cancelled", "Cancelled"),
+        ],
+        default="new",
+    )
+    active = fields.Boolean(default=True)
+
+    _check_expected_price = models.Constraint("CHECK(expected_price > 0)")
+
+    _check_selling_price = models.Constraint("CHECK(selling_price > 0)")
```
Mathilde411: You should add a massage on your constraints, as the second parameter, it'll display if the constraint is violated

estate/models/estate_property.py
```diff
@@ -0,0 +1,133 @@
...
+            record.total_area = record.garden_area + record.living_area
+
+    @api.depends("offer_ids.price")
+    def _compute_best_offer(self):
+        for record in self:
+            offers = [
+                offer.price for offer in record.offer_ids if offer.status != "refused"
+            ]
+            best = max([0, *offers])
+            record.best_offer = best
```
Mathilde411:
```suggestion
            offers = record.offer_ids.filtered(lambda o: o.status != "refused")
            record.best_offer = max(offers.mapped('price'), default=0)
```
Use filtered to filter recordset, then mapped to get a list of the prices.
Also max has a default parameter that no one knows about, so use that :)

estate/models/estate_property.py
```diff
@@ -0,0 +1,133 @@
...
+                )
+
+    @api.depends("state", "offer_ids")
+    def _compute_state(self):
+        for record in self:
+            if record.state not in ["sold", "cancelled"]:
+                if record.offer_ids:
+                    record.state = "offer_received"
+                else:
+                    record.state = "new"
```
Mathilde411: Wouldn't have done it that way. Here you're using an inverse when you should really change the parent's state when the state of a child offer is changed, either by overriding the write method, or most probably by doing it in the method that changes the state of the offers

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,68 @@
+import datetime
+
+from odoo import models, fields, api
+from odoo.exceptions import UserError
+
+
+class PropertyOffer(models.Model):
+    _name = "estate.property.offer"
+    _description = "Estate Property Offer"
+    create_date = fields.Date(default=datetime.date.today())
```
Mathilde411: lambda is missing here

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,68 @@
...
+    price = fields.Float()
+    status = fields.Selection(
+        [
+            ("accepted", "Accepted"),
+            ("refused", "Refused"),
+        ],
+        copy=False,
+    )
```
Mathilde411, line 25: So the offer has no status if it is new ?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,68 @@
...
+    validity = fields.Integer(default=7)
+    deadline = fields.Date(compute="_compute_deadline", inverse="_compute_validity")
+    partner_id = fields.Many2one("res.partner", required=True)
+    property_id = fields.Many2one("estate.property", required=True)
+
+    _check_price = models.Constraint(
+        "CHECK(price > 0)",
+        message="Check the price",
```
Mathilde411: Message should be informative as it's what the user will see when the error triggers. Here:
```suggestion
        message="The price should be positive.",
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,133 @@
...
+            if record.state == "cancelled":
+                raise UserError("A cancelled property cannot be sold")
+            record.state = "sold"
+
+    def action_cancel(self):
+        for record in self:
+            if record.state == "sold":
+                raise UserError("A sold property cannot be cancelled")
+
+            record.state = "cancelled"
```
Mathilde411, line 128: Mind [attribute order](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions) too :)

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,68 @@
...
+            )
+            record.deadline = create_date + datetime.timedelta(days=record.validity)
+
+    @api.depends("deadline", "create_date")
+    def _compute_validity(self):
+        for record in self:
+            create_date = (
+                record.create_date if record.create_date else datetime.date.today()
+            )
+            record.validity = (record.deadline - create_date).days
```
Mathilde411: This should be `_inverse_deadline` also inverse functions don't need depends

estate/models/estate_property.py
Mathilde411: Just saw your offer file, this definitely should be when a record is created

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class PropertyTag(models.Model):
+    _name = 'estate.property.tag'
+    _description = 'Estate Property Tag'
+
+    name = fields.Char(string="Type", required=True)
+
+    _check_name = models.Constraint(
+        "unique(name)",
```
Mathilde411: Message missing here too

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,12 @@
+from odoo import models, fields
+
+
+class PropertyTag(models.Model):
+    _name = 'estate.property.tag'
+    _description = 'Estate Property Tag'
+
+    name = fields.Char(string="Type", required=True)
```
Mathilde411: The string is not correct

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,36 @@
...
+<odoo>
+
+    <record id='estate_property_offer_view_form' model="ir.ui.view">
+        <field name="name">estate.property.offer.view.form</field>
+        <field name="model">estate.property.offer</field>
+        <field name="arch" type="xml">
+            <form>
+                <group>
+                    <field name="partner_id"/>
+                    <field name="status" placeholder="status currently unknown"/>
```
Mathilde411: That's what I was talking about here you are forced to use a placeholder, if you make status required and add a default, you don't need the placeholder here

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,110 @@
...
+                        </h1>
+                    </div>
+                    <group>
+                        <field name="tag_ids" widget="many2many_tags" placeholder="Select a tag."/>
+                    </group>
+
+
+                    <group>
+                        <group>
+                            <field name="state" readonly="1" />
```
Mathilde411: For this you can lookup widget statusbar, it'll display as states usually do in odoo

Mathilde411 (changes requested):
Quite a lot of comment but it's pretty good overall !
Good job !

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,102 @@
...
+        for record in self:
+            if record.price < best_price:
+                raise UserError("You cannot go under the price of the current best offer")
+
+            if record.property_id.state not in ["sold", "cancelled"]:
+                record.property_id.state = "offer_received" if record.status != "accepted" else "offer_accepted"
+            else:
+                raise AccessError("You cannot create an offer on a sold or cancelled property.")
+
+        return super().write(vals)
```
Mathilde411, line 95: This wouldn't trigger on creation of the offer, you need to override create for that

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,102 @@
...
+
+    # Action methods
+    def action_accept(self):
+        for record in self:
+            if record.status == "refused":
+                raise UserError("A refused offer cannot be accepted")
+            if record.property_id.state == "cancelled":
+                raise UserError("An offer on a cancelled property cannot be accepted")
+            if record.property_id.state == "sold":
+                raise UserError("An offer on a sold property cannot be accepted")
```
Mathilde411: Maybe group those ? The button shouldn't be displayed to begin with if the condition are not met, so I'd make the code cleaner by gouping them into only one exit condition

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,102 @@
...
+            record.property_id.state = "offer_accepted"
+
+    def action_refuse(self):
+        for record in self:
+            if record.status == "accepted":
+                raise UserError("An accepted offer cannot be refused")
+            if record.property_id.state == "cancelled":
+                raise UserError("An offer on a cancelled property cannot be refused")
+            if record.property_id.state == "sold":
+                raise UserError("An offer on a sold property cannot be refused")
```
Mathilde411: Same

estate_account/models/estate_account.py
```diff
@@ -0,0 +1,47 @@
...
+                    for x in order.offer_ids
+                    if x.status != "refused"
+                ),
+                default=(None, None),
+                key=lambda x: x[0],
+            )
+
+            if not buyer_id or not selling_price:
+                raise UserError("You cannot sell without a buyer")
+
```
Mathilde411: Elegant but
```suggestion
            best_offer = order.offer_ids.filtered(lambda x: x.status != "refused").sorted('price DESC')[:1]

            if not best_offer:
                raise UserError("You cannot sell without a buyer")
```
And then use `best_offer.partner_id`.
Also shouldn't it be `x.status == "accepted"` ?

Mathilde411 (changes requested):
Congratulations ! This is very good, only some logic fixes ! :D

PR #1261, opened 2026-05-08
[ADD] estate: add empty module shell

patja-odoo:
Create the initial Odoo tutorial addon structure so the module is recognized in 
Apps and can be installed as an empty shell.

This PR only adds the minimal required files (manifest.py and init.py) and does 
not introduce any models/views yet, following the Odoo 
“Chapter 2: A New Application” tutorial step.


awesome_estate/models/awesome_estate_property_offer.py
```diff
@@ -0,0 +1,28 @@
...
+        'res.partner',
+        string='Buyer',
+        required=True,
+    )
+    property_id = fields.Many2one(
+        'awesome.estate.property',
+        string='Property',
+        required=True,
+        ondelete='cascade',
+        index=True,
```
mash-odoo, line 33: What difference will it make if I am writing `index=True`?

awesome_estate/models/awesome_estate_property_offer.py
```diff
@@ -0,0 +1,28 @@
...
+class AwesomeEstatePropertyOffer(models.Model):
+    _name = 'awesome.estate.property.offer'
+    _description = 'Real Estate Property Offer'
+    _order = 'price desc, id desc'
+
+    price = fields.Float()
+    status = fields.Selection(
+        [
+            ('accepted', 'Accepted'),
+            ('refused', 'Refused'),
```
mash-odoo:
```suggestion
            ('accepted', "Accepted"),
            ('refused', "Refused"),
```
Always try to keep the key in single quotes and value in double quotes

awesome_estate/models/awesome_estate_property_offer.py
```diff
@@ -0,0 +1,28 @@
...
+        ],
+        copy=False,
+    )
+    partner_id = fields.Many2one(
+        'res.partner',
+        string='Buyer',
```
mash-odoo:
```suggestion
        string="Buyer",
```
The string/value/text which are going to be shown on UI or displayed to the users, should always be wrapped in double quotes.

awesome_estate/views/awesome_estate_property_type_views.xml
```diff
@@ -0,0 +1,36 @@
...
+        <field name="arch" type="xml">
+            <form>
+                <sheet>
+                    <group>
+                        <field name="name"/>
+                    </group>
+                </sheet>
+            </form>
+        </field>
+    </record>
```
mash-odoo, line 53: There's just one field here so I don't think there's a need to open a form view here. Can we just add the value directly in the list view?

awesome_estate/views/awesome_estate_property_type_views.xml
```diff
@@ -0,0 +1,36 @@
...
+
+    <record id="awesome_estate_property_type_action" model="ir.actions.act_window">
+        <field name="name">Property Types</field>
+        <field name="res_model">awesome.estate.property.type</field>
+        <field name="view_mode">list,form</field>
+    </record>
+
+    <menuitem id="awesome_estate_property_type_menu" name="Property Types" parent="awesome_estate_first_level_menu" action="awesome_estate_property_type_action"/>
```
mash-odoo:
```suggestion
    <menuitem
        id="awesome_estate_property_type_menu"
        name="Property Types"
        parent="awesome_estate_first_level_menu"
        action="awesome_estate_property_type_action"
    />
```
Try to keep it readable directly through code editor, otherwise we would need to scroll in order to read the code. Make a practice that each line should be <100 chars... if it's more, then try breaking it.

awesome_estate/views/awesome_estate_property_views.xml
```diff
@@ -0,0 +1,131 @@
...
+                        <page string="Description">
+                            <field name="description"/>
+                        </page>
+                        <page string="Offers">
+                            <field name="offer_ids">
+                                <list editable="bottom">
+                                    <field name="price"/>
+                                    <field name="partner_id"/>
+                                    <field name="status"/>
+                                </list>
```
mash-odoo: What if we do not write this list view here?

awesome_estate/views/awesome_estate_property_views.xml
```diff
@@ -0,0 +1,131 @@
...
+                <field name="name"/>
+                <field name="property_type_id"/>
+                <field name="buyer_id"/>
+                <field name="salesperson_id"/>
+                <field name="tag_ids"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <separator/>
```
mash-odoo, line 228: What is a separator?

awesome_estate/views/awesome_estate_property_views.xml
```diff
@@ -0,0 +1,131 @@
...
+                <filter name="available" string="Available"
+                        domain="[('state', 'in', ['new', 'offer_received'])]"/>
```
mash-odoo:
```suggestion
                <filter
                    name="available" string="Available"
                    domain="[('state', 'in', ['new', 'offer_received'])]"
                />
```
Keep a proper indentation and format. Apply this wherever needed.

awesome_estate/views/awesome_estate_property_views.xml
```diff
@@ -0,0 +1,131 @@
...
+                <group name="groupby">
+                    <filter name="groupby_postcode" string="Postcode"
+                            context="{'group_by': 'postcode'}"/>
```
mash-odoo: What's the difference between a `domain` and a `context`?

awesome_estate/models/awesome_estate_property.py
```diff
@@ -0,0 +1,60 @@
...
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    property_type_id = fields.Many2one('awesome.estate.property.type')
+    buyer_id = fields.Many2one('res.partner', string='Buyer', copy=False)
+    salesperson_id = fields.Many2one(
+        'res.users',
+        string='Salesperson',
+        default=lambda self: self.env.user,
+    )
+    tag_ids = fields.Many2many('awesome.estate.property.tag')
```
mash-odoo: From where can I view the tags associated with the specific property from the database?

awesome_estate/models/awesome_estate_property.py
```diff
@@ -0,0 +1,60 @@
...
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    property_type_id = fields.Many2one('awesome.estate.property.type')
+    buyer_id = fields.Many2one('res.partner', string='Buyer', copy=False)
+    salesperson_id = fields.Many2one(
+        'res.users',
+        string='Salesperson',
+        default=lambda self: self.env.user,
```
mash-odoo, line 66: What is a lambda function?

awesome_estate/views/awesome_estate_property_views.xml
```diff
@@ -0,0 +1,131 @@
...
+                            <field name="garden"/>
+                        </group>
+                    </group>
+                    <group>
+                        <group string="Garden">
+                            <field name="garden_area"/>
+                            <field name="garden_orientation"/>
+                        </group>
+                        <group string="Tags">
+                            <field name="tag_ids" widget="many2many_tags"/>
```
mash-odoo: What's a widget?

awesome_estate/views/awesome_estate_property_views.xml
```diff
@@ -0,0 +1,131 @@
...
+                <field name="state"/>
+            </list>
+        </field>
+    </record>
+
+    <!-- Form View -->
+    <record id="awesome_estate_property_view_form" model="ir.ui.view">
+        <field name="name">awesome.estate.property.form</field>
+        <field name="model">awesome.estate.property</field>
+        <field name="arch" type="xml">
```
mash-odoo, line 34: Have you explore how this is stored in DB?

awesome_estate/models/awesome_estate_property.py
```diff
@@ -0,0 +1,60 @@
+from datetime import date
+
+from dateutil.relativedelta import relativedelta
+
+from odoo import fields, models
+
+
+class AwesomeEstateProperty(models.Model):
```
mash-odoo, line 6: Why have you used here models.Model specifically?

mash-odoo (changes requested):
Hello!
Good going on the task..
I have left a few questions...please have a look!

awesome_estate/models/awesome_estate_property_offer.py
patja-odoo: Noted!

awesome_estate/views/awesome_estate_property_type_views.xml
patja-odoo: Okay, I'll keep track of that.

awesome_estate/views/awesome_estate_property_type_views.xml
patja-odoo: Yes, I'll add `editable="bottom"` so users can type the value directly in the list without opening a form. Since there's only one field, the form view doesn't add anything extra.

awesome_estate/models/awesome_estate_property_offer.py
patja-odoo: It creates a b-tree index on that column such that when ever a search happens,
(example: WHERE property_id = 5) instead of a sequential scan across every row, 
the db uses the index to jump directly to matching rows.

awesome_estate/views/awesome_estate_property_views.xml
patja-odoo: A widget is an XML attribute that tells Odoo which JavaScript component to use for rendering a field in the UI.

awesome_estate/models/awesome_estate_property.py
patja-odoo: `models.Model` because I need to store property data in database storage.

And as 

`models.TransientModel`, also creates a database table, but records are temporary
as the ORM's garbage collector periodically deletes old transient records.

Lastly

`models.AbstractModel`, does NOT create a database table by default(_auto=false), provides code reuse through mixin inheritance.

Properties, offers, tags, and property types all represent real business entities that need to persist. So `models.Model` is the correct choice for all of them.

awesome_estate/views/awesome_estate_property_views.xml
patja-odoo: Yes, each `<record model="ir.ui.view">` becomes a row in the `ir_ui_view` table.

awesome_estate/models/awesome_estate_property.py
patja-odoo: A Many2many field creates a junction table in PostgreSQL. Odoo names it `{table1}_{table2}_rel` with table names sorted alphabetically. Here in this case it'll be:

```sql
SELECT * FROM awesome_estate_property_awesome_estate_property_tag_rel
WHERE awesome_estate_property_id = 1;
```

awesome_estate/models/awesome_estate_property.py
patja-odoo: A lambda is Python's syntax for an anonymous inline function. `lambda self: self.env.user` means "a function that takes `self` as its argument and returns `self.env.user`."

awesome_estate/models/awesome_estate_property.py
```diff
@@ -0,0 +1,108 @@
+from datetime import date
+
+from dateutil.relativedelta import relativedelta
```
mash-odoo:
```suggestion
from datetime import date
from dateutil.relativedelta import relativedelta
```

awesome_estate/views/awesome_estate_property_views.xml
```diff
@@ -0,0 +1,175 @@
...
+        id="awesome_estate_first_level_menu"
+        name="Properties"
+        parent="awesome_estate_root_menu"
+    />
+    <menuitem
+        id="awesome_estate_property_menu"
+        name="Properties"
+        parent="awesome_estate_first_level_menu"
+        action="awesome_estate_property_action"
+    />
```
mash-odoo: This works but you can separate the menus in a different file for a better architectural perspective.

awesome_estate/models/awesome_estate_property.py
```diff
@@ -0,0 +1,229 @@
...
+        copy=False,
+        default=lambda self: fields.Date.add(fields.Date.context_today(self), months=3),
+    )
+    expected_price = fields.Float(string="Expected Price", required=True)
+    selling_price = fields.Float(string="Selling Price", copy=False)
+    bedrooms = fields.Integer(string="Bedrooms", default=2)
+    living_area = fields.Integer(string="Living Area (sqm)")
+    facades = fields.Integer(string="Facades")
+    garage = fields.Boolean(string="Garage")
+    garden = fields.Boolean(string="Garden")
```
mash-odoo: You don't need to explicitly define string here as it will take from the field name itself.

awesome_estate/models/awesome_estate_property.py
```diff
@@ -0,0 +1,229 @@
...
+            ('south', "South"),
+            ('east', "East"),
+            ('west', "West"),
+        ],
+        string="Garden Orientation",
+    )
+    property_type_id = fields.Many2one(
+        'awesome.estate.property.type',
+        string="Property Type",
+        ondelete='restrict',
```
mash-odoo: Why is this `ondelete='restrict'` used?

awesome_estate/models/awesome_estate_property.py
```diff
@@ -0,0 +1,229 @@
...
+    total_area = fields.Integer(
+        string="Total Area (sqm)",
+        compute='_compute_total_area',
+        store=True,
+        help="Total area computed by summing the living area and the garden area.",
+    )
+    best_price = fields.Float(
+        string="Best Offer",
+        compute='_compute_best_price',
+        store=True,
```
mash-odoo, line 102: When should `store=True` be used?

awesome_estate/models/awesome_estate_property.py
```diff
@@ -0,0 +1,229 @@
...
+        help="Best offer received.",
+    )
+
+    # -----------------------------------------------------------------------
+    # SQL Constraints
+    # -----------------------------------------------------------------------
+    _check_living_area = models.Constraint(
+        'CHECK (living_area >= 0 AND living_area <= 100000)',
+        'Living area must be between 0 and 100,000 sqm. Please enter a realistic value.',
```
mash-odoo:
```suggestion
        "Living area must be between 0 and 100,000 sqm. Please enter a realistic value.",
```

awesome_estate/models/awesome_estate_property.py
```diff
@@ -0,0 +1,229 @@
...
+    _check_bedrooms = models.Constraint(
+        'CHECK (bedrooms >= 0)',
+        'Bedrooms cannot be negative.',
+    )
+    _check_facades = models.Constraint(
+        'CHECK (facades >= 0)',
+        'Facades cannot be negative.',
+    )
+
+    @api.constrains('selling_price', 'expected_price')
```
mash-odoo, line 152: What is the difference between SQL and Python constraints?
Do you know when to use which?

awesome_estate/models/awesome_estate_property.py
```diff
@@ -0,0 +1,229 @@
...
+        # Guard: must have a selling price
+        if not self.selling_price:
+            raise UserError(_("Set a selling price before selling the property."))
+        self.state = 'sold'
+        return True
+
+    def action_cancel(self):
+        self.ensure_one()
+        if self.state == 'sold':
+            raise UserError(_("Sold properties cannot be canceled."))
```
mash-odoo, line 279: Why is this `_` used?

awesome_estate/models/awesome_estate_property.py
```diff
@@ -0,0 +1,229 @@
...
+            raise UserError(_("Property is already canceled."))
+        # Refuse accepted offers via action_refuse() with force_refuse context
+        # to bypass the "already accepted" guard in the offer model.
+        accepted = self.offer_ids.filtered(lambda o: o.status == 'accepted')
+        if accepted:
+            accepted.with_context(force_refuse=True).action_refuse()
+        # Refuse any remaining pending offers (status is falsy, so direct write works)
+        pending = self.offer_ids.filtered(lambda o: not o.status)
+        if pending:
+            pending.write({'status': 'refused'})
```
mash-odoo:
```suggestion
        if pending:
            pending.status = 'refused'
```
This can also be written like this. 
So wdyt? Which approach is better and why?

awesome_estate/models/awesome_estate_property.py
```diff
@@ -0,0 +1,229 @@
...
+            raise UserError(_("Only sold or canceled properties can be reset."))
+        was_sold = self.state == 'sold'
+        self.write({
+            'state': 'new',
+            'selling_price': 0.0,
+            'buyer_id': False,
+        })
+        if was_sold:
+            self.offer_ids.write({'status': False})
+        return True
```
mash-odoo, line 324: So you have added a new button to reset the property values.
But can you explain me the business case where this feature would be efficient....
Like in what conditions would this be useful?

awesome_estate/models/awesome_estate_property.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def action_accept_best_offer(self):
+        self.ensure_one()
```
mash-odoo, line 328: Why is this needed?

awesome_estate/models/awesome_estate_property.py
```diff
@@ -0,0 +1,229 @@
...
+        if self.state == 'sold':
+            raise UserError(_("Property is already sold."))
+        # Auto-fill selling_price and buyer_id from the accepted offer
+        if self.state == 'offer_accepted' and not self.selling_price:
+            accepted = self.offer_ids.filtered(lambda o: o.status == 'accepted')
+            if accepted:
+                self.write({
+                    'selling_price': accepted[0].price,
+                    'buyer_id': accepted[0].partner_id.id,
+                })
```
mash-odoo: Will we ever get multiple offers in `accepted`?
Also can you help me understand why we should have this `accepted[0]`?

awesome_estate/models/awesome_estate_property_offer.py
```diff
@@ -0,0 +1,194 @@
...
+
+    # -----------------------------------------------------------------------
+    # CRUD Methods
+    # -----------------------------------------------------------------------
+    def write(self, vals):
+        # Lightweight guard: block changing to a *different* status once set.
+        # Prevents corrupting an accepted/refused offer via inline editable list.
+        # Clearing to False is always allowed (resets from action_reset()).
+        # The 'force_refuse' context bypass allows property cancellation to work.
+        if 'status' in vals and vals.get('status') and not self.env.context.get('force_refuse'):
```
mash-odoo: Can you explain me the concept of using `force_refuse`?

mash-odoo (changes requested):
Hello,
I have some questions regarding your PR.
1. We might be having some dead code too, so maybe you can go through your entire PR once and remove stuff which is not needed anymore.
And if you want to keep it for reference maybe you can comment it out.
2. If you are trying to implement something new, you need to understand it's purpose before using and not just implement it just for the sake of it
3. Your task is not to complete the chapters but rather understand it

awesome_estate/models/awesome_estate_property.py
patja-odoo: Alright noted!

awesome_estate/views/awesome_estate_property_views.xml
patja-odoo: Yes, I have created a separate XML for menus

awesome_estate/models/awesome_estate_property.py
patja-odoo: `_()` is Odoo's translation function. It marks user-visible strings for translation into different languages.

awesome_estate/models/awesome_estate_property.py
patja-odoo: `store=True` should be used when the computed field needs to persistent in database operations, for this case - `best_price` is used in search filters which requires it to be in the db hence, `store=True`

awesome_estate/models/awesome_estate_property.py
patja-odoo: `ensure_one()` validates that the method is being called on exactly one record, not a recordset of 0 or many.

awesome_estate/models/awesome_estate_property.py
patja-odoo: Initially the implementation had no guards, so `accepted[0]` was necessary to safely pick the first from a recordset that could've contained multiple accepted offers. After adding the `write()` guard, `_accept_process()`, and the property state check, we would almost never get more than one accepted offer. 

Since the functionality changed, we should trace back the full dependency flow and clean up accordingly, accepted[0] is now effectively redundant and shall be marked for clean up.

awesome_estate/models/awesome_estate_property.py
patja-odoo: `ondelete='restrict'` prevents the deletion of a property type if any property references it. If a user tries to delete "House" type while 50 properties are typed "House", they won't be able to do so due to the applied restriction.

awesome_estate/models/awesome_estate_property.py
patja-odoo: Okay for a business use case:

Let's say a property manager accidentally clicks "Sold" while exploring the module. Now without reset, that property is stuck, user cannot do anything in that property now, no new offers, no accept, no refuse, all actions are restricted in the cancelled state right, that's where we use reset, they revert to 'new' state where use can edit offers, and perform just the same property flow action normally, so that's one scenario I thought of while implementing the reset flow.

awesome_estate/models/awesome_estate_property.py
patja-odoo: SQL constraints (`models.Constraint`), they are at the database level by PostgreSQL. They are defined on the model class and called automatically on every INSERT/UPDATE. Used for simple single-field checks like (`price > 0`).

Python constraints (`@api.constrains`), they are at the ORM level in Python. They are fields dependent, they are called when when any of those fields are modified. Used when we need multi-field logic (e.g., `selling_price >= 90% of expected_price`).

From the [official Odoo 19 docs](https://www.odoo.com/documentation/19.0/developer/tutorials/server_framework_101/10_constraints.html):

 *"SQL constraints are usually more efficient than Python constraints. When performance matters, always prefer SQL over Python constraints."*

awesome_estate/views/awesome_estate_property_views.xml
patja-odoo: `<separator/>`  is a horizontal divider line that is used to visually separate certain UI elements.

awesome_estate/views/awesome_estate_property_views.xml
patja-odoo: If we remove this inline `<list>`, Odoo falls back to the default list view  (via ` _get_default_list_view()`)

awesome_estate/views/awesome_estate_property_views.xml
patja-odoo: __`domain`__ filters which records appear or say which record to retrieve/filter whereas, generally a WHERE query with conditions like `>`, `<`, `!=` and etc

__`context`__ says how records are displayed/sorted (grouping, language, timezone, defaults, view selection).

PR #1263, opened 2026-05-08
[IMP] mrp_bom_overview: clean forecast view with consolidated status

times-odoo:
The BOM overview forecast view is simplified by removing visual redundancy and consolidating availability information into the Status column.

Changes introduced via patch on BomOverviewLine and inheritance of report.mrp.report_bom_structure:

- Remove the standalone BOM name header section as it duplicates the breadcrumb information already visible in the page header
- Remove the Availability column to reduce redundancy and consolidate all stock information into the Status column
- Display quantity as integer to avoid unnecessary decimal places on whole-number quantities
- Ensure 'Ready To Produce' tag is only shown when producible quantity is greater than zero
- Always render 'Ready To Produce' tag in green for visual consistency regardless of other conditions
- Each component now independently displays its availability state as a colored tag: green for available, dark for estimated, orange for expected, and red for not available
- When components are missing, top-level BOM line shows estimated production date instead of 'No Ready To Produce'
- All status tags are clickable and navigate to the respective product forecast report

task-6204743

### DIFF
--- a/mrp_bom_overview_forecast/__manifest__.py
+++ b/mrp_bom_overview_forecast/__manifest__.py
@@ -1,0 +1,14 @@
+{
+    "name": "bom_overview",
+    "version": "1.0",
+    "depends": ["mrp", "purchase"],
+    "author": "times",
+    "category": "Tutorials",
+    "license": "LGPL-3",
+    'installable': True,
+    "assets": {
+        "web.assets_backend": [
+            "mrp_bom_overview_forecast/static/src/**/*",
+        ],
+    },
+}
--- a/mrp_bom_overview_forecast/__manifest__.py
+++ b/mrp_bom_overview_forecast/__manifest__.py
@@ -1,11 +1,12 @@
 {
-    "name": "bom_overview",
+    "name": "mrp_bom_overview_forecast",
     "version": "1.0",
     "depends": ["mrp", "purchase"],
     "author": "times",
     "category": "Tutorials",
     "license": "LGPL-3",
     'installable': True,
+    'auto_install': True,
     "assets": {
         "web.assets_backend": [
             "mrp_bom_overview_forecast/static/src/**/*",
--- a/mrp_bom_overview_forecast/static/src/bom_overview.xml
+++ b/mrp_bom_overview_forecast/static/src/bom_overview.xml
@@ -9,27 +9,27 @@
     </t>
 
     <t t-name="mrp.BomOverviewLine" t-inherit="mrp.BomOverviewLine" t-inherit-mode="extension">
-        <xpath expr="//td[@class='text-center' and .//div[hasclass('o_field_badge')]]" position="replace">
+        <xpath expr="//td[.//div[@t-if=&quot;data.hasOwnProperty('status')&quot;]]" position="replace">
             <td t-if="forecastMode" class="text-center">
                 <div t-if="data.is_storable" class="o_field_badge" role="button" t-on-click.prevent="goToForecast">
                     <span t-attf-class="badge rounded-pill o_mrp_overview_badge {{statusBackgroundClass}}" t-out="statusData"/>
                 </div>
             </td>
         </xpath>
-        <xpath expr="//td[@class='text-center' and .//t[@t-if=&quot;data.hasOwnProperty('availability_state')&quot;]]" position="replace"/>
+        <xpath expr="//td[.//t[@t-if=&quot;data.hasOwnProperty('availability_state')&quot;]]" position="replace"/>
     </t>
 
     <t t-name="mrp.BomOverviewTable.headers" t-inherit="mrp.BomOverviewTable" t-inherit-mode="extension">
-        <xpath expr="//th[@title='Reception time estimation.']" position="replace"/>
-        <xpath expr="//tfoot//tr[1]/td[@t-if='forecastMode'][3]" position="replace"/>
+        <xpath expr="//th[@name='th_mrp_bom_h']/following-sibling::th[@t-if='forecastMode'][3]" position="replace"/>
+        <xpath expr="//td[@name='td_mrp_bom_f']/following-sibling::td[@t-if='forecastMode'][3]" position="replace"/>
     </t>
 
     <t t-name="mrp.BomOverviewTable.byproducts_footer" t-inherit="mrp.BomOverviewTable" t-inherit-mode="extension">
-        <xpath expr="//tfoot//t//tr/td[@t-if='forecastMode'][3]" position="replace"/>
+        <xpath expr="//td[@name='td_mrp_bom_b']/following-sibling::td[@t-if='forecastMode'][3]" position="replace"/>
     </t>
 
     <t t-name="mrp.BomOverviewSpecialLine" t-inherit="mrp.BomOverviewSpecialLine" t-inherit-mode="extension">
-        <xpath expr="//td[@t-if='forecastMode'][3]" position="replace"/>
+        <xpath expr="//td[@name='bom_cost']/preceding-sibling::td[@t-if='forecastMode'][last()]" position="replace"/>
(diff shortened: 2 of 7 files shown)


mrp_bom_overview_forecast/static/src/bom_overview_line.js
```diff
@@ -0,0 +1,35 @@
+import { patch } from "@web/core/utils/patch";
+import { BomOverviewLine } from "@mrp/components/bom_overview_line/mrp_bom_overview_line";
+
+patch(BomOverviewLine.prototype, {
+
+    get statusData() {
+        if (this.data.hasOwnProperty('components_available') && this.data.status && this.data.status !== "No Ready To Produce") {
```
mash-odoo: This will definitely break in translated environments. 
For eg:
No Ready To Produce (EN) → Aucune production prête (FR)
So the comparison will fail entirely though it means the same

mrp_bom_overview_forecast/static/src/bom_overview_line.js
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        return "Not Available";
+    },
+
+    get statusBackgroundClass() {
+        if (!this.statusData) {
+            return "text-bg-danger";
+        }
+        if (this.statusData === "Available" || this.statusData.includes("Ready To Produce")) {
```
mash-odoo: Same as above, this will not work in the translated environment

mrp_bom_overview_forecast/static/src/bom_overview_line.js
```diff
@@ -0,0 +1,35 @@
...
+            return "text-bg-success";
+        }
+        if (this.statusData.includes("Expected")) {
```
mash-odoo: Same for this as well

mrp_bom_overview_forecast/static/src/bom_overview_line.js
```diff
@@ -0,0 +1,35 @@
...
+            return "text-bg-warning";
+        }
+        if (this.statusData.includes("Estimated")) {
```
mash-odoo: and this too

mrp_bom_overview_forecast/static/src/bom_overview.xml
```diff
@@ -0,0 +1,41 @@
...
+                <div t-if="data.is_storable" class="o_field_badge" role="button" t-on-click.prevent="goToForecast">
+                    <span t-attf-class="badge rounded-pill o_mrp_overview_badge {{statusBackgroundClass}}" t-out="statusData"/>
+                </div>
+            </td>
+        </xpath>
+        <xpath expr="//td[@class='text-center' and .//t[@t-if=&quot;data.hasOwnProperty('availability_state')&quot;]]" position="replace"/>
+    </t>
+
+    <t t-name="mrp.BomOverviewTable.headers" t-inherit="mrp.BomOverviewTable" t-inherit-mode="extension">
+        <xpath expr="//th[@title='Reception time estimation.']" position="replace"/>
```
mash-odoo: check for this as well

mrp_bom_overview_forecast/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "bom_overview",
+    "version": "1.0",
+    "depends": ["mrp", "purchase"],
+    "author": "times",
+    "category": "Tutorials",
+    "license": "LGPL-3",
+    'installable': True,
```
mash-odoo, line 8: This won't be installed directly, and user wouldn't know the features coming along with this modules. So don't you think we should do something such that this module also get installed once it's dependency is installed?

mrp_bom_overview_forecast/static/src/bom_overview.xml
```diff
@@ -0,0 +1,41 @@
...
+        <xpath expr="//tfoot//tr[1]/td[@t-if='forecastMode'][3]" position="replace"/>
```
mash-odoo: What if some more t-ifs are added then maybe the position you are targeting would be compromised

mrp_bom_overview_forecast/static/src/bom_overview.xml
```diff
@@ -0,0 +1,41 @@
...
+
+    <t t-name="mrp.BomOverviewLine" t-inherit="mrp.BomOverviewLine" t-inherit-mode="extension">
+        <xpath expr="//td[@class='text-center' and .//div[hasclass('o_field_badge')]]" position="replace">
+            <td t-if="forecastMode" class="text-center">
+                <div t-if="data.is_storable" class="o_field_badge" role="button" t-on-click.prevent="goToForecast">
+                    <span t-attf-class="badge rounded-pill o_mrp_overview_badge {{statusBackgroundClass}}" t-out="statusData"/>
+                </div>
+            </td>
+        </xpath>
+        <xpath expr="//td[@class='text-center' and .//t[@t-if=&quot;data.hasOwnProperty('availability_state')&quot;]]" position="replace"/>
```
mash-odoo: You can try comparing with more strong attribute like id rather than class

mrp_bom_overview_forecast/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "bom_overview",
```
mash-odoo:
```suggestion
    "name": "mrp_bom_overview_forecast",
```

mrp_bom_overview_forecast/static/src/bom_overview_line.js
```diff
@@ -0,0 +1,35 @@
+import { patch } from "@web/core/utils/patch";
+import { BomOverviewLine } from "@mrp/components/bom_overview_line/mrp_bom_overview_line";
+
+patch(BomOverviewLine.prototype, {
```
mash-odoo, line 4: What happens if another module patches BomOverviewLine.prototype too?

mash-odoo (changes requested):
Hello,
I have done some initial review.
There are a few things which needs to be taken care of..
Please have a look at it 🐣 
Thanks!

mrp_bom_overview_forecast/static/src/bom_overview_line.js
times-odoo: resolved using comparing producible qty

mrp_bom_overview_forecast/static/src/bom_overview_line.js
times-odoo: resolved using state switch

mrp_bom_overview_forecast/static/src/bom_overview.xml
times-odoo: resolved with direct t-if search

mrp_bom_overview_forecast/__manifest__.py
times-odoo: added 'auto_install' : True

mrp_bom_overview_forecast/static/src/bom_overview.xml
times-odoo: added following and preceding sibling search to get exact t-ifs

mrp_bom_overview_forecast/static/src/bom_overview.xml
times-odoo: there is no id in the parent xml, so switched to @name instead of @class

mrp_bom_overview_forecast/static/src/bom_overview_line.js
times-odoo: patch() supports stacking — multiple patches are applied in module load order. Since we depend on mrp, our patch loads after it. If another module needs to extend our logic, they can depend on our module and patch again, using super to call our version.

PR #1266, opened 2026-05-08
[ADD] estate : created new estate module

sadeo-odoo:
Set up the initial Real Estate module by creating the required module structure and manifest file. Successfully installed the module and defined the base models and fields, enabling Odoo ORM to generate the corresponding database tables automatically.


estate/models/estate_property.py
```diff
@@ -0,0 +1,93 @@
...
+
+    property_type_id = fields.Many2one(
+        "estate.property.type",
+        string="Property Type",
+    )
+
+    tag_ids = fields.Many2many(
+        "estate.property.tag",
+        string="Tags",
+    )
```
mash-odoo, line 82: So when i check in the db, I see that there is a new table created for establishing the relation between property and tags. Why is this happening?

estate/models/estate_property.py
```diff
@@ -0,0 +1,93 @@
...
+    offer_ids = fields.One2many(
+        "estate.property.offer",
+        "property_id",
+        string="Offers",
+    )
+    best_offer = fields.Float(string="Best offer", compute="_compute_best_price")
+
+    @api.depends("offer_ids.price")
```
mash-odoo: What are decorators?

estate/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+from . import estate_property
+from . import estate_property_type
+from . import estate_property_tag
+from . import estate_property_offer
```
mash-odoo, line 2: It's better to import in alphabetical order

estate/models/estate_property.py
```diff
@@ -0,0 +1,93 @@
...
+    def _compute_total(self):
+        for record in self:
+            record.total = record.garden_area + record.living_area
+
+    garden_orientation = fields.Selection(
+        [
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
```
```suggestion
        [
            ('north', "North"),
            ('south', "South"),
            ('east', "East"),
            ('west', "West"),
```
It's better to wrap key(or any technical strings) in single quotes and value(or any string to be displayed to user) in double quotes

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,40 @@
...
+        name="Properties"
+        parent="estate_menu_advetisement"
+        action="estate_property_action"
+    />
+
+    <menuitem
+        id="estate_menu_settings"
+        name="Settings"
+        parent="estate_menu_root"
+        sequence="2"
```
mash-odoo: What is this sequence for?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,122 @@
...
+        <field name="view_mode">list,form</field>
+    </record>
+
+    <record id="estate_property_list_view" model="ir.ui.view">
+        <field name="name">Estate.properties.list</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <list string="Properties" create="1" edit="1" delete="1" multi_edit="1"
+                open_form_view="0" default_group_by="state" default_order="expected_price desc"
+                decoration-danger="expected_price>1000000" limit="20" expand="1">
```
mash-odoo: 1. Writing or not writing `open_form_view="0"` wouldn't change anything as it is by default 0 only.
2. What would be changed if we are giving a `limit`?
3. `multi_edit` is used for?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,122 @@
...
+                <filter name="state" string="Available"
+                    domain="['|',('state','=','new'),('state','=','offer_recieved')]" />
+                <filter string="Postcode" name="group_by_postcode" context="{'group_by':'postcode'}" />
+
+            </search>
+        </field>
+
+    </record>
+
+</odoo>
```
mash-odoo, line 276: Always leave an extra line at the end of the file.

estate/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "estate",
+    "version": "1.0",
+    "depends": ["base"],
+    "category": "tutorials",
+    "author": "sadeo-odoo",
+    "license": "LGPL-3",
+    "description": "A real estate module",
+    "installable": True,
+    "application": True,
```
mash-odoo, line 10: What's the use case of writing both of these?

estate/models/estate_property.py
```diff
@@ -0,0 +1,93 @@
...
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer()
+    facades = fields.Integer()
+
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    total = fields.Float(compute="_compute_total")
+
+    @api.depends("garden_area", "living_area")
```
mash-odoo: What are decorators?

estate/models/estate_property.py
```diff
@@ -0,0 +1,93 @@
...
+
+    garden_orientation = fields.Selection(
+        [
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
+        ]
+    )
+
```
mash-odoo: Do proper ordering of fields, functions, etc. 
You can refer to [this](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions)

estate/models/estate_property.py
```diff
@@ -0,0 +1,93 @@
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate Property"
```
mash-odoo, line 12: Where can I view this?

mash-odoo (changes requested):
Hello!
Going good with the task..
I have added some comments.. please have a look!
Also, please update your PR description and title

estate/models/__init__.py
sadeo-odoo: Thanks for the review. I've fixed it and will keep this in mind for future changes.

estate/models/estate_property.py
sadeo-odoo: Yes, I've fixed it and will keep this in mind for future changes.

estate/models/estate_property.py
sadeo-odoo: A Many2many relationship cannot be stored directly in either of the two model tables because each property can have multiple tags and each tag can belong to multiple properties. To represent these associations, Odoo automatically creates an intermediate relation table that stores pairs of IDs (property_id, tag_id). This table acts as a bridge between the two models and allows Odoo to efficiently manage the Many2many relationship.

estate/models/estate_property.py
```diff
@@ -0,0 +1,207 @@
...
+
+    description = fields.Text()
+
+    postcode = fields.Char()
+
+    date_available = fields.Date(
+        default=lambda self: fields.Date.today() + relativedelta(months=3),
+    )
+
+    expected_price = fields.Float(required=True)
```
mash-odoo, line 22:
```suggestion
    name = fields.Char(required=True)
    description = fields.Text()
    postcode = fields.Char()
    date_available = fields.Date(
        default=lambda self: fields.Date.today() + relativedelta(months=3),
    )
    expected_price = fields.Float(required=True)
```
You don't need to add extra blank lines in between the fields

estate/models/estate_property.py
```diff
@@ -0,0 +1,207 @@
...
+                float_compare(
+                    record.selling_price,
+                    record.expected_price * 0.9,
+                    precision_digits=2,
+                )
+                < 0
+            ):
+                raise ValidationError(
+                    "Selling price cannot be lower than 90% of expected price!",
+                )
```
mash-odoo, line 197:
```suggestion
                raise ValidationError(_(
                    "Selling price cannot be lower than 90% of expected price!",
                ))
```
You can use translation wherever you are displaying something to the customer to make it language independent

estate/models/estate_property.py
```diff
@@ -0,0 +1,207 @@
...
+                raise UserError("A sold property cannot be cancelled")
+
+            record.state = "cancelled"
+
+        return True
+
+    def action_sold(self):
+        for record in self:
+            if record.state == "cancelled":
+                raise UserError("A cancelled property can't be sold")
```
mash-odoo: What's the difference between `UserError` and `ValidationError`?

estate/models/estate_property.py
```diff
@@ -0,0 +1,207 @@
...
+
+            record.state = "sold"
+
+        return True
+
+    @api.ondelete(at_uninstall=False)
```
mash-odoo, line 208: How does `at_uninstall=False` work?

estate/models/estate_property.py
```diff
@@ -0,0 +1,207 @@
...
+    def ondelete(self):
+        for record in self:
+            if record.state != "new" or record.state != "cancelled":
```
```suggestion
            if record.state != 'new' or record.state != 'cancelled':
```
Try keeping technical string in single quote

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,16 @@
...
+    _name = "estate.property.tag"
+    _description = "Estate Property Tag"
+    _order = "name"
+
+    name = fields.Char(required=True)
+
+    color = fields.Integer(string="Color")
+
+    _name_unique = models.Constraint(
+        "UNIQUE(name)",
```
mash-odoo: Will this `unique` work in all cases?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,243 @@
...
+
+    <record id="estate_property_list_view" model="ir.ui.view">
+        <field name="name">Estate.properties.list</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <list
+                string="Properties"
+                create="1"
+                edit="1"
+                delete="1"
```
mash-odoo, line 20: What difference will it make if you don't write this?

estate/models/estate_property.py
```diff
@@ -0,0 +1,207 @@
...
+                raise UserError("This property can't be deleted")
+
+    @api.depends("offer_ids", "offer_ids.partner_id", "offer_ids.create_date")
+    def _compute_is_suspicious(self):
+        now = fields.Datetime.now()
+        five_minutes_ago = now - timedelta(minutes=5)
+
+        Offer = self.env["estate.property.offer"]
```
mash-odoo: No need to declare a new variable if it's been used only once.

estate/models/estate_property.py
```diff
@@ -0,0 +1,207 @@
...
+
+        for property_record in self:
+            property_record.is_suspicious = False
+
+            partners = property_record.offer_ids.mapped("partner_id")
+
+            for partner in partners:
+                offer_count = Offer.search_count(
```
```suggestion
                offer_count = self.env["estate.property.offer"].search_count(
```
You can write it here directly

estate_account/__manifest__.py
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "Estate Account",
+    "author": "sadeo-odoo",
+    "license": "LGPL-3",
+    "depends": ["estate", "account"]
```
mash-odoo: Your `estate_account` module doesn't create a new application, but rather add features to the existing one. But the clients won't know about the existence of this module. So what can be done so that this module gets installed if estate and account both are installed?

estate/views/res_users_views.xml
```diff
@@ -0,0 +1,48 @@
...
+
+        <field name="model">res.users</field>
+
+        <field name="inherit_id"
+               ref="base.view_users_form"/>
+
+        <field name="arch" type="xml">
+
+            <xpath expr="//page[@name='preferences']"
+                   position="after">
```
mash-odoo: Can you re-format this file?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,243 @@
...
+            <search>
+                <field name="name" string="Title" />
+                <field name="postcode" />
+                <field name="expected_price" />
+                <field name="bedrooms" />
+
+                <field
+                    name="living_area"
+                    string="Living Area"
+                    filter_domain="[('living_area', '>=', self)]"
```
mash-odoo, line 202: What's the difference between `domain` and `filtered_domain`?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,243 @@
...
+                            <group>
+                                <field name="description" />
+                                <field name="name" />
+                                <field name="postcode" />
+                                <field name="bedrooms" />
+                                <field name="living_area" />
+                                <field name="garden" />
+                                <field
+                                    name="garden_area"
+                                    invisible="garden == False"
```
```suggestion
                                    invisible="not garden"
```

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,68 @@
...
+    <record id="estate_property_type_form_view" model="ir.ui.view">
+        <field name="name">estate.property.type.form</field>
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <form string="Property Type">
+                <sheet>
+                    <div class="oe_button_box" name="button_box">
+                        <button
+                            name="%(estate.action_open_related_offers)d"
+                            type="action"
```
mash-odoo, line 25: when do you use type="action" and type="object"?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,107 @@
...
+
+        for offer in self.property_id.offer_ids:
+            if self != offer:
+                # raise UserError("An offer is already accepted.")
+                offer.status = "rejected"
+
+        self.status = "accepted"
+        self.property_id.buyer_id = self.partner_id
+        self.property_id.selling_price = self.price
+        self.property_id.state = "offer_accepted"
```
mash-odoo: You can optimize this using `write`

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,107 @@
...
+
+    validity = fields.Integer(
+        string="Days",
+        default=7,
+    )
+
+    date_deadline = fields.Date(
+        string="Deadline Date",
+        compute="_compute_date_deadline",
+        inverse="_inverse_date_deadline",
```
mash-odoo: When should we use inverse function?

mash-odoo (changes requested):
Hello,
Good going on the task
I have some suggestions:
1. Please reformat all the files and take care of the spaces between lines. Refer to our codebase for the same. Avoid leaving empty lines between fields and in the views.
2. Please update the PR description
3. While pushing the commit also mention the chapter number.

estate/models/estate_property_offer.py
sadeo-odoo: We use an inverse function when a computed field should also be editable by the user. The compute method calculates the field automatically, while the inverse method tells Odoo how to update the original fields if the computed value is changed manually.

estate/models/estate_property_offer.py
sadeo-odoo: Yes, I'll update it using a single write() call. The current implementation invokes write() three times, which isn't optimal. I'll optimize it by combining the updates into one write() call to make the code cleaner and more efficient.

estate/views/estate_property_type_views.xml
sadeo-odoo: type="object" is used when you want to call a Python method defined on the current model. It's typically used for business logic, such as changing a record's state, validating data, or performing custom operations.

type="action" is used when you want to execute an existing Odoo action, such as opening another view, a wizard, or a report. Instead of calling a Python method directly, it references an action record.

estate/views/estate_property_views.xml
sadeo-odoo: domain is used to restrict the records that can be selected or displayed. For example, on a Many2one field, it limits the records shown in the dropdown.

filter_domain is used in search views to define how a user's input should be applied as a search filter. It doesn't restrict the available records; instead, it builds the search domain dynamically based on the value entered by the user.

estate/views/res_users_views.xml
sadeo-odoo: Sure, I'll reformat the file and apply the requested changes.

estate_account/__manifest__.py
sadeo-odoo: I used **auto_install = True** in the module manifest. This ensures that the estate_account module is installed automatically whenever both the estate and account modules are installed, so users don't need to install it manually.

estate/models/estate_property.py
sadeo-odoo: Yes, I've fixed it by using self.env["estate.property.offer"] directly.

estate/models/estate_property.py
sadeo-odoo: Yes, I've fixed it. I'll keep this in mind for future changes as well.

estate/models/estate_property.py
sadeo-odoo: Yes, I've fixed it. I'll keep this in mind for future changes as well.

estate/models/estate_property.py
sadeo-odoo: at_uninstall=False ensures that the @api.ondelete check is applied only during normal deletions. During module uninstallation, Odoo bypasses this method so the module and its data can be removed without raising validation errors.

estate/views/estate_property_views.xml
sadeo-odoo: These attributes are enabled by default, so they aren't strictly required. I added them explicitly to make the view configuration clearer, but they can be removed without changing the behavior. I'll remove them to keep the XML cleaner.

estate/models/estate_property_tag.py
sadeo-odoo: Not in all cases. UNIQUE(name) only checks for exact matches at the database level. It won't handle cases like different letter casing (e.g. House vs house) or leading/trailing spaces. In Odoo, it's better to normalize the value before saving or add a Python constraint if those cases also need to be treated as duplicates.

estate/models/estate_property.py
sadeo-odoo: Sure, I'll make the required changes.

estate/models/estate_property.py
sadeo-odoo: Sure, I'll make the required changes.

estate/models/estate_property.py
sadeo-odoo: UserError is used to stop an operation and show a message to the user when a business rule is violated. For example, trying to accept an offer on a property that is already sold.

ValidationError is used when the data itself is invalid or doesn't satisfy a constraint. It's commonly used in SQL or Python constraints to prevent invalid data from being saved, such as a selling price lower than the allowed minimum.

estate/models/estate_property.py
sadeo-odoo: Decorators are used to tell Odoo when and how a method should be executed. They add special behavior to a method without changing its core logic.

estate/views/estate_menus.xml
sadeo-odoo: The sequence field is used to control the order in which records are displayed. Records with a lower sequence value appear first, while those with a higher value appear later.

estate/views/estate_property_views.xml
sadeo-odoo: Sure, I'll make the required changes.

estate/__manifest__.py
sadeo-odoo: installable=True tells Odoo that the module can be installed. If it's set to False, the module won't be available for installation.

application=True marks the module as an application, so it appears in the Apps menu with its own entry.

PR #1269, opened 2026-05-15
19.0 bom forecast sopat

sopat-odoo:
Introduce a custom module that improves the BOM Overview
forecast view for cleaner status display.

WHY this change is being done:
Currently, the BOM Overview shows Estimated badge for both
main product and sub-components without distinction, making
it hard to differentiate availability status at each level.

Technical decisions taken:
1. Inherited report.mrp.report_bom_structure to add
   status field with producible qty at level 0.
2. Used patch() on BomOverviewLine to add statusBackgroundColor
   and statusDisplay getters for dynamic badge styling.
3. Replaced Estimated text with Expected for sub-components
   (level > 0) to clearly distinguish from main product.
4. All status badges are clickable and lead to product
   forecast report for quick navigation.
5. Removed redundant Availability column to keep the
   view clean and focused on Status column.
   
   Task ID- 6213345


mash-odoo:
Hello,
Before I start reviewing, I am getting this unnecessary diff:
image
You just needed to take the clone of tutorials branch and push the new diff. But here it seems that you have checked out from your estate branch and then removed it. The unnecessary diff is because in the main branch, they had some code for awesome_dashboard and other files which you have removed. 
Please make this PR clean.

bom_forecast/static/src/mrp_bom_overview_table.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+    <t t-name="mrp.change_forecast_view" t-inherit="mrp.BomOverviewTable" t-inherit-mode="extension">
+        <xpath expr="//a[@t-on-click.prevent='goToProduct']/ancestor::div[2]" position="replace"/>
```
mash-odoo: What if the parent div[2] which you are currently pointed is changed by someone and some more divs are added above? You should always try giving meaningful xpaths like div with id="something" so there is no space left for errors

bom_forecast/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "bom_forecast",
+    "version": "1.0",
+    "depends": ["mrp", "purchase"],
+    "author": "soham",
+    "category": "Tutorials",
+    "license": "LGPL-3",
+    'installable': True,
+    'auto_install': True,
+    'application': True,
```
mash-odoo: Do we really need this here?

bom_forecast/static/src/bom_overview_line.js
```diff
@@ -0,0 +1,28 @@
...
+    get statusBackgroundColor() {
+        switch (this.data.availability_state) {
+            case "available": return "text-bg-success";
+            case "expected":
+            case "estimated":
+                if (this.data.level === 0) return "text-bg-dark";
+                return "border border-warning text-warning";
+            case "unavailable": return "text-bg-danger";
+            default: return "text-bg-dark";
+        }
```
mash-odoo: This is wrong in many ways.
Why do you need to handle this from JS? Do you know what patch does? Can you have a look in codebase for similar examples?

bom_forecast/static/src/bom_overview_line.js
```diff
@@ -0,0 +1,28 @@
...
+    },
+
+    get statusDisplay() {
+        if (this.data.level > 0 &&
+            this.data.availability_display &&
+            this.data.availability_display.includes("Estimated")) {
+            return this.data.availability_display.replace(
+                "Estimated", "Expected"
+            );
```
mash-odoo: I don't think this should be done because if this is not in English and in some other language, your logic will completely break. WDYT?

bom_forecast/static/src/bom_overview_line.xml
```diff
@@ -0,0 +1,38 @@
...
+        <xpath expr="//td//div[hasclass('o_field_badge')]" position="replace">
+            <div class="o_field_badge">
+
+                <t t-if="data.type == 'operation'">
+                    <span t-attf-class="badge rounded-pill o_mrp_overview_badge {{statusBackgroundColor}}"
+                        t-esc="data.availability_display"/>
+                </t>
+
+                <t t-elif="data.level === 0">
+                    <t t-if="data.status == 'No Ready To Produce'">
```
mash-odoo: How will this behave in translated environments?

bom_forecast/static/src/bom_overview_line.xml
```diff
@@ -0,0 +1,38 @@
...
+
+        <xpath expr="//td[t[@t-if=&quot;data.hasOwnProperty('availability_state')&quot;]]"
+            position="replace"/>
+
+        <xpath expr="//td//div[hasclass('o_field_badge')]" position="replace">
+            <div class="o_field_badge">
+
+                <t t-if="data.type == 'operation'">
+                    <span t-attf-class="badge rounded-pill o_mrp_overview_badge {{statusBackgroundColor}}"
+                        t-esc="data.availability_display"/>
```
mash-odoo: t-esc is deprecated, you can use t-out instead

mash-odoo (changes requested):
Hello!
I have done some initial review and maybe you need to work again on the approach and maybe have a deeper understanding of your task and the approach you are using.

You have pushed the entire diff of the dashboard again in this task which is not required. 
Only the changes you have done regarding this task should be added in the files changed.

bom_forecast/static/src/bom_overview_line.js
sopat-odoo: I removed the `bom_overview_line.js` patch completely and moved the badge display and color logic to the inherited XML template. 
patch modifies an existing JS component/class.

bom_forecast/static/src/mrp_bom_overview_table.xml
sopat-odoo: Done. I replaced the  `ancestor::div[2]` XPath with a meaningful XPath based on the div class and the contained `goToProduct` action, so it no longer depends on the parent div position.

bom_forecast/__manifest__.py
sopat-odoo: You are right, this is not required for this customization module.

bom_forecast/static/src/bom_overview_line.js
sopat-odoo: Right, I removed the JS patch completely. Replacing 
"Estimated"` with `"Expected" in JS is not translation-safe.
I moved the display preparation to the backend instead of using a JS patch. 

`_set_ready_to_produce_status()` sets the Ready To Produce status only for the main product and keeps it False when there is no producible quantity, so the XML does not need to compare translated strings.

`_set_forecast_status_display()` prepares a separate `forecast_status_display` value for the XML. It keeps the main product as Estimated and shows components as Expected, while using Odoo’s _() translation function for the labels.

bom_forecast/static/src/bom_overview_line.xml
sopat-odoo: Done. I removed the `No Ready To Produce` string comparison. The backend now sets `data.status` to False when there is no Ready To Produce quantity, and XML only checks `data.status`, so it works safely in translated environments.

PR #1270, opened 2026-05-15
19.0 second uom adpaw

adpaw-odoo:
Task ID- 4485602
Introduce a new feature in the Point of Sale that allows cashiers to input quantities using a secondary Unit of Measure (UoM). This is specifically designed for products sold in larger units (e.g., Dozens) where the cashier needs a quick shortcut to enter sub-units (e.g., individual Units) without manually calculating decimals.

Currently, if a product is configured in a large UoM like "Dozens or kgs" a cashier wanting to sell 6 individual pieces or half kg etc, must manually calculate and type "0.5". This is prone to human error and slows down the checkout process. This module introduces a dedicated conversion shortcut to automate that calculation while maintaining the primary UoM on the order line.

Technical Decisions:
1. Schema: Added pos_second_uom_id and a computed pos_uom_reference_id to product.template to anchor units to their Category Roots.
2. Loader Integration: Appended the new field to _load_pos_data_fields to push metadata to the POS Reactive Store.
3. Data Integrity: Used an @api.onchange guard with relative_uom_id to auto-clear fields if the UoM Categories no longer match.
4. OWL:  Patched ControlButtons.prototype with a Reactive Getter to show the button only when a secondary UoM exists.
5. Inherited from NumberPopup to create SecondUomPopup, preserving native keypad logic.
6. Conversion Logic: Applied the formula (enteredQty * secondUom.factor) / mainUom.factor within the JS action to sync the orderline.


mash-odoo (changes requested):
Hello,
This PR is currently having unnecessary diff of real_estate and dashboard which is not the purpose of this PR. Please clean this up.

second_uom/models/product_product.py
```diff
@@ -0,0 +1,11 @@
+from odoo import api, models
+
+
+class InheritedProductProduct(models.Model):
```
mash-odoo:
```suggestion
class ProductProduct(models.Model):
```

second_uom/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+{
+    'name': 'POS Second Unit of Measure',
+    'author': 'Aditi (adpaw)',
+    'category': 'Point of Sale',
+    'summary': 'Allow cashier to enter quantity using a second UOM in POS',
+    'depends': ['point_of_sale', 'product'],
```
mash-odoo: Do we need to add 'product' in depends?

second_uom/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    'license': 'LGPL-3',
+    'data': [
```
mash-odoo: I have a question here... how would a customer know that something like this exists?
Shouldn't it be installed when the pos is installed?

second_uom/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+{
+    'name': 'POS Second Unit of Measure',
```
mash-odoo: maybe you can also give a proper name for the module.
`second_uom` seems very vague.

second_uom/views/product_template_views.xml
```diff
@@ -0,0 +1,19 @@
...
+                <field
+                    name="pos_second_uom_id"
+                    string="POS second UOM"
+                    domain="[('id', '=', pos_uom_reference_id)]"
+                    options="{'no_create': True, 'no_open': True}"
+                />
+            </xpath>
+        </field>
+    </record>
+</odoo>
```
mash-odoo: Please add a line at the end of the file

second_uom/static/src/second_uom_popup/second_uom_popup.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates id="template" xml:space="preserve">
+    <t t-name="second_uom.SecondUomPopup" t-inherit="point_of_sale.NumberPopup" t-inherit-mode="primary">
+        <xpath expr="//div[hasclass('input-symbol')]" position="before">
+            <div class="text-center mb-2 fw-bold text-primary">
+                Unit: <t t-esc="props.uomName"/>
```
mash-odoo: Use `t-out` instead of `t-esc` as it is deprecated

pos_second_uom/static/src/second_uom_button/second_uom_button.js
```diff
@@ -0,0 +1,32 @@
...
+        const result = await makeAwaitable(this.dialog, SecondUomPopup, {
+            title: "Secondary Unit Entry",
+            uomName: secondUom.name,
+            startingValue: 0,
+        });
+
+        if (result !== null && result !== undefined) {
+            const enteredQty = parseFloat(result);
+            if (isNaN(enteredQty)) return;
+            const newMainQty = (enteredQty * secondUom.factor) / mainUom.factor;
```
mash-odoo, line 29: Has this formula been tested against all UoM configurations

pos_second_uom/static/src/second_uom_button/second_uom_button.js
```diff
@@ -0,0 +1,32 @@
+import { ControlButtons } from "@point_of_sale/app/screens/product_screen/control_buttons/control_buttons";
+import { SecondUomPopup } from "@second_uom/second_uom_popup/second_uom_popup";
+import { patch } from "@web/core/utils/patch";
+import { makeAwaitable } from "@point_of_sale/app/utils/make_awaitable_dialog";
+
+patch(ControlButtons.prototype, {
```
mash-odoo, line 7: Do we want to apply this patch ControlButtons globally?

pos_second_uom/static/src/second_uom_button/second_uom_button.js
```diff
@@ -0,0 +1,32 @@
...
+    displaySecondUomBtn() {
+        const selectedLine = this.currentOrder?.getSelectedOrderline();
```
mash-odoo, line 9: Does changing selected orderline automatically rerender ControlButtons?

pos_second_uom/models/product_template.py
```diff
@@ -0,0 +1,28 @@
+from odoo import api, fields, models
+
+
+class InheritedProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    pos_second_uom_id = fields.Many2one('uom.uom', string="POS Second UoM")
+    pos_uom_reference_id = fields.Many2one(
+        'uom.uom', string="UoM Root", compute='_compute_pos_uom_reference_id', store=True
+    )
```
mash-odoo, line 10: Why persist this?
It seems to exist only for domain filtering.
Could possibly be related/computed non-stored?
unless it needs searching. 
WDYT?

second_uom/static/src/second_uom_button/second_uom_button.js
```diff
@@ -0,0 +1,32 @@
...
+
+    async clickSecondUom() {
+        const selectedLine = this.currentOrder.getSelectedOrderline();
+        if (!selectedLine) return;
+
+        const secondUom = selectedLine.product_id.pos_second_uom_id;
+        const mainUom = selectedLine.product_id.uom_id;
+
+        const result = await makeAwaitable(this.dialog, SecondUomPopup, {
+            title: "Secondary Unit Entry",
```
mash-odoo: I guess this should be translated

pos_second_uom/static/src/second_uom_popup/second_uom_popup.js
```diff
@@ -0,0 +1,14 @@
...
+    static template = "second_uom.SecondUomPopup";
+    static props = {
+        ...NumberPopup.props,
+        uomName: { type: String },
+    };
+
+    confirm() {
+    this.props.getPayload(this.state.buffer);
+    this.props.close();
+    }
```
mash-odoo, line 13:
```suggestion
    confirm() {
        this.props.getPayload(this.state.buffer);
        this.props.close();
    }
```

second_uom/views/product_template_views.xml
```diff
@@ -0,0 +1,19 @@
...
+        <field name="inherit_id" ref="product.product_template_form_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='uom_id']" position="after">
+                <field name="pos_uom_reference_id" invisible="1"/>
+                <field
+                    name="pos_second_uom_id"
+                    string="POS second UOM"
+                    domain="[('id', '=', pos_uom_reference_id)]"
+                    options="{'no_create': True, 'no_open': True}"
+                />
```
mash-odoo: Why is it made invisible first and then visible?

second_uom/views/product_template_views.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <record id="product_template_form_view_inherit_pos_second_uom" model="ir.ui.view">
+        <field name="name">product.template.form.inherit.second.uom</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='uom_id']" position="after">
+                <field name="pos_uom_reference_id" invisible="1"/>
+                <field
```
mash-odoo: You might need to do something as the string isn't being shown on the form.
You can maybe do something like this instead.

Image

second_uom/views/product_template_views.xml
```diff
@@ -0,0 +1,19 @@
...
+                    name="pos_second_uom_id"
+                    string="POS second UOM"
+                    domain="[('id', '=', pos_uom_reference_id)]"
+                    options="{'no_create': True, 'no_open': True}"
```
mash-odoo: Do we need `no_open` true here? As for uom, we have the option to open it.

mash-odoo (changes requested):
Hello @adpaw-odoo,
Good work done on the task... 🐣 
Please checkout my comments!

second_uom/__manifest__.py
adpaw-odoo: yes, by adding auto install, this module will be installed with pos.

second_uom/__manifest__.py
adpaw-odoo: not really, I've added it explicitly even though pos loads products on its own.

pos_second_uom/static/src/second_uom_button/second_uom_button.js
adpaw-odoo: yes,it works with all configured uoms.

pos_second_uom/static/src/second_uom_button/second_uom_button.js
adpaw-odoo: using patch this second uom features is added on all the sessions like retail, restaurant, etc.

pos_second_uom/static/src/second_uom_button/second_uom_button.js
adpaw-odoo: yes, it does. Changing the selected line triggers an update on the parent screen. And as contrilbuttons is a child of that screen, it also re renders along with it. The displaysecondBtn() checks the newly clicked orderline and shows/hides the button depending on the whether the product has a second uom or not.

second_uom/static/src/second_uom_button/second_uom_button.js
adpaw-odoo: Okayy, I've imported _t and wrapped the title string in it so it can be translated properly.

second_uom/views/product_template_views.xml
adpaw-odoo: Line 9 is actually the helper field which provides the domain for pos_second_uom_id.
But, doing so is of no use i see, cuz it still works the same if the line is omitted.

second_uom/views/product_template_views.xml
adpaw-odoo: oh  okay, for the users to atually see the units configuration. 
i will remove it.

PR #1278, opened 2026-05-19
[ADD] estate: estate module added for real estate businesses

sreedevk:


estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
+from dateutil.relativedelta import relativedelta
+
+from odoo.tools.float_utils import float_compare, float_is_zero
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError, ValidationError
```
Megaaaaaa, line 4: Almost got it :smile:

```suggestion
from dateutil.relativedelta import relativedelta

from odoo import _, api, fields, models
from odoo.exceptions import UserError, ValidationError
from odoo.tools.float_utils import float_compare, float_is_zero
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+            ("offer_accepted", "Offer Accepted"),
+            ("sold", "Sold"),
+            ("cancelled", "Cancelled"),
+        ],
+    )
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        default=(fields.Date.today() + relativedelta(months=3)), copy=False
```
Megaaaaaa: Just making sure the history gets preserved when someone else want to add a line :+1:

```suggestion
        default=(fields.Date.today() + relativedelta(months=3)), copy=False,
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+    seller_id = fields.Many2one(
+        "res.users", string="Sales Person", default=lambda self: self.env.user
```
Megaaaaaa:
```suggestion
        "res.users", string="Sales Person", default=lambda self: self.env.user,
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+    buyer_id = fields.Many2one("res.partner", string="Buyer")
+
+    expected_price = fields.Float(required=True)
+    _check_expected_price = models.Constraint(
+        "CHECK(expected_price > 0)", "Expected price should be positive."
+    )
```
Megaaaaaa: Usually in a model definition we put first the "underscore fields" (_name, _desc, etc) then all the fields, then all the constrains and finally all the methods so this would go after all you fields declarations :+1:

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+    selling_price = fields.Float(readonly=True, copy=False)
+    _check_selling_price = models.Constraint(
+        "CHECK(selling_price > 0)", "Selling price must be positive."
+    )
```
Megaaaaaa: And same here :+1:

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+                raise UserError(_("Cancelled properties cannot be sold."))
+            else:
+                property.state = "sold"
+                return True
+
+    @api.constrains("selling_price")
+    def _check_selling_price(self):
+        for property in self:
+            sp = property.selling_price
+            ep = property.expected_price
```
Megaaaaaa: These are at the very limit of being okay in odoo's coding style :smile: We really don't like the one letter per word style appart from a few very obvious exceptions for recurring stuff (like in accounting, we very often use 'aml' for account_move_line because it's used everywhere, very long otherwise and everyone is used to it) and since you only use those once, i wouldn't declare a variable for them. If you really want to though, call them fully `selling_price` and `expected_price`. It helps in longer methods when you use a variable that's been declared many lines before to quickly know what it is without having to jump around in the code.


```suggestion
            selling_price = property.selling_price
            expected_price = property.expected_price
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,68 @@
...
+class EstatePropertyOffer(models.Model):
+    _name = "estate.property.offer"
+    _description = "Offer made on properties"
+    _order = "price desc"
+
+    property_type_id = fields.Many2one(related="property_id.property_type_id")
+    price = fields.Float()
+    _check_price = models.Constraint(
+        "CHECK(price > 0)", "Offer price must be positive."
+    )
```
Megaaaaaa: after all your fields 😉

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,68 @@
...
+            ("refused", "Refused"),
+        ],
+        help="Offer Status",
+    )
+
+    partner_id = fields.Many2one("res.partner", required=True)
+    property_id = fields.Many2one("estate.property", required=True)
+    validity = fields.Integer(default=7)
+    date_deadline = fields.Date(
+        compute="_compute_date_deadline", inverse="_compute_validity"
```
```suggestion
        compute="_compute_date_deadline", inverse="_compute_validity",
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,68 @@
...
+        for offer in self:
+            offer.date_deadline = (
+                offer.create_date or fields.Date.today()
+            ) + relativedelta(days=offer.validity)
+
+    def action_accept_offer(self):
+        for offer in self:
+            is_not_actionable = any(
+                pstat == "accepted"
+                for pstat in offer.property_id.offer_ids.mapped("status")
```
Megaaaaaa: Kinda want to be annoying here and say that pstat is not the best choice for a name but that's rather understandable thanks to the `for in`. I would still just call it `status`
```suggestion
                status == "accepted"
                for status in offer.property_id.offer_ids.mapped("status")
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,68 @@
...
+            else:
+                offer.status = "accepted"
+                offer.property_id.buyer_id = offer.partner_id
+                offer.property_id.selling_price = offer.price
+                offer.property_id.state = "sold"
+                return True
+
+    def action_reject_offer(self):
+        for offer in self:
+            offer.status = "refused"
```
Megaaaaaa: When you write the same value on all the records of a recordset, you can directly assign it on the recordset. It might depend on the version though but in 19.0, you can do `self.status = "refused"` and all the records in self will be updated :+1:

```suggestion
        self.status = "refused"
```

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,27 @@
+from odoo import api, fields, models
+
+
+class EstatePropertyType(models.Model):
+    _name = "estate.property.type"
+    _description = "type of the estate property"
+    _order = "name"
+
+    offer_ids = fields.One2many(
+        "estate.property.offer", "property_type_id", string="Offers"
```
```suggestion
        "estate.property.offer", "property_type_id", string="Offers",
```

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,27 @@
...
+    offer_count = fields.Integer(compute="_compute_offer_count")
+    active = fields.Boolean(default=True)
+    name = fields.Char(required=True)
+    sequence = fields.Integer(string="Sequence", default=1, help="Used for ordering")
+    _name_uniq = models.Constraint(
+        "unique (name)",
+        "Each property type name must be unique.",
+    )
+    property_ids = fields.One2many(
+        "estate.property", "property_type_id", string="Properties"
```
```suggestion
        "estate.property", "property_type_id", string="Properties",
```

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,27 @@
...
+    _order = "name"
+
+    offer_ids = fields.One2many(
+        "estate.property.offer", "property_type_id", string="Offers"
+    )
+    offer_count = fields.Integer(compute="_compute_offer_count")
+    active = fields.Boolean(default=True)
+    name = fields.Char(required=True)
+    sequence = fields.Integer(string="Sequence", default=1, help="Used for ordering")
+    _name_uniq = models.Constraint(
```
Megaaaaaa, line 24: after all the fields

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<odoo>
+    <menuitem id="estate.app_menu" name="Estate">
+        <menuitem id="advertisements_list_menu" name="Advertisements">
+            <menuitem 
+                id="estate.advertisements_properties_list_menu_item"
+                name="Properties"
+                action="estate.estate_property_list_form" />
```
Megaaaaaa: Cleaner without the space 😉
```suggestion
                action="estate.estate_property_list_form"/>
```
There are mny places where this can be applied 👍

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
+        string="State",
+        copy=False,
+        default="new",
+        required=True,
+        selection=[
+            ("new", "New"),
+            ("offer_recd", "Offer Received"),
```
Megaaaaaa: Make it the full name `offer_received`. Clearer for everyone
```suggestion
            ("offer_received", "Offer Received"),
```

estate/views/estate_property.xml
```diff
@@ -0,0 +1,115 @@
...
+                            <field name="property_type_id" string="Property Type" options="{'no_create': true}" />
+                            <field name="seller_id" string="Sales Person" />
+                            <field name="buyer_id" string="Buyer" />
+                        </group>
+                    </page>
+                    <page string="Offers">
+                        <group>
+                            <field 
+                                name="offer_ids"
+                                context="{'list_view_of': 'estate.estate_property_offer_list_form'}"
```
Megaaaaaa: typo ? 
```suggestion
                                context="{'list_view_ref': 'estate.estate_property_offer_list_form'}"
```

estate/views/estate_property_type.xml
```diff
@@ -0,0 +1,38 @@
...
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <form>
+                <header>
+                    <field name="name" />
+                </header>
+                <notebook>
+                    <page string="Properties">
+                        <group>
+                            <field name="property_ids" context="{'list_view_of': 'estate.estate_property_list_form'}" />
```
Megaaaaaa: again ?
```suggestion
                            <field name="property_ids" context="{'list_view_ref': 'estate.estate_property_list_form'}" />
```

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "Estate",
+    "depends": ["base"],
+    "author": "Sreedev Kodichath <srkod@odoo.com>",
+    "license": "LGPL-3",
+    "data": [
+        # datafiles
```
Megaaaaaa: Don't keep the comments :smile:
```suggestion
```

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+        "security/ir.model.access.csv",
+        # views
```
Megaaaaaa:
```suggestion
```

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "Estate",
+    "depends": ["base"],
+    "author": "Sreedev Kodichath <srkod@odoo.com>",
```
Megaaaaaa: As an Odoo employee, you cannot sign any work in your own name, the author of this module is "Odoo S.A." 😉

Megaaaaaa (changes requested):
Hello 👋

Here's a first (official) review for you, really nice PR for a first, gj :+1:

There's not much and even amongst those there quite a lot of things that repeat themselves so not a whole lot to say :smile:

Keep in mind that a review is not the utlimate solution, it's only someone seeing things he would have done differently and offering an alternative. You're always free to answer with an other alternative or even say you disagree and bring your arguments. For stuff that's not just styling or conventions, don't tak the wrong habit of just blindly applying comments, you spent the time developping the thing, everything you've tried doesn't show up in the diff, only the last one. You probably have very interesting things to say to the reviewer about the decisions you made.

What I would suggest you do and continue to do while you're not the most comfortable with odoo's structure is to not apply all the changes at once. For example, changing a technical value inside a Selection field has a lot of impact and if you change everything at once, you might end up with a lot of errors when running the database. Splitting the review in multiple steps and trying to run your db in between those can save you a lot of time in the long run.

Also a quick tip for the methodology when applying reviews changes: What I like to do it put a reaction like a thumbsup on a message once I have made the change locally. Then when everything is marked with something or answered, I push, go through my diff and mark as resolved the things that are now outdated so nothing is left forgotten.

estate/models/estate_property.py
Megaaaaaa: Here's [the official doc](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#imports) about the imports order which is very much easier to understand when you see it than when someone tries to explain :smile:

Megaaaaaa:
Also since you've already squashed your work into a single commit, you can spend the extra time to make it compliant to [the git guidelines](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html#commit-message-structure). We'll explain it next week anyway but you've done the hardest part, having a single commit :smile:
If you do so, don't forget to update the PR title and message to remain the same as your commit's because those won't update automatically.
(and if you're not sure about the task-123, here it would be task-6229508, 6229508 being the id of your Server 101 task you can find in its url: https://www.odoo.com/odoo/project/26755/tasks/6229508)

estate/models/estate_property_type.py
sreedevk: this particular constraint was already at the bottom of other fields, that's why it doesn't show up as "outdated", resolving the issue.

sreedevk:
@Megaaaaaa Thank you for the detailed review! 😄 

I finished the backend framework 101 chapter entirely, addressed all your comments and the CI has passed as well.

PR #1281, opened 2026-05-19
[ADD] estate: doing the tutorial

ROM4N0:


estate/models/est_property_offers.py
```diff
@@ -0,0 +1,51 @@
+from odoo import models, fields, api, exceptions
+import datetime as dt
```
aboo-odoo: It is preferable to only import the method you need when you can. Here you only use `timedelta`

 
```suggestion
from datetime import timedelta
```

estate/models/est_property_offers.py
```diff
@@ -0,0 +1,51 @@
+from odoo import models, fields, api, exceptions
```
aboo-odoo:
```suggestion
from odoo import models, fields, api
from odoo.exceptions import UserError, ValidationError
```

estate/models/est_property_offers.py
```diff
@@ -0,0 +1,51 @@
...
+import datetime as dt
+
+class test(models.Model):
```
aboo-odoo:
```suggestion
class EstatePropertyOffer(models.Model):
```

estate/models/est_property_offers.py
```diff
@@ -0,0 +1,51 @@
...
+    _name = "est_property_offers"
```
aboo-odoo: Singular over plural [When defining an Odoo Model : use singular form of the name](https://www.odoo.com/documentation/17.0/contributing/development/coding_guidelines.html#symbols-and-conventions)
```suggestion
    _name = "est_property_offer"
```

estate/models/est_property_offers.py
aboo-odoo: Imports should follow RUFF specification, i.e; the ASCII order and be grouped into 3 categories:
1. General python library
2. Imports of odoo
3. Imports from Odoo modules

estate/models/est_property_offers.py
aboo-odoo: There is a tacit convention that strings viewed by the users are surrounded by `"`, otherwise it is `'` (of course, each team dos a bit differently, they'll let you know)
```suggestion
    _name = 'est_property'
```

estate/models/est_property_offers.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+
+    name = fields.Char("name")
+    price = fields.Integer("price",required=True)
+    status = fields.Selection(string = "status", default = "Waiting",
+                              selection = [
+                                  ("Accepted","Accepted"),
+                                  ("Refused","Refused"),
+                                  ("Waiting","Waiting")])
```
aboo-odoo:
```suggestion
    status = fields.Selection(
        string='status',  # Single quote + no space
        default='Waiting',  # Single quote + no space
        selection = [
            ('accepted', "Accepted"),  # Imagine selection like a dict
            ('refused', "Refused"),    # You normally set the keys in 
            ('waiting', "Waiting"),    # lower letters
        ],
    )
```

estate/models/est_property_offers.py
```diff
@@ -0,0 +1,51 @@
...
+    partner_id = fields.Many2one("res.partner")
+    property_id = fields.Many2one("est_test",required=True)
+
+
+    validity_date = fields.Date(compute="_update_validity_date",inverse="_update_validity_duration")
```
aboo-odoo: compute methods should be prefixed by _compute and inverse methods should be prefixed by _inverse
```suggestion
    validity_date = fields.Date(compute="_compute_validity_date",inverse="_inverse_validity_duration")
```

estate/models/est_property_offers.py
```diff
@@ -0,0 +1,51 @@
...
+    validity_duration = fields.Integer(default=7)
+    
+    @api.depends("validity_duration")
+    def _update_validity_date(self):
+        for record in self:
```
aboo-odoo: `record` is not really explicit as variable name. It works fine for small method, but when you will have huge methods, it's better to use a more meaningful variable name
```suggestion
        for property in self:
```

estate/models/est_property_offers.py
```diff
@@ -0,0 +1,51 @@
...
+    _name = "est_property_offers"
+    _description = "Property Offers"
+    
+    _check_price = models.Constraint(
+        'CHECK (price >= 0)',
+        'The Price should be positive!',
+    )
+
+
+    name = fields.Char("name")
```
aboo-odoo, line 17: The string attribute defaults to the name of the field - suffix such as `_id.s`. Thus, when you want the string to be the same as the field name, you don't need to add it 
```suggestion
    description = fields.Text()
```

estate/models/est_property_offers.py
```diff
@@ -0,0 +1,51 @@
...
+    @api.depends("validity_duration")
+    def _update_validity_date(self):
+        for record in self:
+            record.validity_date = fields.Date.today() + dt.timedelta(days=record.validity_duration)
+    
+    def _update_validity_duration(self):
+        for record in self:
+            record.validity_duration = (record.validity_date - fields.Date.today()).days
+    
+    def action_confirm(self):
```
aboo-odoo, line 54: `self` is a recordset, i.e. a collection of records. Your method would failed if called on several records when trying to perform `self.state`.

Currently, through the UI, you cannot trigger the error, but from the command line or a server action, users could get the traceback.

side note: you can run odoo in terminal mode by adding `shell` after `odoo-bin`: `python odoo-bin shell --addons-path ...`. From there, you could perform 
```
estates = self.env['estate.property'].create([{'name': 'TEST1', ADD the other relevant key-value pairs}, {'name': 'TEST2', ADD the other relevant key-value pairs}])
estates.action_confirm()  # This should trigger the traceback
```

There are two ways to solve the issue:
- begin your method with `self.ensure_one()`, quite self explanatory 
- or use a for loop like in _compute methods, we did but only in the else condition, the traceback will be triggered when executing `self.status == "Refused"` - a read of a value on the recordset

aboo-odoo, line 54: Functionally speaking, could you confirm an offer if the property was cancelled ?

estate/models/est_property_offers.py
```diff
@@ -0,0 +1,51 @@
...
+        if self.property_id.state == "Sold":
+            raise exceptions.UserError("Property was already sold!")
```
aboo-odoo: Dome strings are not automatically translated, such as fields names, selection values, etc. However, this is not the case for exceptions. To specify you want a string to be translateable, you need to add the `_` method like so 
```suggestion
            raise UserError(self.env._("Property was already sold!"))
```

estate/models/est_property_offers.py
```diff
@@ -0,0 +1,51 @@
...
+        else:
+            for record in self:
+                record.status = "Accepted"
+                self.property_id.selling_price = self.price
+                self.property_id.partner_id = self.partner_id
+                self.property_id.state = "Sold"
+
+    def action_reject(self):
+        for record in self:
+            record.status = "Refused"
```
aboo-odoo: writes can be performed on a recordset 
```suggestion
        self.status = "Refused"
```

aboo-odoo: You need to end all new files with an empty line. This minimizes the diff next time someone adds lines to your file; easier to review.
```suggestion
        for record in self:
            record.status = "Refused"

```

estate/models/est_property_tags.py
```diff
@@ -0,0 +1,12 @@
+from odoo import models, fields
+
+class test(models.Model):
```
aboo-odoo: All you classes are named test, please name them accordingly to the object they refer to

estate/models/est_test.py
```diff
@@ -0,0 +1,112 @@
...
+    _check_selling_price = models.Constraint(
+        'CHECK (selling_price >= 0)',
+        'The Selling Price should be positive!',
+    )
+    
+    
+    name = fields.Char('Property Name', required=True)
+    description = fields.Text('Description', required=True)
+    post_code = fields.Char("Postal Code",required=True)
+    date_availability = fields.Date("Date Availability",default=(fields.Date.today() + dt.timedelta(days=90))) #add 90 days which is about 3 months
```
aboo-odoo: When declaring the default value like so, it will get computed once you start the database. The computed value will then be used until your database is restarted. This means that if you start your db today 26/05 and let it run for a month and that, on the 26/06 you create a new property record, the default `date_availibility` will be 26/08 (mai + 3 month) while we expect it to be 26/09 (june + 3 month). 

To make the default value recomputed at each call; you need to call a function

estate/models/est_test.py
```diff
@@ -0,0 +1,112 @@
...
+        string = "Garden Orientation",
+        selection=[("North","North"),("South","South"),("East","East"),("West","West")])
+    active = fields.Boolean("Active",default=True)
+    state = fields.Selection(
+        string = "State",
+        selection = [("New","New"),("Offer Received","Offer Received"),("Offer Accepted","Offer Accepted"),("Sold","Sold"),("Cancelled","Cancelled")]
+    )
+    partner_id = fields.Many2one("res.partner", string="Buyer")
+    user_id = fields.Many2one("res.users", string="Salesman")
+    property_type = fields.Many2one("est_property_types", string="Property Type")
```
aboo-odoo: Naming convention for related field is that they all end with the suffix `_id` or `_ids` depending on whether it refer to one or many items


```suggestion
    property_type_id = fields.Many2one("est_property_types", string="Property Type")
```

estate/models/est_test.py
```diff
@@ -0,0 +1,112 @@
+from odoo import models, fields, api, exceptions, tools
```
aboo-odoo:
```suggestion
from odoo import models, fields, api, exceptions
from odoo.tools import float_compare, float_is_zero
```

estate/models/est_test.py
```diff
@@ -0,0 +1,112 @@
...
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offers_ids")
+    def _compute_best_offer(self):
+        
+        for record in self:
+            if record.offers_ids:
+                record.max_offer = max(record.offers_ids.mapped("price"))
+            else:
+                record.max_offer = 0
```
aboo-odoo:
```suggestion
            record.max_offer = max(record.offers_ids.mapped("price"), default=0)
```

estate/models/est_test.py
```diff
@@ -0,0 +1,112 @@
...
+            else:
+                self.state = "Cancelled"
+
+class testModelLine(models.Model):
+    _name = "test_model_line"
+    _description="Test Model Line"
+    model_id = fields.Many2one("est_test")
+    name = fields.Char()
+    price = fields.Char()
+    status = fields.Char()
```
aboo-odoo: Don't forget to remove this 😄

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_test_model,access_test_model,model_est_test,base.group_user,1,1,1,1
+access_types,access_types,model_est_property_types,base.group_user,1,1,1,1
+access_tags,access_tags,model_est_property_tags,base.group_user,1,1,1,1
+access_offers,access_offers,model_est_property_offers,base.group_user,1,1,1,1
```
aboo-odoo: new line

estate/views/estate_property_offers_views.xml
```diff
@@ -0,0 +1,36 @@
...
+        <list string="Tests">
+            
+            <field name="name"/>
+            <field name="price"/>
+            <field name="status"/>
+            <field name="partner_id"/>
+
+
+
+
```
aboo-odoo: That's a lot of empty space 👀

estate/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
```
aboo-odoo: You may see this in some file because it was compulsary before, note that is not anymore and we are trying to slowly remove them all

```suggestion
```

estate/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+{
+    'name':"estate",
+    'author':"nutav",
```
aboo-odoo: Everything you create for Odoo belongs to "Odoo S.A." hehe

```suggestion
    "author": "Odoo S.A.",
```

aboo-odoo (commented):
Good work but still for more to do 🥳  I'm being extra-picky so that I can transmit as much information as possible. If there is anything you disagree with, please share it without fear. 

A generic comment:
```
PROBLEM - what is the problem 

task-XXXXXX
```
- can you squash all your commits: come to me and I'll show you how 😄

estate/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+from . import est_property
+from . import est_property_offers
+from . import est_property_tags
+from . import est_property_types
+from . import inherited_model
```
aboo-odoo, line 5: Better call the file the same you would if it wasn't an override 😄 It's more understandable than `inherited`
```suggestion
from . import res_users
```

estate/models/est_property.py
```diff
@@ -0,0 +1,110 @@
...
+            property.total_area = property.living_area + property.garden_area
+
+    @api.depends("offers_ids")
+    def _compute_best_offer(self):
+        
+        for property in self:
+            
+            property.max_offer = max(property.offers_ids.mapped("price"),default=0)
+            
+
```
aboo-odoo, line 75: That's a whole lot of spaces

estate/models/est_property.py
```diff
@@ -0,0 +1,110 @@
...
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = None
+
+    @api.ondelete(at_uninstall=False)
+    def _ondelete(self):
+        for property in self:
+            if not(property.state == "new" or property.state == "cancelled"):
+                raise UserError("You can only delete properties that are 'New' or 'Cancelled'.")
```
aboo-odoo, line 90: Don't forget to set the error messages as translateable

estate/models/est_property.py
```diff
@@ -0,0 +1,110 @@
...
+                self.state = "sold"
+
+    def cancel_action(self):
+        for property in self:
+            if self.state =="sold":
+                #Raise error
+                raise UserError("Can't change to Cancel if current state is Sold!")
+            else:
+                self.state = "cancelled"
+
```
aboo-odoo, line 110: End of file line breaks

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,27 @@
...
+    _inherit = "est.property"
+
+    def sell_action(self):
+
+        super().sell_action()
+        
+        self.env['account.move'].create({
+            'partner_id': self.partner_id.id,
+            'move_type': 'out_invoice',
+            'line_ids':[
```
aboo-odoo, line 14: Isn't that missing the price of the house ?

aboo-odoo (commented):
Few small extra remarks 😄

PR #1282, opened 2026-05-19
[ADD] estate, estate_account: create Real Estate and Real Estate Account modules for managing real estate business.

abdelgad:
PROBLEM
Odoo does not have a built-in module to handle real estate properties, track buyer offers, or manage the property sales pipeline.

GOAL
Create Real Estate module and Real Estate Account modules to manage properties from the moment they hit the market until the final sale and invoice.

SOLUTION
Added Real Estate and Real Estate Account modules

task-6229482

### DIFF
--- a/estate/models/__init__.py
+++ b/estate/models/__init__.py
@@ -1,0 +1,5 @@
+from . import estate_property
+from . import estate_property_type
+from . import estate_property_tag
+from . import estate_property_offer
+from . import res_users
--- a/estate/models/estate_property.py
+++ b/estate/models/estate_property.py
@@ -1,0 +1,147 @@
+from odoo import api, fields, models
+from odoo.exceptions import UserError
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "It's free real estate"
+
+    # Order attributes
+    _order = "id desc"
+    
+    name = fields.Char(required=True,)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        default=lambda self: fields.Date.add(fields.Date.today(), months=3),
+        copy=False,)
+    expected_price = fields.Float(required=True,)
+    selling_price = fields.Float(readonly=True, copy=False,)
+    bedrooms = fields.Integer(default=2,)
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        string='Garden Orientation',
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West'),],
+        )
+    state = fields.Selection(
+            string='State',
+            selection=[
+                ('new', 'New'),
+                ('offer_received', 'Offer Received'),
+                ('offer_accepted', 'Offer Accepted'),
+                ('sold', 'Sold'),
+                ('canceled', 'Cancelled'),
+            ],
+            default='new',
+            required=True,
+            copy=False,
+        )
+
+    # Reserved fields
+    active = fields.Boolean(default=True, string='Active',)
+
+    # Relations
+    property_type_id = fields.Many2one("estate.property.type",)
+    property_tag_ids = fields.Many2many("estate.property.tag",)
+    salesperson_id = fields.Many2one('res.users', string='Salesman', default=lambda self: self.env.user,)
+    buyer_id = fields.Many2one('res.partner', copy=False,)
+    offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers",)
+
+    # -------------------------------------------------------------------------
+    # COMPUTED FIELDS
+    # -------------------------------------------------------------------------
+    total_area = fields.Integer(
+        string="Total Area (sqm)", 
+        compute="_compute_total_area",
+    )
+
+    best_price = fields.Float(
+        string="Best Offer", 
+        compute="_compute_best_price",
+    )
+
+    # -------------------------------------------------------------------------
+    # SQL Constraints
+    # -------------------------------------------------------------------------
+    _check_expected_price = models.Constraint(
+        'CHECK(expected_price > 0)', 
+        'A property expected price must be strictly positive.',
+    )
+    
+    _check_selling_price = models.Constraint(
+        'CHECK(selling_price >= 0)', 
+        'A property selling price must be positive.',
+    )
+
+    # -------------------------------------------------------------------------
+    # COMPUTE METHODS
+    # -------------------------------------------------------------------------
+    @api.depends("living_area", "garden_area",)
+    def _compute_total_area(self,):
+        for property in self:
+            property.total_area = property.living_area + property.garden_area
+
+    @api.depends("offer_ids.price",)
+    def _compute_best_price(self,):
+        for property in self:
+            if property.offer_ids:
+                property.best_price = max(property.offer_ids.mapped("price"), default=0.0)
+            else:
+                property.best_price = 0.0
+
+    # -------------------------------------------------------------------------
+    # CONSTRAINTs METHODS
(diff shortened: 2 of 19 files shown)


estate/models/estate_property.py
```diff
@@ -0,0 +1,132 @@
+from odoo import api, fields, models
+from odoo.exceptions import UserError
+from odoo.exceptions import ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "It's free real estate"
```
aboo-odoo, line 7: 😆

Image

estate/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+from . import estate_property
+from . import estate_property_type
+from . import estate_property_tag
+from . import estate_property_offer
```
aboo-odoo: You need to end all new files with an empty line. This minimizes the diff next time someone adds lines to your file; easier to review.
```suggestion
from . import estate_property_offer

```

estate/models/estate_property.py
```diff
@@ -0,0 +1,132 @@
...
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(default=fields.Date.add(fields.Date.today(), months=3), copy=False)
```
aboo-odoo: When declaring the default value like so, it will get computed once you start the database. The computed value will then be used until your database is restarted. This means that if you start your db today 26/05 and let it run for a month and that, on the 26/06 you create a new property record, the default `date_availibility` will be 26/08 (mai + 3 month) while we expect it to be 26/09 (june + 3 month). 

To make the default value recomputed at each call; you need to call a function, i.e. `lambda`

estate/models/estate_property.py
```diff
@@ -0,0 +1,132 @@
...
+            selection=[
+                ('new', 'New'),
+                ('offer_received', 'Offer Received'),
+                ('offer_accepted', 'Offer Accepted'),
+                ('sold', 'Sold'),
+                ('canceled', 'Cancelled'),
+            ],
+            default='new',
+            required=True,
+            copy=False
```
aboo-odoo: Adding a comma at the end allows for smother future PR when someone need to add some parameter below
```suggestion
            copy=False,
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,132 @@
...
+        )
+
+    # Relations
+    property_type_id = fields.Many2one("estate.property.type")
```
aboo-odoo: Nice use of `_id`/`_ids` 👍

estate/models/estate_property.py
```diff
@@ -0,0 +1,132 @@
...
+    def _check_selling_price(self):
+        for record in self:
+            # Selling price = 0 => Do nothing
+            if float_is_zero(record.selling_price, precision_digits=2):
+                continue
+
+            minimum_price = record.expected_price * 0.90
+
+            if float_compare(record.selling_price, minimum_price, precision_digits=2) < 0:
+                raise ValidationError("The selling price cannot be lower than 90% of the expected price.")
```
aboo-odoo: Dome strings are not automatically translated, such as fields names, selection values, etc. However, this is not the case for exceptions. To specify you want a string to be translateable, you need to add the `_` method like so 
```suggestion
            raise UserError(self.env._("The selling price cannot be lower than 90% of the expected price."))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,132 @@
...
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for record in self:
+            if record.offer_ids:
+                record.best_price = max(record.offer_ids.mapped("price"))
+            else:
+                record.best_price = 0.0
```
aboo-odoo: Alternatively:
```suggestion
            record.best_price = max(record.offer_ids.mapped("price"), default=0.0)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,132 @@
...
+
+            if float_compare(record.selling_price, minimum_price, precision_digits=2) < 0:
+                raise ValidationError("The selling price cannot be lower than 90% of the expected price.")
+
+    # -------------------------------------------------------------------------
+    # COMPUTE METHODS
+    # -------------------------------------------------------------------------
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
```
aboo-odoo: `record` is not really explicit as variable name. It works fine for small method, but when you will have huge methods, it's better to use a more meaningful variable name
```suggestion
        for property in self:
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,61 @@
...
+    # -------------------------------------------------------------------------
+    _check_price = models.Constraint(
+        'CHECK(price > 0)', 
+        'An offer price must be strictly positive.'
+    )
+
+    @api.depends("create_date", "validity")
+    def _compute_date_deadline(self):
+        for offer in self:
+            base_date = offer.create_date.date() if offer.create_date else fields.Date.today()
```
aboo-odoo: Python trick 😄 
```suggestion
            base_date = (offer.create_date or fields.Date.today()).date()
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,61 @@
...
+            if offer.date_deadline:
+                offer.validity = (offer.date_deadline - base_date).days
+
+    # -------------------------------------------------------------------------
+    # ACTIONS
+    # -------------------------------------------------------------------------
+    def action_accept(self):
+        for record in self:
+            if "accepted" in record.property_id.offer_ids.mapped("status"):
```
aboo-odoo: Alternatively, you can use a filtered here:

```suggestion
            if record.property_id.offer_ids.filtered(lambda offer: offer.status == "accepted":
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,61 @@
...
+                raise UserError("An offer has already been accepted for this property.")
+            
+            record.status = 'accepted'
+            record.property_id.selling_price = record.price
+            record.property_id.buyer_id = record.partner_id
```
aboo-odoo: Shoudln't you refuse all the other offers ? + Should be able to accept/refuse an offer if the property is sold/cancelled ?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,61 @@
...
+        return True
+
+    def action_refuse(self):
+        for record in self:
+            record.status = 'refused'
+        return True
```
aboo-odoo: unlike read operations, write operations can be performed on a recordset
```suggestion
        self.status = 'refused'

```

aboo-odoo (commented):
Phenomenal work 🥳  I'm being extra-picky so that I can transmit as much information as possible. If there is anything you disagree with, please share it without fear. 

A generic comment:
```
PROBLEM - what is the problem 

task-XXXXXX
```
- can you squash all your commits: come to me and I'll show you how 😄

estate/models/estate_property_offer.py
abdelgad: Good catch :+1:  but `fields.Date.today()` returns a `date` object which doesn't have a `.date()` method. Luckily, `fields.Datetime` exists, so the correct line would be: `base_date = (offer.create_date or fields.Datetime.now()).date()`

aboo-odoo (approved):
Well done, complete and clean PR 🥳

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def action_sold(self):
+        res = super().action_sold()
+ 
+        for property in self:
+            self.env["account.move"].create(
+                {
+                    "partner_id": property.buyer_id.id,
+                    "move_type": "out_invoice",
+                    "invoice_line_ids": [
```
aboo-odoo, line 16: One small thing: isn't this missing the price of the house ?

PR #1284, opened 2026-05-19
[ADD] estate, estate_account: modules for managing real estate businesses

sidum-odoo:
Problem:
No existing module for managing real estate businesses.

Solution:
2 new modules : "estate" and "estate_account".
The main module "estate" allows property management with Kanban, List and Form views. Types, tags and offers can be added/assigned to a property. The flow is managed through automatic state changes when offers are received/accepted.
Additionnaly, the module "estate_account" allows automatic invoice creation when a property is sold.

Task-6229416

### DIFF
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -1,0 +1,16 @@
+{
+    "name": "Real Estate",
+    "application": True,
+    "depends": ["base"],
+    "author": "Odoo S.A.",
+    "license": "LGPL-3",
+    "data": [
+        "security/ir.model.access.csv",
+        "views/estate_property_views.xml",
+        "views/estate_property_type_views.xml",
+        "views/estate_property_tag_views.xml",
+        "views/estate_property_offer_views.xml",
+        "views/estate_menus.xml",
+        "views/res_users_views.xml",
+    ],
+}
--- a/estate/models/__init__.py
+++ b/estate/models/__init__.py
@@ -1,0 +1,5 @@
+from . import estate_property
+from . import estate_property_offer
+from . import estate_property_tag
+from . import estate_property_type
+from . import res_users
--- a/estate/models/estate_property.py
+++ b/estate/models/estate_property.py
@@ -1,0 +1,132 @@
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models, _
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate Property"
+    _order = "id desc"
+
+    name = fields.Char(string="Title", required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        string="Available From",
+        copy=False,
+        default=fields.Date.today() + relativedelta(days=3),
+    )
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float(readonly=True, copy=False)
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer(string="Living Area (sqm)")
+    facades = fields.Integer()
+    garages = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer(string="Garden Area (sqm)")
+    garden_orientation = fields.Selection(
+        selection=[
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
+        ],
+    )
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
+        readonly=False,
+        selection=[
+            ("new", "New"),
+            ("offer_received", "Offer Received"),
+            ("offer_accepted", "Offer Accepted"),
+            ("sold", "Sold"),
+            ("canceled", "Canceled"),
+        ],
+        string="Status",
+        default="new",
+        required=True,
+    )
+    type_id = fields.Many2one("estate.property.type", string="Property Type")
+    buyer_id = fields.Many2one("res.partner", string="Buyer", copy=False, readonly=True)
+    salesperson_id = fields.Many2one(
+        "res.users",
+        string="Salesman",
+        default=lambda self: self.env.user,
+    )
+    tag_ids = fields.Many2many("estate.property.tag", string="Tags")
+    offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers")
+    total_area = fields.Integer(compute="_compute_total_area")
+    best_price = fields.Float(compute="_compute_best_price")
+
+    _check_expected_price = models.Constraint(
+        "CHECK(expected_price > 0)",
+        "A property expected price should be strictly positive.",
+    )
+
+    _check_selling_price = models.Constraint(
+        "CHECK(selling_price >= 0)", "A property selling price should be positive."
+    )
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for record in self:
+            record.best_price = (
+                max(record.offer_ids.mapped("price")) if record.offer_ids else 0
+            )
+
+    @api.constrains("selling_price", "expected_price")
+    def _check_selling_price(self):
+        for record in self:
+            if not float_is_zero(record.selling_price, precision_digits=2):
+                if (
+                    float_compare(
+                        record.selling_price,
+                        0.9 * record.expected_price,
+                        precision_digits=2,
+                    )
+                    < 0
+                ):
+                    raise ValidationError(
+                        _(
+                            "The selling price must be at least 90% of the expected price"
+                        )
+                    )
+
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = None
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_except_new_or_canceled(self):
+        if any(record.state == "new" or record.state == "canceled" for record in self):
+            raise UserError(_("Can't delete a new or canceled property."))
+
+    def action_set_sold(self):
+        for record in self:
+            if record.state == "canceled":
+                raise UserError(_("Canceled properties cannont be sold."))
+            if not any(offer.status == "accepted" for offer in record.offer_ids):
+                raise UserError(
+                    _("You can't sell a property with no accepted offers on it")
+                )
+            record.state = "sold"
+        return True
+
+    def action_set_canceled(self):
+        for record in self:
+            if record.state == "sold":
+                raise UserError(_("Sold properties cannont be canceled."))
+            record.state = "canceled"
+        return True
--- a/estate/models/estate_property_offer.py
+++ b/estate/models/estate_property_offer.py
@@ -1,0 +1,87 @@
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models, _
+from odoo.exceptions import UserError
+
+
+class EstatePropertyOffer(models.Model):
+    _name = "estate.property.offer"
+    _description = "Estate Property Offer"
+    _order = "price desc"
+
+    price = fields.Float()
+    status = fields.Selection(
+        selection=[("accepted", "Accepted"), ("refused", "Refused")],
+        copy=False,
+    )
+    partner_id = fields.Many2one("res.partner", required=True)
+    property_id = fields.Many2one("estate.property", required=True)
+    validity = fields.Integer(default=7, string="Validity (days)")
+    date_deadline = fields.Date(
+        compute="_compute_date_deadline",
+        inverse="_inverse_date_deadline",
+        string="Deadline",
+    )
+    property_type_id = fields.Many2one(
+        string="Property Type",
+        related="property_id.type_id",
+        store=True,
+    )
+
+    _check_price = models.Constraint(
+        "CHECK(price > 0)", "An offer price should be strictly positive."
+    )
+
+    @api.depends("create_date", "validity")
+    def _compute_date_deadline(self):
+        for record in self:
+            record.date_deadline = (
+                fields.Date.to_date(record.create_date) or fields.Date.today()
+            ) + relativedelta(days=record.validity)
+
+    def _inverse_date_deadline(self):
+        for record in self:
+            delta = record.date_deadline - (
+                fields.Date.to_date(record.create_date) or fields.Date.today()
+            )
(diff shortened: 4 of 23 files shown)


sidum-odoo:
@Megaaaaaa ready for review

estate/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+from . import (
+    estate_property,
+    estate_property_offer,
+    estate_property_tag,
+    estate_property_type,
+    res_users,
+)
```
Megaaaaaa: I agree it looks better but we write the `from . import x` on each line except maybe for test files :smile:
```suggestion
from . import estate_property
from . import estate_property_offer
from . import estate_property_tag
from . import estate_property_type
from . import res_users
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,123 @@
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
```
Megaaaaaa, line 5: Nice ordering

estate/models/estate_property.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate Property"
+    _order = "id desc"
```
Megaaaaaa, line 11: Nice naming

estate/models/estate_property.py
```diff
@@ -0,0 +1,123 @@
...
+
+    name = fields.Char(string="Title", required=True)
```
Megaaaaaa, line 13: And nice indentation, spacing, decision making on when to oneline or multiline a field. I already sense this review is going to go very well :+1:

estate/models/estate_property.py
```diff
@@ -0,0 +1,123 @@
...
+    garages = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer(string="Garden Area (sqm)")
+    garden_orientation = fields.Selection(
+        selection=[
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
+        ]
```
Megaaaaaa: The trailing coma so that people can add attributes to the field without modifying your lines :+1:
```suggestion
        ],
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,123 @@
...
+            ("canceled", "Canceled"),
+        ],
+        string="Status",
+        default="new",
+        required=True,
+    )
+    type_id = fields.Many2one("estate.property.type", string="Property Type")
+    buyer_id = fields.Many2one("res.partner", string="Buyer", copy=False, readonly=True)
+    salesperson_id = fields.Many2one(
+        "res.users", string="Salesman", default=lambda self: self.env.user
```
Megaaaaaa:
```suggestion
        "res.users", string="Salesman", default=lambda self: self.env.user,
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,123 @@
...
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        best = max(self.offer_ids.mapped("price")) if self.offer_ids else 0
+        for record in self:
+            record.best_price = best
```
Megaaaaaa: Becareful here that it's going to behave weirdly when self containts multiple records as you are setting on each property the best price to the value of the best offer amongst ALL the offers of ALL the properties.

```suggestion
        for record in self:
            record.best_price = max(record.offer_ids.mapped("price")) if record.offer_ids else 0
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,123 @@
...
+        for record in self:
+            if not float_is_zero(record.selling_price, precision_digits=2):
+                if (
+                    float_compare(
+                        record.selling_price,
+                        0.9 * record.expected_price,
+                        precision_digits=2,
+                    )
+                    == -1
+                ):
```
Megaaaaaa, line 95: Just a tiny bit more readable imo. Up to you :+1:

```suggestion
                if float_compare(
                    record.selling_price,
                    0.9 * record.expected_price,
                    precision_digits=2,
                ) < 0:
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,123 @@
...
+                    raise ValidationError(
+                        "The selling price must be at least 90% of the expected price"
+                    )
```
Megaaaaaa, line 100: Any string that will be shown to the user needs to be translatable as we have clients all around the world. For that we use the `odoo._` method which marks a string as requiring translations for our translators.
```suggestion
                    # Needs `from odoo import _`
                    raise ValidationError(_("The selling price must be at least 90% of the expected price"))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,123 @@
...
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = None
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_except_new_or_canceled(self):
+        if any(record.state == "new" or record.state == "canceled" for record in self):
+            raise UserError("Can't delete a new or canceled property.")
```
Megaaaaaa: UserError -> translations

estate/models/estate_property.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def action_set_sold(self):
+        for record in self:
+            if record.state == "canceled":
+                raise UserError("Canceled properties cannont be sold.")
```
Megaaaaaa: Translation

estate/models/estate_property.py
```diff
@@ -0,0 +1,123 @@
...
+            record.state = "sold"
+        return True
+
+    def action_set_canceled(self):
+        for record in self:
+            if record.state == "sold":
+                raise UserError("Sold properties cannont be canceled.")
```
Megaaaaaa: Translation

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class EstatePropertyOffer(models.Model):
+    _name = "estate.property.offer"
+    _description = "Estate Property Offer"
+    _order = "price desc"
+
+    price = fields.Float()
+    status = fields.Selection(
+        selection=[("accepted", "Accepted"), ("refused", "Refused")], copy=False
```
```suggestion
        selection=[("accepted", "Accepted"), ("refused", "Refused")], copy=False,
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,82 @@
...
+    partner_id = fields.Many2one("res.partner", required=True)
+    property_id = fields.Many2one("estate.property", required=True)
+    validity = fields.Integer(default=7, string="Validity (days)")
+    date_deadline = fields.Date(
+        compute="_compute_date_deadline",
+        inverse="_inverse_date_deadline",
+        string="Deadline",
+    )
+    property_type_id = fields.Many2one(
+        string="Property Type", related="property_id.type_id", store=True
```
```suggestion
        string="Property Type", related="property_id.type_id", store=True,
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,82 @@
...
+
+    @api.depends("create_date", "validity")
+    def _compute_date_deadline(self):
+        for record in self:
+            base_date = (
+                fields.Date.to_date(record.create_date)
+                if record.create_date
+                else fields.Date.today()
+            )
+            record.date_deadline = base_date + relativedelta(days=record.validity)
```
Megaaaaaa: Why not just:
```suggestion
            record.date_deadline = (record.create_date or fields.Date.today()) + relativedelta(days=record.validity)
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def _inverse_date_deadline(self):
+        for record in self:
+            base_date = (
+                fields.Date.to_date(record.create_date)
+                if record.create_date
+                else fields.Date.today()
+            )
+            delta = record.date_deadline - base_date
```
Megaaaaaa: Same here just go for `create_date or today`, makes it much cleaner :+1:

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,82 @@
...
+    def create(self, vals_list):
+        for vals in vals_list:
+            prop = self.env["estate.property"].browse(vals["property_id"])
+            best_price = prop.best_price
+            if vals["price"] < best_price:
+                raise UserError(
+                    "You can't create an offer with a lower amount than an existing offer."
+                )
+            prop.state = "offer_received"
+        return super().create(vals_list)
```
Megaaaaaa, line 67: Ok so multiple important things here:

The first one that got me started is that you do a `.browse()` inside a `for` loop. Methods that read the database like `.search()`, `.browse()`, etc have a very high impact on performance. Not because of the fetching itself but more because of the entire process around it so if you can do your fetch in a single request BEFORE looping, you gain a ton of performances.

Then, you have a UserError so you need to mark it as translatable.

And finally, don't be shy with variable names. Use full words so that it's very clear what everything is even a thousand lines after a declaration. No need to scroll.
 
```suggestion
    def create(self, vals_list):
        properties = self.env["estate.property"].browse([vals["property_id"] for vals in vals_list]) # Might need a `if vals.get("property_id")
        properties_by_id = {prop.id: prop for prop in properties}
        for vals in vals_list:
            property = properties_by_id.get(vals["property_id"])
            best_price = property.best_price
            if vals["price"] < best_price:
                raise UserError(_("You can't create an offer with a lower amount than an existing offer."))
            property.state = "offer_received"
        return super().create(vals_list)
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def action_set_accepted(self):
+        states = self.property_id.offer_ids.mapped("status")
+        for record in self:
+            if "accepted" in states:
+                raise UserError("Une offre a déjà été acceptée.")
```
Megaaaaaa: UserError so must be translatable AND you cannot write in French. Every single word you write in the code must be in english (if odoo controlled, some api's we call have their own naming convention, don't alias those :smile:)
```suggestion
                raise UserError(_("An offer has already been accepted."))
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,82 @@
...
+            record.property_id.state = "offer_accepted"
+        return True
+
+    def action_set_refused(self):
+        for record in self:
+            previous_status = record.status
+            record.status = "refused"
+            if previous_status == "accepted":
+                record.property_id.buyer_id = None
+                record.property_id.selling_price = 0
```
Megaaaaaa, line 86: Even if the previous state was not accepted, now that it's refused, you want no `buyer_id` and no `selling_price`, no ?
```suggestion
            record.property_id.buyer_id = None
            record.property_id.selling_price = 0
```

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,25 @@
...
+class EstatePropertyType(models.Model):
+    _name = "estate.property.type"
+    _description = "Estate Property Type"
+    _order = "sequence,name"
+
+    name = fields.Char(required=True)
+    property_ids = fields.One2many("estate.property", "type_id", string="Properties")
+    sequence = fields.Integer()
+    offer_ids = fields.One2many(
+        "estate.property.offer", "property_type_id", string="Offers"
```
```suggestion
        "estate.property.offer", "property_type_id", string="Offers",
```

estate/models/res_users.py
```diff
@@ -0,0 +1,12 @@
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    property_ids = fields.One2many(
+        "estate.property",
+        "salesperson_id",
+        domain="['|',('state', '=', 'new'),('state', '=', 'offer_received')]",
```
Megaaaaaa: No space around operators but definitely a space after commas :smile:
```suggestion
        domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]",
```

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,52 @@
...
+                    <div class="oe_title">
+                        <h1><field name="name"/></h1>
+                    </div>
+                    <notebook>
+                        <page string="Properties">
+                            <field name="property_ids">
+                                <list>
+                                    <field name="name"/>
+                                    <field name="expected_price"/>
+                                    <field name ="state"/>
```
Megaaaaaa: You did that to check if I was paying attention right ? Well I AM ! :smile:
```suggestion
                                    <field name="state"/>
```

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "Real Estate",
+    "application": True,
+    "depends": ["base"],
+    "author": "Simon Dumoulin",
```
Megaaaaaa: As an Odoo employee, you can't sign your work in your own name. The author of this module is "Odoo S.A." :smile:

.gitignore
```diff
@@ -6,6 +6,10 @@ __pycache__/
 # C extensions
 *.so
 
+# IDE
+.vscode/
+ruff.toml
+
 # Distribution / packaging
```
Megaaaaaa: Shouldn't be part of your diff :smile:

estate_account/__manifest__.py
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "Property Invoicing",
+    "depends": ["estate", "account"],
+    "author": "Simon Dumoulin",
```
Megaaaaaa: Odoo S.A.

Megaaaaaa (changes requested):
Hello 👋

I must say I'm impressed, this is a really really good PR 👍 

There is very few things to say and very few reccurring comments. Only the trailing commas and the translations. Otherwise it's rather comment about technical stuff than about guielines. Tat's the kind of thing you want to have on your PRs later. That's nice :+1:

So, please keep in mind that a review is not the utlimate solution, it's only someone seeing things he would have done differently and offering an alternative. You're always free to answer with an other alternative or even say you disagree and bring your arguments. For stuff that's not just styling or conventions, don't tak the wrong habit of just blindly applying comments, you spent time developping the thing, everything you've tried doesn't show up in the diff, only the final one. You probably have very interesting things to say to the reviewer about the decisions you made.

What I would suggest you do and continue to do while you're not the most comfortable with odoo's structure is to not apply all the changes at once. For example, changing multiple compute methods at once might lead you to a lot of errors if you don't check that the new implementation doesn't break anything before working on another one. Splitting the review in multiple steps and trying to run your db in between those can save you a lot of time in the long run.

Also a quick tip for the methodology when applying reviews changes: What I like to do it put a reaction like a thumbsup on a message once I have made the change locally. Then when everything is marked with something, I push, go through my diff and mark as resolved the things that are now outdated so nothing is left forgotten.

PR #1285, opened 2026-05-19
[ADD] estate,estate_account: technical onboarding commit 

jupir-odoo:
[ADD] estate,estate_accountant: technical onboarding commit
    
Problem:
There is currently no modules to manage real estate businesses in odoo.

Solution:
Implement estate properties, offer, tags, and types models with the associated views.
    
task-6229399

### DIFF
--- a/awesome_owl/static/src/playground.js
+++ b/awesome_owl/static/src/playground.js
@@ -1,5 +1,13 @@
-import { Component } from "@odoo/owl";
+import { Component, useState } from "@odoo/owl";
 
 export class Playground extends Component {
     static template = "awesome_owl.playground";
+
+    setup() {
+        this.state = useState({ value: 0 })
+    }
+
+    increment() {
+        this.state.value++
+    }
 }
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -1,0 +1,18 @@
+{
+    "name": "Estate",
+    "summary": "Track your real estate business",
+    "depends": ["base"],
+    "data": [
+        "security/ir.model.access.csv",
+        "views/estate_property_views.xml",
+        "views/estate_property_offer_views.xml",
+        "views/estate_property_tag_views.xml",
+        "views/estate_property_type_views.xml",
+        "views/res_users_views.xml",
+        "views/estate_property_menus.xml",
+    ],
+    "installable": True,
+    "application": True,
+    "author": "Odoo S.A.",
+    "license": "LGPL-3",
+}
--- a/estate/models/estate_property.py
+++ b/estate/models/estate_property.py
@@ -1,0 +1,131 @@
+from datetime import datetime
+
+from dateutil.relativedelta import relativedelta
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+
+class EstatePropery(models.Model):
+    _name = 'estate.property'
+    _description = "Estate property"
+    _order = 'id desc'
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        copy=False,
+        default=lambda self: self.__default_date_availability(),
+    )
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float(readonly=True, copy=False)
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        string="Orientation",
+        selection=[
+            ('north', "North"),
+            ('east', "East"),
+            ('soutch', "South"),
+            ('west', "West"),
+        ],
+    )
+    total_area = fields.Integer(compute='_compute_total_area')
+    best_price = fields.Float(compute='_compute_best_price')
+    property_type_id = fields.Many2one('estate.property.type')
+    tag_ids = fields.Many2many('estate.property.tag')
+    salesperson_id = fields.Many2one('res.users', default=lambda self: self.env.user)
+    buyer_id = fields.Many2one('res.partner', copy=False)
+    offer_ids = fields.One2many('estate.property.offer', 'property_id')
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
+        selection=[
+            ('new', "New"),
+            ('offer_received', "Offer Received"),
+            ('offer_accepted', "Offer Accepted"),
+            ('sold', "Sold"),
+            ('cancelled', "Cancelled"),
+        ],
+        copy=False,
+        default='new',
+    )
+
+    _expected_price_constraint = models.Constraint(
+        'CHECK(expected_price >= 0)',
+        "Expected price must be positive.",
+    )
+    _selling_price_constraint = models.Constraint(
+        'CHECK(selling_price >= 0)',
+        "Selling price must be positive.",
+    )
+
+    @api.depends('living_area', 'garden_area')
+    def _compute_total_area(self):
+        for property in self:
+            property.total_area = property.living_area + property.garden_area
+
+    @api.depends('offer_ids.price')
+    def _compute_best_price(self):
+        for property in self:
+            property.best_price = max(property.offer_ids.mapped('price'), default=0)
+
+    @api.constrains('selling_price', 'expected_price')
+    def _check_selling_price_and_expected_price(self):
+        for property in self:
+            if (
+                not float_is_zero(property.selling_price, 0)
+                and float_compare(
+                    property.expected_price * 0.9,
+                    property.selling_price,
+                    0,
+                )
+                > 0
+            ):
+                raise ValidationError(
+                    _(
+                        "You cannot accept an offer lower than 90% of the expected price. Lower the expected price if you want to accept it.",
+                    ),
+                )
+
+    @api.onchange('garden')
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = 'north'
+        else:
+            self.garden_area = None
+            self.garden_orientation = None
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_except_valid_state(self):
+        if any(property.state not in ['new', 'cancelled'] for property in self):
+            raise UserError(
+                _("Can't delete a property if status is not New or Cancelled"),
+            )
+
+    def action_cancel_property(self):
+        for property in self:
+            if property.state == 'sold':
+                raise UserError(_("Sold properties cannot be cancelled."))
+            property.state = 'cancelled'
+
+    def action_mark_sold_property(self):
+        for property in self:
+            if property.state == 'cancelled':
+                raise UserError(_("Cancelled properties cannot be sold."))
+
+            if not any(offer.status == 'accepted' for offer in property.offer_ids):
+                raise UserError(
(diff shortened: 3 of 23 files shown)


jupir-odoo:
@aboo-odoo kind reminder :D

estate/models/estate_property.py
aboo-odoo, line 1: In a Model attribute order should be:

3. Field declarations
9. Action methods

estate/models/estate_property.py
```diff
@@ -0,0 +1,111 @@
...
+    _order = "id desc"
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        copy=False, default=datetime.today() + relativedelta(months=3)
+    )
+    expected_price = fields.Float(required=True)
+    _expected_price_constraint = models.Constraint(
```
aboo-odoo: The two constraints should be just after the `_order` attribute 😄

estate/models/estate_property.py
```diff
@@ -0,0 +1,111 @@
+from dateutil.relativedelta import relativedelta
+from datetime import datetime
+
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+
+class EstatePropery(models.Model):
+    _name = "estate.property"
```
aboo-odoo: There is a tacit convention that strings viewed by the users are surrounded by `"`, otherwise it is `'`
```suggestion
    _name = 'estate.property'
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        string="Orientation",
+        selection=[("N", "North"), ("E", "East"), ("S", "South"), ("W", "West")],
```
aboo-odoo: Consider selection and a kind of dictionary, for dictionaries, you would write the keys in lower letters
```suggestion
        selection=[('north', "North"), ('east', "East"), ('south', "South"), ('west', "West")],
```
Also, don't hesitate to be exhaustive in your name; `estate.garden_orientation = 'north'` is a bit clearer than `estate.garden_orientation = 'n'`

estate/models/estate_property.py
```diff
@@ -0,0 +1,111 @@
...
+            ("offer_received", "Offer Received"),
+            ("offer_accepted", "Offer Accepted"),
+            ("sold", "Sold"),
+            ("cancelled", "Cancelled"),
+        ],
+        copy=False,
+        default="new",
+    )
+
+    def action_cancel_property(self):
```
aboo-odoo: `self` is a recordset, i.e. a collection of records. Your method would failed if called on several records when trying to perform `self.state`.

Currently, through the UI, you cannot trigger the error, but from the command line or a server action, users could get the traceback.

side note: you can run odoo in terminal mode by adding `shell` after `odoo-bin`: `python odoo-bin shell --addons-path ...`. From there, you could perform 
```
estates = self.env['estate.property'].create([{'name': 'TEST1', ADD the other relevant key-value pairs}, {'name': 'TEST2', ADD the other relevant key-value pairs}])
estates.action_cancel_property()  # This should trigger the traceback
```

There are two ways to solve the issue:
- begin your method with `self.ensure_one()`, quite self explanatory 
- or use a for loop like in _compute methods

estate/models/estate_property.py
```diff
@@ -0,0 +1,111 @@
...
+        if self.state == "sold":
+            raise UserError("Sold properties cannot be cancelled")
```
aboo-odoo: Dome strings are not automatically translated, such as fields names, selection values, etc. However, this is not the case for exceptions. To specify you want a string to be translateable, you need to add the `_` method like so 
```suggestion
            raise UserError(self.env._("Sold properties cannot be cancelled"))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,111 @@
...
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for property in self:
+            property.total_area = property.living_area + property.garden_area
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for property in self:
+            property.best_price = (
+                max(property.offer_ids.mapped("price")) if property.offer_ids else 0
```
aboo-odoo: This is fine but you could also:

```suggestion
                max(property.offer_ids.mapped("price"), default=0)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,111 @@
...
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = "N"
+        else:
+            self.garden_area = None
+            self.garden_orientation = None
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_except_valid_state(self):
+        if any(property.state not in ["new", "cancelled"] for property in self):
```
aboo-odoo: This works well! An alternative could be

```suggestion
        if self.filtered(lambda property: property.state not in ["new", "cancelled"]):
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+
+    price = fields.Float()
+    _check_price = models.Constraint("CHECK(price >= 0)", "Price must be positive")
+    status = fields.Selection(
+        selection=[("accepted", "Accepted"), ("refused", "Refused")], copy=False
+    )
+    partner_id = fields.Many2one("res.partner", required=True)
+    property_id = fields.Many2one("estate.property", required=True)
+    property_type_id = fields.Many2one(
+        related="property_id.property_type_id", store=True
```
aboo-odoo: Why do you need to store this related field ?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+    )
+    validity = fields.Integer(default=7)
+    date_deadline = fields.Date(
+        compute="_compute_deadline", inverse="_inverse_deadline"
+    )
+
+    def action_accept(self):
+        if self.property_id.state == "sold":
```
aboo-odoo: What happens if the property is cancelled ?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+        self.status = "accepted"
+        self.property_id.state = "offer_accepted"
+        self.property_id.buyer_id = self.partner_id
+        self.property_id.selling_price = self.price
+
+    def action_refuse(self):
+        self.status = "refused"
+
+    @api.model
+    def create(self, vals):
```
aboo-odoo: It would be preferable that you create the records in a batch rather than one by one, better performances 😄 
```suggestion
    @api.model_create_multi
    def create(self, vals_list):
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+        for record in vals:
+            property = self.env["estate.property"].browse(record["property_id"])
+
+            if float_compare(property.best_price, record["price"], 2) == 1:
+                raise UserError("You already have a higher offer")
+
+            if hasattr(property, "state"):
```
aboo-odoo: The property model always has the state attribute (you defined it as field), in what scenario could you not have one ?

estate/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+        "views/estate_property_views.xml",
+        "views/estate_property_offer_views.xml",
+        "views/estate_property_tag_views.xml",
+        "views/estate_property_type_views.xml",
+        "views/estate_user_views.xml",
+        "views/estate_menus.xml",
+    ],
+    "installable": True,
+    "application": True,
+    "author": "Julien (jupir)",
```
aboo-odoo: Everything you create for Odoo belongs to "Odoo S.A." hehe

```suggestion
    "author": "Odoo S.A.",
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @api.depends("create_date", "validity")
+    def _compute_deadline(self):
+        for offer in self:
+            delta = timedelta(days=offer.validity)
+            offer.date_deadline = (
+                offer.create_date.date() + delta
+                if offer.create_date
+                else today() + delta
+            )
```
aboo-odoo: Alternatively (one-liner)
```suggestion
            offer.date_deadline = (offer.create_date or today()).date() + timedelta(days=offer.validity)
```

aboo-odoo (commented):
Phenomenal work 🥳  I'm being extra-picky so that I can transmit as much information as possible. If there is anything you disagree with, please share it without fear. 

A generic comment:
```
PROBLEM - what is the problem 

task-XXXXXX
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+class EstatePropery(models.Model):
+    _name = "estate.property"
+    _description = "Estate property"
+    _order = "id desc"
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        copy=False, default=datetime.today() + relativedelta(months=3)
```
aboo-odoo: I've just noticed an issue here:

when declaring the default value like so, it will get computed once you start the database. The computed value will then be used until your database is restarted. This means that if you start your db today 26/05 and let it run for a month and that, on the 26/06 you create a new property record, the default `date_availibility` will be 26/08 (mai + 3 month) while we expect it to be 26/09 (june + 3 month). 

To make the default value recomputed at each call; you need to call a function

jupir-odoo:
Hello @aboo-odoo, j'ai appliqué tes conseils. Merci beaucoup. J'ai aussi fait le tuto sur les unit tests. Je ne sais pas si tu préfères que je les push aussi ici ou que je crée une deuxième PR. Merci d'avance !

awesome_owl/static/src/playground.js
```diff
@@ -1,5 +1,13 @@
...
 export class Playground extends Component {
     static template = "awesome_owl.playground";
+
+    setup() {
+        this.state = useState({ value: 0 })
+    }
+
+    increment() {
+        this.state.value++
+    }
```
aboo-odoo, line 12: What is this doing here ? Can you remove it ahah?

estate/tests/test_estate_property.py
```diff
@@ -0,0 +1,84 @@
+from odoo.exceptions import UserError
+from odoo.tests import Form, new_test_user, tagged
+from odoo.tests.common import TransactionCase
+
+
+@tagged("post_install", "-at_install", "estate")
```
aboo-odoo, line 6: The estate tag is not really necessary since you can always run the test with the following tag: `/estate` 

```suggestion
@tagged("post_install", "-at_install")
```

estate/tests/test_estate_property.py
```diff
@@ -0,0 +1,84 @@
...
+                ]
+            )
+
+    def test_properties_cannot_be_sold_without_accepted_offer(self):
+        # when property has no offer it cannot be marked sold
+        with self.assertRaises(UserError):
+            self.properties.action_mark_sold_property()
+
+        # when property has no accepted offer it cannot be marked sold
+        buyer = new_test_user(self.env, login="mark_baier@example.com")
```
aboo-odoo, line 67: You use that buyer in several tests, it could be a good idea to add it to the setupclass 👍 

You could also create another property with offers. So that you have one with and one without

aboo-odoo (commented):
A nice and clean PR! Congrats 🥳

PR #1286, opened 2026-05-19
maorb - Technical Training

Mark-Orban:


estate/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+from . import estate_property
+from . import estate_property_type
+from . import estate_property_tag
+from . import estate_property_offer
+from . import res_users
```
Megaaaaaa: You're missing the **E**nd **O**f **L**ine linebreak character (also known as EOL). Every file in Odoo must end with an EOL so that if someone wants to add something to your file, they don't override the history of your last line :+1:
```suggestion
from . import res_users

```

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
+from odoo import models, fields, api
```
Megaaaaaa: The [coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#imports) will show you how to order your imports. Simply put it's split in three groups: external imports, odoo imports and odoo.addons import. Each group has modules imported alphabetically. Each module as attributes imported asciibetically (All uppercases before all lowercases) :+1: I recommend to keep this piece of information close to you while you dev in the beginning, it's very (very very) usefull.

```suggestion
from odoo import api, fields, models
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _check_expected_price = models.Constraint("CHECK (expected_price >= 0)", "The expected price must be strictly positive.")
+    _check_selling_price = models.Constraint("CHECK (selling_price >=0)", "The selling price must be strictly positive.")
```
Megaaaaaa: So despite being well ordered, there is a grouping rule that split the attributes first before any alphabetical sort. In every model we group attributes: reserved names (`_name`, `_description`,...), then the fields, then the constrains (like these two) and then the methods (sometimes also grouped: constrains, then computes and inverses, then helpers, then business methods). So these two lines need to go between all your fields and all your methods :+1:

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+        string = "Property Name",
+        required = True,
+    )
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        copy = False, 
+        default = fields.Date.today().replace(month=(fields.Date.today().month + 3) % 12),
+    )
+    expected_price = fields.Float(required = True)
```
Megaaaaaa: So that's something the runbot (and ore precisely the ci/style) will always remind you of but inside fields declarations, we don't put spaces around the operators. There are many occurences of this in your code so I will not flag them all. The runbot already does so anyway :+1:

```suggestion
    expected_price = fields.Float(required=True)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = None
+    
+    @api.constrains("selling_price")
+    def _check_minimum_selling_price(self):
+        for record in self:
+            if record.selling_price and float_compare(record.expected_price * 0.9, record.selling_price, precision_digits=2) > 0:
+                raise ValidationError("The selling price must be at least 90%% of the expected price. You must reduce the expected price in order to accept this offer.")
```
Megaaaaaa: Every piece of text that will be shown to the user needs to be translatable because we have clients all over the world and you never know who will stumble upon your message. To do so, in Odoo, we have the `odoo._()` method. That methods flags the string as needing translations and then the odoo translators army takes care of the rest :+1:


```suggestion
                # put from odoo import _ in your imports at the top of the file
                raise ValidationError(_("The selling price must be at least 90%% of the expected price. You must reduce the expected price in order to accept this offer."))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_except_active_property(self):
+        for record in self:
+            if record.state in ("offer_received", "offer_accepted", "sold"):
+                raise UserError("Only new or cancelled properties may be deleted.")
```
Megaaaaaa: Translation
```suggestion
                raise UserError(_("Only new or cancelled properties may be deleted."))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def action_sold_button(self):
+        if self.state != "cancelled":
+            self.state = "sold"
+        else:
+            raise UserError("A cancelled property cannot be sold!")
```
Megaaaaaa: Translation
```suggestion
            raise UserError(_("A cancelled property cannot be sold!"))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+        return True
+    
+    def action_cancel_button(self):
+        if self.state != "sold":
+            self.state = "cancelled"
+        else:
+            raise UserError("A sold property cannot be cancelled!")
```
Megaaaaaa: Translation
```suggestion
            raise UserError(_("A sold property cannot be cancelled!"))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
+from odoo import models, fields, api
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
```
Megaaaaaa, line 7: This isn't an issue here, just for your information: Regarding quotes and double quotes in Odoo, there is no absolute rule and depending on the team you join, they might ask you to use one or the other. The important thing is to be consistent though. I see you use double quotes everywhere, that is fine and even good. Much better than using one in some places and the other the rest of the time. There are just some cases in which you won't be able to use one because you need it inside a string then you don't really have the choice if you want it to look good.

To give you an example, in the accounting team, we use single quotes everywhere except for messages shown to the user (errors, notifications, etc) where we use double quotes.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,78 @@
+from odoo import models, fields, api
+from odoo.exceptions import UserError
+from odoo.tools import float_compare
```
```suggestion
from odoo import api, fields, models
from odoo.exceptions import UserError
from odoo.tools import float_compare
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,78 @@
...
+
+class EstatePropertyOffer(models.Model):
+    _name = "estate.property.offer"
+    _check_price_positive = models.Constraint("CHECK (price >= 0)", "Offer price must be strictly positive.")
```
Megaaaaaa: after all your fields

estate/models/estate_property.py
```diff
@@ -0,0 +1,126 @@
...
+    bedrooms = fields.Integer(default = 2)
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        string = "Orientation",
+        selection = [("north","North"), ("east", "East"), ("south", "South"), ("west", "West")],
+    )
```
Megaaaaaa, line 40: Just a unspoken rule we follow, for selection fields, we always put the selection list in first. I'm not even sure any reviewer will tell you anything about that but we always do it :)
```suggestion
    garden_orientation = fields.Selection(
        selection=[("north","North"), ("east", "East"), ("south", "South"), ("west", "West")],
        string="Orientation",
    )
```
(just like the related model for relational fields for example)

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,78 @@
...
+    def _compute_date_deadline(self):
+        for record in self:
+            if record.create_date:
+                record.date_deadline = fields.Date.add(record.create_date, days=record.validity)
+            else:
+                record.date_deadline = fields.Date.add(fields.Date.today(), days=record.validity)
+    
+    @api.model
+    def create(self, vals):
+        vals = vals[0]
```
Megaaaaaa: I know you do this because vals is a list of dicts with only one dict inside but this would not be accepted imo. What you can do to make it look much better is change your `@api.model` into `@api.model_create_multi` to flag this create method as usable to create multiple records at once and then your second argument can become a list of multiple dicts. Then you can loop in it and you'll have a much cleaner `create` :+1:


```suggestion
    @api.model_create_multi
    def create(self, vals_list):
        for vals in vals_list:
            ...
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,78 @@
...
+        if vals.get("property_id") and vals.get("price"):
+            prop = self.env["estate.property"].browse(vals["property_id"])
```
Megaaaaaa: We try to put as few comments in the code and let it speak for itself. Following that direction, we need variables to have the clearest names possible as it should be easy to understand what variable holds which piece of data even a thousand lines after declaration so don't be shy on the naming. Except if it wayyy to long, you can name them with full words :+1:

```suggestion
            property = self.env["estate.property"].browse(vals["property_id"])
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,78 @@
...
+            if prop.offer_ids:
+                highest_offer = max(prop.mapped("offer_ids.price"))
+                if float_compare(vals["price"], highest_offer, precision_rounding=0.01) <= 0:
+                    raise UserError(f"The offer must be higher than {highest_offer}")
```
Megaaaaaa: Translation special case: Single variable usage

When you need to put one or multiple variables inside a translated string, you can use `'%s' % var`
```suggestion
                    raise UserError(_("The offer must be higher than %s." % highest_offer))
```

I put it here in case there is no example because it's easily doable in the wrong way: When you have multiple variables, you need to specify the each one because some other languages might reorder them in the trnaslation so it woul be:
```suggestion
                    raise UserError(_("The offer must be higher than %(minimum)s and lower than %(maximum)s.", minimum=minimum, maximum=maximum))
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,78 @@
...
+            prop.state = "offer_received"
+        return super().create(vals)
+    
+    def _inverse_date_deadline(self):
+        for record in self:
+            if record.date_deadline:
+                record.validity = (record.date_deadline - record.create_date.date()).days
+            else:
+                record.validity = (fields.Date.today() - record.create_date.date()).days
```
Megaaaaaa: So it's fine this way but when you have an `if else` that you don't expect to become an `if elif else` at any point like this one (either you have a deadline or you don't.), then it's very often nicer to oneline it. It depends on how verbose you want it to be but this one is pretty simple so you can: 
```suggestion
            record.validity = ((record.date_deadline or fields.Date.today()) - record.create_date.date()).days
```

There were other places where this could have been done before and I suppose I'm going to see more but since it's very minimal, just do what you prefer :smile:

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,78 @@
...
+    
+    def action_status_accepted(self):
+        if self.property_id.state == "new" or self.property_id.state == "offer_received":
```
Megaaaaaa: When we check one selection field for multiple values, we tend to use `in` instead of many `or` so the diffs look better when we add or remove values :+1:

```suggestion
        if self.property_id.state in ("new", "offer_received"):
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,78 @@
...
+            self.status = "accepted"
+            self.property_id.state = "offer_accepted"
+            self.property_id.selling_price = self.price
+            self.property_id.buyer_id = self.partner_id
+        else:
+            raise UserError("This property has already accepted an offer, been sold, or is cancelled!")
```
Megaaaaaa: Translation

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,78 @@
...
+        return True
+
+    def action_status_refused(self):
+        self.status = "refused"
+        return True
```
Megaaaaaa: EOL

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,12 @@
+from odoo import models, fields
```
Megaaaaaa:
```suggestion
from odoo import fields, models
```

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,12 @@
...
+
+class EstatePropertyTag(models.Model):
+    _name = "estate.property.tag"
+    _check_name_unique = models.Constraint("UNIQUE(name)", "Tag must be unique.")
```
Megaaaaaa: After the fields

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,12 @@
...
+    _description = "Estate Property Tag Model"
+    _order = "name"
+
+    name = fields.Char(
+        string = "Property Tag",
+        required = True)
+    color = fields.Integer()
```
Megaaaaaa: Missing EOL

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,30 @@
+from odoo import models, fields
```
```suggestion
from odoo import fields, models
```

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,30 @@
...
+
+class EstatePropertyType(models.Model):
+    _name = "estate.property.type"
+    _check_name_unique = models.Constraint("UNIQUE(name)", "Property type must be unique.")
```
Megaaaaaa: After your fields

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,30 @@
...
+    )
+    offer_ids = fields.One2many(
+        "estate.property.offer",
+        "property_type_id",
+    )
+    offer_count = fields.Integer(compute="_compute_total_offers")
+
+    def _compute_total_offers(self):
+        for record in self:
+            self.offer_count = len(self.offer_ids)
```
Megaaaaaa: Missing EOL

estate/models/res_users.py
```diff
@@ -0,0 +1,10 @@
+from odoo import fields, models
+
+class ResUsers(models.Model):
+    _inherit = ["res.users"]
```
Megaaaaaa: When you inherit from a single model, you can use a string directly, no need for a list :+1:
```suggestion
    _inherit = "res.users"
```

estate/models/res_users.py
```diff
@@ -0,0 +1,10 @@
...
+
+    property_ids = fields.One2many(
+        "estate.property",
+        "salesperson_id",
+        domain=[("state", "in", ["new", "offer_received"])],
+    )
```
Megaaaaaa: EOL

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
+access_estate_property_type,access_estate_property_type,model_estate_property_type,base.group_user,1,1,1,1
+access_estate_property_tag,access_estate_property_tag,model_estate_property_tag,base.group_user,1,1,1,1
+access_estate_property_offer,access_estate_property_offer,model_estate_property_offer,base.group_user,1,1,1,1
```
Megaaaaaa: EOL

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,12 @@
...
+    <menuitem id="estate_menu_root" name="Real Estate">
+        <menuitem id="estate_first_level_menu" name="Advertisements">
+            <menuitem id="estate_property_menu_action" name="Properties" action="estate_property_action"/>
+        </menuitem>
+        <menuitem id="estate_settings_menu" name="Settings">
+            <menuitem id="estate_property_type_menu_action" name="Property Types" action="estate_property_type_action"/>
+            <menuitem id="estate_property_tag_menu_action" name="Property Tags" action="estate_property_tag_action"/>
+        </menuitem>
+    </menuitem>
+</odoo>
```
Megaaaaaa: EOL

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+        'base',
+    ],
+    'application': True,
+    'data' : [
+        'views/estate_property_offer_views.xml',
+        'security/ir.model.access.csv',
+        'views/estate_property_views.xml',
+        'views/estate_menus.xml',
+        'views/estate_property_type_views.xml',
+        'views/estate_property_tag_views.xml',
```
Megaaaaaa: You're gonna have a small issue here because your menu uses actions that are defined in `estate_property_type_views` and `estate_property_tags_views` so when you dropdb and reinitialize it from scratch, you will get an error trying to access actions that do not exist yet :smile:
```suggestion
        'views/estate_property_type_views.xml',
        'views/estate_property_tag_views.xml',
        'views/estate_menus.xml',
```

estate/models/estate_property_type.py
Megaaaaaa: Very few exceptions to this but when you are inside a `for x in self:` loop, you never use `self` anymore. This good practice will prevent a lot of errors and unpredicted behaviors :+1:

```suggestion
            record.offer_count = len(record.offer_ids)
```

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,42 @@
...
+        </field>
+    </record>
+
+    <record id="estate_property_offer_action" model="ir.actions.act_window">
+        <field name="name">Property Offers</field>
+        <field name="res_model">estate.property.offer</field>
+        <field name="domain">[('property_type_id','=', active_id)]</field>
+        <field name="view_mode">list,form</field>
+    </record>
+</odoo>
```
Megaaaaaa: EOL

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,32 @@
...
+            </list>
+        </field>
+    </record>
+
+    <record id="estate_property_tag_action" model="ir.actions.act_window">
+        <field name="name">Tags</field>
+        <field name="res_model">estate.property.tag</field>
+        <field name="view_mode">list,form</field>
+    </record>
+</odoo>
```
Megaaaaaa: EOL

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,53 @@
...
+                        <group class="text-end">
+                            <button name="%(estate.estate_property_offer_action)d" type="action" class="oe_stat_button" icon="fa-money">
+                                <field name="offer_count" widget="statinfo" string="Offers"/>
+                            </button>
+                        </group>
+                    </group>
+                    <notebook>
+                        <page string="Properties">
+                            <group>
+                                <field name="property_ids" string="">
```
Megaaaaaa, line 22: If you wanna stick with the default one, i don't think you need to keep the `string=""`
```suggestion
                                <field name="property_ids">
```

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,53 @@
...
+            </list>
+        </field>
+    </record>
+
+    <record id="estate_property_type_action" model="ir.actions.act_window">
+        <field name="name">Property Types</field>
+        <field name="res_model">estate.property.type</field>
+        <field name="view_mode">list,form</field>
+    </record>
+</odoo>
```
Megaaaaaa: EOL :smile:

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,108 @@
...
+        <field name="name">estate.property.search</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <search>
+                <field name="name" string="Title"/>
+                <field name="postcode"/>
+                <field name="property_type_id" string="Property Type"/>
+                <field name="expected_price" string="Expected Price"/>
+                <field name="bedrooms"/>
+                <field name="living_area" string="Living Area (sqm)" filter_domain="[('living_area', '>', self)]"/>
```
Megaaaaaa: this is a weird way to check for positive living area :smile: Have you tried `'>', 0` ?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,108 @@
...
+                <field name="facades"/>
+                <field name="tag_ids" string="Tags"/>
+
+                <filter name="available_filter" string="Available" domain="['|',('state', 'ilike', 'new'), ('state', 'ilike', 'offer_received')]"/>
```
Megaaaaaa:
```suggestion
                <filter name="available_filter" string="Available" domain="['|', ('state', 'ilike', 'new'), ('state', 'ilike', 'offer_received')]"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,108 @@
...
+    </record>
+
+    <record id="estate_property_view_form" model="ir.ui.view">
+        <field name="name">estate.property.form</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form>
+                <header>
+                    <button name="action_sold_button" type="object" string="Sold" invisible="state in ('sold')"/>
+                    <button name="action_cancel_button" type="object" string="Cancel" invisible="state in ('sold')"/>
```
Megaaaaaa: If you don't need multiple values to make it invisible, just use `==` :smile:

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,108 @@
...
+                    <field name="state" widget="statusbar" statusbar_visible="new,offer_received,offer_accepted,sold"/>
```
Megaaaaaa:
```suggestion
                    <field name="state" widget="statusbar" statusbar_visible="new, offer_received, offer_accepted, sold"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,108 @@
...
+                <sheet>
+                    <group colspan="2">
+                        <h1><field name="name"/></h1>
+                    </group>
+                    <group>
+                        <field name="tag_ids" widget="many2many_tags" options="{'color_field': 'color'}"/>
+                    </group>
+                    <group>
+                        <group>
+                            <field name="property_type_id" string="Property Type" options="{'no_create':true}"/>
```
Megaaaaaa:
```suggestion
                            <field name="property_type_id" string="Property Type" options="{'no_create': true}"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,108 @@
...
+                                <field name="garage"/>
+                                <field name="garden"/>
+                                <field name="garden_area" string="Garden Area (sqm)" invisible="not garden"/>
+                                <field name="garden_orientation" string="Garden Orientation" invisible="not garden"/>
+                                <field name="total_area" string="Total Area (sqm)"/>
+                            </group>
+                        </page>
+                        <page string="Offers">
+                            <group>
+                                <field name="offer_ids" string="" context="{'list_view_ref' : 'estate.estate_property_offer_view_list'}" readonly="state in ('offer_accepted','sold','cancelled')"/>
```
Megaaaaaa:
```suggestion
                                <field name="offer_ids" string="" context="{'list_view_ref' : 'estate.estate_property_offer_view_list'}" readonly="state in ('offer_accepted', 'sold', 'cancelled')"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,108 @@
...
+        </field>
+    </record>
+    
+    <record id="estate_property_action" model="ir.actions.act_window">
+        <field name="name">Properties</field>
+        <field name="res_model">estate.property</field>
+        <field name="view_mode">list,form</field>
+        <field name="context">{"search_default_available_filter": True}</field>
+    </record>
+</odoo>
```
Megaaaaaa: EOL once again :smile:

estate/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
Megaaaaaa: EOL

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+        'views/res_users_views.xml',
+    ]
+}
```
Megaaaaaa: EOL

estate_account/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_account
```
Megaaaaaa: EOL

estate_account/models/estate_account.py
```diff
@@ -0,0 +1,26 @@
...
+                        }),
+                        Command.create({
+                            "name" : "Administrative fees",
+                            "quantity" : 1,
+                            "price_unit" : 100,
+                        }),
+                    ],
+                }
+            )
+        return super().action_sold_button()
```
Megaaaaaa: Don't use `self` inside a `for x in self` loop :+1:

Megaaaaaa: and EOL :smile:

estate_account/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
Megaaaaaa: EOL

estate_account/__manifest__.py
```diff
@@ -0,0 +1,9 @@
+{
+    'name': 'Estate Account',
+    'depends': [
+        'estate',
+        'account',
+    ],
+    'application': True,
+    'data' : []
+}
```
Megaaaaaa: And the last one: EOL :wink:

Megaaaaaa (changes requested):
Hello 👋

Very nice PR, good job on this! :+1:

There's quite a lot of stuff but don't worry about it, it's a lot of the same small things that repeat themselves.

There is a lot of stuff on your ci/style so usually you would start with that but I would recommend you have a look at my review first because I might explain why the ci/style flags things that you would otherwise think are fine.

Keep in mind that a review is not the utlimate solution, it's only someone seeing things he would have done differently and offering an alternative. You're always free to answer with an other alternative or even say you disagree and bring your arguments. For stuff that's not just styling or conventions, don't tak the wrong habit of just blindly applying comments, you spent time developping the thing, everything you've tried doesn't show up in the diff, only the final one. You probably have very interesting things to say to the reviewer about the decisions you made.

What I would suggest you do and continue to do while you're not the most comfortable with odoo's structure is to not apply all the changes at once. For example, changing a compute method or even just indentation can break stuff and you change everything at once, you might end up with a lot of errors when running the database. Splitting the review in multiple steps and trying to run your db in between those can save you a lot of time in the long run. Also don't be shy with dropping the databse. It's a very nice way to make sure everything works well from scratch.

Also a quick tip for the methodology when applying reviews changes: What I like to do it put a reaction like a thumbsup on a message once I have made the change locally. Then when everything is marked with something, I push, go through my diff and mark as resolved the things that are now outdated with the button under each comment so nothing is left forgotten.

PR #1287, opened 2026-05-20
[ADD] estate: link properties to accounting and finalize business logic

manalelk:


aboo-odoo:
@manalelk Could you push the work you've done so that I can do a first round of review :smile:

estate/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+from . import estate_property
+from . import estate_property_type
+from . import estate_property_tag
+from . import estate_property_offer
```
aboo-odoo: You need to end all new files with an empty line. This minimizes the diff next time someone adds lines to your file; easier to review.

```suggestion
from . import estate_property_offer

```

estate/models/estate_property.py
```diff
@@ -0,0 +1,104 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
```
aboo-odoo: Those are remains from the past, they are not needed anymore

estate/models/estate_property.py
```diff
@@ -0,0 +1,104 @@
...
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "textext"
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    property_type_id = fields.Many2one("estate.property.type")
+    postcode = fields.Char()
+    date_availability = fields.Date(string= 'Available From', default= fields.Date.add(fields.Date.today(), months=3), copy=False)
```
aboo-odoo: When declaring the default value like so, it will get computed once you start the database. The computed value will then be used until your database is restarted. This means that if you start your db today 26/05 and let it run for a month and that, on the 28/06 you create a new property record, the default `date_availibility` will be 28/08 (mai + 3 month) while we expect it to be 28/09 (june + 3 month). 

To make the default value recomputed at each call; you need to make the parameter a function, i.e. `lambda`

estate/models/estate_property.py
```diff
@@ -0,0 +1,104 @@
...
+    selling_price = fields.Float(readonly=True, copy=False)
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer(string='Living Area (sqm)')
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
+    )
```
aboo-odoo, line 24: There is a tacit convention (in the accounting team at least) for quotes: strings the use seems are double quotes `"`, other strings are single quoted `'`. Of course, it is not that simple, each team has its own guidelines regarding this. 

```suggestion
    garden_orientation = fields.Selection(
        selection=[
            ('north', "North"),
            ('south', "South"),
            ('east', "East"),
            ('west', "West"),
        ]
    )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,104 @@
...
+            ('cancelled', 'Cancelled'),
+        ],
+        required=True,
+        copy=False,
+        default='new',
+    )
+    salesperson_id = fields.Many2one('res.users', string= 'Salesman', index=True, default=lambda self: self.env.user)
+    buyer_id = fields.Many2one('res.partner', index=True)
+    tag_ids = fields.Many2many('estate.property.tag', string="Tags")
+    offer_ids = fields.One2many('estate.property.offer', 'property_id', string="Offers")
```
aboo-odoo: Good use of `_id.s` suffixes here! 🥳

estate/models/estate_property.py
```diff
@@ -0,0 +1,104 @@
...
+    total_area = fields.Integer(string='Total Area (sqm)', compute="_compute_total_area")
+    best_price = fields.Float(string='Best Offer', compute="_compute_highest_offer")
+
+    @api.depends('living_area', 'garden_area')
+    def _compute_total_area(self):
+        for record in self:
```
aboo-odoo: `record` is not really explicit as variable name. It works fine for small method, but when you will have huge methods, it's better to use a more meaningful variable name
```suggestion
        for property in self:
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,104 @@
...
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids.price")
+    def _compute_highest_offer(self):
+        for record in self:
+            prices = record.offer_ids.mapped("price")
+            record.best_price = max(prices) if prices else 0.0
```
```suggestion
            record.best_price = max(prices, default=0.0)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,104 @@
...
+    @api.constrains("expected_price", "selling_price")
+    def _check_selling_price(self):
+        for record in self:
+            if float_is_zero(record.selling_price, precision_digits=2):
+                continue
+                
+            limit_price = record.expected_price * 0.9
+            if float_compare(record.selling_price, limit_price, precision_digits=2) == -1:
+                raise ValidationError(
+                    "The selling price cannot be lower than 90% of the expected price!"
```
aboo-odoo: Some strings are not automatically translated, such as fields names, selection values, etc. However, this is not the case for exceptions. To specify you want a string to be translateable, you need to add the `_` method like so 
```suggestion
            raise ValidationError(self.env._("The selling price cannot be lower than 90% of the expected price!"))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,104 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import fields, models, api
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
```
aboo-odoo: Imports should follow RUFF specification, i.e; the ASCII order and be grouped into 3 categories:
1. General python library
2. Imports of odoo
3. Imports from Odoo modules


```suggestion
from odoo import api, fields, models
from odoo.exceptions import UserError, ValidationError
from odoo.tools.float_utils import float_compare, float_is_zero
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,59 @@
...
+            ('refused', 'Refused'),
+        ],
+        copy=False,
+    )
+    property_type_id = fields.Many2one("estate.property.type", related="property_id.property_type_id", store=True)
+
+    @api.depends('create_date', 'validity')
+    def _compute_date_deadline(self):
+        for record in self:
+            base_date = record.create_date.date() if record.create_date else fields.Date.today()
```
aboo-odoo: Python hack:

```suggestion
            base_date = (record.create_date or fields.Date.today()).date()
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,59 @@
...
+    # CONSTRAINTS
+    _check_price = models.Constraint(
+        "CHECK(price > 0)",
+        "An offer price must be strictly positive."
+    )
+
+    # ACTIONS
+    def action_accept(self):
+        for record in self:
+            if record.property_id:
```
aboo-odoo: Functionaly speaking, what happens if the house is cancelled, it seems like I'm still able to accept an offer no ?

aboo-odoo (commented):
Amazing work 🥳  I'm being extra-picky so that I can transmit as much information as possible. If there is anything you disagree with, please share it without fear. 

A generic comment:
```
PROBLEM - what is the problem 

task-XXXXXX
```
- can you squash all your commits: if you don't remember how, come to me and I'll show you how 😄

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,28 @@
...
+    def action_sold(self):
+        res =  super().action_sold()
+
+        for prop in self:
+
+            self.env["account.move"].create({
+                "partner_id": prop.buyer_id.id,
+                "move_type": "out_invoice",
+                "line_ids": [
+                    Command.create({
```
aboo-odoo, line 15: Isn't this missing a line with the price of the house ?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,81 @@
...
+    # CRUD methods       
+    @api.model
+    def create(self, vals):   
+        for val in vals:
+            property_id = val.get('property_id')
+            if property_id:
+                property_record = self.env['estate.property'].browse(property_id)
+
+                if property_record.state == 'cancelled':
+                    raise UserError("An offer cannot be made for a cancelled property!")
```
aboo-odoo, line 51: Don't forget the translations for all the error messages

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,81 @@
...
+        return True
+
+    def action_refuse(self):
+        for offer in self:
+            offer.status = 'refused'
+            if offer.property_id and offer.property_id.buyer_id == offer.partner_id:
+                    offer.property_id.buyer_id = False
+                    offer.property_id.selling_price = 0.0
+        return True
+
```
aboo-odoo, line 81: Don't forget the end of line return of the end of every file - it seems your editor add spaces everytime :S 
```suggestion

```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,81 @@
...
+                        if val.get('price', 0) < existing_offer.price:
+                            raise UserError("The offer amount cannot be lower than an existing offer!")
+
+                property_record.state = 'offer_received'
+
+        return super().create(val)
+
+    # Action methods
+    def action_accept(self):
+        for offer in self:
```
aboo-odoo, line 63: Functional question:  should you be able to accept/refuse offers for properties that are already sold/refused

aboo-odoo (commented):
Just a few last comments 😄

PR #1292, opened 2026-05-25
[ADD] zero_stock_sale_approval: add zero stock approval field on sale order

moahi-odoo:
Sales users were able to confirm orders even when stock was not available. This could lead to wrong commitments, stock issues, and lack of proper control.

Introduced a zero stock approval mechanism in sale orders. Added a Boolean field zero_stock_approval to control confirmation behavior.

- Created a new module for zero stock approval
- Added zero_stock_approval field in sale.order model
- Restricted field editing to Admin/Sales Manager only
- Updated action_confirm to validate stock before confirming
- Blocked confirmation when stock is insufficient unless approval is enabled
- Extended sale order form view to display the new field

### DIFF
--- a/zero_stock_sale_approval/__manifest__.py
+++ b/zero_stock_sale_approval/__manifest__.py
@@ -1,0 +1,10 @@
+{
+    'name': 'Sale Zero Stock Approval',
+    'description': 'Grant permission to confirm approved orders without stock.',
+    'depends': ['sale_management', 'sale_stock'],
+    'author': 'moahi',
+    'license': 'LGPL-3',
+    'data': [
+        'views/sale_order_views.xml',
+    ]
+}
--- a/zero_stock_sale_approval/models/sale_zero_stock.py
+++ b/zero_stock_sale_approval/models/sale_zero_stock.py
@@ -1,0 +1,23 @@
+from odoo import fields, models
+from odoo.exceptions import AccessError
+
+
+class SaleZeroStock(models.Model):
+    _inherit = 'sale.order'
+
+    zero_stock_approval = fields.Boolean(string="Approval by Manager", default=False)
+
+    def action_confirm(self):
+        for record in self:
+            for item in record.order_line:
+                if (
+                    not record.zero_stock_approval
+                    and not self.env.user.has_group("sales_team.group_sale_manager")
+                    and item.product_id.type == "consu"
+                    and item.product_id.is_storable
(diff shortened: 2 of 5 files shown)


zero_stock_sale_approval/__manifest__.py
```diff
@@ -0,0 +1,10 @@
+{
+    'name': 'Sale Zero Stock Approval',
+    'description': 'Grant permission to confirm approved orders without stock.',
+    'depends': ['sale_management', 'sale_stock'],
```
mash-odoo, line 4: This won't be installed directly, and user wouldn't know the features coming along with this modules. So don't you think we should do something such that this module also get installed once it's dependency is installed?

zero_stock_sale_approval/models/sale_zero_stock.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def action_confirm(self):
+        for record in self:
+            for item in record.order_line:
+                if (
+                    not record.zero_stock_approval
+                    and not self.env.user.has_group("sales_team.group_sale_manager")
+                    and item.product_id.type == "consu"
+                    and item.product_id.is_storable
+                    and item.product_id.qty_available < item.product_uom_qty
```
mash-odoo, line 18: Why use qty_available and not free_qty? Can you walk me through your approach here?

Also, I think `item.product_id.qty_available` this will give global stock?
Should availability be checked against the order's warehouse instead of global stock? 🤔

mash-odoo, line 18: Is this comparison correct? Like will it work fine acroos all the UOM's?

zero_stock_sale_approval/models/sale_zero_stock.py
```diff
@@ -0,0 +1,23 @@
+from odoo import fields, models
+from odoo.exceptions import AccessError
+
+
+class SaleZeroStock(models.Model):
+    _inherit = 'sale.order'
+
+    zero_stock_approval = fields.Boolean(string="Approval by Manager", default=False)
```
mash-odoo, line 8: Should approval reset automatically when: order lines change, quantities change, product changes?
Like what if its approved but the SOL is changed later?

Also maybe you can keep a track history on approval to know who approved and when

mash-odoo (changes requested):
Hello!
Good work on the task..
I have added some questions...please have a look at it..

PR #1300, opened 2026-05-27
[ADD] owl framework 101

sreedevk:


awesome_owl/static/src/counter/counter.js
```diff
@@ -0,0 +1,14 @@
+import { Component, useState } from '@odoo/owl'
+
+export class Counter extends Component {
+    static template = "awesome_owl.counter";
```
Megaaaaaa, line 4: This component expects its parent to pass an `onChange` callback, so it is worth making that contract explicit in `static props`. It's helpful to everyone and you will avoid the "no explicit props declared" warning too :+1:

```suggestion
    static template = "awesome_owl.counter";
    static props = {
        onChange: Function,
    };
```

awesome_owl/static/src/todo/todo_item.xml
```diff
@@ -0,0 +1,31 @@
...
+
+            <div class="d-flex align-items-center gap-2">
+                <t t-if="props.todo.isCompleted">
+                    <span class="badge bg-success rounded-pill px-2 py-1">Done</span>
+                </t>
+                <t t-else="">
+                    <span class="badge bg-warning text-dark rounded-pill px-2 py-1">Pending</span>
+                </t>
+
+                <span class="fa fa-remove text-danger cp ms-2"
```
Megaaaaaa: This is a clickable control, so it should be a button rather than a span. Buttons are keyboard-focusable by default so it's preferable to use them in these situations :+1:

```suggestion
                <button type="button" class="btn btn-link text-danger p-0 ms-2" t-on-click="() => props.removeTodo(props.todo.id)">
                    <i class="fa fa-remove"/>
                </button>
```

awesome_owl/static/src/todo/todo_list.js
```diff
@@ -0,0 +1,49 @@
...
+        this.nextId = 2;
+        this.inputRef = useRef("input");
+
+        onMounted(() => {
+            if (this.inputRef.el) this.inputRef.el.focus();
+        });
+    }
+
+    createItem(ev) {
+        if (ev.keyCode === 13) {
```
Megaaaaaa: `keyCode` works in many browsers but it is less self-explanatory and we like self-explanatory. Checking the key name directly reads like the behavior we want: create an item when the user presses Enter.
```suggestion
        if (ev.key === "Enter") {
```

awesome_owl/static/src/todo/todo_list.xml
```diff
@@ -0,0 +1,26 @@
...
+                    t-ref="input"
+                    class="form-control form-control-lg border-primary shadow-sm" 
+                    placeholder="Add a new task.." 
+                    t-on-keyup="createItem" />
+            </div>
+
+            <t t-if="items.length > 0">
+                <t t-foreach="items" t-as="todo" t-key="todo.id">
+                    <TodoItem todo="todo" 
+                        toggleState="toggleItem.bind(this)"
```
Megaaaaaa: You can use Owl dedicated `.bind` syntax for passing component methods as callbacks. It's the Owl-native way to wire child events back to the parent. It's just to show you :+1:
```suggestion
                        toggleState.bind="toggleItem"
```

awesome_owl/static/src/todo/todo_list.xml
```diff
@@ -0,0 +1,26 @@
...
+                    class="form-control form-control-lg border-primary shadow-sm" 
+                    placeholder="Add a new task.." 
+                    t-on-keyup="createItem" />
+            </div>
+
+            <t t-if="items.length > 0">
+                <t t-foreach="items" t-as="todo" t-key="todo.id">
+                    <TodoItem todo="todo" 
+                        toggleState="toggleItem.bind(this)" 
+                        removeTodo="deleteItem.bind(this)"/>
```
Megaaaaaa: And so same here if you want :+1:

Megaaaaaa (commented):
Hello :wave:

Mostly looks good to my not-so-js-trained eye. Just some tiny things we like to do another way and I wanted to show you. It works in both cases anyway but so you see the Odoo way :)

Tell me if anything is not clear enough or if you have more questions!

PR #1303, opened 2026-05-28
[ADD] awesome_owl, awesome_dashboard: partially implement web framework tutorial

daloe-odoo:
add card, counter, and todo components to awesome_owl. add almost reactive pie chart to awesome_dashboard.

task-6253605

### DIFF
--- a/awesome_dashboard/static/src/dashboard.js
+++ b/awesome_dashboard/static/src/dashboard.js
@@ -1,8 +1,36 @@
-import { Component } from "@odoo/owl";
+import { Component, onWillStart, useState } from "@odoo/owl";
 import { registry } from "@web/core/registry";
+import { useService } from "@web/core/utils/hooks";
+import { _t } from "@web/core/l10n/translation";
+
+import { DashboardItem } from "./dashboard_item/dashboard_item";
+import { Layout } from "@web/search/layout"
+import { PieChart } from "./pie_chart/pie_chart";
+
 
 class AwesomeDashboard extends Component {
     static template = "awesome_dashboard.AwesomeDashboard";
+    static components = { Layout, DashboardItem, PieChart };
+
+    setup() {
+        this.action = useService("action");
+        this.state = useState({
+            stats: useService("awesome_dashboard.statistics"),
+        });
+    }
+
+    async openCustomersKanban() {
+        this.action.doAction('base.action_partner_form');
+    }
+
+    async openLeads() {
+        this.action.doAction({
+            type: 'ir.actions.act_window',
+            name: _t('Lots of Leads'),
+            res_model: 'crm.lead',
+            views: [[false, 'list'], [false, 'form']],
+        });
+    }
 }
 
 registry.category("actions").add("awesome_dashboard.dashboard", AwesomeDashboard);
--- a/awesome_dashboard/static/src/dashboard.xml
+++ b/awesome_dashboard/static/src/dashboard.xml
@@ -2,7 +2,20 @@
 <templates xml:space="preserve">
 
     <t t-name="awesome_dashboard.AwesomeDashboard">
-        hello dashboard
+        <t t-set-slot="layout-buttons">
+            <button t-on-click="openCustomersKanban">Customers</button>
+            <button t-on-click="openLeads">Leads</button>
+        </t>
+
+        <Layout display="{controlPanel: {} }" className="'o_dashboard h-100'">
+            <DashboardItem size="2">
+                         <p>Counter: <t t-esc="this.state.stats.nb_new_orders"/></p>
+            </DashboardItem>
+
+            <DashboardItem>
+                <PieChart data="this.state.stats.orders_by_size"/>
+            </DashboardItem>
+        </Layout>
     </t>
 
 </templates>
--- a/awesome_dashboard/static/src/dashboard.xml
+++ b/awesome_dashboard/static/src/dashboard.xml
@@ -12,7 +12,7 @@
                          <p>Counter: <t t-esc="this.state.stats.nb_new_orders"/></p>
             </DashboardItem>
 
-            <DashboardItem>
+            <DashboardItem size="1">
                 <PieChart data="this.state.stats.orders_by_size"/>
             </DashboardItem>
         </Layout>
--- a/awesome_dashboard/static/src/pie_chart/pie_chart.js
+++ b/awesome_dashboard/static/src/pie_chart/pie_chart.js
@@ -1,0 +1,46 @@
+import { Component, onMounted, onWillStart, onWillUnmount, useRef } from "@odoo/owl"
+import { loadJS } from "@web/core/assets";
+import { getColor } from "@web/core/colors/colors";
+
+
+export class PieChart extends Component {
+    static template = "awesome_dashboard.pie_chart"
+    static props = {
+        data: Object,
+        optional: true
+    }
+
+    setup() {
+        this.state =
+        onWillStart(async () => {
+            await loadJS("/web/static/lib/Chart/Chart.js")
+        })
+
+        this.canvasRef = useRef("canvas");
+
+        if (this.props.data) {
+            onMounted(() => {
+                const keys = Object.keys(this.props.data)
+                const values = Object.values(this.props.data)
+                const colors = keys.map((_, index) => getColor(index));
+
+                this.chart = new Chart(this.canvasRef.el, {
+                    type: "pie",
+                    data: {
+                        labels: keys,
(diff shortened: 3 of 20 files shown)


daloe-odoo:
@Megaaaaaa kind of ready for review

awesome_dashboard/static/src/dashboard.js
```diff
@@ -1,8 +1,36 @@
-import { Component } from "@odoo/owl";
+import { Component, onWillStart, useState } from "@odoo/owl";
```
Megaaaaaa, line 1: Being extra annoying here, this import is not used :smile:

awesome_dashboard/static/src/dashboard.xml
```diff
@@ -2,7 +2,20 @@
 <templates xml:space="preserve">
 
     <t t-name="awesome_dashboard.AwesomeDashboard">
-        hello dashboard
+        <t t-set-slot="layout-buttons">
+            <button t-on-click="openCustomersKanban">Customers</button>
+            <button t-on-click="openLeads">Leads</button>
+        </t>
+
+        <Layout display="{controlPanel: {} }" className="'o_dashboard h-100'">
```
Megaaaaaa, line 10: Its preferable to declare the `layout-buttons` slot inside the `<Layout>` component it belongs to. As written, the buttons are defined before the component that should receive the slot. It doesn't make much a difference but we prefer it that way.

```suggestion
        <Layout display="{ controlPanel: {} }" className="'o_dashboard h-100'">
            <t t-set-slot="layout-buttons">
```

awesome_dashboard/static/src/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,54 @@
+import { Component, onMounted, onPatched, onWillStart, onWillUnmount, useRef, useState } from "@odoo/owl"
```
Megaaaaaa, line 1: Currently import but not used :smile:
```suggestion
import { Component, onMounted, onPatched, onWillStart, onWillUnmount, useRef } from "@odoo/owl"
```

awesome_dashboard/static/src/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,54 @@
...
+import { loadJS } from "@web/core/assets";
+import { getColor } from "@web/core/colors/colors";
+
+
+export class PieChart extends Component {
+    static template = "awesome_dashboard.pie_chart"
+    static props = {
+        data: Object,
+        optional: true
+    }
```
Megaaaaaa, line 11: This declares `data` as a required prop and then adds `optional: true` as a separate prop definition, which is not the shape Owl expects. That matters here because the dashboard statistics are loaded asynchronously, so `orders_by_size` can be missing on the first render. The current version can fail before the data arrives, even though the chart is naturally allowed to start empty. 

```suggestion
        data: {
            type: Object,
            optional: true,
        },
```

awesome_dashboard/static/src/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,54 @@
...
+
+        this.canvasRef = useRef("canvas");
+
+        if (this.props.data) {
+            onMounted(() => {
+                this.renderChart()
+            })
+
+            onPatched(() => {
+                this.chart.destroy()
```
Megaaaaaa, line 26: `onPatched` can run before a chart instance exists, especially sine `data` is optional and the first render may be empty :+1:

```suggestion
                this.chart?.destroy();
```

awesome_dashboard/static/src/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    setup() {
+        onWillStart(async () => {
+            await loadJS("/web/static/lib/Chart/Chart.js")
+        })
+
+        this.canvasRef = useRef("canvas");
+
+        if (this.props.data) {
```
Megaaaaaa, line 20: Oh but actually it won't happend since you only reach that `if (this.props.data)`. I think that's another isue. Your hooks should be assigned anytime and then it's the job of `renderChart` to not do anything if props is empty. Does that make sense ? The lifecycle hooks should be always registered in the setup. :+1:

```suggestion
```

awesome_dashboard/static/src/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,54 @@
...
+
+            onWillUnmount(() => {
+            this.chart.destroy()
+        });
+        }
+    }
+
+    renderChart() {
+        const keys = Object.keys(this.props.data)
+        const values = Object.values(this.props.data)
```
Megaaaaaa, line 38: Because `data` is loaded asynchronously, `this.props.data` can be undefined when the chart first renders. The chart component should handle that calmly and render an empty chart state instead of throwing. This is a useful tutorial lesson: components that receive async service data should be defensive at their boundary. What I would do instead is normalize the prop to an empty object before reading its keys and values. Does that make sense ? :)

```suggestion
        const data = this.props.data || {};
        const keys = Object.keys(data)
        const values = Object.values(data)
```

awesome_owl/static/src/counter/counter.js
```diff
@@ -0,0 +1,21 @@
+import { Component, useState } from "@odoo/owl";
+
+
+export class Counter extends Component {
+    static template = "awesome_owl.counter";
+    static prop = {
```
Megaaaaaa, line 6: :wink:

```suggestion
    static props = {
```

awesome_owl/static/src/todo/todo_item.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+
+    <t t-name="awesome_owl.todo_item">
+        <div class="form-check">
+            <input class="form-check-input" type="checkbox" t-on-change="onChange"/>
+            <label t-att-class="props.todo.isCompleted ? 'text-decoration-line-through text-muted' : '' ">
+                <t t-esc="props.todo.id"/>.
+                <t t-esc="props.todo.description"/>
+                    <span class="fa fa-remove" t-on-click="onRemove"/>
```
Megaaaaaa, line 10: Okay so I already said this to Sreedev so I guess this is the way they show it in the tutorial but being an clickable action, this should be a button not a span. It works great with the mouse but keyboard users and assistive technologies, don't forget to make this kind of action a button in the future :smile:

```suggestion
                    <button type="button" class="btn btn-link p-0" t-on-click="onRemove">
                        <i class="fa fa-remove"/>
                    </button>
```

awesome_owl/static/src/todo/todo_list.js
```diff
@@ -0,0 +1,41 @@
...
+    static components = { TodoItem };
+
+    setup() {
+        this.nextId = 0;
+        this.todos = useState([]);
+        Autofocus("input")
+    }
+
+    addTodo(event) {
+        if (event.key === 'Enter' && event.target.value !== '') {
```
Megaaaaaa, line 17: This only rejects an actually empty string, so a todo containing only spaces can still be added. It's up to you really but I would consider this a small bug in the feature. Your call :smile:
```suggestion
        if (event.key === "Enter" && event.target.value.trim() !== "") {
```

awesome_owl/static/src/todo/todo_list.js
```diff
@@ -0,0 +1,41 @@
...
+            this.todos.push({
+                id: this.nextId++,
+                description: event.target.value,
```
Megaaaaaa, line 20: Then if you validate the trimmed value, you should save the trimmed value. Still your call :smile:
```suggestion
                description: event.target.value.trim(),
```

awesome_owl/static/src/utils.js
```diff
@@ -0,0 +1,10 @@
+import { onMounted, useRef } from "@odoo/owl"
+
+
+export function Autofocus(refName) {
```
Megaaaaaa, line 4: So this calling Owl hooks, it would be prefereable to name it like a hook so we get the intention clearer. This would be more of a `useAutofocus` in this case. I feel like it help makes the difference with a simple utility :+1:

```suggestion
export function useAutofocus(refName) {
```

Megaaaaaa (commented):
Hellooo :wave:

Looks pretty good already, here's your review :+1: My comments are mostly about conventions that I wanted to show you or very minimal improvements to the logic. 

Don't hesitate to tell me if anything is unclear or if you have question :wink:

PR #1305, opened 2026-05-28
maorb - Web Framework Tutorial

Mark-Orban:


awesome_dashboard/static/src/dashboard/dashboard.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+
+    <t t-name="awesome_dashboard.AwesomeDashboard">
+        <t t-set-slot="layout-buttons">
+            <button class="btn btn-primary" t-on-click="openCustomers">Customers</button>
+            <button class="btn btn-primary" t-on-click="openLeads">Leads</button>
+        </t>
+        <Layout className="'o_dashboard h-100'">
```
Megaaaaaa: Any slot meant for a component is usually declared inside the component. As this is not the first time I see this, I suppose this is how they tell you to do it in the tutorial but in my experience, we will always put it **inside** the component. Just so you know :+1:

```suggestion
        <Layout className="'o_dashboard h-100'">
            <t t-set-slot="layout-buttons">
```

awesome_dashboard/static/src/dashboard/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,42 @@
+import { Component, onMounted, onPatched, onWillStart, onWillUnmount, useRef } from "@odoo/owl"
+import { loadJS } from "@web/core/assets";
+
+export class PieChart extends Component {
+    static template = "awesome_dashboard.PieChart";
+    static props = {
+        title: String,
+        data: Object,
```
Megaaaaaa: The chart data comes from a statistics service that is filled asynchronously, so it is normal for `statistics.orders_by_size` to be missing on the first render. Declaring `data` as required makes the child component stricter than the parent data flow can guarantee. Let's make it optional for safety measures :+1:

```suggestion
        data: {
            type: Object,
            optional: true,
        },
```

awesome_dashboard/static/src/dashboard/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,42 @@
...
+    };
+
+    setup() {
+        this.canvasRef = useRef("canvas");
+        onWillStart(() => loadJS("/web/static/lib/Chart/Chart.js"));
+        onMounted(() => {
+            this.renderChart();
+        });
+        onPatched(() => {
+            this.chart.destroy();
```
Megaaaaaa: Don't overlook that `onPatched` can run before a chart exists, especially while the async statistics are still loading. Leaving this unguarded accepts a runtime error path during a normal loading state.
```suggestion
            this.chart?.destroy();
```

awesome_dashboard/static/src/dashboard/pie_chart/pie_chart.js
```diff
@@ -0,0 +1,42 @@
...
+            this.renderChart();
+        });
+        onWillUnmount(() => {
+            this.chart.destroy();
+        });
+    }
+
+    renderChart() {
+        const labels = Object.keys(this.props.data);
+        const data = Object.values(this.props.data);
```
Megaaaaaa: Because the dashboard statistics are loaded asynchronously, `this.props.data` can be undefined when the component first renders and calling `Object.keys` on undefined would crash. Just normalize the props and we're good to go :+1:

```suggestion
        const chartData = this.props.data || {};
        const labels = Object.keys(chartData);
        const values = Object.values(chartData);
```

awesome_owl/static/src/playground.js
```diff
@@ -1,5 +1,17 @@
-import { Component } from "@odoo/owl";
+import { Component, markup, useState } from "@odoo/owl";
```
Megaaaaaa: unused import :wink:
```suggestion
import { Component, useState } from "@odoo/owl";
```

awesome_owl/static/src/playground.xml
```diff
@@ -2,8 +2,15 @@
 <templates xml:space="preserve">
 
     <t t-name="awesome_owl.playground">
-        <div class="p-3">
+        <div>
             hello world
+            <Counter onChange="() => this.incrementSum()"/>
```
Megaaaaaa: Yo ucan use Owl's `.bind` syntax for passing component methods to children. It works just as good but looks a bit better and also, it's the odo oway :smile:

```suggestion
            <Counter onChange.bind="incrementSum"/>
```

awesome_owl/static/src/todo_list/todo_item.xml
```diff
@@ -0,0 +1,17 @@
...
+    <t t-name="awesome_owl.TodoItem">
+        <div class="form-check">
+            <input type="checkbox" t-att-id="props.todo.id" t-att-checked="props.todo.isCompleted" t-on-change="onChange"/>
+            <span t-if="props.todo.isCompleted" class="text-muted text-decoration-line-through">
+                <t t-out="props.todo.id"/>. <t t-out="props.todo.description"/>
+            </span>
+            <span t-else="">
+                <t t-out="props.todo.id"/>. <t t-out="props.todo.description"/>
+            </span>
+            <span class="fa fa-remove" t-on-click="onRemove"/>
```
Megaaaaaa: I suppose this is the way the tutorial tells you to do it as I've seen it multiple times but for your information, when we make an action clickable, we really want to put it in a <button> because it works just as good with the mouse but helps a lot with the keyboard usage and also some assistive controls. Don't make much a difference here but now you know :+1:

```suggestion
            <button type="button" class="btn btn-link p-0" t-on-click="onRemove">
                <i class="fa fa-remove"/>
            </button>
```

awesome_owl/static/src/todo_list/todo_list.js
```diff
@@ -0,0 +1,41 @@
...
+        this.nextId = 1;
+        this.todos = useState([]);
+        this.inputRef = useRef("input");
+        onMounted(() => {
+            this.inputRef.el.focus();
+        });
+    }
+
+    addTodo(event) {
+        if (event.keyCode === 13 && event.target.value != "") {
```
Megaaaaaa: We prefer using the key name instead of the key code as it is much more self explanatory and we like self explanatory. "When a user clicks 'Enter' then this happens." :)

Also I would have trimmed the input to make sure we do no accept only spaces as an input but that's a preference thing. Altough if you decide to do that, then you need to save `event.target.value.trim()` instead of `event.target.value` right below. :+1:

```suggestion
        if (event.key === "Enter" && event.target.value.trim() !== "") {
```

Megaaaaaa (commented):
Hello hello :wave:

Here's a review for you! Looks pretty good, it's only a few comments about stuff that we tend to do another way and that I thought it'd be best to show you. :+1:

Tell me if anything is unclear or if you have any question!

PR #1307, opened 2026-05-29
[ADD] awesome_owl, awesome_dashboard: basic client-side modules for todos and t-shirt sales dashboard

sidum-odoo:
Problem:
- No existing module for handling simple todo list.
- No existing module for analysing t-shirt sales.

Solution:
- New module (100% client-side) with basic components such as todo list and counter.
- New module (100% client-side) implementing a dashboard for t-shirt sales analysis.

Task-6241079

### DIFF
--- a/awesome_dashboard/static/src/dashboard/dashboard.js
+++ b/awesome_dashboard/static/src/dashboard/dashboard.js
@@ -1,0 +1,52 @@
+import { Component, useState } from "@odoo/owl";
+import { _t } from "@web/core/l10n/translation"
+import { Layout } from "@web/search/layout"
+import { registry } from "@web/core/registry";
+import { useService } from "@web/core/utils/hooks";
+import { DashboardItem } from "./dashboard_item";
+import { PieChart } from "./pie_chart";
+import { DashboardConfigDialog } from "./dashboard_config_dialog";
+
+class AwesomeDashboard extends Component {
+    static template = "awesome_dashboard.AwesomeDashboard";
+
+    static components = { Layout, DashboardItem, PieChart, DashboardConfigDialog };
+
+    setup() {
+        this.action = useService("action");
+        this.statisticsService = useService("awesome_dashboard.statistics")
+        this.dialog = useService("dialog");
+        this.statistics = useState(this.statisticsService);
+        this.configState = useState({
+            disabledItems: JSON.parse(localStorage.getItem("disabled_dashboard_items") || "[]")
+        });
+    }
+
+    get items() {
+        const allItems = registry.category("awesome_dashboard").getAll();
+        return allItems.filter(item => !this.configState.disabledItems.includes(item.id));
+    }
+
+    openConfiguration() {
+        this.dialog.add(DashboardConfigDialog, {
+            onConfigSaved: () => {
+                this.configState.disabledItems = JSON.parse(localStorage.getItem("disabled_dashboard_items") || "[]");
+            }
+        });
+    }
+
+    openCustomers() {
+        this.action.doAction("base.action_partner_form");
+    }
+
+    openLeads() {
+        this.action.doAction({
+            type: "ir.actions.act_window",
+            name: _t("Leads"),
+            res_model: "crm.lead",
+            views: [[false, "list"], [false, "form"]],
+        });
+    }
+}
+
+registry.category("lazy_components").add("AwesomeDashboard", AwesomeDashboard);
--- a/awesome_dashboard/static/src/dashboard/dashboard_config_dialog.xml
+++ b/awesome_dashboard/static/src/dashboard/dashboard_config_dialog.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates xml:space="preserve">
+    <t t-name="awesome_dashboard.DashboardConfigDialog">
+        <Dialog title="'Dashboard items configuration'">
+            <div class="p-3">
+                <p class="text-muted">Which cards do you wish to see?</p>
+                <div t-foreach="allItems" t-as="item" t-key="item.id" class="form-check my-2">
+                    <input 
+                        class="form-check-input" 
+                        type="checkbox" 
+                        t-model="state[item.id]" 
+                        t-att-id="'check_' + item.id"
+                    />
+                    <label class="form-check-label" t-att-for="'check_' + item.id" t-out="item.description"/>
+                </div>
+            </div>
+            <t t-set-slot="footer">
+                <button class="btn btn-primary" t-on-click="onApply">Apply</button>
+                <button class="btn btn-secondary" t-on-click="props.close">Cancel</button>
+            </t>
+        </Dialog>
+    </t>
+</templates>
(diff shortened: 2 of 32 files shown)


sidum-odoo:
@Megaaaaaa ready for review

awesome_dashboard/static/src/dashboard/dashboard.js
```diff
@@ -0,0 +1,51 @@
+import { Component, onWillStart, useState } from "@odoo/owl";
```
Megaaaaaa: Imported but unused :wink:

```suggestion
import { Component, useState } from "@odoo/owl";
```

awesome_dashboard/static/src/dashboard/dashboard.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    openCustomers() {
+        this.action.doAction("base.action_partner_form");
+    }
+
+    openLeads() {
+        this.action.doAction({
+            type: "ir.actions.act_window",
+            name: "Leads",
```
Megaaaaaa: This action label is user-facing, so it should go through `_t`, the JS translation method (import { _t } from "@web/core/l10n/translation"). 

```suggestion
            name: _t("Leads"),
```

awesome_dashboard/static/src/dashboard/dashboard_config_dialog.xml
```diff
@@ -0,0 +1,23 @@
...
+            <div class="p-3">
+                <p class="text-muted">Which cards do you wish to see?</p>
+                <div t-foreach="allItems" t-as="item" t-key="item.id" class="form-check my-2">
+                    <input 
+                        class="form-check-input" 
+                        type="checkbox" 
+                        t-model="state[item.id]" 
+                        t-att-id="'check_' + item.id"
+                    />
+                    <label class="form-check-label" t-att-id="'check_' + item.id" t-out="item.description"/>
```
Megaaaaaa: Currently  you have duplicated ids. It looks like its works but it's not the way 😅 No big deal thoug, just assign the `for` attribute of the label and you'll be good to go :+1:

```suggestion
                    <label class="form-check-label" t-att-for="'check_' + item.id" t-out="item.description"/>
```

awesome_dashboard/static/src/dashboard/number_card.js
```diff
@@ -0,0 +1,11 @@
+import { Component, xml } from "@odoo/owl";
+
+export class NumberCard extends Component {
+    static template = xml`
```
Megaaaaaa, line 9: We always make our props explicit. Or at least as much as possible. It helps everyone reading the component and debugging. It's probably the only piece of documentation we try to write on every single component :smile:

```suggestion
    static props = {
        title: String,
        value: { optional: true },
    };
    static template = xml`
```

awesome_dashboard/static/src/dashboard/pie_chart.js
```diff
@@ -0,0 +1,47 @@
+import { Component, onWillStart, useRef, onMounted, useEffect } from "@odoo/owl";
+import { loadJS } from "@web/core/assets";
+
+export class PieChart extends Component {
+    static template = "awesome_dashboard.PieChart";
```
Megaaaaaa, line 9: Same here, let's make the props explicit :+1:

awesome_dashboard/static/src/dashboard/pie_chart.js
```diff
@@ -0,0 +1,47 @@
...
+            this.renderChart();
+        }, () => [this.props.data]);
+    }
+
+    renderChart() {
+        if (!this.props.data) {
+            return; 
+        }
+        if (this.chart) {
+            this.chart.destroy();
```
Megaaaaaa, line 33: The chart is destroyed before re-rendering, which is good, but it is never destroyed when the component is unmounted. Chart.js keeps an object attached to the canvas, so cleaning it up teaches the right lifecycle habit for external libraries. Leaving it as-is accepts a small memory/resource leak when the dashboard is left or reloaded. What I would do instead is import `onWillUnmount` and destroy the chart there to clean it up a bit. :+1:

```py
# awesome_dashboard/static/src/dashboard/pie_chart.js (setup): destroy the Chart.js instance when Owl removes the component.
        onWillUnmount(() => {
            this.chart?.destroy();
        });
```

awesome_dashboard/static/src/dashboard/pie_chart_card.js
```diff
@@ -0,0 +1,13 @@
+import { Component, xml } from "@odoo/owl";
+import { PieChart } from "./pie_chart";
+
+export class PieChartCard extends Component {
+    static components = { PieChart };
```
Megaaaaaa, line 9: `props` :smile:

awesome_owl/static/src/todo/todoitem.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+    <t t-name="awesome_owl.TodoItem">
+        <div class="list-group-item d-flex gap-3 align-items-center justify-content-between p-2" t-att-class="{'text-muted text-decoration-line-through': props.todo.isCompleted}">
+            <div class="d-flex align-items-center flex-grow-1 gap-3">
+                <input type="checkbox" t-att-checked="props.todo.isCompleted" t-on-change="() => props.toggleState(props.todo.id)"/>
+                <t t-esc="props.todo.id"/>. <t t-esc="props.todo.description"/>
+            </div>
+            <span class="fa fa-remove" t-on-click="() => props.deleteTodo(props.todo.id)"/>
```
Megaaaaaa: This is a clickable action, so it should be a button rather than a span. From seeing everyone do it, I understand this is the way the tutorial tells you to do but it's prefereable to use `<button>`. The current version works but it is less clear and weaker for keyboard users and assistive technologies. Just keep in mind: Callable action -> Button.

```suggestion
            <button type="button" class="btn btn-link p-0" t-on-click="() => props.deleteTodo(props.todo.id)">
                <i class="fa fa-remove"/>
            </button>
```

awesome_owl/static/src/todo/todolist.js
```diff
@@ -0,0 +1,40 @@
...
+    static components = { TodoItem };
+
+    setup() {
+        this.todos = useState([]);
+        this.nextId = 1;
+        this.inputRef = useAutoFocus('input');
+    }
+
+    addTodo(ev) {
+        if (ev.keyCode == 13) {
```
Megaaaaaa: And this too comes from the tutorial I would guess but we try to use the key names a lot more than the key codes to get a more self explanatory code. In opposition to `ev.keyCode == 13`, here, you understand instantaneously that this triggers on "Enter" input :+1:

```suggestion
        if (ev.key === "Enter") {
```

Megaaaaaa (commented):
Hello hello :wave:

Nice pr already, here 's a review for you :+1: These are mostly conventions that I wanted to show you but there is also a lot of what we usually call "useless diff" in reviews like changing quotes to double quotes or the other way around or removing or adding an empty line where nothing else changes. There are also a few EOL that you're missing but I guess the ci/style already told you that :smile:

Don't hesitate to tell me if anything is unclear or if you have any question :+1:

PR #1312, opened 2026-06-01
[ADD] Zero_quantity_account_move: added zero quantity for invoice lines

moahi-odoo:
Added a boolean field 'zero_move'(Is Zero Quantity) on account move line.

For rejected or scrap goods, quantity should be sent as 0 even if a credit note is created.

* Added zero_move field in account.move.line
* Overridden EDI method to send quantity as 0 when ticked
* Added test cases for validation

### DIFF
--- a/zero_quantity_account_move/__init__.py
+++ b/zero_quantity_account_move/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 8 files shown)


moahi-odoo:
qwertyu

moahi-odoo:
mohit

PR #1333, opened 2026-06-05
[ADD] estate: initialize estate module

vrushibh:
Description
This PR implements the estate and estate_account modules for Odoo 19 (Chapters 1 to 14).

Changes Made:
- Models & Relations: Created property, type, tag, offer models, and extended res.users with relational fields.
- Computes & Constraints: Added @api.depends for total_area and best_price, @api.onchange for garden defaults, and price validation constraints.
- Business Logic: Implemented offer processing (action_accept, action_refuse), property state buttons (action_sold, action_cancel), and batch write updates.
- Accounting Integration: Created estate_account module to automatically generate customer invoices (account.move) upon property sale.
- Views & Security: Created XML views (Form, List, Search) with statusbars, filters, and defined security rules in ir.model.access.csv.

Chapter Reference:
Odoo 19 Tutorial (Chapters 1 to 14)


estate/models/estate_property.py
```diff
@@ -0,0 +1,45 @@
...
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        selection=[
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
+        ],
```
mash-odoo, line 34:
```suggestion
        selection=[
            ('north', "North"),
            ('south', "South"),
            ('east', "East"),
            ('west', "West"),
        ],
```
Keep technical keys in single quotes and the value or any string which would be visible on UI in double quotes

estate/security/estate_security.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
```
mash-odoo: Why do we write this tag here?

estate/security/estate_security.xml
```diff
@@ -0,0 +1,29 @@
...
+
+        <record id="module_category_real_estate" model="ir.module.category">
+            <field name="name">Real Estate</field>
+            <field name="sequence">10</field>
```
mash-odoo: What's the purpose of this sequence field?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,97 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <data>
+        <record id="action_estate_property" model="ir.actions.act_window">
+            <field name="name">Properties</field>
+            <field name="res_model">estate.property</field>
+            <field name="view_mode">list,form,kanban</field>
+            <field name="context">{'search_default_available': 1}</field>
```
mash-odoo: What's the difference between domain and context?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,97 @@
...
+            <field name="arch" type="xml">
+                <search string="Properties">
+                    <field name="name" string="Title" />
+                    <field name="postcode" />
+                    <field name="expected_price" />
+                    <field name="bedrooms" />
+                    <field name="living_area" />
+                    <field name="facades" />
+                    <separator />
+                    <filter name="available" string="Available" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]" />
```
mash-odoo: Can you optimize this domain?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,97 @@
...
+            <field name="model">estate.property</field>
+            <field name="arch" type="xml">
+                <search string="Properties">
+                    <field name="name" string="Title" />
+                    <field name="postcode" />
+                    <field name="expected_price" />
+                    <field name="bedrooms" />
+                    <field name="living_area" />
+                    <field name="facades" />
+                    <separator />
```
mash-odoo: What's the working of a separator?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,97 @@
...
+                    <field name="bedrooms" />
+                    <field name="living_area" />
+                    <field name="expected_price" />
+                    <field name="selling_price" />
+                    <field name="date_availability" />
+                </list>
+            </field>
+        </record>
+
+        <record id="view_estate_property_form" model="ir.ui.view">
```
mash-odoo: Follow [this](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html?highlight=coding%20guideli#security-view-and-action) for naming your views
```suggestion
        <record id="estate_property_view_form" model="ir.ui.view">
```
`<model_name>_view_<view_type>`, where `view_type` is kanban, form, list, search,...

estate/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "Real Estate",
+    "author": "Odoo S.A.",
+    "license": "LGPL-3",
+    "depends": ["base"],
```
mash-odoo, line 5: What if we don't write `base` here?

mash-odoo (changes requested):
Hello!
Good start on the task..
Here are a few comments...Have a look!
Also, mention the chapter number in the commits.

mash-odoo (changes requested):
Hello!
Good start on the task..
Here are a few comments...Have a look!

estate/models/estate_property.py
vrushibh: Correct. I have updated all selection fields (both garden_orientation and state) to use single quotes for technical keys and double quotes for their UI labels.

estate/security/estate_security.xml
vrushibh: The `<data>` tag was not needed because we were not using attributes like noupdate="1". So, I removed the unnecessary <data> tags from the security and view files and kept only the <odoo> root tag.

estate/security/estate_security.xml
vrushibh: The sequence field decides the order of the categories. A smaller number shows the category first.

estate/views/estate_property_views.xml
vrushibh: -Domain is used to filter records at the database level (like a SQL WHERE clause). It decides "which" records to display.

-Context is used to pass information to the view, such as grouping records (like {'group_by': 'postcode'} ) or setting default values for new records.

estate/views/estate_property_views.xml
vrushibh: Yes, I have optimized it to:
domain= [('state', 'in', ('new', 'offer_received'))]

It is better because it is much easier to read and makes it simple to add more states in the future.

estate/views/estate_property_views.xml
vrushibh: The <separator/> tag shows a vertical line between filters in the UI. 
Logically, it joins filters with an AND operator instead of OR,

 I have practiced this in our search view (estate_property_view_search) by using <separator/> to split the State filters and Feature filters (Garage/Garden).

estate/views/estate_property_views.xml
vrushibh: Done. I have renamed all view IDs to follow the <model_name>_view_<view_type> format (e.g., estate_property_view_form).

estate/__manifest__.py
vrushibh: If we don't write base, Odoo will still load it automatically. However, writing it explicitly is a best practice to make the module dependency clear.

estate/models/estate_property.py
```diff
@@ -0,0 +1,132 @@
...
+        else:
+            self.garden_area = 0
+            self.garden_orientation = False
+
+    def action_cancel(self):
+        for record in self:
+            if record.state == "sold":
+                raise UserError("A sold property cannot be cancelled.")
+            record.state = "cancelled"
+        return True
```
mash-odoo: Here the function is just for a single record so there is need for a loop. You can directly use self.ensure_one().

estate/models/estate_property.py
```diff
@@ -0,0 +1,132 @@
...
+                )
+            record.state = "sold"
+        return True
+
+    def action_accept_best_offer(self):
+        for record in self:
+            if not record.offer_ids:
+                raise UserError("This property has no offers to accept.")
+            best_offer = max(record.offer_ids, key=lambda o: o.price)
+            best_offer.action_accept()
```
mash-odoo: What if there are multiple functions with the same name like `action_accept`.
How do you think this is handled by Odoo?

estate/models/estate_property_maintenance.py
```diff
@@ -0,0 +1,68 @@
+from odoo import fields, models
+from odoo.exceptions import UserError
+
+
+class EstatePropertyMaintenance(models.Model):
+    _name = "estate.property.maintenance"
+    _description = "Property Maintenance"
+
+    name = fields.Char(required=True, string="Title", translate=True)
```
mash-odoo, line 9: What will be affected by adding this `translate=True`?

estate/models/estate_property_maintenance.py
```diff
@@ -0,0 +1,68 @@
...
+        for record in self:
+            if not record.technician_id:
+                raise UserError("Please select a technician first.")
+            record.state = "assigned"
+        return True
+
+    def action_start(self):
+        for record in self:
+            record.state = "started"
+        return True
```
mash-odoo, line 72: Do we need to have for loops here?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+class EstatePropertyOffer(models.Model):
+    _name = "estate.property.offer"
+    _description = "Real Estate Property Offer"
+
+    price = fields.Float(required=True)
+    status = fields.Selection(
+        selection=[
+            ('accepted', 'Accepted'),
+            ('rejected', 'Rejected'),
+        ],
```
mash-odoo, line 21:
```suggestion
        selection=[
            ('accepted', "Accepted"),
            ('rejected', "Rejected"),
        ],
```

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,19 @@
+from odoo import api, fields, models
+
+
+class EstatePropertyType(models.Model):
+    _name = "estate.property.type"
+    _description = "Real Estate Property Type"
+    _rec_names_search = ["name", "code"]
```
mash-odoo, line 7: What's `_rec_names_search` used for?

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,21 @@
+from random import randint
+
+from odoo import fields, models
+
+
+class EstatePropertyTag(models.Model):
+    _name = "estate.property.tag"
+    _description = "Real Estate Property Tag"
+    _rec_name = 'name'
```
mash-odoo, line 9: Does it make difference with/without writing this?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,161 @@
...
+    </record>
+
+    <record id="estate_property_view_form" model="ir.ui.view">
+        <field name="name">estate.property.view.form</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form string="Properties">
+                <header>
+                    <button name="action_accept_best_offer" type="object" string="Accept Best Offer" class="oe_highlight" invisible="state in ('sold', 'cancelled', 'offer_accepted')"/>
+                    <button name="action_sold" type="object" string="Sold" invisible="state in ('sold', 'cancelled')"/>
```
mash-odoo, line 51: What's the difference between `type="object"` and `type="action"`?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,161 @@
...
+                <sheet>
+                    <div class="oe_title">
+                        <h1>
+                            <field name="name" placeholder="Property Name"/>
+                        </h1>
+
+                    </div>
+                    <group>
+                        <group>
+                            <field name="tag_ids" widget="many2many_tags" options="{'color_field': 'color'}" class="o_color_only_tags"/>
```
mash-odoo, line 74: Do you know where do these attributes come from?
What if you want to create a generic attribute which can be used anywhere in Odoo..do you have any idea regarding this?

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,44 @@
...
+        <field name="model">estate.property.offer</field>
+        <field name="arch" type="xml">
+            <list editable="bottom">
+                <field name="property_id"/>
+                <field name="partner_id"/>
+                <field name="price"/>
+                <field name="validity"/>
+                <field name="date_deadline"/>
+                <field name="status"/>
+                <field name="create_date"/>
```
mash-odoo, line 23: You are using this `create_date` field, but it's not declared anywhere in python.
So how is this rendered?

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,43 @@
...
+<odoo>
+
+    <record id="estate_property_type_action" model="ir.actions.act_window">
+        <field name="name">Property Types</field>
+        <field name="res_model">estate.property.type</field>
+        <field name="view_mode">list,form</field>
+        <field name="help" type="html">
+            <p class="o_view_nocontent_smiling_face">
+                Create your first property type!
+            </p>
```
mash-odoo, line 11: Can you explain why we need to have this?

estate/views/estate_property_maintenance_views.xml
```diff
@@ -0,0 +1,71 @@
...
+                        </group>
+                    </group>
+                    <group string="Description">
+                        <field name="description" placeholder="Describe the maintenance issue..."/>
+                    </group>
+                </sheet>
+            </form>
+        </field>
+    </record>
+</odoo>
```
mash-odoo: Leave an extra line at the end of the file

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,73 @@
...
+    @api.onchange("date_deadline")
+    def _onchange_date_deadline(self):
+        self._inverse_date_deadline()
+
+    def action_accept(self):
+        for record in self:
+            if record.property_id.state in ("sold", "cancelled"):
+                raise UserError("You cannot accept an offer for a sold or cancelled property.")
+            if any(offer.status == "accepted" for offer in record.property_id.offer_ids):
+                raise UserError("An offer has already been accepted for this property.")
```
mash-odoo: Will this error ever be raised? WDYT?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,73 @@
...
+            record.status = "accepted"
+            record.property_id.buyer_id = record.partner_id
+            record.property_id.selling_price = record.price
+            record.property_id.state = "offer_accepted"
+            other_offers = record.property_id.offer_ids - record
```
mash-odoo: You can optimize this

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,73 @@
...
+            other_offers.write({'status': 'rejected'})
```
mash-odoo: What's the difference between these two?
1. other_offers.write({'status': 'rejected'})
2. other_offers.status = 'rejected'

mash-odoo (changes requested):
Hello,
Good progress on the task.
I have some suggestions..
1. I think the price should be made mandatory [here](https://www.awesomescreenshot.com/video/54590261?key=684a93e7ecf66b3d6d2d321ecaffa4f2) because any offer without price doesn't make sense. WDYT?
2.Write a proper description for your PR
3. For your commits, currently it's fine not to have that entire format, you can just write in your words what you did and include the chapter number.

vrushibh:
> 1. I think the price should be made mandatory [here](https://www.awesomescreenshot.com/video/54590261?key=684a93e7ecf66b3d6d2d321ecaffa4f2) because any offer without price doesn't make sense. WDYT?

> ->  Agreed! `fields.Float` in Odoo maps to PostgreSQL `double precision`, which defaults to `0.0` (or `0.00`). Because of this, `required=True` alone would still allow saving an offer with `0.00`.
> -> To solve this completely, I have set `required=True` along with the `models.Constraint("CHECK(price > 0)")` constraint.   This ensures that zero (`0.00`) and negative values trigger a validation error upon save.

>2.Write a proper description for your PR

>-> I have also updated the PR description with the summary of changes and chapter number.

estate/models/estate_property.py
vrushibh: Thanks for pointing this out! Since button actions in form views operate on single records, I have replaced the `for` loops with `self.ensure_one()`.

estate/models/estate_property.py
vrushibh: Since `best_offer` is a record of `estate.property.offer`, Python specifically calls the `action_accept` method of this model.

estate/models/estate_property_maintenance.py
vrushibh: Adding `translate=True` enables multi-language support (i18n) for this field. Odoo stores translations for user-facing values per language. Users logged in with different languages can see and input translated text independently for the field.

estate/models/estate_property_maintenance.py
vrushibh: Thanks for pointing this out! Since button actions in form views operate on single records, I have replaced the `for` loops with `self.ensure_one()`.

estate/models/estate_property_offer.py
vrushibh: `ruff format` changed the quotes automatically, but I have fixed it back to double quotes following Odoo guidelines.

estate/models/estate_property_type.py
vrushibh: `_rec_names_search` specifies which fields Odoo searches against during a name search or Many2one dropdown autocomplete. Adding `["name", "code"]` allows users to search property types by both name and code in search dropdowns.

estate/models/estate_property_tag.py
vrushibh: No difference. Odoo uses the `name` field as `_rec_name` by default. We only need to write `_rec_name` if we want to use a different field (like `_rec_name = 'code'`).

estate/views/estate_property_views.xml
vrushibh: - `type="object"` executes a Python method on the model matching the button's `name` attribute.
- `type="action"` triggers an Odoo action record (like an `ir.actions.act_window` XML ID) matching the `name` attribute.

estate/views/estate_property_views.xml
vrushibh: These attributes and widgets are defined in Odoo's core `web` addon (under `web/static/src/views/fields/`).

To create a generic widget or attribute that can be used anywhere in Odoo:
1. Create a custom OWL JavaScript component extending standard field props.
2. Register it globally in the field registry:
   `registry.category("fields").add("my_custom_widget", MyCustomWidget);`

Once registered, `widget="my_custom_widget"` can be used generically across any XML view in Odoo.

estate/views/estate_property_offer_views.xml
vrushibh: `create_date` is a built-in magic field automatically created by Odoo ORM on all models. Therefore, it works in XML views without needing python declaration.

estate/views/estate_property_type_views.xml
vrushibh: This provides an empty-state helper message when there are no records in the view, guiding the user to create their first record.

estate/views/estate_property_maintenance_views.xml
vrushibh: Done! Added an extra line at the end.

estate/models/estate_property_offer.py
vrushibh: I simplified this by checking `if self.property_id.buyer_id:` instead of looping over `offer_ids`.

estate/models/estate_property_offer.py
vrushibh: Optimized by combining the updates into a single `property_id.write({...})` call.

estate/models/estate_property_offer.py
vrushibh: `.write({})` calls the ORM write method using a dictionary (useful for updating multiple fields at once), while `.status = 'rejected'` uses direct field assignment, which is cleaner for updating a single field.

PR #1339, opened 2026-06-15
Server framework 101

LucasDaniOV:


estate/__manifest__.py
```diff
@@ -0,0 +1,14 @@
...
+    'depends': [
+        'base_setup'
+    ],
+    'installable': True,
+    'application': True,
+    'data': [
+        'security/ir.model.access.csv',
+
+        'views/estate_property_views.xml',
+        'views/estate_menus.xml'
```
ziriraha:
```suggestion
        'views/estate_menus.xml',
```
We usually like to add the comma at the end, so that one day if you add a new item to the list, we dont create a diff in the last line (btw ruff suggests this)

estate/models/estate_property.py
```diff
@@ -0,0 +1,39 @@
+from odoo import fields, models
+from dateutil.relativedelta import relativedelta
+
```
ziriraha, line 2:
```suggestion


```
two empty lines on top of class definitions

ziriraha (changes requested):
looks good, one thing to keep in mind is that you should add a new link at the end of each file. i recommend setting up your editor to add it for you.

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+        else:
+            self.garden_area = 0
+            self.garden_orientation = False
+
+    """
+    Button functions
+    """
+
+    def sell_property(self):
+        for record in self:
```
ziriraha:
```suggestion
        for property in self:
```
if we name it record it makes it hard to know what we're working with.

ziriraha (changes requested):
nice work

LucasDaniOV:
> nice work

thank you

estate/models/estate_property.py
```diff
@@ -0,0 +1,171 @@
...
+    def sell_property(self):
+        for property in self:
+            if property.state == "cancelled":
+                msg = "Cancelled properties cannot be sold"
+                raise exceptions.UserError(msg)
+            if property.state in ["offer_accepted", "sold"]:
+                msg_0 = "Property is already sold"
+                raise exceptions.UserError(msg_0)
+            property.state = "sold"
+        return True
```
ziriraha, line 104: fyi (i dont know if its explained in the tutorial) actions can return action objects so you can say things like: open this popup, reload the page, redirect to here

estate_account/security/ir.model.access.csv
ziriraha, line 1: useless file

estate/models/estate_property.py
```diff
@@ -0,0 +1,171 @@
...
+            ("sold", "Sold"),
+            ("cancelled", "Cancelled"),
+        ],
+        copy=False,
+        default="new",
+    )
+
+    """
+    Relational fields
+    """
```
ziriraha: I personally dont like these separation comments, some people add them but then others when changing the code they put everything wherever they want and they become useless

nitpick: Also I prefer # instead of """ for multiline comments

estate/models/estate_property.py
```diff
@@ -0,0 +1,171 @@
...
+        "CHECK(garden_area >= 0)",
+        "The garden area should be at least 0",
+    )
+
+    @api.constrains("selling_price")
+    def _check_selling_price(self):
+        for property in self:
+            if property.selling_price < 0.9 * property.expected_price:
+                msg = "Selling price cannot be lower than ninety percent of expected price"
+                raise exceptions.ValidationError(msg)
```
ziriraha:
```suggestion
                raise exceptions.ValidationError(self.env._("Selling price cannot be lower than ninety percent of expected price"))
```
You can inline it and you should make it translatable

Another nitpick: but i prefer if you import `from odoo.exceptions import ValidationError` instead of using it this way

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,80 @@
...
+        return True
+
+    """
+    Method extensions
+    """
+
+    @api.model
+    def create(self, vals_list):
+        for vals in vals_list:
+            property = self.env["estate.property"].browse(vals["property_id"])
```
ziriraha: this is going to explode query count, you should do a browse outside the for loop and the use that recordset to browse it (again) for the correct id, something like:
```py
properties = self.env["estate.property"].browse([vals["property_id"] for vals in vals_list])

for vals in vals_list:
    property = properties.browse(vals["property_id"])
```

The first call saves them on cache, the second one just retrieves the one needed

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,32 @@
...
+        <field name="name">estate.property.tag.list</field>
+        <field name="model">estate.property.tag</field>
+        <field name="arch" type="xml">
+            <list string="Tags" editable="bottom">
+                <field name="name"/>
+                <field name="color" widget="color_picker"/>
+            </list>
+        </field>
+    </record>
+</odoo>
```
ziriraha: new line eof

estate_account/__manifest__.py
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "Real Estate Account",
+    "depends": ["base_setup", "estate", "account"],
+    "installable": True,
+    "application": True,
+    "data": [
+        "security/ir.model.access.csv",
+    ],
```
ziriraha:
```suggestion
```
not really as its a module that will connect both apps (not an app itself), in any case we would put `auto_install: True` so it automatically installs when you have estate and account apps.

estate_account/__manifest__.py
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "Real Estate Account",
+    "depends": ["base_setup", "estate", "account"],
```
ziriraha:
```suggestion
    "depends": ["estate", "account"],
```
`base_setup` is implied from the other modules

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "Real Estate",
+    "depends": ["base_setup"],
```
ziriraha:
```suggestion
    "depends": ["base"],
```
You dont need base_setup here

ziriraha (changes requested):
Good job, a couple small comments

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate Property Model"
+    _order = "id desc"
+
+    name = fields.Char("Title", required=True, translate=True)
+    description = fields.Text(string="Description", required=True, translate=True)
+    postcode = fields.Char("Postcode", required=True, translate=False)
```
ziriraha: no need to specify translate=False

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate Property Model"
+    _order = "id desc"
+
+    name = fields.Char("Title", required=True, translate=True)
```
ziriraha, line 12: translate=True will translate the value of the field, not the field name if thats what you want. field strings are always translated. putting translate=True in a model that will not have xml data is not useful (afaik)

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+    description = fields.Text(string="Description", required=True, translate=True)
+    postcode = fields.Char("Postcode", required=True, translate=False)
+    expected_price = fields.Float("Expected Price", required=True)
+    selling_price = fields.Float("Selling Price", readonly=True, copy=False)
+    bedrooms = fields.Integer("Bedrooms", required=True, default=2)
+    living_area = fields.Integer("Living Area (sqm)", required=True)
+    facades = fields.Integer("Facades", required=True)
+    garage = fields.Boolean("Garage", required=True)
+    garden = fields.Boolean("Garden", required=True)
+    garden_area = fields.Integer("Garden Area", required=False)
```
ziriraha: no need for required false

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,79 @@
...
+        return True
+
+    @api.model
+    def create(self, vals_list):
+        property_ids = [vals["property_id"] for vals in vals_list]
+        properties = self.env["estate.property"].browse(property_ids)
+        properties_dict = {p.id: p for p in properties}
+
+        for vals in vals_list:
+            property = properties_dict.get(vals["property_id"])
```
ziriraha:
```suggestion
            property = properties.browse(vals["property_id"])
```

thats enough, no need for making a dict

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,33 @@
...
+                {
+                    "partner_id": property.buyer_id.id,
+                    "move_type": "out_invoice",
+                    "invoice_line_ids": [
+                        Command.create(
+                            {
+                                "name": "Commission",
+                                "price_unit": property.selling_price
+                                if property.selling_price
+                                else property.expected_price,
```
ziriraha:
```suggestion
                                "price_unit": property.selling_price or property.expected_price,
```

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,33 @@
...
+                            {
+                                "name": "Admin fee",
+                                "price_unit": 100,
+                                "quantity": 1,
+                            },
+                        ),
+                    ],
+                },
+            )
+        return super().sell_property()
```
ziriraha: In sell_property you have some check that raise errors. In this case, you are creating the account.move and then you are performing the checks. you should place the super() call above the creation of the account move, as there might a case where you create a move and then the sell errors out for some reason.

We usually do:
```py
def func():
    res = super().func()
    ....
    return res
```

Dont take this as a rule of thumb, there are cases where we do want to run our logic before the super() (for example to add an extra check to raise an error)

ziriraha (commented):
i dont know if i said this before, but can we squash all commits into a single one please?

PR #1341, opened 2026-06-16
chean - Technical Onboarding

anchevalier:


estate/models/estate_property.py
```diff
@@ -0,0 +1,138 @@
...
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = ""
+
+    def action_sold(self):
```
ziriraha, line 148: actions can also be run in a recordset, dont forget the `for estate in self` if the action is prohibited from running in more than one record you should add `self.ensure_one()`

estate/models/estate_property.py
```diff
@@ -0,0 +1,138 @@
...
+        "A property expected price should be strictly positive.",
+    )
+
+    _check_selling_price = models.Constraint(
+        "CHECK(selling_price >= 0)", "A property selling price should be positive."
+    )
+
+    @api.constrains("selling_price", "expected_price")
+    def _check_selling_price(self):
+        for record in self:
```
ziriraha:
```suggestion
        for estate in self:
```
so we know with which type of record we are working

ziriraha (changes requested):

estate/models/estate_property.py
anchevalier: I'll add `ensure_one()` here because the buttons are only available from the form view, so the action is intended to run on a single record.

If I understood correctly, another option would be to expose the action from the list view and handle it on a full recordset with a for estate in self.

estate/models/estate_property.py
anchevalier: Got it, that makes sense. I'll keep that in mind so it's immediately clear which type of record is being handled.

estate/models/estate_property.py
```diff
@@ -0,0 +1,162 @@
...
+            ("canceled", "Canceled"),
+        ],
+        required=True,
+        copy=False,
+        default="new",
+    )
+    property_type_id = fields.Many2one(
+        comodel_name="estate.property.type",
+        string="Property Type",
+    )
```
ziriraha, line 59:
```suggestion
    property_type_id = fields.Many2one("estate.property.type", "Property Type")
```
No need to apply this change, but just so you know, this is possible to do

estate/models/estate_property.py
```diff
@@ -0,0 +1,162 @@
...
+            if (
+                float_compare(
+                    property_.selling_price,
+                    minimum_price,
+                    precision_digits=2,
+                )
+                < 0
+            ):
+                raise ValidationError(
+                    _(
```
ziriraha:
```suggestion
                    self.env._(
```
we no longer use the _ import, its better to use self.env._ (basically because its horrible when you want to do a for _ in ... and you cant)

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,110 @@
...
+        return True
+
+    def action_refuse_offer(self):
+        self.ensure_one()
+
+        if self.status == "accepted":
+            raise UserError(_("You cannot refuse an offer once it has been accepted."))
+
+        self.status = "refused"
+        return True
```
ziriraha, line 113: fyi (i dont know if its explained in the tutorial) actions can return action objects so you can say things like: open this popup, reload the page, redirect to here

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
+access_estate_property_type,access_estate_property_type,model_estate_property_type,base.group_user,1,1,1,1
+access_estate_property_tag,access_estate_property_tag,model_estate_property_tag,base.group_user,1,1,1,1
+access_estate_property_offer,access_estate_property_offer,model_estate_property_offer,base.group_user,1,1,1,1
```
ziriraha, line 5: new line at eof

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,43 @@
...
+        sequence="10"
+    />
+
+    <menuitem
+        id="estate_property_tag_menu_action"
+        action="estate_property_tag_action"
+        parent="estate_settings_menu"
+        sequence="20"
+    />
+</odoo>
```
ziriraha: new line at eof (on the others too)

estate/views/res_users_views.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0"?>
+<odoo>
+<data>
+    <record id="res_users_view_form" model="ir.ui.view">
+        <field name="name">res.users.view.form.inherit.estate</field>
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" position="inside">
```
ziriraha:
```suggestion
            <notebook position="inside">
```
I think you can also do overrides like this without xpath

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,31 @@
...
+    _inherit = "estate.property"
+
+    def action_sold(self):
+        for record in self:
+            self.env["account.move"].create(
+                {
+                    "partner_id": record.buyer_id.id,
+                    "move_type": "out_invoice",
+                    "invoice_line_ids": [
+                        Command.create(
```
ziriraha, line 16: you can also do

```suggestion
                        (0, 0,
```
some people hate Command... but i dont know why

ziriraha (commented):
Good work, can you please squash all the commits into a single one?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,110 @@
...
+        for offer in self:
+            create_date = offer.create_date or fields.Datetime.now()
+            offer.validity = (offer.date_deadline - create_date.date()).days
+
+    @api.model
+    def create(self, vals_list):
+        properties = self.env["estate.property"].browse(
+            [vals["property_id"] for vals in vals_list]
+        )
+        properties_by_id = {property_.id: property_ for property_ in properties}
```
ziriraha: this is good but isnt needed, you just call properties.browser(id) after and it would be the same effect

Something like:
```py
properties = self.env["estate.property"].browse([vals["property_id"] for vals in vals_list])

for vals in vals_list:
    property = properties.browse(vals["property_id"])
```

estate_account/__manifest__.py
```diff
@@ -0,0 +1,4 @@
+{
+    "name": "Estate Account",
+    "depends": ["estate", "account"],
```
ziriraha, line 3: Sometimes we add `auto_install: True` to tell it to install automatically when all the depends modules are installed.

ziriraha (commented):
forgot to add

estate/views/res_users_views.xml
anchevalier: Yes, it works !

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,31 @@
...
+                            {
+                                "name": "Administrative Fees",
+                                "quantity": 1.0,
+                                "price_unit": 100.00,
+                            }
+                        ),
+                    ],
+                }
+            )
+        return super().action_sold()
```
ziriraha: In action_sold you have some check that raise errors. In this case, you are creating the account.move and then you are performing the checks. you should place the super() call above the creation of the account move, as there might a case where you create a move and then the sell errors out for some reason.

We usually do:
```py
def func():
    action = super().func()
    ....
    return action
```
Dont take this as a rule of thumb, there are cases where we do want to run our logic before the super() (for example to add an extra check to raise an error)

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,31 @@
+from odoo import Command, models
+
+
+class EstateProperty(models.Model):
+    _inherit = "estate.property"
+
+    def action_sold(self):
+        for record in self:
```
ziriraha, line 10: super().action_sold is ensure_one, then this one must be ensure_one too

PR #1361, opened 2026-07-03
[19.0] Real Estate module tutorial

hagav-odoo:
This PR contains the following: 

- Models: `estate`, `estate.type`, `estate.tag`, `estate.offer`
- Security access rights
- List/form/search views + menus
- Computed fields, garden onchange, offer deadline inverse
- Sold/cancel and accept/refuse actions


estate/models/estate.py
```diff
@@ -0,0 +1,83 @@
...
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        selection=[
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
+        ]
+    )
```
mash-odoo, line 31:
```suggestion
    garden_orientation = fields.Selection(
        selection=[
            ('north', "North"),
            ('south', "South"),
            ('east', "East"),
            ('west', "West"),
        ]
    )
```

estate/models/estate.py
```diff
@@ -0,0 +1,83 @@
...
+            ("cancelled", "Cancelled"),
+        ],
+        required=True,
+        copy=False,
+        default="new",
+    )
+    active = fields.Boolean(default=True)
+    estate_type_id = fields.Many2one("estate.type", string="Estate Type")
+    seller_id = fields.Many2one("res.users", string="Seller", default=lambda self: self.env.user)
+    buyer_id = fields.Many2one("res.partner", string="Buyer", copy=False)
```
mash-odoo, line 47: Why do you need to give string here?
What will appear on the UI if we do not write this?

estate/models/estate.py
```diff
@@ -0,0 +1,83 @@
...
+    tag_ids = fields.Many2many("estate.tag", string="Tags")
+    offer_ids = fields.One2many("estate.offer", "property_id", string="Offers")
+    total_area = fields.Integer(compute="_compute_total_area", store=True)
```
mash-odoo, line 50: What's the use case of making this field stored?

estate/models/estate_offer.py
```diff
@@ -0,0 +1,42 @@
+from dateutil.relativedelta import relativedelta
+from odoo import api, fields, models
+
+
+class EstateOffer(models.Model):
+    _name = 'estate.offer'
+    _description = 'Estate Offer'
+
+    price = fields.Float()
```
mash-odoo, line 9: The data type of the field is given as Float, but when you look into the database, it's coming different. WDYT? How's that happening?

estate/models/estate_offer.py
```diff
@@ -0,0 +1,42 @@
...
+        for record in self:
+            record.date_deadline = (record.create_date or fields.Date.today()) + relativedelta(days=record.validity)
+
+    def _inverse_date_deadline(self):
+        for record in self:
+            record.validity = (record.date_deadline - record.create_date.date()).days
+
+    def action_accept(self):
+        for record in self:
+            record.status = 'accepted'
```
mash-odoo, line 30: What's the difference between these two lines?
1. record.status = 'accepted'
2. record.write({'status': 'accepted'})

estate/views/estate_offers.xml
```diff
@@ -0,0 +1,37 @@
...
+    <field name="name">Estate Offer List</field>
+    <field name="model">estate.offer</field>
+    <field name="arch" type="xml">
+        <list>
+            <field name="price" />
+            <field name="partner_id" />
+            <field name="validity" />
+            <field name="date_deadline" />
+            <field name="status" />
+            <button name="action_accept" string="Accept" type="object" icon="fa-check" />
```
mash-odoo, line 13: Can you explain the purpose of `type` attribute?

estate/views/estate_views.xml
```diff
@@ -0,0 +1,126 @@
...
+            <field name="bedrooms" />
+            <field name="living_area" />
+            <field name="facades" />
+            <field name="postcode" />
+            <field name="expected_price" />
+            <field name="estate_type_id" />
+            <filter
+                    name="available"
+                    string="Available"
+                    domain="[('state', 'in', ('new', 'offer_received'))]"
```
mash-odoo, line 115: What is a domain?

estate/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+{  # pyright: ignore[reportUnusedExpression]
+    "name": "estate",
+    "depends": ["base"],
```
mash-odoo, line 3: What if we remove this depends?

estate/models/estate.py
```diff
@@ -0,0 +1,83 @@
...
+            record.best_offer = max(record.offer_ids.mapped("price"), default=0)
+
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = False
```
mash-odoo, line 71: Can we perform this onchange with compute?
If yes.... then why are we using here onchange and not compute?

estate/models/estate_offer.py
```diff
@@ -0,0 +1,42 @@
...
+
+    price = fields.Float()
+    status = fields.Selection(
+        selection=[('accepted', 'Accepted'), ('refused', 'Refused')],
+        copy=False,
+    )
+    partner_id = fields.Many2one('res.partner', required=True)
+    property_id = fields.Many2one('estate', required=True)
+    validity = fields.Integer(default=7)
+    date_deadline = fields.Date(compute='_compute_date_deadline', inverse='_inverse_date_deadline')
```
mash-odoo, line 17: How does an inverse function work?

estate/models/estate.py
```diff
@@ -0,0 +1,83 @@
+from dateutil.relativedelta import relativedelta
+from odoo import api, fields, models, exceptions
+
+
+class Estate(models.Model):
```
mash-odoo, line 5: Is it mandatory to use `models.Model`?

mash-odoo (changes requested):
Hello,
Good beginning on the task.
Some points to be looked at:
- Do not push unnecessary files in the commit. Try to keep commits clean.
- Add description and chapter name while pushing the commits in order to understand the changes done for that particular commit. 
- Update the PR title and description.

estate/models/estate.py
hagav-odoo: yeah makes sense, will switch the keys to single quotes 👍

estate/models/estate.py
hagav-odoo: kept `string=` on purpose as good practice. 
without it odoo still builds the label from the field name (strip `_id`, capitalize) so it'd show "Seller" / "Buyer" either way doesn't break if we remove it.

estate/models/estate.py
hagav-odoo: used `store=True` while testing whether the compute could skip the for-loop (f315778). it worked for most cases, but broke on duplicate/import (only the first record got computed), so i switched to a for-loop and just forgot to drop `store=True`.  (removing it now)

btw with `store=True`, odoo reads the value straight from the db on list instead of recomputing every time.

estate/models/estate_offer.py
hagav-odoo: i used `fields.Float()` for price, but in postgres it shows as `double precision`. that's because odoo maps a Float  to postgres `float8` / `double precision`, not a type literally called "float".

estate/models/estate_offer.py
hagav-odoo: for a saved record they're basically the same,
 `record.status = 'accepted'` goes through the field setter and ends up calling `write({'status': 'accepted'})` under the hood ( `Field.__set__` in orm/fields.py).

main differences:
- assignment = nicer for one field
- `write` = better when updating several fields at once, or the whole recordset in one go

estate/views/estate_offers.xml
hagav-odoo: `type` tells odoo what `name` refers to on click:

- `object` → call a python method on the current record (here `action_accept`)
- `action` → load and run an `ir.actions.*` with the current record context

estate/views/estate_views.xml
hagav-odoo: domain is odoo’s filter for searching records.

estate/__manifest__.py
hagav-odoo: if we remove `depends` (or set `[]`), odoo injects `base` when loading the manifest (`modules/module.py`) .

if we do `depends: [" "]`, the list is non-empty so that fallback is skipped, and it fails later because `" "` isn’t a real module

estate/models/estate.py
hagav-odoo: we could use compute here, but then UX becomes very bad as computed fields are readonly, and even with an inverse the user has to click on save before it reflects changes (inverse runs on save).

onchange is a better fit: it updates client-side instantly, and the user can still edit garden_area / garden_orientation freely

estate/models/estate_offer.py
hagav-odoo: inverse runs when the user edits `date_deadline`,  it writes back into `validity` so both stay in sync. without inverse the computed field would be readonly.

inverse only fires on save.

estate/models/estate.py
hagav-odoo: yes, by inheriting `models.Model`, odoo intercepts the class via its metaclass and registers it as a real model. it's  like link that enables table creation, fields, ORM, etc.

if we drop `models.Model`, odoo can’t intercept/register the class, so it’s mandatory.

PR #1362, opened 2026-07-03
19.0 tutorials mrgor

mrgor-odoo:


estate/models/estate_property.py
```diff
@@ -0,0 +1,41 @@
+from odoo import models, fields
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate Property"
+    name = fields.Char(required=True)
```
```suggestion
    _description = "Real Estate Property"

    name = fields.Char(required=True)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,46 @@
+from odoo import models, fields
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
```
```suggestion
    _name = 'estate.property'
```
wrap the technical strings in single quotes and the strings which appear on UI and visible to customer in double quotes

estate/models/estate_property.py
```diff
@@ -0,0 +1,46 @@
...
+    garden_orientation = fields.Selection(
+        selection=[
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
+        ]
+    )
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
```
mash-odoo: Arrange all the fields alphabetically so that it's easier to locate it later

estate/models/estate_property.py
```diff
@@ -0,0 +1,46 @@
...
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        selection=[
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
+        ]
```
mash-odoo, line 27:
```suggestion
        selection=[
            ('north', "North"),
            ('south', "South"),
            ('east', "East"),
            ('west', "West"),
        ]
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,46 @@
...
+            ("cancelled", "Cancelled"),
+        ],
+        required=True,
+        copy=False,
+        default="new",
+    )
+    property_type_id = fields.Many2one("estate.property.type", string="Property Type")
+    buyer_id = fields.Many2one("res.partner", string="Buyer", copy=False)
+    salesperson_id = fields.Many2one(
+        "res.users", string="Salesperson", default=lambda self: self.env.user
```
mash-odoo: Why do we need to use here lambda function?
Do we have any other approach here?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,108 @@
...
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="facades"/>
+
+                <filter
+                    string="Available"
+                    name="available"
+                    domain="[('state','in',['new','offer_received'])]"/>
```
```suggestion
                    domain="[('state', 'in', ['new', 'offer_received'])]"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,108 @@
...
+
+</odoo>
+
+
+
+
+
+
+
+
```
```suggestion
</odoo>
```

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "Real Estate",
+    "version": "1.0",
+    "author": "Mrunmayee",
```
mash-odoo: When you are working for a company, you should keep the author name as `Odoo S.A.` or just skip writing it.

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "depends": ["base"],
```
mash-odoo, line 5: What's the purpose of `depends`?

estate/models/estate_property.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate Property"
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        copy=False, default=lambda self: fields.Date.add(fields.Date.today(), months=3)
```
mash-odoo, line 13: Can this be done in any other way?

estate/models/estate_property.py
```diff
@@ -0,0 +1,48 @@
...
+        copy=False,
+        default="new",
+    )
+    property_type_id = fields.Many2one("estate.property.type", string="Property Type")
+    buyer_id = fields.Many2one("res.partner", string="Buyer", copy=False)
+    salesperson_id = fields.Many2one(
+        "res.users", string="Salesperson", default=lambda self: self.env.user
+    )
+    tag_ids = fields.Many2many("estate.property.tag", string=" Property Tags")
+    offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers")
```
mash-odoo: Where can I find the records of offers connected to a property?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,113 @@
...
+                <filter
+                    string="Available"
+                    name="available"
+                    domain="[('state','in',['new','offer_received'])]"/>
+
+                <filter
+                    string="Postcode"
+                    name="group_postcode"
+                    context="{'group_by':'postcode'}"/>
```
mash-odoo: WDYM by context?

mash-odoo (changes requested):
Hello,
Good start on the task..
Here are a few comments and suggestions.. 🐣 
Also please change your PR title and description!!

estate/models/estate_property.py
mrgor-odoo: I used a lambda because the logic is only one line, so it keeps the code short and readable. Another approach would be to define a method like _default_salesperson() that returns self.env.user and use that method as the default. Both work, but I felt lambda was simpler here.

estate/__manifest__.py
mrgor-odoo: depends specifies which modules my module relies on. Odoo uses it to ensure those modules are installed and loaded before installing or upgrading my module. In this case, I added base because my module uses the core Odoo framework, such as models.Model, field types, and models like res.users and res.partner.

estate/models/estate_property.py
mrgor-odoo: Yes, another approach is to create a separate method, for example _default_date_availability(), which returns the default date, and then pass that method to default.

estate/models/estate_property.py
mrgor-odoo: The offers are stored in the estate.property.offer table. They are connected to a property using the property_id field. When I access offer_ids, Odoo automatically fetches all offer records whose property_id matches the current property.

estate/views/estate_property_views.xml
mrgor-odoo: context passes extra information to Odoo. In this case, it tells Odoo to group the records by postcode.

PR #1365, opened 2026-07-20
19.0: Onboarding bepro

bepro-odoo:


.vscode/tutorials.code-workspace
YassinWalid, line 1: I believe you could add this file to your ignores so it's not pushed to the repository?

estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import property, property_offer, property_tag, property_type, res_user
```
YassinWalid:
```suggestion
from . import property
from . import property_offer
from . import property_tag
from . import property_type
from . import res_user
```
We prefer having each import on a single line, as this allows better traceability (with `git blame` for example)

estate/models/property.py
```diff
@@ -0,0 +1,115 @@
...
+    _name = "estate.property"
+    _description = "Properties of our managed estates"
+    _order = "id desc"
+
+    _check_positive_amounts = models.Constraint(
+        'CHECK ('
+        'expected_price >= 0 AND selling_price >= 0'
+        ' AND living_area >= 0 AND facades >= 0'
+        ' AND garden_area >= 0'
+        ')', 'Values must be positive!')
```
YassinWalid: I believe the constraints should be split, because a user editing one of these fields, along with another field, might get this generic error message, but will not know which value needs to be positive. Therefore having one constraint per field is better for more meaningful error messages.

estate/models/property.py
```diff
@@ -0,0 +1,115 @@
...
+
+    name = fields.Char(string='Title', default="Unknown", required=True)
```
YassinWalid:
```suggestion
    name = fields.Char(string='Title', required=True)
```
I don't think we need the default, since having it with no default forces the user to fill in this required field.

estate/models/property.py
```diff
@@ -0,0 +1,115 @@
...
+    ])
+    property_type_id = fields.Many2one("estate.property.type", string="Property Type")
+    salesman_id = fields.Many2one('res.users', string="Salesman", default=lambda self: self.env.user)
+    buyer_id = fields.Many2one('res.partner', string="Buyer", copy=False)
+    tag_ids = fields.Many2many("estate.property.tag")
+    offer_ids = fields.One2many("estate.property.offer", "property_id")
+
+    description = fields.Text(string='description')
+    postcode = fields.Char(string='postcode')
+    date_availability = fields.Date(string='Available from', default=lambda _: fields.Date.add(fields.Date.today(), months=3), copy=False)
```
YassinWalid, line 26:
```suggestion
    date_availability = fields.Date(string='Available from', default=fields.Date.add(fields.Date.today(), months=3), copy=False)
```
We don't need the lambda here. We could just put the value as it is.

estate/models/property.py
```diff
@@ -0,0 +1,115 @@
...
+
+    name = fields.Char(string='Title', default="Unknown", required=True)
+    active = fields.Boolean(default=True)
+    state = fields.Selection(required=True, default='new', copy=False, selection=[
+        ('new', 'New')
+        , ('offer_received', 'Offer Received')
+        , ('offer_accepted', 'Offer Accepted')
+        , ('sold', 'Sold')
+        , ('cancelled', 'Cancelled'),
+    ])
```
YassinWalid, line 18:
```suggestion
    state = fields.Selection(required=True, default='new', copy=False, selection=[
        ('new', 'New'),
        ('offer_received', 'Offer Received'),
        ('offer_accepted', 'Offer Accepted'),
        ('sold', 'Sold'),
        ('cancelled', 'Cancelled'),
    ])
```

estate/models/property.py
```diff
@@ -0,0 +1,115 @@
...
+    facades = fields.Integer(string='# facades')
+    garage = fields.Boolean(string='Has garage')
+    garden = fields.Boolean(string='Has garden')
+    garden_area = fields.Integer(string='garden area size')
+    garden_orientation = fields.Selection(string='garden orientation',
+    selection=[
+        ('north', 'North')
+        , ('south', 'South')
+        , ('east', 'East')
+        , ('west', 'West')])
```
YassinWalid:
```suggestion
    garden_orientation = fields.Selection(string='garden orientation',
        selection=[
            ('north', 'North'),
            ('south', 'South'),
            ('east', 'East'),
            ('west', 'West'),
        ],
    )
```

estate/models/property.py
```diff
@@ -0,0 +1,115 @@
...
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for record in self:
+            if record.offer_ids:
+                record.best_price = max(record.offer_ids.mapped("price"))
+            else:
+                record.best_price = 0
```
YassinWalid, line 67:
```suggestion
            record.best_price = max(record.offer_ids.mapped("price")) if record.offer_ids else 0
```
Just a suggestion, feel free to ignore it if you want 👌

estate/models/property.py
```diff
@@ -0,0 +1,115 @@
...
+
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        for record in self:
+            record.garden_area = 10 if record.garden else 0
+            record.garden_orientation = 'north' if record.garden else None
```
YassinWalid:
```suggestion
            record.garden_orientation = 'north' if record.garden else False
```
When deassigning a value, we prefer using 'False' since it's a bit more explicit, regardless of the type

estate/models/property.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def action_sold(self):
+        for record in self:
+            if record.state == 'cancelled':
+                raise exceptions.UserError(_("Cancelled properties cannot be sold!"))
+            #
```
YassinWalid:
```suggestion
```

estate/models/property.py
```diff
@@ -0,0 +1,115 @@
...
+            record.state = "sold"
+        return True
+
+    def action_cancelled(self):
+        for record in self:
+            if record.state == "sold":
+                raise exceptions.UserError(_("Sold properties cannot be cancelled!"))
+            #
```
YassinWalid:
```suggestion
```

estate/models/property_offer.py
```diff
@@ -0,0 +1,77 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError, ValidationError
+
+
+class PropertyOffer(models.Model):
+    _name = "estate.property.offer"
+    _description = "Bids for a property"
+    _order = "price desc"
+
+    _check_positive_amounts = models.Constraint('CHECK(price >= 0)')
```
YassinWalid:
```suggestion
    _check_positive_amounts = models.Constraint('CHECK(price >= 0)', 'some message')
```
Here we can display a message to tell the user what the expected value should follow

estate/models/property_offer.py
```diff
@@ -0,0 +1,77 @@
...
+
+    price = fields.Float(string="Price", required=True)
+    status = fields.Selection(string="Status", copy=False
+        , selection=[('accepted', 'Accepted'), ('refused', 'Refused')])
```
YassinWalid:
```suggestion
    status = fields.Selection(string="Status", copy=False,
        selection=[('accepted', 'Accepted'), ('refused', 'Refused')])
```

estate/models/property_offer.py
```diff
@@ -0,0 +1,77 @@
...
+    property_type_id = fields.Many2one(related="property_id.property_type_id", store=True)
+    validity = fields.Integer(string="Validity of offer", default=7)
+    date_deadline = fields.Date(string="Offer expiry", compute="_compute_date_deadline", inverse="_inverse_date_deadline")
+
+    @api.depends("create_date", "validity")
+    def _compute_date_deadline(self):
+        for record in self:
+            record.date_deadline = None
+            if record.create_date and record.validity:
+                record.date_deadline = fields.Date.add(record.create_date, days=record.validity)
```
YassinWalid: Maybe instead of assigning `None` as a fallback, we can add the validity period to today (`fields.Date.today()`). This way, at time of creation, the user will see the deadline, and it will be recomputed anyway once it is created and `create_date` is filled

YassinWalid:
```suggestion
            record.date_deadline = None
            if record.create_date and record.validity:
                record.date_deadline = fields.Date.add(record.create_date.date(), days=record.validity)
```
Since the `create_date` field is of type datetime, we need only the date part of it

estate/models/property_offer.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _inverse_date_deadline(self):
+        for record in self:
+            if record.date_deadline:
+                raw_difference = record.date_deadline - (record.create_date.date() or fields.Date.today())
+                difference = raw_difference.total_seconds()
+                if difference > 0:
+                    record.validity = floor(difference / dt.timedelta(days=1).total_seconds())
```
YassinWalid:
```suggestion
                record.validity = (record.date_deadline - (record.create_date.date() or fields.Date.today())).days
```
If we only take the date part from the `create_date` in the compute, we can ignore the fractions of the day, and have it as a difference of days as well.

estate/models/property_offer.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        # Precondition: price and property_id are required fields
+        for vals in vals_list:
+            property = self.env["estate.property"].browse(vals["property_id"])
```
YassinWalid:
```suggestion
        property_ids = self.env['estate.property'].browse({vals['property_id'] for vals in vals_list})
        for vals in vals_list:
            property_id = property_ids.filtered(lambda p: p.id == vals['property_id'])
```
Instead of doing a browse in every iteration, we could do one browse for all of them, and then filter each time getting the property we want.

estate/models/property_offer.py
```diff
@@ -0,0 +1,77 @@
...
+            property.ensure_one()
```
YassinWalid, line 53:
```suggestion
```
Since the id is unique, there will only be at most one property matching `vals["property_id"]`, so I think the `ensure_one` is not needed.

estate/models/property_offer.py
```diff
@@ -0,0 +1,77 @@
...
+            # Hook-approach (for composability)
+            property._set_offer_received()
+            max_peer_offer_price = max(property.offer_ids.mapped('price'), default=None)
```
YassinWalid:
```suggestion
            max_peer_offer_price = property.best_price
```
Since we already have a field carrying the current best price, we can use that right away.

estate/models/property_offer.py
```diff
@@ -0,0 +1,77 @@
...
+            if max_peer_offer_price and max_peer_offer_price > vals['price']:
+                raise UserError(_("New offer price must be higher than those of pre-existing offers!"))
+        #
```
YassinWalid:
```suggestion
```

estate/models/property_offer.py
```diff
@@ -0,0 +1,77 @@
...
+        return super().create(vals_list)
+
+    @api.constrains('status')
+    def _check_maximum_one_offer_accepted(self):
+        for record in self:
+            peer_offers = record.property_id.offer_ids
+            accepted_peers = peer_offers.filtered(lambda r: r.status == 'accepted')
+            if len(accepted_peers) > 1:
+                raise ValidationError(_("A single offer can be accepted at a time!"))
```
YassinWalid:
```suggestion
        accepted_offers = self.filtered(lambda r: r.status == "accepted")
        properties = accepted_offers.mapped('property_id')
        for property in properties:
            peer_offers = property.offer_ids
            accepted_peers = peer_offers.filtered(lambda r: r.status == 'accepted')
            if len(accepted_peers) > 1:
                raise ValidationError(_("A single offer can be accepted at a time!"))
```
I think you can filter the records before doing the check on them, since if the `status` is changed to any value other than accepted, we don't need to recheck that.
We can also get their properties right away, since if two offers in one check belong to the same property, you'll run two iterations for them

estate/models/property_offer.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def action_confirm(self):
+        for record in self:
+            if record.status == "refused":
+                raise UserError(_("Offer is already refused!"))
```
YassinWalid, line 64: why are we not allowing the user to revise their decision regarding an offer?

estate/models/property_offer.py
```diff
@@ -0,0 +1,77 @@
...
+            #
+            record.status = 'accepted'
+            record.property_id.confirm_offer()
+        return True
+
+    def action_cancel(self):
+        for record in self:
+            if record.status == "accepted":
+                raise UserError(_("Offer is already accepted!"))
```
YassinWalid, line 73: here too

estate/models/res_users.py
YassinWalid, line 1: please rename the file to match the name of the model (`res_users`)

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property_model,access_estate_property_model,estate.model_estate_property,base.group_user,1,1,1,1
+access_estate_property_type_model,access_estate_property_type_model,estate.model_estate_property_type,base.group_user,1,1,1,1
+access_estate_property_tag_model,access_estate_property_tag_model,estate.model_estate_property_tag,base.group_user,1,1,1,1
+access_estate_property_offer_model,access_estate_property_offer_model,estate.model_estate_property_offer,base.group_user,1,1,1,1
```
YassinWalid:
```suggestion
access_estate_property_offer_model,access_estate_property_offer_model,estate.model_estate_property_offer,base.group_user,1,1,1,1

```
Add an empty line at the end of the file

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,14 @@
...
+        <menuitem id="estate_properties_menu_action" action="estate_properties_view"/>
+    </menuitem>
+
+    <menuitem id="estate_settings_menu" name="Settings">
+        <menuitem id="estate_properties_types_menu_action" action="estate_properties_types_view"/>
+        <menuitem id="estate_properties_tags_menu_action" action="estate_properties_tags_view"/>
+    </menuitem>
+</menuitem>
+
+</odoo>
```
YassinWalid:
```suggestion
</odoo>

```

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,14 @@
...
+<menuitem id="estate_menu_root" name="Estate">
+    <menuitem id="estate_first_level_menu" name="Advertisements">
+        <menuitem id="estate_properties_menu_action" action="estate_properties_view"/>
+    </menuitem>
+
+    <menuitem id="estate_settings_menu" name="Settings">
+        <menuitem id="estate_properties_types_menu_action" action="estate_properties_types_view"/>
+        <menuitem id="estate_properties_tags_menu_action" action="estate_properties_tags_view"/>
+    </menuitem>
+</menuitem>
```
YassinWalid:
```suggestion
    <menuitem id="estate_menu_root" name="Estate">
        <menuitem id="estate_first_level_menu" name="Advertisements">
            <menuitem id="estate_properties_menu_action" action="estate_properties_view"/>
        </menuitem>

        <menuitem id="estate_settings_menu" name="Settings">
            <menuitem id="estate_properties_types_menu_action" action="estate_properties_types_view"/>
            <menuitem id="estate_properties_tags_menu_action" action="estate_properties_tags_view"/>
        </menuitem>
    </menuitem>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,264 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="estate_properties_view" model="ir.actions.act_window">
```
YassinWalid:
```suggestion
    <record id="estate_property_action" model="ir.actions.act_window">
```
The id of the action should reflect that it's an action. Having the word view in it might be misleading.
(also applies for the other actions)

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,264 @@
...
+
+    <record id="estate_properties_offers_view" model="ir.actions.act_window">
+        <field name="name">Property Offers</field>
+        <field name="res_model">estate.property.offer</field>
+        <field name="view_mode">list,form</field>
+        <field name="domain">[('property_type_id', '=', active_id)]</field>
+    </record>
+
+    <record id="estate_property_type_view_tree" model="ir.ui.view">
+        <field name="name">estate.property.types.list</field>
```
YassinWalid:
```suggestion
    <record id="estate_property_type_view_list" model="ir.ui.view">
        <field name="name">estate.property.type.list</field>
```
Please check the coding guidelines linked in the overall comment

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,264 @@
...
+                <field name="property_type_id"/>
+                <templates>
+                    <t t-name="card" class="flex-row">
+                        <main class="ms-2">
+                            <field name="name" class="fw-bold"/>
+                            <div>
+                                Expected price:
+                                <field name="expected_price"/>
+                            </div>
+                            <t t-if="record.state == 'offer_received'">
```
YassinWalid:
```suggestion
                            <t t-if="record.state.raw_value == 'offer_received'">
```
We want to compare the raw value as stored, without the formatting using user preferences.
(This should apply for other usages as well)

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,264 @@
...
+    </record>
+
+    <record id="estate_property_types_view_form" model="ir.ui.view">
+        <field name="name">estate.property.types.form</field>
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <form string="Estate Property Type">
+                <sheet>
+                    <div class="oe_button_box" name="button_box">
+                        <button name="%(estate.estate_properties_offers_view)d" type="action" class="oe_stat_button" invisible="not offer_count">
```
YassinWalid: but why do we want to make it invisible if there are no offers? The user might use this button to navigate to the offers view and create an offer from there right away. Please let me know what you think

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,264 @@
...
+                                </group>
+                            </group>
+                        </page>
+                    </notebook>
+                </sheet>
+            </form>
+        </field>
+    </record>
+
+    <record id="estate_property_offers_view_tree" model="ir.ui.view">
```
YassinWalid: why are we not showing the deadline and validity here in this view as suggested by the "Goal" section of the tutorial? I think they're important fields to show to the user to be able to assess the different offers

estate/views/res_user_views.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="name">estate.res.user.form</field>
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//notebook/page[@name='preferences']" position="after">
+                <page string="Real Estate Properties" name="estate_properties">
+                    <field name="property_ids" readonly="True"/>
+                    <group>
+                        <group></group>
+                    </group>
```
YassinWalid: I don't understand the need for this?

estate/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+{
+    'name': "Real Estate",
+    'depends': ['base'],
+    'author': 'bepro',
```
YassinWalid: this is okay for this module since it's just for onboarding, but normally it should be `Odoo S.A.`

estate_account/__manifest__.py
YassinWalid, line 1: one small nitpick: try to make the use of ' and " uniform.

YassinWalid (changes requested):
Thanks for the impressive work. I have left a few comments in the PR, so feel free to check them (and if you don't agree with sth we can always discuss it, since in many cases there is no "absolute truth". Please also check the coding guidelines below.

Patterns for names and ids for records in xml files:
please check this link: https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#xml-ids-and-naming

Also, please split your views into one file per model, so they can be found easier.

estate/models/property.py
bepro-odoo: I tested the approach without lambda and fieldtype datetime. The form value is stuck at a datetime around startup of odoo server, meaning that the default value is generated *once* and reused as default. The purpose of the field is to have a date set 3 months into the future at point of record creation, implying the date must be calculated at that point in time, hence the lambda is required.

Info @SaddemAmine

estate/models/property.py
bepro-odoo: ignored

estate/models/property_offer.py
bepro-odoo: instructed by tutorial

estate/models/property_offer.py
bepro-odoo: instructed by tutorial

estate/views/estate_property_views.xml
bepro-odoo: instructed by tutorial

estate/models/property_offer.py
bepro-odoo: kept for reason that filter is constructed with untrusted user input and property could contain empty recordset

estate/models/property_offer.py
bepro-odoo: skipped suggestion because the invariant body would depend on an effect of another conceptual property, opted to work with the conceptual property directly

estate/models/property.py
```diff
@@ -0,0 +1,120 @@
+from odoo import _, api, exceptions, fields, models
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+
+class Property(models.Model):
+    _name = "estate.property"
+    _description = "Properties of our managed estates"
+    _order = "id desc"
+
+    _check_positive_expected_price = models.Constraint(
```
YassinWalid: According to the coding guidelines, the SQL constraints should be after the field declarations in the file.

Image

estate/security/ir.model.access.csv
YassinWalid, line 1: Please check the guidelines regarding the name of the records here. They should be the same as the id, but replacing the underscores with dots.

estate/views/estate_property_views.xml
YassinWalid, line 1: the names of view files should include the suffix `_views.xml` 
https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#file-naming
This applies to the other files as well.

YassinWalid (commented):
Thanks for your work. This looks perfect for me, and I left some minor comments.

SaddemAmine:
f9788864ebe0b9fc0782c042d4692afb4fd7fe9e has a "wrong" title btw 🤓
I'll take look at the PR soon!

Nice work!

SaddemAmine:
@robodoo help

PR #1367, opened 2026-07-20
PR Onboarding toher

Tom-Hermann:
[ADD] Chap-2


estate/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+from . import estate_property
+from . import estate_type
+from . import estate_tag
+from . import estate_offer
+from . import inhertied_user
```
SaddemAmine: We use the same model name even when we inherit
```suggestion
from . import res_user
```

estate/models/estate_offer.py
```diff
@@ -0,0 +1,64 @@
...
+    property_type_id = fields.Many2one(related="property_id.property_type_id", store=True)
+
+    @api.model
+    def create(self, vals_list):
+        for vals in vals_list:
+            property = self.env["estate.property"].browse(vals["property_id"])
+            property.set_offer_received()
+            max_price = max(property.proterty_offer_ids.mapped('price'), default=None)
+            if max_price and max_price > vals['price']:
+                raise exceptions.UserError("New offer can't be lower than an other offer")
```
SaddemAmine: You don't need a loop here since we're ordering the records using `_order = "price desc"`
Grabbing `offer_ids[0]` is enough!

SaddemAmine: You can directly import `UserError` from odoo.exceptions to avoid needing to access it
I don't think it's costly to keep your current pattern, but the one I mentioned is the common way

Also don't forget about translation!

```suggestion
                raise UserError(self.env._("New offer can't be lower than an other offer"))
```

estate/models/estate_offer.py
```diff
@@ -0,0 +1,64 @@
...
+            return super().create(vals_list)
+
+    @api.depends("validity")
+    def _compute_dead_line(self):
+        for offer in self:
+            offer.date_deadline = fields.Date.today() + timedelta(days=offer.validity)
+
+    def _inverse_validity(self):
+        for offer in self:
+            offer.validity = (offer.date_deadline - fields.Date.today()).days
```
SaddemAmine: What if the offer was already created in the past? It's best to rely on a combination of `create_date` and `fields.Date.today()` if the record is still not created.

estate/models/estate_offer.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @api.model
+    def create(self, vals_list):
+        for vals in vals_list:
+            property = self.env["estate.property"].browse(vals["property_id"])
+            property.set_offer_received()
+            max_price = max(property.proterty_offer_ids.mapped('price'), default=None)
+            if max_price and max_price > vals['price']:
+                raise exceptions.UserError("New offer can't be lower than an other offer")
+            return super().create(vals_list)
```
SaddemAmine: Notice how the super call is inside the loop, this means you only check on the first set of values

estate/models/estate_offer.py
```diff
@@ -0,0 +1,64 @@
...
+
+    price = fields.Float(string="Price")
+    status = fields.Selection(
+        string="Status",
+        selection=[
+            ("accepted", "Accepted"),
+            ("refused", "Refused")
+        ],
+        copy=False
+    )
```
SaddemAmine, line 23: This means that offers are either accepted or refused by default
What about the third state, when they're received but not handled yet?

IIRC the documentation mentions a third state `new` which is the default for offers

estate/models/estate_offer.py
```diff
@@ -0,0 +1,64 @@
...
+    @api.depends("validity")
+    def _compute_dead_line(self):
+        for offer in self:
+            offer.date_deadline = fields.Date.today() + timedelta(days=offer.validity)
+
+    def _inverse_validity(self):
+        for offer in self:
+            offer.validity = (offer.date_deadline - fields.Date.today()).days
+
+    def action_accepte_offer(self):
```
SaddemAmine: We always use English!
```suggestion
    def action_accept_offer(self):
```

estate/models/estate_offer.py
```diff
@@ -0,0 +1,64 @@
...
+        for offer in self:
+            if offer.status == "refused":
+                raise exceptions.UserError("Can't accept an refused offer")
```
SaddemAmine: What about an offer that was already accepted?

estate/models/estate_offer.py
```diff
@@ -0,0 +1,64 @@
...
+            offer.status = "accepted"
+            offer.property_id.accept_offer()
+        return True
+
+    def action_refuse_offer(self):
+        for offer in self:
+            if offer.status == "accepted":
+                raise exceptions.UserError("Can't refuse an accepted offer")
```
SaddemAmine: I'm not sure if the spec mentions this but I feel like it would be better to also not refuse an order that is already refused

estate/models/estate_offer.py
```diff
@@ -0,0 +1,64 @@
+from datetime import timedelta
+
+from odoo import api, exceptions, fields, models
+
+
+class EstateOffer(models.Model):
```
SaddemAmine: This is a nitpick but usually we match the class name to the model name (`_name`)
```suggestion
class EstatePropertyOffer(models.Model):
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+    _order = "id desc"
+
+    _check_positif_expected_price = models.Constraint(
+        'CHECK(expected_price >= 0)',
+        'The expected price must be positive'
+    )
+    _check_positif_selling_price = models.Constraint(
+        'CHECK(selling_price >= 0)',
+        'The selling price must be positive'
+    )
```
SaddemAmine, line 20: I might be wrong but I remember one of these checks has to have strict positivity

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+    description = fields.Text(string="Description")
+    postcode = fields.Char(string="Postcode")
+    expected_price = fields.Float(string="Expected Price", required=True)
+    selling_price = fields.Float(string="Selling Price", copy=False, readonly=True)
+    best_price = fields.Integer(string="Best Price", compute="_compute_best_price")
+    expected_date = fields.Date(
+        string="Expected Date",
+        required=True,
+        copy=False,
+        default=(fields.Date.today() + timedelta(days=90)),
```
SaddemAmine: I think it's better to use `months` instead of days to account for different month lengths and leap years

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+    living_area = fields.Integer(string="Living area (square metter)")
+    facades = fields.Integer(string="Number of facades")
+    garage = fields.Boolean(string="Have a garage?")
+    garden = fields.Boolean(string="Have a garden?")
+    garden_area = fields.Integer(string="Garden area (square metter)")
+    garden_orientation = fields.Selection(
+        string="Garden's orientation",
+        selection=[
+            ("north", "North"),
+            ("sud", "Sud"),
```
SaddemAmine:
```suggestion
            ("south", "South"),
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+    @api.depends('garden_area', 'living_area')
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends('proterty_offer_ids.price')
+    def _compute_best_price(self):
+        for record in self:
+            prices = record.proterty_offer_ids.mapped('price')
+            record.best_price = max(prices, default=0)
```
SaddemAmine: Same here for the best offer, we know they're sorted by desc price, so we can just use the first offer

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = 'north'
+        else:
+            self.garden_area = 0
+            self.garden_orientation = ''
+        # return {'warning': {'title': "Test warning", 'message': "foo", 'type': 'notification'},}
```
SaddemAmine: You should make sure not to leave any unintentional comments lying around

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+
+    @api.constrains("selling_price", "expected_price")
+    def _check_price_expectation(self):
+        for record in self:
+            if float_utils.float_is_zero(record.selling_price, 0):
```
SaddemAmine: This is inconsistent with the rest of the method
```suggestion
            if float_utils.float_is_zero(record.selling_price, 2):
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+                return True
+            if float_utils.float_compare(record.selling_price, record.expected_price * 0.9, 2) == -1:
+                raise exceptions.ValidationError("Selling price can't be smaller than 90% of the expected price")
```
SaddemAmine: Translation
```suggestion
                raise exceptions.ValidationError(self.env._("Selling price can't be smaller than 90% of the expected price"))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+        return True
+
+    @api.ondelete(at_uninstall=False)
+    def delete(self):
+        for record in self:
+            if record.state not in ["new", "cancelled"]:
+                raise exceptions.UserError(f"Can't delete an advertise in {record.state} state")
```
SaddemAmine: Same here about translation

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+        for advertisements in self:
+            if advertisements.state == "cancelled":
+                raise exceptions.UserError("Can't sold an canceled advertise")
+            advertisements.state = "sold"
+        return True
+
+    def action_cancel_adv(self):
+        for advertisements in self:
+            if advertisements.state == "sold":
+                raise exceptions.UserError("Can't cancel an sold advertise")
```
SaddemAmine: Here too

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+    @api.ondelete(at_uninstall=False)
+    def delete(self):
+        for record in self:
+            if record.state not in ["new", "cancelled"]:
+                raise exceptions.UserError(f"Can't delete an advertise in {record.state} state")
+
+    def action_sold_adv(self):
+        for advertisements in self:
+            if advertisements.state == "cancelled":
+                raise exceptions.UserError("Can't sold an canceled advertise")
```
SaddemAmine: Ditto

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+            if float_utils.float_is_zero(record.selling_price, 0):
+                return True
+            if float_utils.float_compare(record.selling_price, record.expected_price * 0.9, 2) == -1:
+                raise exceptions.ValidationError("Selling price can't be smaller than 90% of the expected price")
+        return True
+
+    @api.ondelete(at_uninstall=False)
+    def delete(self):
+        for record in self:
+            if record.state not in ["new", "cancelled"]:
```
SaddemAmine: It's slightly faster to use a set for membership checks
```suggestion
            if record.state not in {"new", "cancelled"}:
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+                    offer.status = 'refused'
+
+            accepted_offer = accepted_offer[0]
+            adrivertise.buyer_id = accepted_offer.partner_id
+            adrivertise.selling_price = accepted_offer.price
+            adrivertise.state = 'offer_accepted'
+
+    def set_offer_received(self):
+        for record in self:
+            record.state = "offer_recieved"
```
SaddemAmine: I'd rethink the variable names btw
the recordset contains estate.property records, not advertisements, so I'd say
`for property in self:` makes a lot more sense

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+                raise exceptions.UserError("Can't cancel an sold advertise")
+            advertisements.state = "cancelled"
+        return True
+
+    def accept_offer(self):
+        for adrivertise in self:
+            accepted_offer = []
+            for offer in adrivertise.proterty_offer_ids:
+                if accepted_offer:
+                    offer.status = "refused"
```
SaddemAmine: This probably never triggers as accepted_offer (bad name for a list btw) is Falsy

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+                elif offer.status == 'accepted':
+                    accepted_offer.append(offer)
+                else:
+                    offer.status = 'refused'
```
SaddemAmine: I think you can avoid this loop by just passing the accepted offer as a parameter and assigning refused the rest

`(self.offer_ids - accepted_offer).state = 'refused'`

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+        return True
+
+    def action_cancel_adv(self):
+        for advertisements in self:
+            if advertisements.state == "sold":
+                raise exceptions.UserError("Can't cancel an sold advertise")
+            advertisements.state = "cancelled"
+        return True
+
+    def accept_offer(self):
```
SaddemAmine, line 123: IMO since this method contains offer logic, it makes more sense for it to live on the offer model instead.
Keeping a method here that does the property side effect of accepting an offer makes more sense, then you call that method when you set the offer to accepted

estate/models/estate_offer.py
```diff
@@ -0,0 +1,64 @@
...
+    _name = "estate.property.offer"
+    _description = "Offer"
+    _order = "price desc"
+
+    price = fields.Float(string="Price")
```
SaddemAmine: The string attribute is useless since by default the label is the capitalized field name

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+        selection=[
+            ("north", "North"),
+            ("sud", "Sud"),
+            ("east", " East"),
+            ("west", "West"),
+        ],
+    )
+    total_area = fields.Integer(string="Total Area", compute="_compute_total_area")
+    state = fields.Selection(
+        string="Estate's state",
```
SaddemAmine, line 52: Just to let you know that this label won't show because we're using the state widget

estate/models/estate_tag.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class PropertyTag(models.Model):
+    _name = "estate.property.tag"
+    _description = "Estate Tag"
+    _order = "name"
+
+    name = fields.Char(string="Tag")
+
+    _name_idx = models.UniqueIndex('(name)', 'Another record already exists with the same name!')
```
SaddemAmine: This is overkill IMO, using `models.Constraint` should be enough

estate/models/estate_property.py
SaddemAmine: This must be wrapped in a lambda

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+        else:
+            self.garden_area = 0
+            self.garden_orientation = ''
+        # return {'warning': {'title': "Test warning", 'message': "foo", 'type': 'notification'},}
+
+    @api.constrains("selling_price", "expected_price")
+    def _check_price_expectation(self):
+        for record in self:
+            if float_utils.float_is_zero(record.selling_price, 0):
+                return True
```
SaddemAmine, line 98: If the first record in a multi-record recordset triggers this, we don't check the rest

estate/models/estate_property.py
```diff
@@ -0,0 +1,141 @@
...
+        string="Garden's orientation",
+        selection=[
+            ("north", "North"),
+            ("sud", "Sud"),
+            ("east", " East"),
+            ("west", "West"),
+        ],
+    )
+    total_area = fields.Integer(string="Total Area", compute="_compute_total_area")
+    state = fields.Selection(
```
SaddemAmine, line 51: This is missing required and no copy IIRC

estate/models/estate_offer.py
```diff
@@ -0,0 +1,64 @@
...
+
+    property_type_id = fields.Many2one(related="property_id.property_type_id", store=True)
+
+    @api.model
+    def create(self, vals_list):
+        for vals in vals_list:
+            property = self.env["estate.property"].browse(vals["property_id"])
+            property.set_offer_received()
+            max_price = max(property.proterty_offer_ids.mapped('price'), default=None)
+            if max_price and max_price > vals['price']:
```
SaddemAmine: Doesn't the spec require the price to be strictly higher?

estate/models/estate_type.py
```diff
@@ -0,0 +1,21 @@
+from odoo import api, fields, models
+
+
+class EstateType(models.Model):
+    _name = "estate.property.type"
+    _description = "Estate Type"
+    _order = "name"
```
SaddemAmine: You need to sort by sequence first to be able to keep the order of the dragged records in the kanban view

estate/models/estate_tag.py
```diff
@@ -0,0 +1,12 @@
...
+    color = fields.Integer(string="Color")
```
SaddemAmine: Useless `string`

estate/models/estate_type.py
```diff
@@ -0,0 +1,21 @@
...
+
+    name = fields.Char(string="Type", required=True)
+    _name_idx = models.UniqueIndex('(name)', 'Another record already exists with the same name!')
```
SaddemAmine: I think this is overkill :/

estate/models/estate_type.py
```diff
@@ -0,0 +1,21 @@
...
+    property_ids = fields.One2many('estate.property', 'property_type_id')
+
+    sequence = fields.Integer('Sequence', default=1, help="Used to order stages. Lower is better.")
+
+    offer_ids = fields.One2many('estate.property.offer', 'property_type_id')
+    offer_count = fields.Integer(compute="_compute_offer_count")
```
SaddemAmine, line 18: Attributes order
https://www.odoo.com/documentation/19.0/nl/contributing/development/coding_guidelines.html

estate/models/res_user.py
SaddemAmine, line 1: We like to keep the same name
This file should still be named res_users.py, same for the class name

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_model,access_estate_model,model_estate_property,base.group_user,1,1,1,1
+access_state_type_model,access_state_type_model,model_estate_property_type,base.group_user,1,1,1,1
+access_state_tag_model,access_state_tag_model,model_estate_property_tag,base.group_user,1,1,1,1
+access_state_offer_model,access_state_offer_model,model_estate_property_offer,base.group_user,1,1,1,1
```
SaddemAmine:
```suggestion
access_estate_type_model,access_state_type_model,model_estate_property_type,base.group_user,1,1,1,1
access_estate_tag_model,access_state_tag_model,model_estate_property_tag,base.group_user,1,1,1,1
access_estate_offer_model,access_state_offer_model,model_estate_property_offer,base.group_user,1,1,1,1
```

estate/view/estate_property_views.xml
```diff
@@ -0,0 +1,163 @@
...
+        <field name="view_mode">list,form</field>
+        <field name="domain">[('property_type_id', '=', active_id)]</field>
+    </record>
+
+    <!-- List view -->
+    <record id="estate_property_view_list" model="ir.ui.view">
+        <field name="name">estate.property.list</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <list string="Properties" decoration-success="state in ['offer_recieved','offer_accepted']" decoration-muted="state=='sold'" decoration-it="state=='offer_accepted'">
```
SaddemAmine: `offer_recieved` careful, there a typo here and I saw it a few more times across the PR if I'm not mistaken

estate/view/estate_property_views.xml
```diff
@@ -0,0 +1,163 @@
...
+                            <field name="property_type_id"/>
+                        </group>
+                    </group>
+                    <notebook>
+                        <page string="Description">
+                            <field name="description" string="Description"/>
+                            <group>
+                                <group>
+                                    <field name="bedroom"/>
+                                    <field name="living_area" string="Living Area (sqm)" filter_domain="[('living_area', '>=', self)]"/>
```
SaddemAmine: What's the use of having a filer_domain here? This is the form view

estate/view/estate_property_views.xml
```diff
@@ -0,0 +1,163 @@
...
+    <record id="estate_property_search" model="ir.ui.view">
+        <field name="name">estate.property.search</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <search string="Estate Type">
+                <field name="name" string="Estate" />
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedroom"/>
+                <field name="living_area"/>
```
SaddemAmine: You need the filter_domain here otherwise this only looks for exact values

estate/models/res_user.py
```diff
@@ -0,0 +1,7 @@
+from odoo import fields, models
+
+
+class InhertiedUser(models.Model):
+    _inherit = "res.users"
+
+    property_ids = fields.One2many("estate.property", "salesman_id")
```
SaddemAmine, line 7: IIRC this needs a domain

estate/view/estate_property_type_view.xml
```diff
@@ -0,0 +1,51 @@
...
+
+    <!-- Form view -->
+    <record id="estate_property_type_view_form" model="ir.ui.view">
+        <field name="name">estate.property.type.form</field>
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <form string="Estate Property Type">
+                <header>
+                    <button type="action" name="%(estate.estate_properties_offers_view)d" string="Offer" class="oe_stat_button" invisible="not offer_count"/>
+                </header>
```
SaddemAmine: What about the count and icon?

estate/view/estate_property_tag_view.xml
```diff
@@ -0,0 +1,19 @@
...
+        <field name="res_model">estate.property.tag</field>
+        <field name="view_mode">list,form</field>
+    </record>
+
+    <!-- List view -->
+    <record id="estate_property_tag_view_list" model="ir.ui.view">
+        <field name="name">estate.property.tag.list</field>
+        <field name="model">estate.property.tag</field>
+        <field name="arch" type="xml">
+            <list string="Properties" editable="bottom">
```
SaddemAmine: The string seems wrong? This is the tags list

estate/view/estate_property_type_view.xml
```diff
@@ -0,0 +1,51 @@
...
+                <sheet>
+                    <h1>
+                        <field name="name" placeholder="e.g House"/>
```
SaddemAmine: Never used placeholder before but that's cool!
```suggestion
                        <field name="name" placeholder="e.g. House"/>
```

estate/view/estate_property_views.xml
```diff
@@ -0,0 +1,163 @@
...
+                <field name="facades"/>
+                <separator/>
+                <filter string="Available" name="available" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_recieved')]"/>
```
SaddemAmine: Keep in mind that you can write this in a more readable way
```suggestion
                <filter string="Available" name="available" domain="[('state', '=', {'new', 'offer_recieved'}]"/>
```

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+{
+    'name': 'Estate',
+    'depends':  ['base'],
+    'application': True,
+    'author': 'toher',
```
SaddemAmine, line 5: Keep in mind that this is usually Odoo, since we work here.

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def action_sold_adv(self):
+        self._create_property_invoice()
+        return super().action_sold_adv()
+
+    def _create_property_invoice(self):
+        for property in self:
+            self.env["account.move"].with_context(default_move_type='out_invoice').create({
+                "partner_id": property.buyer_id.id,
+                "move_type": "out_invoice",
```
SaddemAmine, line 15: I guess you don't need the context since you specify the move_type in the dict

estate_account/__manifest__.py
```diff
@@ -0,0 +1,9 @@
+{
+    'name': 'Estate Account',
+    'depends':  ['estate', 'account'],
+    'author': 'toher',
+    'license': 'LGPL-3',
+
+    'data': [
+    ],
```
SaddemAmine: I wonder if you can just remove this

SaddemAmine (changes requested):
Thank you for the PR Tom!

I invite you to read the git guidelines to understand more how to structure your commit messages, since you're currently violating how it is usually done. 

Reading the coding guidelines is also a very good idea, I noticed you have lots of typos and naming issues throughout, focusing on them will help you integrate better in teams with big codebases where your diffs will need to be consistent with the rest.

estate/models/estate_offer.py
Tom-Hermann: If the offer has already been rejected or accepted, the button is disabled. In the worst-case scenario, the value is automatically overwrite by himself.

PR #1369, opened 2026-07-20
[ADD] estate: Initial setup for module

sdemeesterde:
Hello,

Happy to make my first PR in odoo.


estate/models/estate_property.py
```diff
@@ -0,0 +1,136 @@
...
+    _order = "id desc"
+
+    active = fields.Boolean(default=True)
+    name = fields.Char("Title", required=True, translate=True)
+    description = fields.Text("Notes")
+    postcode = fields.Char("Postcode", required=True)
+    date_availability = fields.Date(
+        "Availability date",
+        copy=False,
+        default=date.today() + relativedelta(months=3),
```
YassinWalid:
```suggestion
        default=fields.Date.today() + relativedelta(months=3),
```
you should use odoo's `fields.Date.today()` instead of `date.today()`
You could also check out `fields.Date.add()`, which might eliminate the need for `relativedelta`

estate/models/estate_property.py
```diff
@@ -0,0 +1,136 @@
...
+from odoo.tools import float_compare
+
+
+class EstateProperty(models.Model):
+    _name = "realestate.properties"
+    _description = "Real estate properties"
+    _order = "id desc"
+
+    active = fields.Boolean(default=True)
+    name = fields.Char("Title", required=True, translate=True)
```
YassinWalid:
```suggestion
    name = fields.Char("Title", required=True)
```
This is not needed, as it is translated by default

estate/models/estate_property.py
```diff
@@ -0,0 +1,136 @@
...
+    def _compute_best_price(self):
+        for record in self:
+            prices = record.offer_ids.mapped("price")
+            record.best_offer = max(prices, default=0)
+
+    @api.depends("offer_ids")
+    def _compute_selling_price(self):
+        for record in self:
+            accepted_offer = record.offer_ids.filtered(lambda o: o.status == "accepted")
+            record.selling_price = accepted_offer.price if accepted_offer else 0.0
```
YassinWalid: This will probably work, but whenever you add a new offer for example, it will be called, which might be unnecessary if this offer will is be accepted. I believe doing this in the action for accepting the offer works best performance wise.

estate/models/estate_property.py
```diff
@@ -0,0 +1,136 @@
...
+
+    @api.onchange("offer_ids")
+    def _onchange_offer(self):
+        for record in self:
+            if record.state == "new" and len(record.offer_ids) > 0:
+                record.state = "offer received"
```
YassinWalid: Same thing here, we could do that when creating a new offer as well

estate/models/estate_property.py
```diff
@@ -0,0 +1,136 @@
...
+            if record.state == "cancelled":
+                raise exceptions.UserError(_("Cancelled properties cannot be sold"))
+
+            record.state = "sold"
+
+    def cancelled_action_btn(self):
+        for record in self:
+            if record.state == "sold":
+                raise exceptions.UserError(_("Sold properties cannot be cancelled"))
+            record.state = "cancelled"
```
YassinWalid:
```suggestion
    def action_sold(self):
        for record in self:
            if record.state == "cancelled":
                raise exceptions.UserError(_("Cancelled properties cannot be sold"))

            record.state = "sold"

    def action_cancel(self):
        for record in self:
            if record.state == "sold":
                raise exceptions.UserError(_("Sold properties cannot be cancelled"))
            record.state = "cancelled"
```
According to the coding guidelines, the name of action methods should be prefixed with `action_`

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+
+    @api.depends("validity")
+    def _computed_date_deadline(self):
+        for offer in self:
+            create_date = offer.create_date if offer.create_date else datetime.today()
+            if offer.validity:
+                offer.date_deadline = create_date + timedelta(
+                    days=offer.validity,
+                )
```
YassinWalid:
```suggestion
    @api.depends("validity", "create_date")
    def _computed_date_deadline(self):
        for offer in self:
            create_date = offer.create_date if offer.create_date else datetime.today()
            if offer.validity:
                offer.date_deadline = create_date + timedelta(
                    days=offer.validity,
                )
```
Since the compute also depends on `create_date` in the computation, we should also include it in the `@api.depends()`

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,68 @@
...
+            offer.property_id.selling_price = offer.price
+            offer.property_id.buyer_id = offer.partner_id
+
+    def action_refuse(self):
+        for offer in self:
+            offer.status = "refused"
+            if offer.property_id.state == "offer accepted":
+                offer.property_id.state = "new"
+            offer.property_id.selling_price = 0
+            offer.property_id.buyer_id = None
```
YassinWalid: What if another offer was accepted and not this one? Wouldn't this action mess the data up while it was not even the accepted offer. Also, if this was the accepted offer, there might be other offers existing, so the state of the property does it necessarily have to be "new"

estate/models/estate_property.py
```diff
@@ -0,0 +1,136 @@
...
+    )
+    _check_selling_price = models.Constraint(
+        "CHECK(selling_price >= 0)",
+        _("The selling price should be strictly positive"),
+    )
+    state = fields.Selection(
+        [
+            ("new", "New"),
+            ("offer received", "Offer Received"),
+            ("offer accepted", "Offer Accepted"),
```
YassinWalid:
```suggestion
            ("offer_received", "Offer Received"),
            ("offer_accepted", "Offer Accepted"),
```
selection keys should be snake-cased

estate/models/estate_property.py
```diff
@@ -0,0 +1,136 @@
+from datetime import date
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, _, exceptions, fields, models
+from odoo.orm.utils import ValidationError
+from odoo.tools import float_compare
+
+
+class EstateProperty(models.Model):
+    _name = "realestate.properties"
```
YassinWalid: please check the coding guidelines for the naming convention of the models.

estate/security/ir.model.access.csv
YassinWalid, line 1: for the name, it should be similar to the id but with dots instead of underscores. Also, we don't need to specify the module name in the id column, since it is there by default.

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,14 @@
...
+                action="estate_advertisement_property_action" />
+        </menuitem>
+        <menuitem id="settings_menu" name="Settings">
+            <menuitem id="estate_settings_property_type_action_id"
+                action="estate_settings_property_type_action" />
+            <menuitem id="estate_settings_property_tag_action_id"
+                action="estate_settings_property_tag_action" />
+        </menuitem>
+    </menuitem>
+</odoo>
```
YassinWalid:
```suggestion
</odoo>

```
Don't forget to add the empty line at the end of every file, since this allows better code tracability (git blame), in case someone decides to add a line after your last line

estate/views/estate_property_offers_list.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- the root elements of the data file -->
```
YassinWalid:
```suggestion
```
This comment in the documentation was for explanation purposes. You don't need to put it in the files

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,47 @@
...
+        </field>
+    </record>
+
+    <record id="estate_property_tag_views_filter" model="ir.ui.view">
+        <field name="name">realestate.property.tag.filter</field>
+        <field name="model">realestate.properties.tag</field>
+        <field name="arch" type="xml">
+            <search string="Filter string">
+                <field name="name" />
+                <filter string="Group by name" name="name"
```
YassinWalid: why is this needed? Didn't we make the name field unique anyway?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,115 @@
...
+    <record id="estate_property_custom_view" model="ir.ui.view">
+        <field name="name">realestate.property.form</field>
+        <field name="model">realestate.properties</field>
+        <field name="arch" type="xml">
+            <form string="Custom description form">
+                <header>
+                    <button name="sold_action_btn" type="object" string="Property sold"
+                        invisible="state == 'sold'" />
+                    <button name="cancelled_action_btn" type="object" string="Property cancelled"
+                        invisible="state == 'sold'" />
```
YassinWalid: Do we want to show the action buttons even when the property is cancelled? in my opinion, trying to sell/cancel an already cancelled property does not make sense (this is still just an opinion, if you think otherwise let me know)

YassinWalid (changes requested):
Thanks for the impressive work. I have done an initial review, and left some comments in the PR and other general ones below. Please check them out, especially the coding guidelines and how to format the name/id of python fields and methods, xml record names and ids, as well as the model names themselves

Instead of date and datetime, you should use `fields.Date` and `fields.Datetime`

For the naming convention of records' name/id, name of models, fields, the order of methods for a model, and much more, check the [coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html) for useful tips/rules.

estate/models/estate_property.py
sdemeesterde: Initially I did something similar to what you describe, but ran into some issue:
There is a constraint between "expected price" and "selling price" (the selling price must be at least 90% of the expected price). If the user modifies the expected price, the constraint might be broke and in that case if the use decides to update its price for the same offer, then the selling price should also be updated.

But in reality later on we disable the option to modify an already accepted offer. Which makes more sense business logic wise.

So long story short, it was meant to cover an edge case removed by additional business logic ! So I agree :)

sdemeesterde:
Thank you for the many comments ! I really appreciate all the feedback, and tried to fix all the issues.

I still have a warning from the ci about the translation, but should I care about it ?

estate/models/res_users.py
YassinWalid, line 1: The name of the file should match the name of the model, so "res_users.py" (the same applies for the class name as well

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def action_refuse(self):
+        for offer in self:
+            offer.status = "refused"
+            if offer.property_id.state == "offer_received" and any(
+                s != "refused" for s in offer.mapped("status")
+            ):
+                offer.property_id.state = "new"
+            offer.property_id.selling_price = 0
+            offer.property_id.buyer_id = None
```
YassinWalid: What if another offer was accepted, wouldn't these two lines mess the `buyer_id` and  and `selling_price` of the property?

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,6 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_properties,estate.property,model_estate_property,base.group_user,1,1,1,1
+access_estate_properties_type,estate.property.type,model_estate_property_type,base.group_user,1,1,1,1
+access_estate_properties_tag,estate.property_tag,model_estate_property_tag,base.group_user,1,1,1,1
+access_estate_properties_offer,estate.property.offer,model_estate_property_offer,base.group_user,1,1,1,1
+access_estate_res_users,estate.res.users,model_res_users,base.group_user,1,1,1,1
```
YassinWalid: why are we giving full access to the `res.users` model to all base users? (maybe you have a reasoning behind it that I don't know, so you can convince me if you want)

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,90 @@
...
+            offer.property_id.state = "offer_accepted"
+            offer.property_id.selling_price = offer.price
+            offer.property_id.buyer_id = offer.partner_id
+
+    def action_refuse(self):
+        for offer in self:
+            offer.status = "refused"
+            if offer.property_id.state == "offer_received" and any(
+                s != "refused" for s in offer.mapped("status")
+            ):
```
YassinWalid: This `offer.mapped("status")` will only return the status of the `offer` in this iteration, which you have just assigned "refused" to

estate/views/user_properties_views.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<odoo>
+    <record id="user_properties_form" model="ir.ui.view">
+        <field name="name">user.properties.form</field>
```
YassinWalid: the name of the inherited view should include an `.inherit.{details}` according to the [coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#inheriting-xml).

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,36 @@
+from odoo import Command, models
+
+
+class PropertyAccount(models.Model):
```
YassinWalid: The class name should be the same in the inherit as the original model (here it would be "EstateProperty")

estate_account/security/ir.model.access.csv
YassinWalid, line 1: why do we even need this file if we're not creating any `ir.model.access` records?

YassinWalid (commented):
Thanks for your work. I left some final comments on the PR, as well as some general ones below.

For the commit message titles, they should be [TAG] <module_name>: title. So in your case, they should have "estate" instead of "Real Estate". Also, the commit that introduces `estate_account` should have the tag [ADD] not [IMP].

Regardig the warnings, they're because translations should be done as `self.env._` instead of `_` since 18.0

estate/models/estate_property_offer.py
sdemeesterde: Yes and it has become irrelevant/wrong because the business logic changed preventing an already accepted offer to be cancelled.

sdemeesterde:
Thanks for the review, I implemented the necessary changes.

YassinWalid (commented):
Thanks for your work.
One tiny further comment: For 528d7a90d29528b75d4f7248d2d8d541c96e04f6 the title should have "estate_account" instead of "estate", since this is the module you're adding.

PR #1370, opened 2026-07-20
[ADD] estate: add estate module for tutoriel

jeanbfly:


estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
+from odoo import api, fields, models
+from odoo.exceptions import UserError
+from odoo.orm.utils import ValidationError
```
YassinWalid:
```suggestion
from odoo.exceptions import UserError, ValidationError
```
`ValidationError` is alos imported from `odoo.exceptions`

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+    garden_area = fields.Integer("Garden Area (sqm)")
+    garden_orientation = fields.Selection(
+        selection=[
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
+        ],
+    )
+    last_seen = fields.Datetime("Last Seen", default=fields.Datetime.now)
```
YassinWalid, line 56: why is this field needed?

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+    total_area = fields.Integer(compute="_total_area", readonly=True)
+    best_price = fields.Float(compute="_best_price", readonly=True, string="Best Offer")
+
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = "north"
+            return
+        self.garden_orientation = self.garden_area = None
```
YassinWalid:
```suggestion
        self.garden_orientation = 0
        self.garden_area = False
```
Since each field has a different type, each gets cleared in a different way. Also, we assign each value on a different line.

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+        for record in self:
+            record.best_price = max(record.offer_ids.mapped("price"), default=0)
+
+    @api.depends("living_area", "garden_area")
+    def _total_area(self):
+        for record in self:
+            record.total_area = record.garden_area + record.living_area
+
+    @api.depends("selling_price", "expected_price")
+    @api.constrains("selling_price")
```
YassinWalid:
```suggestion
    @api.constrains("selling_price", "expected_price")
```
you don't need to have a depends alongside the constrains. You can put both fields on the `@api.constrains` right away.

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+        if self.state == "sold":
+            e = "A sold property cannot be cancelled"
+            raise UserError(e)
+        self.state = "cancelled"
+
+    def sold(self):
+        if self.state == "cancelled":
+            e = "A cancelled property cannot be sold"
+            raise UserError(e)
+        self.state = "sold"
```
YassinWalid: the action functions are usually named action_<something>, so we could rename these to action_cancel and action_sold.
Also, don't forget to translate the error messages

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,48 @@
...
+    partner_id = fields.Many2one("res.partner", required=True, string="Partner")
+    property_id = fields.Many2one("estate.property", required=True)
+    validity = fields.Integer(default=7, string="Validity (days)")
+    date_deadline = fields.Datetime(
+        compute="_date_deadline",
+        inverse="_set_date_deadline",
+    )
+
+    @api.depends("create_date", "validity")
+    def _date_deadline(self):
```
YassinWalid: the name of the computes should have the word compute in it (for example here, it would be `_compute_date_deadline`)

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,48 @@
...
+        for record in self:
+            record.date_deadline = fields.Datetime.add(
+                record.create_date or fields.Datetime.now(),
+                days=record.validity,
+            )
+
+    def _set_date_deadline(self):
```
YassinWalid: the inverse should be called `_inverse_date_deadline`

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,48 @@
...
+    date_deadline = fields.Datetime(
+        compute="_date_deadline",
+        inverse="_set_date_deadline",
+    )
+
+    @api.depends("create_date", "validity")
+    def _date_deadline(self):
+        for record in self:
+            record.date_deadline = fields.Datetime.add(
+                record.create_date or fields.Datetime.now(),
```
YassinWalid: since `create_date` is a datetime field, we should only get the date part of it to calculate the date_deadline here and the validity in the inverse, to avoid unneeded differences. (it would be `record.create_date.date()`)

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,48 @@
...
+        self.property_id.offer_ids.status = "refused"
+        self.status = "accepted"
+        self.property_id.state = "offer accepted"
+        self.property_id.selling_price = self.price
+        self.property_id.buyer_id = self.partner_id
+
+    def refuse(self):
+        self.status = "refused"
+        self.property_id.selling_price = None
+        self.property_id.buyer_id = None
```
YassinWalid: Since the actions may be called on multiple records, we should loop over them to apply the action to all of them.
For the `action_accept`, what if there is already an accepted offer?
Also, what if this offer was not the best one, wouldn't unassigning the `buyer_id` and `selling_price` in the `action_refuse` yield incorrect data?

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,1
+estate.access_estate_property_type,access_estate_property_type,estate.model_estate_property_type,base.group_user,1,1,1,1
+estate.access_estate_property_tag,access_estate_property_tag,estate.model_estate_property_tag,base.group_user,1,1,1,1
+estate.access_estate_property_offer,access_estate_property_offer,estate.model_estate_property_offer,base.group_user,1,1,1,1
```
YassinWalid:
```suggestion
id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
access_estate_property,access.estate.property,estate.model_estate_property,base.group_user,1,1,1,1
access_estate_property_type,access.estate.property.type,estate.model_estate_property_type,base.group_user,1,1,1,1
access_estate_property_tag,access.estate.property.tag,estate.model_estate_property_tag,base.group_user,1,1,1,1
access_estate_property_offer,access.estate.property.offer,estate.model_estate_property_offer,base.group_user,1,1,1,1
```
you don't need to specify the name of the module for the id, since it's by default the current one.
Also, names should be the same as ids, by just replacing the underscores with dots (refer to the coding guidelines).

estate/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "Real Estate",
+    "author": "Jean (jebou)",
```
YassinWalid: this is okay since it's only a tutorial, but normally for normal tasks, it's `Odoo S.A.`.

YassinWalid (changes requested):
Thanks for the incredible work. As an initial review, I left some comments in the PR, and some general ones below.

Don't forget to add an empty line at the end of each file. We mostly do this for a clean history, since if somebody adds a line at the end of the file, they'll technically be changing your line as well, messing up the git blame or history of that specific line.

For the order of attributes in the model, the naming of the xml ids, and other guidelines, you can follow the [coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html).

One minor comment, why do you put the string of your views "Channel", "Test", etc.?

estate/models/estate_property.py
jeanbfly: it was an exercice in the tutoriel to test the fields.Datetime.now method. Should I remove it ?

estate/__manifest__.py
jeanbfly: good to know, it was to pass to CI checks

estate/models/__init__.py
YassinWalid, line 1: nitpick(feel free to ignore it for now): we usually do the imports as `from . import <model>`

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,79 @@
...
+                )
+
+            self.env["estate.property"].browse(property_id).state = "offer received"
+        return super().create(vals_list)
+
+    @api.depends("create_date", "validity")
+    def _compute_date_deadline(self):
+        for record in self:
+            record.date_deadline = fields.Datetime.add(
+                record.create_date.date() or fields.Datetime.now().date(),
```
YassinWalid, line 61: what if the `create_date` is false? this would cause an error. Also, there is a `fields.Date.today()`, so we don't need to do the `fields.Datetime.now().date()` imo

YassinWalid (commented):
Thanks for the great work. I have one general comment. For the commit titles, the [ADD] tag should be used when adding a new module, but otherwise, for most of the other commits (except for the `estate_account` one), they are improvements to the `estate` module, so they should have the [IMP] tag. Also, don't forget to add the empty line to every file, not just the python files.

PR #1371, opened 2026-07-22
19.0 estate cebul

cezarbulancea:
My first odoo pr!


estate/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+from . import (
+    estate_property,
+    estate_property_type,
+    estate_property_tag,
+    estate_property_offer,
+    estate_property_users,
+)
```
SaddemAmine: This is the more commonly used way to handle this in Odoo but technically this works fine
Up to you to choose if you want to opt for that instead, personally I don't care much but I do still prefer the way we usually write the imports
```suggestion
from . import estate_property
from . import estate_property_type
from . import estate_property_tag
from . import estate_property_offer
from . import estate_property_users
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,115 @@
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+
+class Property(models.Model):
```
SaddemAmine: Nitpick, the class name usually follows `_name`
```suggestion
class EstateProperty(models.Model):
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
+        string="State",
+        selection=[('new', 'New'), ('offer received', 'Offer Received'),
+                   ('offer accepted', 'Offer Accepted'), ('sold', 'Sold'), ('cancelled', 'Cancelled')],
+        required=True,
+        copy=False,
+        default='new'
+    )
```
SaddemAmine, line 53: Another (a bit important though) nit is that we like to use trailing commas, this avoids needing an extra step to see previous commits when blaming code. I won't make more comments for this but feel free to add them for the rest of your diff too!

```suggestion
    garden_orientation = fields.Selection(
        string="Garden Orientation",
        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')],
    )
    active = fields.Boolean(default=True)
    state = fields.Selection(
        string="State",
        selection=[('new', 'New'), ('offer received', 'Offer Received'),
                   ('offer accepted', 'Offer Accepted'), ('sold', 'Sold'), ('cancelled', 'Cancelled')],
        required=True,
        copy=False,
        default='new',
    )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,115 @@
...
+            return
+
+        self.garden_area = None
+        self.garden_orientation = None
+
+    @api.ondelete(at_uninstall=False)
+    def _check_state(self):
+        for record in self:
+            if record.state == "new" or record.state == "cancelled":
+                raise UserError(self.env._("Can't delete a new or cancelled property."))
```
SaddemAmine: This does the opposite of what it's supposed to do

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,69 @@
...
+
+        self.property_id.state = 'offer accepted'
+
+        return True
+
+    def action_refuse_offer(self):
+        self.ensure_one()
+        self.status = 'refused'
+
+        return True
```
SaddemAmine, line 89: What about adding guards here? This logic means we can accept multiple offers on a property
Edit: As per our conversation, it's alright to use ensure_one when we only want to enforce that the method would be only called on one record. I find it makes more sense to allow multi-recordsets here through!

estate/models/estate_property.py
```diff
@@ -0,0 +1,115 @@
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+
+class Property(models.Model):
+    _name = 'estate.property'
+    _description = "Estate Property"
+    _order = "id desc"
```
SaddemAmine: Usually we keep the quotes consistent
Personally I like using single quotes for technical strings and double quotes for user facing ones!
```suggestion
    _name = 'estate.property'
    _description = "Estate Property"
    _order = 'id desc'
```

estate/__manifest__.py
SaddemAmine, line 1: Normally, we would also add author and license keys. Author would be "Odoo S.A."

estate/models/estate_property_users.py
SaddemAmine, line 1: The way you named the file / class conveys that this is a new `estate.property.user` model.
For inheritance, we actually keep the same name of the model we're inheriting from.

The file would be res_users.py, class should also be ResUsers

estate/models/estate_property_users.py
```diff
@@ -0,0 +1,8 @@
+from odoo import fields, models
+
+
+class EstatePropertyUsers(models.Model):
+    _inherit = "res.users"
+
+    property_ids = fields.One2many("estate.property", "salesman",
+                                   domain=[('state', 'in', ('new','offer received'))])
```
SaddemAmine: There's a style issue here, I'll leave it for you to figure out!

estate/security/ir.model.access.csv
SaddemAmine, line 1: Why is there no rule for `model_res_users` in this file?

estate/models/estate_property.py
```diff
@@ -0,0 +1,115 @@
...
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        default=lambda self: fields.Date.add(fields.Date.today(), months=3),
```
SaddemAmine: ~~Just to let you know that this could be done without the lambda~~
```suggestion
        default=fields.Date.add(fields.Date.today(), months=3),
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,115 @@
...
+        copy=False
```
SaddemAmine: Nitpick: it's preferred to use trailing commas to avoid a misleading blame when adding more keys/params later on.
```suggestion
        copy=False,
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float(readonly=True)
```
SaddemAmine: IIRC this needs `copy=False`

estate/models/estate_property.py
```diff
@@ -0,0 +1,115 @@
...
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer()
```
SaddemAmine: I think we can add a label saying this value is in sqm here

estate/models/estate_property.py
```diff
@@ -0,0 +1,115 @@
...
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        string="Garden Orientation",
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
+    )
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
+        string="State",
+        selection=[('new', 'New'), ('offer received', 'Offer Received'),
+                   ('offer accepted', 'Offer Accepted'), ('sold', 'Sold'), ('cancelled', 'Cancelled')],
```
SaddemAmine: Keys should follow Odoo variable naming conventions
Also styling
```suggestion
        selection=[
                ('new', "New"),
                ('offer_received', "Offer Received"),
                ('offer_accepted', "Offer Accepted"),
                ('sold', "Sold"),
                ('cancelled', "Cancelled"),
        ],
```

Read more [here](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions)

estate/models/estate_property.py
```diff
@@ -0,0 +1,115 @@
...
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        string="Garden Orientation",
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
+    )
+    active = fields.Boolean(default=True)
+    state = fields.Selection(
+        string="State",
```
SaddemAmine: This line is useless since by default the label is a capitalized version of the field name

estate/models/estate_property.py
```diff
@@ -0,0 +1,115 @@
...
+        selection=[('new', 'New'), ('offer received', 'Offer Received'),
+                   ('offer accepted', 'Offer Accepted'), ('sold', 'Sold'), ('cancelled', 'Cancelled')],
+        required=True,
+        copy=False,
+        default='new'
+    )
+    property_type_id = fields.Many2one("estate.property.type", string="Property Type")
+    salesman = fields.Many2one("res.users", default=lambda self: self.env.user)
+    buyer = fields.Many2one("res.partner", copy=False)
```
SaddemAmine: Relational fields must be suffixed with `_id`/`_ids`

estate/models/estate_property.py
```diff
@@ -0,0 +1,115 @@
...
+    tag_ids = fields.Many2many("estate.property.tag")
+    offer_ids = fields.One2many("estate.property.offer", "property_id")
+
+    total_area = fields.Integer(compute="_compute_total_area")
+    best_price = fields.Float(compute="_compute_best_offer")
+
+    _check_expected_price = models.Constraint(
+        'CHECK(expected_price > 0)',
+        'The expected price must be strictly positive.',
```
SaddemAmine: If you follow the quotes guideline I mentioned this would be 
```suggestion
        "The expected price must be strictly positive.",
```

As you prefer!

estate/models/estate_property.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+
+    _check_selling_price = models.Constraint(
+        'CHECK(selling_price >= 0)',
+        'The selling price must be positive.',
```
SaddemAmine: Same here

estate/models/estate_property.py
```diff
@@ -0,0 +1,115 @@
...
+    @api.constrains('selling_price')
+    def _check_selling_price(self):
+        for record in self:
+            if not float_is_zero(record.selling_price, precision_digits=2):
+                if float_compare(
+                        record.selling_price,
+                        0.9 * record.expected_price,
+                        precision_digits=2) < 0:
+                    raise ValidationError(self.env._(
+                        "The selling price should be at least 90% the expected price"))
```
SaddemAmine:
```suggestion
                    raise ValidationError(
                        self.env._("The selling price should be at least 90% the expected price")
                    )
```

Or all in one line

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,69 @@
...
+        'An offer price must be strictly positive',
+    )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            property = self.env["estate.property"].browse(vals["property_id"])
+
+            if property.offer_ids:
+                max_offer = max(property.mapped("offer_ids.price"))
```
SaddemAmine: `_order = "price desc"` means we know `offer_ids[0]` is the maximum

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,69 @@
...
+                if float_compare(vals["price"], max_offer, precision_digits=2) <= 0:
+                    raise UserError(f"The offer must be strictly higher than {max_offer:.2f}")
```
SaddemAmine: Translation, make sure you format the string properly
You find example in the docs for this one I think, also obviously in the codebase

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,69 @@
...
+
+            property.state = 'offer received'
+
+        return super().create(vals_list)
+
+
+    @api.depends("create_date", "validity")
+    def _compute_date_deadline(self):
+        for record in self:
+            create_date = record.create_date if record.create_date else fields.Date.today()
```
SaddemAmine:
```suggestion
            create_date = record.create_date or fields.Date.today()
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,69 @@
...
+            record.date_deadline = fields.Date.add(create_date, days=record.validity)
+
+    def _inverse_date_deadline(self):
+        for record in self:
+            create_date = record.create_date if record.create_date else fields.Date.today()
```
SaddemAmine:
```suggestion
            create_date = record.create_date or fields.Date.today()
```

estate/models/estate_property_offer.py
SaddemAmine, line 89: It doesn't make sense to refuse an offer that is already accepted.

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class PropertyType(models.Model):
+    _name = 'estate.property.type'
+    _description = "Estate Property Type"
+    _order = "sequence, name"
+
+    name = fields.Char(required=True)
+    property_ids = fields.One2many("estate.property", "property_type_id")
+    sequence = fields.Integer('Sequence')
```
SaddemAmine, line 11: ~~The doc specifies a default value of 10 on the sequence~~
Nvmm 🥴

It is not a bad idea in some cases to do that though, not particularly useful here

estate/models/estate_property_users.py
SaddemAmine: It is slightly faster to use a set for membership checks!
```suggestion
                                   domain=[('state', 'in', {'new', 'offer received'}))])
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,142 @@
...
+                        <h1>
+                            <field name="name" placeholder="Property Name"/>
+                        </h1>
+                    </div>
+                    <group>
+                        <field name="tag_ids" widget="many2many_tags" options="{'color_field': 'color'}"/>
+                    </group>
+                    <group>
+                        <group>
+                            <field name="property_type_id" string="Property Type"/>
```
SaddemAmine: Isn't this supposed to have no create?
```suggestion
                            <field name="property_type_id" string="Property Type" options="{'no_create': True}"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,142 @@
...
+
+    <record id="estate_property_view_list" model="ir.ui.view">
+        <field name="name">estate.property.list</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <list decoration-success="state in ('offer received', 'offer accepted')"
+                  decoration-bf="state=='offer accepted'"
+                  decoration-muted="state=='sold'" string="Properties">
+                <field name="name" string="Title" width="15%"/>
+                <field name="property_type_id" options="{'no_create': True}" string="Type" width="10%"/>
```
SaddemAmine: I think the no_create here is useless since the list isn't editable anyway...

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,26 @@
...
+                    }),
+                    Command.create({
+                        'name': "administrative fees",
+                        'quantity': 1,
+                        'price_unit': 100
+                    })
+                ]
+            }
+        )
+        return super().action_sold_property()
```
SaddemAmine: From a defensive programming standpoint, it's better to call `action_sold_property` before creating the `account.move`. For the purposes of this tutorial, not doing this has no real drawback, but usually, I'd ask you to reverse the order.

SaddemAmine (commented):
Here's a few comments for you. Good work!

estate/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+from . import estate_property
+from . import estate_property_type
+from . import estate_property_tag
+from . import estate_property_offer
+from . import res_users
```
SaddemAmine, line 5: Nitpick: alphabetically sorting these is also recommended btw

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,81 @@
...
+            record.property_id.state = 'offer_accepted'
+
+        return True
+
+    def action_refuse_offer(self):
+        for record in self:
+            if record.status != 'accepted':
+                record.status = 'refused'
+
+        return True
```
SaddemAmine, line 89: I still think this logic isn't perfect

What happens when you accept an offer that has been already refused?
Shouldn't we guard against accepting/refusing offers that are not in "final" states?

estate/models/estate_property.py
```diff
@@ -0,0 +1,130 @@
...
+        self.garden_area = None
+        self.garden_orientation = None
+
+    @api.ondelete(at_uninstall=False)
+    def _check_state(self):
+        for record in self:
+            if record.state not in ('new', 'canceled'):
+                raise UserError(
+                    self.env._("Can't delete a property if the state is not New or Cancelled.")
+                )
```
SaddemAmine: Nitpick
```suggestion
            if record.state not in {'new', 'canceled'}:
                raise UserError(self.env._("Can't delete a property if the state is not New or Cancelled."))
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            property = self.env['estate.property'].browse(vals['property_id'])
+
+            if property.offer_ids:
+                max_offer = property.offer_ids[0].price
+                if float_compare(vals['price'], max_offer, precision_digits=2) <= 0:
+                    raise UserError(self.env._("The offer must be strictly higher than %.2f") % max_offer)
```
SaddemAmine:
```suggestion
                    raise UserError(self.env._("The offer must be strictly higher than %.2f", max_offer))
```

SaddemAmine (commented):
Almost there, good work!

SaddemAmine (approved):
Good work Cezar!

PR #1376, opened 2026-07-29
19.0 Technical Training Real Estate Module

amtha-odoo:
- Created Estate module and models
- Added list, form, and search views
- Configured menus, actions, and security
- Implemented model relationships (Many2one, One2many, Many2many)
- Added computed fields, inverse methods, and onchange logic
- Implemented property and offer business actions (Sold, Cancel, Accept, Refuse)
- Added validations and state management
- Updated property buyer and selling price on offer acceptance


estate/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+# estate/models/__init__.py
+
+from . import estate_property
```
mash-odoo:
```suggestion
from . import estate_property
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,164 @@
...
+    _description = "Estate Property"
+    _order = "name asc"
+
+    name = fields.Char(required=True)
+    total_area = fields.Integer(compute="_compute_total_area")
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
```
mash-odoo: Please refer to [this](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions) for proper ordering of attributes in a model

Image

estate/models/estate_property.py
```diff
@@ -0,0 +1,164 @@
...
+
+    buyer_id = fields.Many2one(
+        "res.partner",
+        string="Buyer",
+        copy=False,
+    )
+    salesperson_id = fields.Many2one(
+        "res.users",
+        string="Salesperson",
+        default=lambda self: self.env.user,
```
mash-odoo, line 24: You don't need to add string explicitly if you want it to be similar to the field name itself.

estate/models/estate_property.py
```diff
@@ -0,0 +1,164 @@
...
+    garden_area = fields.Integer()
+
+    garden_orientation = fields.Selection(
+        [
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
+        ]
+    )
```
mash-odoo, line 61:
```suggestion
    garden_orientation = fields.Selection(
        [
            ('north', "North"),
            ('south', "South"),
            ('east', "East"),
            ('west', "West"),
        ]
    )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,164 @@
...
+
+            if (
+                float_compare(
+                    record.selling_price,
+                    minimum_price,
+                    precision_rounding=0.01,
+                )
+                < 0
+            ):
+                raise ValidationError(
```
mash-odoo, line 122: When would you raise `UserError` and `ValidationError`?
What is the difference between both of them?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,71 @@
...
+        string="Validity (days)",
+        default=7,
+    )
+    deadline = fields.Date(
+        string="Deadline",
+        compute="_compute_date_deadline",
+        inverse="_inverse_date_deadline",
+        store=True,
+    )
+    create_date = fields.Datetime()
```
mash-odoo: Do you really need to add this field?

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,12 @@
+from odoo import fields, models
+
+
+class EstatePropertyTag(models.Model):
+    _name = "estate.property.tag"
+    _description = "Property Tag"
+
+    name = fields.Char(required=True)
+    _check_tag_name = models.Constraint(
+        "UNIQUE(name)",
```
mash-odoo, line 11: Will this unique work for all the case?

estate/view/estate_property_offer_views.xml
```diff
@@ -0,0 +1,50 @@
...
+                        <field name="deadline"/>
+
+                        <field name="status" readonly="True"/>
+                    </group>
+                </sheet>
+            </form>
+        </field>
+    </record>
+
+</odoo>
```
mash-odoo: Always leave an extra line at the end of the file.

estate/view/estate_property_offer_views.xml
```diff
@@ -0,0 +1,50 @@
...
+        <field name="name">estate.property.offer.list</field>
+        <field name="model">estate.property.offer</field>
+        <field name="arch" type="xml">
+            <list string="Property Offers">
+                <field name="property_id"/>
+                <field name="price"/>
+                <field name="partner_id"/>
+                <field name="validity"/>
+                <field name="deadline"/>
+                <button name="action_confirm" string="Confirm" type="object" icon="fa-check"/>
```
mash-odoo, line 20: When to use type `object` and type `action`?

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "Estate",
+    "version": "1.0",
+    "depends": ["base"],
+    "application": True,
+    "category": "Tutorials",
+    "author": "Thakor Anish",
```
mash-odoo: When you are working for a company, you should keep the author name as Odoo S.A. or just skip writing it.

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "license": "LGPL-3",
```
mash-odoo, line 8: What other options are there for licensing?

estate/models/estate_property.py
```diff
@@ -0,0 +1,164 @@
...
+                raise UserError("Sold property cannot be cancelled.")
+
+            record.state = "cancelled"
+
+        return True
+
+    best_price = fields.Float(
+        string="Best Offer",
+        compute="_compute_best_price",
+        store=True,
```
mash-odoo: Why do we need to store this value?

mash-odoo (changes requested):
Hello,
Good start on the task.
Some points to be looked at:
1. Do not leave unnecessary lines between fields.
2. Reformat the files according to the coding guidelines.
3. Add description and chapter name while pushing the commits in order to understand the changes done for that particular commit.
4. Update the PR title.

estate/models/__init__.py
amtha-odoo: Done

estate/models/estate_property.py
amtha-odoo: Done

estate/models/estate_property.py
amtha-odoo: Remove string attributes from fields where the default field label is already appropriate.

estate/models/estate_property.py
amtha-odoo: Done

estate/models/estate_property.py
amtha-odoo: UserError : use when the user perform invalid action.usually use in button and action methods and stop the current operation
ValidationError : use when data violates a validation.prevents invalid data from being saved.

Both will show an error popup to the user, but UserError is used when a user performs an action that is not allowed, and ValidationError is used when entered or modified data fails a validation

estate/models/estate_property_offer.py
amtha-odoo: No,don't need to add this field manually. create_date is an automatic field provided by Odoo's ORM, and Odoo creates and manages it automatically for every record.

estate/view/estate_property_offer_views.xml
amtha-odoo: Ok

estate/__manifest__.py
amtha-odoo: Ok

estate/__manifest__.py
amtha-odoo: /home/odoo/odoo19/community/odoo/addons/base/models/ir_module.py  == line no. : 296

GPL-2,GPL-3,AGPL-3,LGPL-3,OEEL-1,OPL-1

estate/models/estate_property.py
amtha-odoo: For a calculated field, use store=True to store the value in the database. but here  we don't need to store this value here, so I removed store=True.

estate/models/estate_property.py
amtha-odoo: string
/home/odoo/odoo19/community/odoo/orm/fields.py    line no. = 508
=>Odoo automatically generates the UI label from the field's technical name.

estate/view/estate_property_offer_views.xml
amtha-odoo: type = "object" : used when execute python method defined on the model.
type = "action" : used when execute an odoo action identified by its external ID.

PR #1381, opened 2026-08-07
19.0 Tutorials YACHA

yacha-odoo:
This PR contains the implementation of the Real Estate module from the Odoo development tutorial.

It covers:
- Creating the `estate` module and its models
- Adding fields and model relationships
- Adding security access rights
- Creating menus, actions, and views
- Adding property offers, types, and tags
- Improving the property search and form views

### DIFF
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -10,6 +10,7 @@
     "data": [
         "security/ir.model.access.csv",
         "views/estate_property_views.xml",
+        "views/estate_property_type_views.xml",
         "views/estate_menus.xml",
     ],
 }
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -1,0 +1,10 @@
+{
+    'name': 'Real Estate',
+    'version': '1.0',
+    'author': 'yacha',
+    'depends': ['base'],
+    'application': True,
+    'category': 'Tutorials',
+    'installable': True,
+    'license': 'LGPL-3'
+}
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -12,6 +12,7 @@
         "views/estate_property_views.xml",
         "views/estate_property_type_views.xml",
         "views/estate_property_tag_views.xml",
+        "views/estate_property_offer_views.xml",
         "views/estate_menus.xml",
     ],
 }
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -1,10 +1,11 @@
 {
-    'name': 'Real Estate',
-    'version': '1.0',
-    'author': 'yacha',
-    'depends': ['base'],
-    'application': True,
-    'category': 'Tutorials',
-    'installable': True,
-    'license': 'LGPL-3'
+    "name": "Real Estate",
+    "version": "1.0",
+    "author": "yacha",
+    "depends": ["base"],
+    "application": True,
+    "category": "Tutorials",
+    "installable": True,
+    "license": "LGPL-3",
+    "data": ["security/ir.model.access.csv"],
 }
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -11,6 +11,7 @@
         "security/ir.model.access.csv",
         "views/estate_property_views.xml",
         "views/estate_property_type_views.xml",
+        "views/estate_property_tag_views.xml",
         "views/estate_menus.xml",
     ],
 }
--- a/estate/__manifest__.py
+++ b/estate/__manifest__.py
@@ -7,5 +7,9 @@
     "category": "Tutorials",
     "installable": True,
     "license": "LGPL-3",
-    "data": ["security/ir.model.access.csv"],
+    "data": [
+        "security/ir.model.access.csv",
+        "views/estate_property_views.xml",
+        "views/estate_menus.xml",
+    ],
 }
--- a/estate/models/estate_property.py
+++ b/estate/models/estate_property.py
@@ -6,7 +6,6 @@ from dateutil.relativedelta import relativedelta
 def _default_date_availability(self):
     return fields.Date.today() + relativedelta(months=3)
 
-
 class EstateProperty(models.Model):
     _name = "estate.property"
     _description = "Real Estate Property"
@@ -24,6 +23,7 @@ class EstateProperty(models.Model):
     facades = fields.Integer()
     garage = fields.Boolean()
     garden = fields.Boolean()
+        
     garden_area = fields.Integer()
     garden_orientation = fields.Selection(
         [
@@ -58,3 +58,20 @@ class EstateProperty(models.Model):
     tag_ids = fields.Many2many("estate.property.tag", string="Tags")
 
     offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers")
+
+    total_area = fields.Integer(compute="_compute_total")
+
+    best_price = fields.Float(compute="_compute_best_price")
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for record in self:
+            if record.offer_ids:
+                record.best_price = max(record.offer_ids.mapped("price"))
+            else:
+                record.best_price = 0
--- a/estate/models/estate_property.py
+++ b/estate/models/estate_property.py
@@ -56,3 +56,5 @@ class EstateProperty(models.Model):
     )
 
     tag_ids = fields.Many2many("estate.property.tag", string="Tags")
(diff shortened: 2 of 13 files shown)


estate/models/estate_property.py
```diff
@@ -0,0 +1,77 @@
+from odoo import api, models, fields
+from dateutil.relativedelta import relativedelta
```
mash-odoo, line 2: You can refer to [this](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#imports) for ordering your imports.

estate/models/estate_property.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@api.model
+def _default_date_availability(self):
+    return fields.Date.today() + relativedelta(months=3)
```
mash-odoo, line 7: This should be placed inside the class.
You can follow [this](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions) for ordering your fields, functions, constraints etc. You can check the attribute order followed in a model.

estate/models/estate_property.py
```diff
@@ -0,0 +1,77 @@
...
+        
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        [
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West"),
+        ]
+    )
```
mash-odoo, line 35:
```suggestion
    garden_orientation = fields.Selection(
        [
            ('north', "North"),
            ('south', "South"),
            ('east', "East"),
            ('west', "West"),
        ]
    )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,77 @@
...
+        default="new",
+    )
+
+    property_type_id = fields.Many2one("estate.property.type", string="Property Type")
+
+    buyer_id = fields.Many2one("res.partner", string="Buyer", copy=False)
+
+    salesperson_id = fields.Many2one(
+        "res.users", string="Salesperson", default=lambda self: self.env.user
+    )
```
mash-odoo, line 56:
```suggestion
    property_type_id = fields.Many2one("estate.property.type", string="Property Type")
    buyer_id = fields.Many2one("res.partner", string="Buyer", copy=False)
    salesperson_id = fields.Many2one(
        "res.users", string="Salesperson", default=lambda self: self.env.user
    )
```
Do not leave unnecessary lines between the field declarations.

estate/models/estate_property.py
```diff
@@ -0,0 +1,77 @@
...
+        ],
+        string="State",
+        required=True,
+        copy=False,
+        default="new",
+    )
+
+    property_type_id = fields.Many2one("estate.property.type", string="Property Type")
+
+    buyer_id = fields.Many2one("res.partner", string="Buyer", copy=False)
```
mash-odoo, line 52: What if you do not write the string here?

estate/models/estate_property.py
```diff
@@ -0,0 +1,77 @@
...
+
+    salesperson_id = fields.Many2one(
+        "res.users", string="Salesperson", default=lambda self: self.env.user
+    )
+
+    tag_ids = fields.Many2many("estate.property.tag", string="Tags")
+
+    offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers")
+
+    total_area = fields.Integer(compute="_compute_total")
```
mash-odoo, line 62:
```suggestion
    total_area = fields.Integer(compute="_compute_total_area")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,77 @@
...
+
+    buyer_id = fields.Many2one("res.partner", string="Buyer", copy=False)
+
+    salesperson_id = fields.Many2one(
+        "res.users", string="Salesperson", default=lambda self: self.env.user
+    )
+
+    tag_ids = fields.Many2many("estate.property.tag", string="Tags")
+
+    offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers")
```
mash-odoo, line 60: How will I be able to view that which offers are linked to which property?

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
+access_estate_property_type,access_estate_property_type,model_estate_property_type,base.group_user,1,1,1,1
+access_estate_property_tag,access_estate_property_tag,model_estate_property_tag,base.group_user,1,1,1,1
+access_estate_property_offer,access_estate_property_offer,model_estate_property_offer,base.group_user,1,1,1,1
```
mash-odoo, line 5: Always leave an extra line at the end of every file.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,14 @@
...
+    _name = "estate.property.offer"
+    _description = "Estate Property Offer"
+
+    price = fields.Float()
+    status = fields.Selection(
+        [("accepted", "Accepted"), ("refused", "Refused")], copy=False
+    )
+
+    partner_id = fields.Many2one("res.partner", required=True)
+    property_id = fields.Many2one("estate.property", required=True)
```
mash-odoo, line 14: What is the difference between setting a field as` required=True` in the model and `required="1"` in the XML view?

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,35 @@
...
+
+        <menuitem
+            id="estate_advertisements_menu"
+            name="Advertisements">
+
+            <menuitem
+                id="estate_property_menu_action"
+                action="estate_property_action"/>
+
+        </menuitem>
```
mash-odoo, line 16:
```suggestion
    <menuitem
        id="estate_menu_root"
        name="Real Estate"
    >
        <menuitem
            id="estate_advertisements_menu"
            name="Advertisements"
        >
            <menuitem
                id="estate_property_menu_action"
                action="estate_property_action"
            />
        </menuitem>
```
Re-format this file a bit.

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mash-odoo, line 1: Is it compulsory to write this?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,118 @@
...
+                        <h1>
+                            <field name="name"/>
+                        </h1>
+                    </group>
+
+                    <group>
+                        
+                        <group>
+                            <field name="property_type_id"/>
+                            <field name="tag_ids" widget="many2many_tags"/>
```
mash-odoo, line 42: Why do we use a widget?

estate/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "Real Estate",
+    "version": "1.0",
+    "author": "yacha",
```
mash-odoo, line 4: When you are working for a company, you should keep the author name as `Odoo S.A. `or just skip writing it.

mash-odoo (changes requested):
Hello,
Good start on the task.
Some points to be looked at:
- Add chapter name while pushing the commits in order to understand the changes done for that particular commit.
- Update the PR title and description according to the guidelines.
- Reformat the necessary files.

estate/models/estate_property.py
yacha-odoo: Odoo automatically derives the string name explicitly from the variable name. So if we don't give this string then also it is deriving string to Property Type. I tested it. So i think i should remove it.

estate/models/estate_property.py
yacha-odoo: The property_id field in estate.property.offer stores which property the offer belongs to. The One2many field uses this inverse field to retrieve all offers linked to that particular property.

estate/models/estate_property_offer.py
yacha-odoo: required=True makes the field mandatory at the model level, so it applies whenever the field is used. required="1" in the XML only makes the field mandatory in that specific view.

estate/views/estate_property_type_views.xml
yacha-odoo: It's the XML declaration that identifies the file as an XML 1.0 document. It's the standard way to start an XML file.
If we don't write it still ODOO parses the XML files and shows no errors
But we still explicitly define the version of that xml file.

estate/views/estate_property_views.xml
yacha-odoo: We use a widget when we want to change the default way a field is displayed or interacted with in the view. It allows us to provide a more suitable UI for that particular field.

PR #1383, opened 2026-08-17
onboarding server framework 

snoek100:


estate/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
yoba-odoo, line 1: Keep in mind to always add a newline in the end of each file

estate/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+        'base',
+    ],
+    'data': [
+        'data/ir.model.access.csv',
+        'views/estate_property_views.xml',
+        'data/estate_menus.xml'
+    ],
+    'views': [
+        
+    ],
```
yoba-odoo, line 14: We can remove this I don't think we need it since we are already defining the views in the data key

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,92 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="estate_property_view_search" model="ir.ui.view">
+        <field name="name">estate.property.search</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <search string="Search">
+                <field name="name" string="Title"/>
```
yoba-odoo, line 8: This will work fine but to avoid setting a string for each field in each view you can add this string to the field definition in python and it will be used everywhere you call this field in XML

estate/models/estate_property.py
```diff
@@ -0,0 +1,30 @@
+from odoo import fields, models
+from dateutil.relativedelta import relativedelta
+
+class TestModel(models.Model):
```
yoba-odoo: I know this is a tutorial 😅 but avoid to name anything test unless it is a real test as this is the convention we use for unit test files

yoba-odoo (commented):
Good work 💪 
Can you just keep it one commit per chapter? That would be easier to follow

Also the commit titles has to follow the guidelines [here](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html)

estate/models/estate_property.py
```diff
@@ -0,0 +1,101 @@
...
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        string='type',
+        selection=[('north', 'North'), ('south', 'South'), ('East', 'east'), ('West', 'west')],
+    )
+    total_area = fields.Integer(string="Total Area", compute="_compute_total_surface")
```
yoba-odoo, line 39:
```suggestion
    total_area = fields.Integer(string="Total Area", compute="_compute_total_area")
```
We normally tend to name the compute method after the exact name of field

estate/models/estate_property.py
```diff
@@ -0,0 +1,101 @@
...
+    def action_property_cancelled(self):
+        for record in self:
+            if record.state == "sold":
+                no_sell_a_sold_property = "Can't cancel a sold property"
+                raise exceptions.UserError(no_sell_a_sold_property)
+            record.state = "cancelled"
+        return True
+
+    @api.ondelete(at_uninstall=False)
+    def ondelete(self):
```
yoba-odoo, line 104: It is kinda a nitpick but we usually name those method as `_unlink_****` for example:
```suggestion
    def _unlink_except_new_cancelled(self):
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,66 @@
...
+    # =========contraints============
+    _check_positive_offer_price = models.Constraint("CHECK (price > 0)", "expected price should be bigger than 0")
+
+    @api.depends('validity', "create_date")
+    def compute_deadline(self):
+        for record in self:
+            if record.create_date:
+                record.date_deadline = record.create_date + relativedelta(days=record.validity)
+            else:
+                record.date_deadline = fields.Datetime.today() + relativedelta(days=record.validity)
```
yoba-odoo, line 33:
```suggestion
        for record in self:
            check_date = record.create_date if record.create_date else fields.Datetime.today()
            record.date_deadline = check_date + relativedelta(days=record.validity)
```
To avoid duplicating some of the code we can do this, wdyt?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def _inverse_deadline(self):
+        for record in self:
+            record.validity = (record.date_deadline - record.create_date).days
+
+    # ===========button actions===========
+    def action_accept(self):
+
```
yoba-odoo, line 41: Nitpick 😇 not a needed line

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,66 @@
...
+            for property in record.property_id:
+                record.status = "refused"
+
+    @api.model
+    def create(self, vals):
+        for to_create in vals:
+            property = self.env["estate.property"].browse(to_create["property_id"])
+            new_bid = to_create["price"]
+            for offer in property.offer_ids:
+                if offer.price > new_bid:
```
yoba-odoo: since you introduced an order for the offers here https://github.com/odoo/tutorials/pull/1383/changes#diff-8000c23a2907f34f0cc387be9dc4178411fda81b674724854bd4d048f9e11b16R9 the recordset will be ordered upon price by default so normally `offer_ids[0]` will have the offer that has the highest price

estate/models/inhereted_users.py
```diff
@@ -0,0 +1,10 @@
+from odoo import fields, models
+
+class InheritedModel(models.Model):
```
yoba-odoo, line 4: When inherit a model we name the class the same as the parent class

estate/views/estate_property_settings_views.xml
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0"?>
```
yoba-odoo: it is a better practice to have a XML view file for each model.
So here separate the estate_property and the estate_property_type and estate_property_tag and estate_property_offer views each in a separate file

estate/views/estate_property_view_form.xml
```diff
@@ -0,0 +1,180 @@
+<?xml version="1.0"?>
```
yoba-odoo, line 1: same as above ☝️

yoba-odoo (commented):
Nice work 👌 
While applying the changes try to make the ci/* checks pass

PR #1385, opened 2026-08-17
aaboe - Technical Training 

aaboe-odoo:


estate/views/estate_basic_views.xml
```diff
@@ -0,0 +1,98 @@
+<?xml version="1.0" ?>
+<odoo>
+    <!-- Estate Property List -->
+    <record id="estate_property_list" model="ir.ui.view">
+        <field name="name">estate.property.list</field>
+        <field name="model">estate.property</field>
+            <field name="arch" type="xml">
+                <list string="Properties">
+                    <field name="name" string="Title" class="text-center" width="50px"/>
```
yoba-odoo, line 9: Although this works fine, you don't have to set a string for the same field in each XML view. It is a better practice to identify the name of the field in the field definition in the python file

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,16 @@
...
+            </menuitem>
+            <menuitem id="estate_settings_menu" name="Settings">
+                    <menuitem id="estate_property_type_menu_action" action="estate_property_type_action" name="Property Types"/>
+                    <menuitem id="estate_property_tag_menu_action" action="estate_property_tag_action" name="Property Tags"/>
+            </menuitem>
+
+
+        </menuitem>
+    </data>
+</odoo>
```
yoba-odoo, line 16: It is mandatory to have newline at the end of each file

estate/views/estate_basic_views.xml
```diff
@@ -0,0 +1,98 @@
+<?xml version="1.0" ?>
```
yoba-odoo, line 1: I am not sure why do we need this file. We can add those views in the same file as the action. So having it in `estate_property_views.xml` is the right place I would say

yoba-odoo (commented):
Nice work 👌

estate/views/estate_basic_views.xml
aaboe-odoo: I thought would be nice to isolate things a bit, but you are right not worth it since it is not going to be that big anyway..

Thank you will merge these togther

aaboe-odoo:
> Nice work 👌

Thank you @yoba-odoo for the feedback I will be addressing it shortly

estate/models/estate_property.py
```diff
@@ -0,0 +1,106 @@
...
+    DEFAULT_BEDROOM_COUNT = 2
+    DEFAULT_AVAILABILITY_DATE = fields.Date.today() + relativedelta(months=3)
+
+    name = fields.Char(string='Title', required=True, default='New Property')
+    description = fields.Text()
+    active = fields.Boolean(default=True)
+    date_availability = fields.Date(string='Available From', copy=False, default=DEFAULT_AVAILABILITY_DATE)
+
+    expected_price = fields.Float(required=True)
+    _check_expected_price = models.Constraint('check(expected_price > 0)', 'Expected Price must be greater than 0')
```
yoba-odoo: The constraints are usually defined after we define all fields

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,70 @@
...
+from odoo.exceptions import UserError
+
+
+class EstatePropertyOffer(models.Model):
+    _name = 'estate.property.offer'
+    _description = 'Estate Property Offer Model'
+    _order = 'price desc'
+
+    price = fields.Float()
+    _check_price = models.Constraint('check(price > 0)', 'Price must be greater than 0')
```
yoba-odoo: same ☝️

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,50 @@
...
+            <field name="arch" type="xml">
+                <list string="Offers" editable="bottom" decoration-danger="status == 'refused'" decoration-success="status == 'accepted'">
+                    <field name="price"/>
+                    <field name="partner_id"/>
+                    <field name="property_type_id" optional="hide"/>
+                    <field name="validity"/>
+                    <field name="date_deadline"/>
+                    <button name="action_accept" type="object" icon="fa-check" title="Accept" invisible="status"/>
+                    <button name="action_refuse" type="object" icon="fa-times" title="Refuse" invisible="status"/>
+                    <field name="status" column_invisible="True"/> <!-- Used by list decorations and Accept/Refuse buttons -->
```
yoba-odoo: This was how we do it in older versions but currently we can remove this field as when parsing the view the framework itself creates a field for it to use. So you don't have to define a field tag just to use it in a condition or decoration.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,144 @@
...
+                  decoration-muted="state == 'sold'">
+                <field name="name" class="text-center" width="50px"/>
+                <field name="property_type_id" class="text-center" width="50px"/>
+                <field name="postcode" class="text-center" width="50px"/>
+                <field name="bedrooms" class="text-center" width="50px"/>
+                <field name="living_area" class="text-center" width="50px"/>
+                <field name="expected_price" class="text-center" width="50px"/>
+                <field name="selling_price" class="text-center" width="50px"/>
+                <field name="date_availability" class="text-center" width="50px" optional="hide"/>
+                <field name="state" column_invisible="True"/> <!-- Used by list decorations -->
```
yoba-odoo: same ☝️

estate/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    'installable': True,
+    'application': True,
+    'data': [
+        'views/estate_property_views.xml',
+        'views/estate_property_offer_views.xml',
+        'views/estate_property_type_views.xml',
+        'views/estate_property_tags_views.xml',
+        'views/res_users_views.xml',
+        'views/estate_menus.xml',
+        'security/ir.model.access.csv',
```
yoba-odoo: Security file always have to come first, as this is the order in which the files will be loaded and security is important to be there when loading other files

yoba-odoo (commented):
Nice work 💪

aaboe-odoo:
Thank you for the review and feedback, @yoba-odoo 
Feedback addressed in https://github.com/odoo/tutorials/pull/1385/changes/a1dfdb74cf670336e638d7178c1c3977cc4c06d9

PR #1386, opened 2026-08-17
feraz - Technical Training

feraz-odoo:


estate/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
...
+
+class EstateProperty(models.Model):
+
+    _name = "estate.property"
+    _description = "Estate property"
+
+    name = fields.Char("Property Name", required=True)
+    description = fields.Text("Description")
+    postcode = fields.Char("Postcode")
+    date_availability = fields.Date("Availability Date", copy=False, default=fields.Date.to_date(fields.Date.today() + timedelta(days=3 * 30)))
```
qucol-odoo: No need to call `to_date` here. It is used to cast a value (usually a string) to a Date, but here your value is already a date object.

Also, just FYI, an alternative to `timedelta` that allows a `months` argument directly is `relativedelta` (from `dateuil.relativeldelta`).

```suggestion
    date_availability = fields.Date("Availability Date", copy=False, default=fields.Date.today() + relativedelta(months=3)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
...
+    expected_price = fields.Float("Expected Price", required=True)
```
qucol-odoo: By default, the "string" of a field is the name of that field with the `_` replaced by spaces (` `).  In case you're curious, that is defined [here](https://github.com/odoo/odoo/blob/0824dc24665de6bfa805d540e756cdcb006edba6/odoo/orm/fields.py#L506-L511).

In this case (and others below), it's not necessary to specify the string of your field, as it matches the default behavior anyway.
```suggestion
    expected_price = fields.Float(required=True)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
...
+    selling_price = fields.Float("Selling Price", readonly=True, copy=False)
+    bedrooms = fields.Integer("Number of Bedrooms", default=2)
+    living_area = fields.Integer("Living Area m²")
+    facades = fields.Integer("Number of Facades")
+    garage = fields.Boolean("Garage")
+    garden = fields.Boolean("Garden")
+    garden_area = fields.Integer("Garden Area m²")
+    garden_orientation = fields.Selection(string="Garden Orientation", selection=[('north', 'North'), ('east', 'East'), ('south', 'South'), ('west', 'West')])
+    state = fields.Selection(string="State", required=True, copy=False, selection=[('new', 'New'), ('offer_received', 'Offer Received'), ('offer_accepted', 'Offer Accepted'), ('sold', 'Sold'), ('cancelled', 'Cancelled')], default='new')
```
qucol-odoo: Kind of a nitpick, but usually you want the `selection` parameter to be the first or right after the `string` one. Also, when you have a long selection (or a long line in general), it's better to break it down into multiple, shorter lines.
```suggestion
    state = fields.Selection(
    selection=[
        ('new', 'New'),
        ('offer_received', 'Offer Received'),
        ('offer_accepted', 'Offer Accepted'),
        ('sold', 'Sold'), ('cancelled', 'Cancelled'),
    ],
    default='new'
    required=True,
    copy=False,)
```

estate/views/estate_property_menu_views.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
+<odoo>
+
+    <!-- Tutorial menu  -->
+    <menuitem id="estate_property_menu_root" name="Estate Property">
+        <menuitem id="estate_property_first_level_menu" name="Estate Properties">
+            <menuitem id="estate_property_model_menu_action" action="estate_property_action"/>
+            <!--<menuitem id="estate_property_offer_model_menu_action" action="estate_property_offer_action"/>-->
```
qucol-odoo: Not the end of the world here as it is a tutorial, but avoid pushing commented code. In bigger reviews in can easily be forgotten and accidentally end up in the codebase. If you want to keep this line somewhere in case you might need it again later, you can use [git stash](https://git-scm.com/docs/git-stash).

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,109 @@
...
+            <form string="Estate Property">
+                <sheet>
+                    <h1>
+                        <field name="name"/>
+                    </h1>
+
+                    <field name="tags" widget="many2many_tags"/>
+
+                    <br/>
+
```
qucol-odoo: Nitpick, but no need for all those blank lines, they don't really make things easier to read and just take space for "nothing".

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,109 @@
...
+        <field name="arch" type="xml">
+            <search string="Estate Property Search">
+                <field name="name"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="facades"/>
+                <field name="type"/>
+                <!--<filter name="available" string="Available" domain="[('date_availability', '&lt;=', current_date)]" />-->
```
qucol-odoo: Same as above, avoid commented code

estate/__manifest__.py
```diff
@@ -0,0 +1,37 @@
...
+    'website': "https://www.odoo.com",
+
+    # Categories can be used to filter modules in modules listing
+    # Check https://github.com/odoo/odoo/blob/15.0/odoo/addons/base/data/ir_module_category_data.xml
+    # for the full list
+    'category': 'Tutorials',
+    'version': '0.1',
+
+    # any module necessary for this one to work correctly
+    'depends': [],
```
qucol-odoo: Technically it works even without explicitly mentioning "base" as a dependency, but explicit is always better than implicit
```suggestion
    'depends': ['base'],
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class EstatePropertyOffer(models.Model):
+
+    _name = "estate.property.offer"
+    _description = "Estate property offer"
+
+    price = fields.Float("Offer Price")
+    status = fields.Selection(string="Status", copy=False, selection=[('accepted', 'Accepted'), ('refused', 'Refused')])
+    partner = fields.Many2one(string="Buyer", comodel_name="res.partner",required=True)
```
qucol-odoo: Missing a whitespace 😇 
```suggestion
    partner = fields.Many2one(string="Buyer", comodel_name="res.partner", required=True)
```

qucol-odoo (commented):
Thank you for your great work 👌 I wrote a few comments which are mostly nitpicks, try to apply them whenever you find the time :)

I see that your latest commit titles are correctly formatted, that's nice 👍 You could try to rename the previous ones to please the runbot 😇

estate/models/estate_property.py
```diff
@@ -0,0 +1,117 @@
...
+    seller_id = fields.Many2one(string="Salesperson", comodel_name="res.users", default=lambda self: self.env.user)
+
+    tag_ids = fields.Many2many(comodel_name="estate.property.tag")
+
+    offer_ids = fields.One2many(comodel_name="estate.property.offer", inverse_name="property_id")
+
+    total_area = fields.Integer(compute="_compute_total_area")
+
+    best_price = fields.Float("Best offer price", compute="_compute_best_price")
+
```
qucol-odoo: No need for all the blank lines between the fields
```suggestion
    tag_ids = fields.Many2many(comodel_name="estate.property.tag")
    offer_ids = fields.One2many(comodel_name="estate.property.offer", inverse_name="property_id")
    total_area = fields.Integer(compute="_compute_total_area")
    best_price = fields.Float("Best offer price", compute="_compute_best_price")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,117 @@
...
+            ('offer_accepted', 'Offer Accepted'),
+            ('sold', 'Sold'), ('cancelled', 'Cancelled'),
+        ],
+        default='new',
+        required=True,
+        copy=False)
+
+    active = fields.Boolean(default=True)
+
+    type_id = fields.Many2one(comodel_name="estate.property.type")
```
qucol-odoo: Not an issue, more like a piece of info worth sharing: the first argument of a `Many2one` is always the `comodel_name` (see its [`__init__` method](https://github.com/odoo/odoo/blob/189df3538aa07d22f683008522b9143421f411a2/odoo/orm/fields_relational.py#L250)), so there's no need to specify the argument name).

Image

Image

estate/models/estate_property.py
```diff
@@ -0,0 +1,117 @@
...
+        for property in self:
+            property.total_area = property.living_area + property.garden_area
+
+    @api.depends("offer_ids")
+    def _compute_best_price(self):
+        for property in self:
+            if not property.offer_ids:
+                property.best_price = 0
+                continue
+            best_offer = max(property.offer_ids, key=lambda offer: offer.price)
```
qucol-odoo: More odoo-esque way
```suggestion
            best_offer = max(property.offer_ids.mapped('price'))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,117 @@
+from dateutil.relativedelta import relativedelta
+
+from odoo import models, fields, api, exceptions
```
qucol-odoo: Usually we just import the exceptions we need
```suggestion
from odoo import models, fields, api
from odoo.exceptions import UserError
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,117 @@
...
+        for property in self:
+            if property.state == "cancelled":
+                raise exceptions.UserError("Cancelled properties cannot be sold")
+            property.state = "sold"
+        return True
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_only_new_cancelled(self):
+        for property in self:
+            if property.state != 'new' and property.state != 'cancelled':
```
qucol-odoo:
```suggestion
            if property.state not in ('new', 'cancelled'):
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,65 @@
...
+        for offer in self:
+            offer.validity = (offer.date_deadline - offer.create_date.date()).days
+
+    def action_accept_offer(self):
+        # We can only accept one offer at a time
+        self.ensure_one()
+
+        # Ensure there is no existing accepted offer
+        for offer in self.property_id.offer_ids:
+            if offer.state == "accepted":
```
qucol-odoo: Couldn't you use your `_no_accepted_offer` method?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,65 @@
...
+                raise exceptions.UserError("Property already has an accepted offer.")
+
+        self.state = "accepted"
+
+        self.property_id.buyer_id = self.partner_id
+        self.property_id.state = "offer_accepted"
+        self.property_id.selling_price = self.price
```
qucol-odoo: You can batch those writes
```suggestion
        self.property_id.write({
            'buyer_id': self.partner_id.id,
            'state': 'offer_accepted',
            'selling_price': self.price,
        })
```

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,25 @@
...
+
+    _name_uniq = models.Constraint(
+        'unique(name)',
+        'A property type with the same name already exists.',
+    )
+
+    @api.depends("property_ids.offer_ids")
+    def _compute_offer_count(self):
+        for type in self:
+            type.offer_count = sum(len(property.offer_ids) for property in type.property_ids)
```
qucol-odoo: That looks like a good use case for [`search_count`](https://github.com/odoo/odoo/blob/0d44f26d9b0fb1c1a5db463cf1f8dd0d3c72ba26/odoo/orm/models.py#L1347), can you try using it? :)

estate/tests/test_estate.py
```diff
@@ -0,0 +1,108 @@
...
+            {'name': 'property1', 'state': 'cancelled'},
+            {'name': 'property2', 'state': 'cancelled'},
+        ])
+
+        with self.assertRaises(UserError):
+            self.properties.action_sell_property()
+
+
+    def test_offer_creation(self):
+        """Test that everything behaves like it should when creating an offer."""
```
qucol-odoo: That doesn't really say much, don't hesitate to list the flows you're testing. [Here](https://github.com/odoo/enterprise/blob/7b309f2e74f9fa37b109375de6387db3241c2bcb/hr_referral/tests/test_referral.py#L83-L100) is a good example

estate/tests/test_estate.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        # add env on cls and many other things
+        super(EstateTestCase, cls).setUpClass()
+
+        # create the data for each tests. By doing it in the setUpClass instead
+        # of in a setUp or in each test case, we reduce the testing time and
+        # the duplication of code.
+        cls.properties = cls.env['estate.property'].create([
```
qucol-odoo, line 18: That's a very good practice, and creating the properties in a batch is also a good plan, but I'm not fully sure it's the best option. Seeing `self.properties[0]` and `self.properties[1]` in the other tests makes it hard to know the difference between the two. I believe it would be just as good to have two different variables for the properties, that you can eventually join in some tests if you really want to use `assertRecordValues`
```suggestion
        cls.property, cls.property_with_garden = cls.env['estate.property'].create([
```

qucol-odoo (commented):
Good job! I highlighted a few nitpicks :)

estate/tests/test_estate.py
feraz-odoo: Would it be okay to have both properties and the two references you suggested ? Because, to me, in test_creation_area, it makes sense to treat them as a batch. However, I guess keeping and using two different references to the same data might not be good practice either. What do you think ? @qucol-odoo

qucol-odoo: Sounds fine to me, I see that you've created the two properties in your `setUpClass`, don't forget to use them in your tests :ok_hand:

estate/tests/test_estate.py
```diff
@@ -0,0 +1,117 @@
...
+        """Test that the total_area is computed like it should."""
+        self.properties.living_area = 20
+        self.assertRecordValues(self.properties, [
+           {'name': 'property1', 'total_area': 20},
+           {'name': 'property2', 'total_area': 60},
+        ])
+
+
+    def test_action_sell(self):
+        """Test that everything behaves like it should when selling a property."""
```
qucol-odoo, line 37: No need to worry about it as you're pretty much done with the tests now, but once again this comment doesn't really help understanding what behaviors you're testing.

estate/tests/test_estate.py
```diff
@@ -0,0 +1,117 @@
...
+            {'price': 10000, 'partner_id': self.partner.id, 'property_id': self.properties[1].id},
+        ])
+
+        offers.action_accept_offer()
+
+        self.properties.action_sell_property()
+
+        self.assertRecordValues(self.properties, [
+           {'name': 'property1', 'state': 'sold', 'selling_price': 2000},
+           {'name': 'property2', 'state': 'sold', 'selling_price': 10000},
```
qucol-odoo: No need to test the `name` field, unless I'm missing something? You're doing that every time you use `assertRecordValues`

estate/tests/test_estate.py
```diff
@@ -0,0 +1,117 @@
...
+        Test that everything behaves like it should when creating an offer.
+        - The user can create one or several offers for a property
+        - The user can not sell properties with no accepted offers
+        - The user can sell properties with an accepted offer
+        - Offers can't be created for sold properties
+        """
+        offers = self.env['estate.property.offer'].create([
+            {'price': 2000, 'partner_id': self.partner.id, 'property_id': self.properties[0].id},
+            {'price': 1500, 'partner_id': self.partner.id, 'property_id': self.properties[0].id},
+            {'price': 12200, 'partner_id': self.partner.id, 'property_id': self.properties[1].id},
```
qucol-odoo: One of the place where you should use your `self.property` and `self.property_with_garden`

estate/tests/test_estate.py
```diff
@@ -0,0 +1,117 @@
...
+
+        self.properties.action_sell_property()
+
+        self.assertRecordValues(self.properties, [
+           {'name': 'property1', 'state': 'sold', 'selling_price': 2000},
+           {'name': 'property2', 'state': 'sold', 'selling_price': 10000},
+        ])
+
+        with self.assertRaises(UserError):
+            self.properties.action_cancel_property()
```
qucol-odoo, line 59: Not mandatory, but it's a good practice to say *why* an error should be raised. You can do it by either adding a comment somewhere around the `assertRaises` or adding a `msg` argument to the `assertRaises` call.

qucol-odoo (commented):
Few last comments on this PR, it's some very good job 👏 Feel free to apply them or simply read them.

In a real case scenario, it wouldn't make much sense to have commits fixing issues from other commits belonging to the same PR, because it means that you are introducing an issue and fixing it simultaneously: it would be better to simply not introduce it at all. It would be a great exercise for you to try to get rid of all the [FIX] commits in your PR (hint: you should probably use [git rebase -i](https://git-scm.com/docs/git-rebase#_interactive_mode)). Ultimately you could also squash commits to have one per chapter, or even squash everything as one big [ADD] commit, it's kinda up to you, as long as you experiment a bit with the interactive rebasing (mostly the edit and squash options) :)

PR #1387, opened 2026-08-17
yalla - Technical Training

YousefAllam221b:


estate/models/estate_property.py
```diff
@@ -0,0 +1,47 @@
+from datetime import timedelta
+from odoo import models, fields
+
+class EstateProperty(models.Model):
+    _name = 'estate.property'
+    _description = 'Estate Property'
+
+    name = fields.Char(required=True)
+
```
yoba-odoo: readability wise it is not the best practice to have a newline between each two fields

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
```
yoba-odoo: Try to keep in mind to have a newline in the end of each file

estate/models/estate_property.py
```diff
@@ -0,0 +1,47 @@
...
+    description = fields.Text()
+
+    postcode = fields.Char()
+
+    date_availability = fields.Date(copy=False, default=lambda self: fields.Date.today() + timedelta(days=90))
```
yoba-odoo: This will be working completely fine but we normally tend to use `relativedelta` rather than `timedelta`

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,7 @@
+<odoo>
+    <menuitem id="real_estate_menu_root" name="Real Estate">
+        <menuitem id="advertisements_menu" name="Advertisements">
+            <menuitem id="estate_property_menu_action" action="estate_property_action"/>
+        </menuitem>
+    </menuitem>
+</odoo>
```
yoba-odoo: same EOL

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,7 @@
+<odoo>
+    <record id="estate_property_action" model="ir.actions.act_window">
+        <field name="name">Properties</field>
+        <field name="res_model">estate.property</field>
+        <field name="view_mode">list,form</field>
+    </record>
+</odoo>
```
yoba-odoo: same ☝️

yoba-odoo (commented):
Good work 👌 

For the commit titles please follow the guidelines [here](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html)

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+
+    _positive_expected_price = models.Constraint(
+        "CHECK(expected_price > 0)",
+        "The expected price of a property cannot be negative.",
+    )
+    _positive_selling_price = models.Constraint(
+        "CHECK(selling_price >= 0)",
+        "The selling price of a property cannot be negative.",
+    )
+    _order = "id desc"
```
yoba-odoo: Usually this is introduced right below the name of the class and the description

estate/models/estate_property.py
```diff
@@ -0,0 +1,116 @@
...
+            raise UserError("You cannot mark a canceled property as sold.")
+        self.state = "sold"
+
+    def action_canceled(self):
+        if self.state == "sold":
+            raise UserError("You cannot cancel a sold property.")
+        self.state = "canceled"
+
+    @api.ondelete(at_uninstall=False)
+    def _delete_if_new_or_canceled(self):
```
yoba-odoo, line 115: In odoo, delete we call it unlink 😄 

```suggestion
    def _unlink_if_new_or_canceled(self):
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+    date_deadline = fields.Date(
+        compute="_compute_date_deadline", inverse="_inverse_date_deadline", store=True
+    )
+    validity = fields.Integer(default=7)
+
+    _positive_price = models.Constraint(
+        "CHECK(price > 0)", "The price of an offer cannot be negative."
+    )
+    _order = "price desc"
```
yoba-odoo: same

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,89 @@
...
+    def _compute_date_deadline(self):
+        for record in self:
+            base_date = (
+                fields.Date.to_date(record.create_date)
+                if record.create_date
+                else fields.Date.today()
+            )
+            record.date_deadline = fields.Date.add(base_date, days=record.validity)
+
+    @api.depends("date_deadline")
```
yoba-odoo: you don't need to specify a decorator here it is already linked to the field with the compute and the fields definition

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,89 @@
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            estate_property = self.env["estate.property"].browse(
+                vals.get("property_id")
+            )
+            existing_prices = estate_property.offer_ids.mapped("price")
+
+            if existing_prices:
+                min_price = min(existing_prices)
```
yoba-odoo, line 85: Since you defined an `_order` for the offers model with price so any recordset of offers will be sorted according to the price so to get the minimum you can do something like `estate_property.offer_ids[-1]`

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,15 @@
...
+    _description = "Estate Property Tag"
+
+    name = fields.Char(required=True)
+    color = fields.Integer()
+
+    _name_uniq = models.Constraint(
+        "UNIQUE (name)",
+        "The name of the tag must be unique!",
+    )
+    _order = "name"
```
yoba-odoo: same 😇

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,27 @@
...
+
+    property_ids = fields.One2many(
+        "estate.property", "property_type_id", string="Properties"
+    )
+    offer_ids = fields.One2many(
+        "estate.property.offer", "property_type_id", string="Offers"
+    )
+    offer_count = fields.Integer(compute="_compute_offer_count")
+    sequence = fields.Integer("Sequence", default=1)
+    _order = "sequence, name"
```
yoba-odoo: 👀

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,22 @@
+<odoo>
+  <record id="estate_property_offer_view_tree" model="ir.ui.view">
```
yoba-odoo: won't make any difference but just a headsup.
tree is the old name of list so we now tend to use list instead of tree

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,43 @@
...
+            </button>
+          </div>
+          <h1>
+            <field name="name" placeholder="Property Type" />
+          </h1>
+          <notebook>
+            <page string="Properties">
+              <field name="property_ids">
+                <list>
+                  <field name="name" string="Title" />
```
yoba-odoo: Instead of giving a name for the field in each view you can identify the name in the fields definition and it will be used across all views

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    ],
+    "installable": True,
+    "application": True,
+    "author": "me",
+    "license": "LGPL-3",
+    "data": [
+        "security/ir.model.access.csv",
+        "views/estate_property_views.xml",
+        "views/estate_property_actions.xml",
+        "views/estate_menus.xml",
```
yoba-odoo, line 15: Am I missing something? why didn't you introduced the other xml files here?
`estate_property_offer_views`, `estate_property_type_views` and `res_users_views`
I would be surprised if they are working without adding them to the manifest file😂

yoba-odoo (commented):
Good work 💯 
Some comments for enhancements, could you also please try to make the ci/* checks pass

yoba-odoo:
Another thing I just noticed for the commits we normally name them like this
`[tag] module_name : small description about the PR` don't forget to check the git guidelines

estate/models/estate_property.py
YousefAllam221b: > In odoo, delete we call it unlink 😄

yeah i just saw it in the guidelines, will change that

estate/__manifest__.py
YousefAllam221b: i forgot to add them but for some reason the local server didn't break so i thought everything is fine 😂😂

estate/models/estate_property_offer.py
YousefAllam221b: Yeah, that makes sense. But for the sake of readability, wouldn't it be better to leave it as is so the reader doesn't have to guess why I grabbed the last item? specially since its a low cost operation either way

YousefAllam221b:
> Another thing I just noticed for the commits we normally name them like this `[tag] module_name : small description about the PR` don't forget to check the git guidelines

Yeah i have read the git guidelines, i thought this is how you would want it, to know exactly which chapter is each commit. so in this case, the Chapter is the name of the module

estate/views/estate_property_type_views.xml
YousefAllam221b: okay this makes more sense

PR #1388, opened 2026-08-17
valvi - technical training

Topvennie:


estate/models/estate_offer.py
```diff
@@ -0,0 +1,53 @@
...
+    def _compute_date_deadline(self):
+        for record in self:
+            record.date_deadline = add(
+                record.create_date if record.create_date else fields.Date.today(),
+                days=+record.validity,
+            )
+
+    def _inverse_date_deadline(self):
+        for record in self:
+            record.validity = (record.date_deadline - fields.Date.today()).days
```
qucol-odoo, line 49: This is good, but I believe it could be even better 😇 
Right now, if I change the value of the `validity` field, it immediately updates the deadline accordingly, which is nice. On the other hand, if I change the value of the deadline field, it will only update the `validity` after a save the whole `estate.property` record. That's because the `_inverse` method is only executed when writing to the database. You can add the [`api.onchange`](https://www.odoo.com/documentation/19.0/developer/reference/backend/orm.html#odoo.api.onchange) decorator to have the `_inverse` method called everytime we update the `date_deadline` field in the form. That way, we mirror the behavior between `validity` and `date_deadline`.

```suggestion
    @api.onchange("date_deadline")
    def _inverse_date_deadline(self):
        for record in self:
            record.validity = (record.date_deadline - fields.Date.today()).days
```

estate/models/estate_offer.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # Actions
+
+    def set_status_accepted(self):
```
qucol-odoo: Method for actions should start with `action_`. Also, I don't think it should be possible to accept multiple offers at once, so you can replace the loop with [`self.ensure_one()`](https://github.com/odoo/odoo/blob/6fc51dd2bdca9f189247c9fc9f7a12776babb72b/odoo/orm/models.py#L5931). This will raise an error if we try to call the method on multiple records at once (which in fact, wouldn't make sense). Fyi, this is advised in the [coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions) :)
```suggestion
    def action_accept_offer(self):
```

estate/models/estate_offer.py
```diff
@@ -0,0 +1,53 @@
...
+        for record in self:
+            for offer in self.property_id.offer_ids:
+                if offer.status == "accepted":
```
qucol-odoo: This is slightly more pythonic/odoo-esque

```suggestion
            if any(self.property_id.offer_ids.filtered(lambda offer: offer.status == 'accepted')
```

estate/models/estate_offer.py
```diff
@@ -0,0 +1,53 @@
...
+                    raise UserError("Only one offer can be accepted")
+
+            record.status = "accepted"
+
+        return True
+
+    def set_status_refused(self):
```
qucol-odoo: Same as for the `accepted` action ☝️

estate/models/estate_offer.py
```diff
@@ -0,0 +1,53 @@
...
+    # Actions
+
+    def set_status_accepted(self):
+        for record in self:
+            for offer in self.property_id.offer_ids:
+                if offer.status == "accepted":
+                    raise UserError("Only one offer can be accepted")
+
+            record.status = "accepted"
+
```
qucol-odoo, line 71: Unnecessary blank lines, might as well remove them 😇

estate/models/estate_property.py
```diff
@@ -0,0 +1,120 @@
...
+        ondelete="restrict",
+    )
+    tag_ids = fields.Many2many("estate.tag")
+    offer_ids = fields.One2many("estate.offer", "property_id")
+
+    # Computed fields
+    total_area = fields.Integer(compute="_compute_total_area")
+    best_price = fields.Integer(compute="_compute_best_price")
+
+    @api.depends("offer_ids")
```
qucol-odoo: I believe you could be more specific with the dependence (same thing for `_compute_buyer_id` below)
```suggestion
    @api.depends("offer_ids.status")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,120 @@
...
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for record in self:
+            # Fallback to 0 if no records are present
+            record.best_price = max(record.offer_ids.mapped("price") + [0])
```
qucol-odoo: Nitpick, but no need to concatenate the lists here, you can use `or` for the fallback
```suggestion
            record.best_price = max(record.offer_ids.mapped("price") or [0])
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        self.garden_area = 10 if self.garden else None
+        self.garden_orientation = "north" if self.garden else None
```
qucol-odoo: Not the end of the world, but you're checking the same condition twice, you could just use a classic if/else to avoid that.

estate/models/estate_property.py
```diff
@@ -0,0 +1,120 @@
...
+    # Actions
+
+    def set_state_cancel(self):
+        for record in self:
+            if record.state == "sold":
+                raise UserError("A sold property cannot be cancelled")
+
+            record.state = "cancelled"
+
+        return True
```
qucol-odoo: Same comment as for the other actions above (and for the unnecessary blank lines) ☝️

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,97 @@
...
+
+    <record id="estate_property_search" model="ir.ui.view">
+        <field name="name">Search</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <search string="Delivery Carrier">
+                <field name="name" string="Estate" />
+                <field name="selling_price"/>
+                <separator/>
+                <filter string="Available" name="available" domain="['|', ('state', '=', 'new'), ('state', '=', 'received')]"/>
```
qucol-odoo:
```suggestion
                <filter string="Available" name="available" domain="[('state', 'in', ('new', 'received'))]"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,97 @@
...
+                </header>
+                <sheet>
+                    <group class="mb-5">
+                        <field name="name" string="" required="True" placeholder="My new house" class="fs-1" />
+                        <field name="tag_ids" string="" widget="many2many_tags"/>
+                    </group>
+                    <group>
+                        <group>
+                            <field name="state" readonly="True" />
+                            <field name="property_type_id" string="Property Type" />
```
qucol-odoo: The string here is unnecessary. Just like for the python fields, the name of the field is used by default (with the `_` replaced by spaces, and the `_id(s)` removed.

estate/__manifest__.py
qucol-odoo, line 1: I get the following warnings when initializing your db:

```
2026-08-19 09:47:19,989 38008 WARNING rd-demo-valvi odoo.modules.module: Missing `author` key in manifest for 'estate', defaulting to '' 
2026-08-19 09:47:19,989 38008 WARNING rd-demo-valvi odoo.modules.module: Missing `license` key in manifest for 'estate', defaulting to LGPL-3 
```

You can easily fix those warnings by adding the missing keys. The author should be `"Odoo S.A."`, and you can use the default license `LGPL-3` (it's still better to specify it in the manifest, as explicit is always better than implicit).

qucol-odoo (commented):
Helloo, thank you for your great work 🙏 I left a few comments, try to apply them whenever you find the time :) It's also a good practice to try to have a green runbot, you can see the current issues on the [runbot build page](https://runbot.odoo.com/runbot/batch/2710772/build/122148919) :)

estate/views/estate_property_views.xml
Topvennie: Fixed in 952067812b58c3287ca32de1db7aa947eb0e7aa7

estate/__manifest__.py
Topvennie: Fixed in [62d84fc](https://github.com/odoo/tutorials/pull/1388/commits/62d84fc31c42d6de2758153ebb1671f6b6f90898)

estate/views/estate_property_views.xml
Topvennie: Fixed in [62d84fc](https://github.com/odoo/tutorials/pull/1388/commits/62d84fc31c42d6de2758153ebb1671f6b6f90898)

estate/models/estate_property.py
Topvennie: Fixed in [62d84fc](https://github.com/odoo/tutorials/pull/1388/commits/62d84fc31c42d6de2758153ebb1671f6b6f90898)

estate/models/estate_property.py
Topvennie: Fixed in [62d84fc](https://github.com/odoo/tutorials/pull/1388/commits/62d84fc31c42d6de2758153ebb1671f6b6f90898)

estate/models/estate_offer.py
Topvennie: Fixed in [62d84fc](https://github.com/odoo/tutorials/pull/1388/commits/62d84fc31c42d6de2758153ebb1671f6b6f90898)

estate/models/estate_offer.py
Topvennie: Fixed in [62d84fc](https://github.com/odoo/tutorials/pull/1388/commits/62d84fc31c42d6de2758153ebb1671f6b6f90898)

estate/models/estate_offer.py
Topvennie: Fixed in [62d84fc](https://github.com/odoo/tutorials/pull/1388/commits/62d84fc31c42d6de2758153ebb1671f6b6f90898)

estate/models/estate_offer.py
Topvennie: Fixed in [62d84fc](https://github.com/odoo/tutorials/pull/1388/commits/62d84fc31c42d6de2758153ebb1671f6b6f90898)

estate/models/estate_offer.py
Topvennie: Fixed in [62d84fc](https://github.com/odoo/tutorials/pull/1388/commits/62d84fc31c42d6de2758153ebb1671f6b6f90898)

estate/models/estate_property.py
Topvennie: Fixed in [62d84fc](https://github.com/odoo/tutorials/pull/1388/commits/62d84fc31c42d6de2758153ebb1671f6b6f90898)

estate/models/estate_property.py
Topvennie: Fixed in [62d84fc](https://github.com/odoo/tutorials/pull/1388/commits/62d84fc31c42d6de2758153ebb1671f6b6f90898)

estate/models/estate_offer.py
```diff
@@ -0,0 +1,101 @@
...
+    def _inverse_date_deadline(self):
+        for record in self:
+            record.validity = (record.date_deadline - fields.Date.today()).days
+
+    # Actions
+
+    def action_status_accepted(self):
+        self.ensure_one()
+
+        if self.status == "accepted" or self.status == "refused":
```
qucol-odoo:
```suggestion
        if self.status in ('accepted', 'refused'):
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,129 @@
...
+    </record>
+
+    <record id="estate_property_view_search" model="ir.ui.view"> 
+        <field name="name">estate.property.view.search</field> 
+        <field name="model">estate.property</field> 
+        <field name="arch" type="xml">
+        <search>
+            <field name="name" string="Estate" />
+            <field name="selling_price" />
+            <field name="living_area" string="Living Area (sqm)" filter_domain="[('living_area', '>=', self)]" />
```
qucol-odoo: [Optional]
Although it technically doesn't break, the XML parser can sometimes struggle with some characters, such as `>`. 
```suggestion
            <field name="living_area" string="Living Area (sqm)" filter_domain="[('living_area', '&gt;=', self)]" />
```

estate/tests/test_estate_property.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def test_sell_no_accepted_offers(self):
+        with self.assertRaises(UserError):
+            self.property.action_state_sold()
+
+
+@tagged("post_install", "-at_install")
+class TestEstatePropertyHttp(HttpCase):
+    def test_garden_reset(self):
+        f = Form(self.env["estate.property"])
```
qucol-odoo: This is ok because the test/method is short, but you should try to use self-explanatory variable name. An easy and short alternative here was `form` or even `property_form`.

qucol-odoo (commented):
Very good job overall 👍 👏 I just left a few comments

In a real case scenario, it wouldn't make much sense to have commits fixing issues from other commits belonging to the same PR, because it means that you are introducing an issue and fixing it simultaneously: it would be better to simply not introduce it at all. It would be a great exercise for you to try to get rid of all the [FIX]/[CLN]/... commits in your PR (hint: you should probably use [git rebase -i](https://git-scm.com/docs/git-rebase#_interactive_mode)). Ultimately you could also squash commits to have one per chapter, or even squash everything as one big [ADD] commit, it's kinda up to you, as long as you experiment a bit with the interactive rebasing (mostly the edit and squash options) :)

estate/tests/test_estate_property.py
Topvennie: Fixed in [eb53f3a](https://github.com/odoo/tutorials/pull/1388/commits/eb53f3ae5c2ce6d6c9f99fb71962c6ce829731d3)

estate/views/estate_property_views.xml
Topvennie: Fixed in [eb53f3a](https://github.com/odoo/tutorials/pull/1388/commits/eb53f3ae5c2ce6d6c9f99fb71962c6ce829731d3)

estate/models/estate_offer.py
Topvennie: Fixed in [eb53f3a](https://github.com/odoo/tutorials/pull/1388/commits/eb53f3ae5c2ce6d6c9f99fb71962c6ce829731d3)

PR #1389, opened 2026-08-17
aryep - Technical training 

aryep-odoo:


estate/models/__init__.py
qucol-odoo, line 1: You're missing an empty line at the end of this file (and all the others too). It's a good practice to leave one. This is something you can setup in your config if you don't want to have to think about it. On VSCode, you can add the following to your `settings.json`:

```json
    "files.insertFinalNewline": true,
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,46 @@
+from odoo import fields, models
+from dateutil.relativedelta import relativedelta
```
qucol-odoo: Nitpick, but there's a guideline for the order of the imports: https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#imports

1. External libraries
2. Imports of odoo submodules
3. Imports from Odoo addons (rarely, and only if necessary)

Also, the lines should be alphabetically sorted. A good example is [here](https://github.com/odoo/odoo/blob/0824dc24665de6bfa805d540e756cdcb006edba6/addons/account/controllers/portal.py#L3-L14).

It's not always 100% respected, but it's a good and easy habit to have :)
```suggestion
from dateutil.relativedelta import relativedelta

from odoo import fields, models
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,94 @@
...
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="facades"/>
+
+                <!-- Filters and groups -->
+                <filter name="available_properties" string="Available Properties" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"/>
+                <filter name="postcode_grouping" string="Postcode" context="{'group_by': 'postcode'}"/>
+                <!-- <group> -->
+                <!-- </group> -->
```
qucol-odoo: Be careful not to push commented code. In bigger reviews it might be forgotten and end up in the codebase, which is useless and a bit unfortunate.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,94 @@
...
+
+
+    <!-- Main Form View -->
+     <record id="estate_property_form_view" model="ir.ui.view">
+        <field name="name">estate.property.view.form</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form string="Estate Property">
+                <sheet>
+                    <!-- Property Title -->
```
qucol-odoo, line 35: Nitpick, the comment doesn't hurt but it's not really necessary. We can already easily tell that this is the title of the property. Usually you want to add a comment when the code itself is not explicit enough, or to separate big segments.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,94 @@
...
+                <!-- Search fields -->
+                <field name="name"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="facades"/>
+
+                <!-- Filters and groups -->
+                <filter name="available_properties" string="Available Properties" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"/>
```
qucol-odoo:
```suggestion
                <filter name="available_properties" string="Available Properties" domain="[('state', 'in', ('new', 'offer_received'))]"/>
```

estate/__manifest__.py
qucol-odoo, line 1: I have a warning when initializing your db:
```
2026-08-18 12:01:13,565 81195 WARNING rd-demo-aryep odoo.modules.module: Missing `license` key in manifest for 'estate', defaulting to LGPL-3 
```
You probably have the same. It's usually a good idea to try to remove all warnings. In this case, you can just add the `'license'` key with the value `'LGPL-3'` in the `__manifest__.py`. Technically it won't change anything since it is the license used by default, but explicit is always better than implicit 👍 

```diff
+ 'license': 'LGPL-3',
```

Also, the ordering of the keys doesn't matter, but `author` and `license` tend to appear at the end of the manifest (e.g. [here](https://github.com/odoo/odoo/blob/19.0/addons/barcodes/__manifest__.py)). It's up to  you to decide whether you want to reorganize the keys to match the other manifests or not.

estate/models/estate_property.py
```diff
@@ -0,0 +1,46 @@
...
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate Property Management"
+
+    name = fields.Char(string="Title", required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        string="Available From", 
+        default=lambda _ : fields.Date.today() + relativedelta(months=3)
```
qucol-odoo, line 14: You have a few styling issues in your code, this is one of them. You can see them all on your runbot build [here](https://runbot.odoo.com/runbot/batch/2708285/build/122012677). It would be great to fix them to get a green runbot :)

```suggestion
        default=lambda _: fields.Date.today() + relativedelta(months=3)
```

qucol-odoo (commented):
Thank you for your great work 🙏 I left a few comments, try to apply the suggestions if you agree with them :) 

Could you try to rename your commits to match [our guideline](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html#commit-message-header)?

estate/models/estate_property.py
aryep-odoo: Will do! In regards to styling, is there any tool or linter (both in recommendations or pre-made config) that you recommend for the day-to-day stuff?

estate/__manifest__.py
aryep-odoo: Good to know, I have updated those keys and values to try and stick to those "guidelines" as well. Thanks for the comment

aryep-odoo:
@qucol-odoo thank you so much for your first batch of comments! There were definitely several things and details I had missed initially but most of them should have been resolved by now

I left some follow-ups in some of your recommendations with some questions. Let me know what you think, and once again, thanks for taking the time to check the PR

estate/models/estate_property.py
qucol-odoo: Yes there is :) We use [Ruff](https://docs.astral.sh/ruff/) to apply our styling guidelines. There's a configuration file in the Odoo repository [here](https://github.com/odoo/odoo/blob/19.0/ruff.toml), and you can also find the exact configuration used by the runbot on the [runbot build](https://runbot.odoo.com/runbot/batch/2709013/build/122026845).

image

If you use VSCode, you can use the [Ruff extension](https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff) to use Ruff as a linter :innocent:

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = False
+
+    @api.constrains("selling_price")
+    def _check_selling_price(self):
+        for record in self:
+            if record.selling_price and float_compare(record.selling_price, 0.9 * record.expected_price, precision_digits=2) < 0:
+                raise ValidationError(_("Selling price cannot be lower than 90% of the expected price."))
```
qucol-odoo, line 121: Good job finding the translation function! It is a better practice to use it as self.env._(...) than simply _, as explained in [its docstring](https://github.com/odoo/odoo/blob/189df3538aa07d22f683008522b9143421f411a2/odoo/orm/environments.py#L318-L324), for the reasons explained in [this commit](https://github.com/odoo/odoo/commit/b794f0f332f473deb2c04eba60baf4761db3b508).

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,86 @@
...
+            record.property_id.state = "offer_accepted"
+            record.property_id.partner_id = record.partner_id
+            record.property_id.selling_price = record.price
+
+            # Refuse other offers for the same property
+            other_offers = self.search([
+                ("property_id", "=", record.property_id.id),
+                ("id", "!=", record.id),
+            ])
+            other_offers.write({"status": "refused"})
```
qucol-odoo, line 82: Few things to say here:
1. An action should (usually) only operate on one record, so you don't need to loop over `self`, and instead you should call [`self.ensure_one()`](https://github.com/odoo/odoo/blob/6a3655c8efd2de68754ace38bd1ab5eb6499faca/odoo/orm/models.py#L5931) at the very start of the action.
2. When updating multiple fields of a same record, you can batch the writes using the `\write` method (as you did further down the method actually)
3. You should avoid using `search` in a loop. In this case it ends up being fine because we're removing the loop, but otherwise it would have been a bad practice. An alternative is to use [`_read_group`](https://github.com/odoo/odoo/blob/6a3655c8efd2de68754ace38bd1ab5eb6499faca/odoo/orm/models.py#L1861) and then to loop over the result ([example](https://github.com/odoo/odoo/blob/6a3655c8efd2de68754ace38bd1ab5eb6499faca/addons/hr_timesheet/models/project_project.py#L68-L79))
```suggestion
    def action_accept(self):
        self.ensure_one()
        self.status = "accepted"
        self.property_id.write({
            'state': 'offer_accepted',
            'partner_id': self.partner_id.id,
            'selling_price': self.price,
        })

        # Refuse other offers for the same property
        self.search([
            ("property_id", "=", self.property_id.id),
            ("id", "!=", record.id),
        ]).write({"status": "refused"})
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def action_refuse(self):
+        for record in self:
```
qucol-odoo, line 85:
```suggestion
        self.ensure_one()
```

estate/tests/common.py
```diff
@@ -0,0 +1,27 @@
...
+        super().setUpClass()
+
+        cls.estate_property_types = cls.env["estate.property.type"].create([
+            {"name": "House"},
+            {"name": "Apartment"},
+        ])
+        cls.estate_property_tags = cls.env["estate.property.tag"].create([
+            {"name": "tag1"},
+            {"name": "tag2"},
+        ])
```
qucol-odoo, line 17: If you don't use them, do not create them 👌

estate/tests/common.py
```diff
@@ -0,0 +1,27 @@
...
+        cls.estate_properties = cls.env["estate.property"].create([
```
qucol-odoo, line 18: Creating in batch is a great practice, but do you really need two properties? Their only difference is the price, which I don't think really changes anything in your tests. I would either:
- Create only one property
- Create two properties that have significant differences (for example, one with a garden and one without)

As a side note, keeping the two properties together doesn't sound like the best idea here, since it forces you to call `search' at the start of each test, which is unnecessarily  costly. Instead, you could do the following:
```suggestion
        cls.property_1, cls.property_2 = cls.env["estate.property"].create([
```

with more appropriate names if your properties have significant differences (e.g., `property_with_garden`, ...).

estate/tests/test_estate_property.py
```diff
@@ -0,0 +1,31 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        main_property = cls.estate_properties.search([("name", "=", "Property 1")])
+        main_property.living_area = 90
+        main_property.garden = True
+        main_property.garden_area = 100
+        main_property.garden_orientation = "south"
```
qucol-odoo, line 17: You could just create a property with those values from the start, instead of searching and editing an existing one

estate/tests/test_estate_property_offer.py
```diff
@@ -0,0 +1,55 @@
...
+    def test_sell_property_without_offer(self):
+        """Test that you can't sell properties that do not have any offer"""
+        main_property = self.estate_properties.search([("name", "=", "Property 1")])
+
+        for offer in self.estate_property_offers:
+            if offer.property_id.id == main_property.id:
+                offer.unlink()
+
+        with self.assertRaises(ValidationError):
+            self._sell_main_property()
```
qucol-odoo, line 55: The idea of the `_sell_main_property` is good, but here you search for the `main_property` twice because of it. You can fix that by applying my proposition above (separate the properties at creation), or by updating the `_sell_main_property` to something like `_sell_property(property)`.

estate/tests/test_estate_property_offer.py
```diff
@@ -0,0 +1,55 @@
...
+            },
+        ])
+
+    def test_sell_property_without_offer(self):
+        """Test that you can't sell properties that do not have any offer"""
+        main_property = self.estate_properties.search([("name", "=", "Property 1")])
+
+        for offer in self.estate_property_offers:
+            if offer.property_id.id == main_property.id:
+                offer.unlink()
```
qucol-odoo, line 52: Couldn't you just use the other property, that doesn't have any offers? Unless you want to test the deletion of offers, but then you should precise it in the docstring.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,156 @@
...
+        <field name="name">estate.property.view.search</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <search>
+                <!-- Search fields -->
+                <field name="name"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area" filter_domain="[('living_area', '>=', self)]"/>
```
qucol-odoo, line 146: Some characters can be misinterpreted by the XML parser, it's always better to be cautious
```suggestion
                <field name="living_area" filter_domain="[('living_area', '&gt;=', self)]"/>
```

qucol-odoo (commented):
Very nice work, good job 👌 I left a few comments :)

Regarding your commit history, I have two comments to share:
1. You should only have two `[ADD]` commits, because you're only adding two modules: `estate` and `estate_account`. All the other commits that are not adding new modules should (probably) be `[IMP]`. As a reminder, you can see all the tags [here](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html#tag-and-module-name).
2. In a real case scenario, it wouldn't make much sense to have commits fixing logic/syntax issues from other commits belonging to the same PR, because it means that you are introducing an issue and fixing it simultaneously: it would be better to simply not introduce it at all. It would be a great exercise for you to try to get rid of all the [FIX] commits in your PR (hint: you should probably use [git rebase -i](https://git-scm.com/docs/git-rebase#_interactive_mode)). Ultimately you could also squash commits to have one per chapter, or even squash everything as one big [ADD] commit, it's kinda up to you, as long as you experiment a bit with the interactive rebasing (mostly the edit and squash options) :)

PR #1390, opened 2026-08-17
Server technical tutorial (josiv)

jozefsivek:


estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,100 @@
...
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="facades"/>
+
+                <filter
+                    string="Available"
+                    name="available"
+                    help="display only available estates"
+                    domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"/>
```
qucol-odoo: You can use the `in` operator here :)

```suggestion
                    domain="[('state', 'in', ('new', 'offer_received'))]"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,100 @@
...
+                                <field name="bedrooms"/>
+                                <field name="living_area"/>
+                                <field name="facades"/>
+                                <field name="garage"/>
+                                <field name="garden"/>
+                                <field name="garden_area"/>
+                                <field name="garden_orientation"/>
+                            </group>
+                        </page>
+                    </notebook>
```
qucol-odoo, line 217: Not necessarily important, but you could have added the `state` field somewhere in there too. Without adding the field anywhere, I don't know how you could test the behavior of your "Available" filter.

estate/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+{
+    'name': "A Real Estade Advestisement Demo",
+    'version': '1.0',
+    'depends': ['base'],
+    'author': "Jozef Sivek (josiv)",
```
qucol-odoo: Totally fine for a tutorial module, but just so you know, in a real case scenario, you'd have to set the author to `'Odoo S.A.'`.

estate/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    'data': [
+        'security/ir.model.access.csv',
+
+        'views/estate_property_views.xml',
+
+        'views/estate_menus.xml',
+    ],
+    # data files containing optionally loaded demonstration data
+    'demo': [
+        #'demo/demo_data.xml',
```
qucol-odoo: A general rule of thumb is to avoid pushing commented code. In a bigger review, it can always be forgotten and end up in the database, which is unfortunate.

estate/models/estate_property.py
```diff
@@ -0,0 +1,36 @@
...
+    state = fields.Selection(
+            selection=[
+                ('new', 'New'),
+                ('offer_received', 'Offer Received'),
+                ('offer_accepted', 'Offer Accepted'),
+                ('sold', 'Sold'),
+                ('cancelled', 'Cancelled')],
+            default='new',
+            help="State of the estate property")
+
```
qucol-odoo: You have a few styling errors, including this one (you should only have one blank line at the end of your file, not two).

```suggestion
```

qucol-odoo (commented):
Very good job so far 👌 I wrote a few comments, try to have a look at them and apply them when you find the time :) 

Also, it would be good if you could follow our [git commit guidelines](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html#tag-and-module-name). Your commit name should always follow the pattern `[TAG] <module_name>: <description>`. You should try to rename your first commits ([hint](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html#tag-and-module-name)), and apply that guideline for all the future ones 😇

estate/views/estate_property_views.xml
jozefsivek: Good point, I think I added at first, but removed it because it did not looked like breadcrumbs we used to see. I was thinking it will be added back later in the chapter as better UI element. I will add it again.

estate/models/estate_property.py
jozefsivek: Indeed. I did not install linter yet. I will do at as soon as possible, as there is no single one to choose blindly.

qucol-odoo: My advice would be to use [Ruff](https://docs.astral.sh/ruff/), as it is the one we use on the Runbot, which will block your PRs if you don't respect the styling guidelines. The configuration file we use is available on the github repo [here](https://github.com/odoo/odoo/blob/19.0/ruff.toml). There's also a [VSCode extension](https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff).

PR #1391, opened 2026-08-18
Onboarding Juica Server 101

JuanICasareski:


estate/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+from . import estate_property
+from . import estate_property_type
+from . import estate_property_tag
+from . import estate_property_offer
```
yoba-odoo: We should always have a newline by the end of each file

estate/models/estate_property.py
```diff
@@ -0,0 +1,76 @@
...
+    name = fields.Char('Property Name', required=True, translate=True)
+    description = fields.Text('Description', translate=True)
+    postcode = fields.Char('Post Code', required=True)
+    date_availability = fields.Date(
+        'Availability Date', 
+        required=True, 
+        copy=False,
+        default=fields.Date.today() + relativedelta(months=3)
+    )
+
```
yoba-odoo: Not the best practice to have a newline in between the fields' definition, readability wise

estate/models/estate_property.py
```diff
@@ -0,0 +1,76 @@
+from odoo import fields, models
+from dateutil.relativedelta import relativedelta
+
```
yoba-odoo, line 5: We normally should always have two blank lines before the definition of the class

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,8 @@
+from odoo import fields, models
+
+class EstatePropertyType(models.Model):
```
yoba-odoo: heads up for the typo here. This is EstatePropertyTag not Type 😅

estate/models/estate_property.py
```diff
@@ -0,0 +1,102 @@
...
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for property in self:
+            property.best_price = max(property.offer_ids.mapped("price"), default=0)
+
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        self.garden_area = 10 if self.garden else 0
+        # Nasty magic string. I should turn the option into a variable and then reference it
+        self.garden_orientation = "north" if self.garden else None
```
yoba-odoo, line 96:
```suggestion
        if self.garden:
                self.garden_area = 10
                # Nasty magic string. I should turn the option into a variable and then reference it
                self.garden_orientation = "north"
        else:
                self.garden_area = 0
                self.garden_orientation = None
```
better to have it like this instead of checking twice, wdyt?

yoba-odoo (commented):
Good work 💪 
Just a few comments, and always add a newline by the end of the file

estate/models/estate_property_tag.py
JuanICasareski: haha I shouldn't be copying file so freely, fixed

estate/models/estate_property.py
JuanICasareski: I would normally prefer a `return` inside the `if self.garden` and what would've been the else block as a fallback.

Something along the lines of:
```python
if self.garden:
   self.garden_area = 10
   return

self.garden_area = 0
```

idk enough about the company's guidelines or maybe what you usually do, so....

yoba-odoo: That would be a good approach as well :ok_hand:

estate/models/__init__.py
JuanICasareski: Damn I should've set up the linter haha

estate/models/estate_property.py
JuanICasareski: Done!

estate/models/estate_property.py
JuanICasareski: Done!

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,90 @@
...
+            offer.date_deadline = fields.Date.today() + relativedelta(days=offer.validity)
+
+    def _inverse_deadline(self):
+        for offer in self:
+            delta = offer.date_deadline - fields.Date.today()
+            offer.validity = delta.days
+
+    def action_offer_accept(self):
+        for offer in self:
+            if offer.property_state in ('offer_accepted', 'sold', 'cancelled'):
```
yoba-odoo: If we are introducing the field `property_state` just for this check so I guess we can remove it and we do something like

```suggestion
            if offer.property_id.state in ('offer_accepted', 'sold', 'cancelled'):
```
It is better not to add a field if we can access it in an easier way

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,90 @@
...
+        for v in values:
+            pid = v['property_id']
+            prop_min_offer[pid] = min(prop_min_offer.get(pid, float('inf')), v.get('price', 0))
+
+        # Browse the properties referenced by the new offers
+        properties = self.env['estate.property'].browse(prop_min_offer.keys())
+
+        for prop in properties:
+            # No new offer may be lower than the best existing one
+            best_existing = max(prop.offer_ids.mapped('price'), default=0)
```
yoba-odoo: Since you are already sorting the offers upon their price in line 10 so you won't have to do this with max

```suggestion
            best_existing = prop.offer_ids[0]
```

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,58 @@
...
+            >
+                <field name="price" />
+                <field name="partner_id" />
+                <field name="validity" />
+                <field name="date_deadline" />
+                <button 
+                    name="action_offer_accept"
+                    string="Accept"
+                    icon="fa-check"
+                    invisible="property_state in ('offer_accepted', 'sold', 'cancelled')"
```
yoba-odoo: Since we are already doing a check in python to avoid accepting a property with these states I think here we can use 'status'

```suggestion
                    invisible="status"
```

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,58 @@
...
+                    type="object"
+                />
+                <button 
+                    name="action_offer_refuse"
+                    string="Refuse"
+                    icon="fa-times"
+                    invisible="status in ('accepted', 'refused')"
```
yoba-odoo:
```suggestion
                    invisible="status"
```
Since status only have two values so it is either gonna have one of those values or it will be false so just checking that it has a value is enough I think

estate_account/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_property
```
yoba-odoo: newline at the end of file 😅

yoba-odoo (commented):
Good work 👏 👏 

Just a few comments. Also try to make the ci/* checks to pass, check what are the errors and fix them

PR #1392, opened 2026-08-18
[EMTAL] - Server Framework 101 tutorial work

Emivvvvv:


estate/models/estate_property.py
```diff
@@ -0,0 +1,88 @@
+from odoo import api, fields, models
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate Property Model"
+
+    name = fields.Char("Property Name", required=True)
+    description = fields.Text("Description")
```
qucol-odoo, line 9: By default, the "string" of a field is the name of that field with the _ replaced by spaces ( ). In case you're curious, that is defined [here](https://github.com/odoo/odoo/blob/0824dc24665de6bfa805d540e756cdcb006edba6/odoo/orm/fields.py#L506-L511).

In this case (and others below), it's not necessary to specify the string of your field, as it matches the default behavior anyway.

estate/models/estate_property.py
```diff
@@ -0,0 +1,88 @@
...
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    best_price = fields.Float("Best Price", compute="_find_best_price")
+
+    @api.depends("offer_ids.price")
+    def _find_best_price(self):
```
qucol-odoo, line 79: Compute methods should always start with `compute_`
```suggestion
    def _compute_best_price(self):
```

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+estate.property,estate.property,model_estate_property,base.group_user,1,1,1,1
+estate.property.type,estate.property.type,model_estate_property_type,base.group_user,1,1,1,1
+estate.property.tag,estate.property.tag,model_estate_property_tag,base.group_user,1,1,1,1
+estate.property.offer,estate.property.offer,model_estate_property_offer,base.group_user,1,1,1,1
```
qucol-odoo, line 5: Missing a line at the end of the file, that will make the runbot red :( If you're using VSCode, you can add the following to your `settings.json` to have that line added automatically:

 ```json
    "files.insertFinalNewline": true,
```
Fyi that blank line seems to be missing in all your non-python files.

File: estate/views/estate_property_tag_views.xml [hunk 4]
```diff
@@ -0,0 +1,21 @@
...
+    </record>
+
+    <record id="estate_property_tag_view_form" model="ir.ui.view">
+        <field name="name">estate.property.tag.form</field>
+        <field name="model">estate.property.tag</field>
+        <field name="arch" type="xml">
+            <form string="Estate Tag Form">
+                <sheet>
+                        <field name="name"/>
+                </sheet>
```
qucol-odoo, line 17: That view is a little sad like this.

Image

 In most/all cases, the child of a `<sheet>` should be a container. In this case, you should use a group (this is what is advised in the [documentation](https://www.odoo.com/documentation/19.0/developer/reference/user_interface/view_architectures.html#sheet-make-the-layout-responsive) aswell).


```suggestion
                <sheet>
                        <group>
                                <field name="name"/>
                        </group>
                </sheet>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,102 @@
...
+        <field name="arch" type="xml">
+            <search string="Estate Property Search">
+                <field name="name"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="facades"/>
+
+                <filter name="available_properties" string="Available Properties" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]" />
```
qucol-odoo, line 94: You could use the `'in'` operator
```suggestion
                <filter name="available_properties" string="Available Properties" domain="[('state', 'in', ('new', 'offer_received'))]" />
```

estate/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+        "security/ir.model.access.csv",
+        "views/estate_property_views.xml",
+        "views/estate_property_type_views.xml",
+        "views/estate_property_tag_views.xml",
+        "views/estate_property_offer_views.xml",
+        "views/estate_menus.xml",
+    ],
+    "installable": True,
+    "application": True,
+    "author": "emtal",
```
qucol-odoo, line 17: Just so you know, the author should be `"Odoo S.A."`, but this is of course fine for a tutorial module :)

qucol-odoo (commented):
Thank you for the great work :) I highlighted a few stuff that you can have a look at whenever you have the time 👌 

Good job also on respecting the [commit naming convention](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html#tag-and-module-name) 👏 Just FYI, theoretically only the first commit should start with [ADD], as the other ones don't add any new module, they just add features into the existing module (thus they should start with [IMP]). In a more realistic setting, your entire PR could be a singular big [ADD] commit.

estate/models/estate_property.py
Emivvvvv: Thanks I'll change them. Btw, is there and documentation for general Odoo naming conventions?

qucol-odoo: Yes there is: https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions :)

estate/security/ir.model.access.csv
Emivvvvv: I don't use vscode but from now on I'll also start formatting xml files, so this should not be a problem. Thanks for letting me know.

estate/models/estate_property_offer.py
```diff
@@ -44,3 +46,23 @@ def _inverse_date_deadline(self):
...
+        for offer in self:
+            # Check property have already been sold or cancelled
+            if offer.property_id.state in ["sold", "cancelled"]:
+                raise UserError("This property has already been sold or cancelled!")
+
+            # Any accepted offer?
+            if any(o.status == "accepted" for o in offer.property_id.offer_ids):
+                raise UserError("An offer have already been accepted!")
+
+            # Accept offer
```
qucol-odoo: IMO, these comments are unnecessary, your code is already very self-explanatory (good job for that 👍)

estate/models/estate_property.py
```diff
@@ -0,0 +1,155 @@
...
+            ("sold", "Sold"),
+            ("canceled", "Canceled"),
+        ],
+        default="new",
+        copy=False,
+    )
+    active = fields.Boolean(default=True)
+
+    # Many2one references
+    type_id = fields.Many2one(comodel_name="estate.property.type")
```
qucol-odoo, line 50: Not an issue, more like a piece of info: the first argument of a `Many2one` field is always the `comodel_name` (see its [`__init__` method](https://github.com/odoo/odoo/blob/189df3538aa07d22f683008522b9143421f411a2/odoo/orm/fields_relational.py#L250)), so its not necessary to specify it.

Image

Image

estate/models/estate_property.py
```diff
@@ -0,0 +1,155 @@
...
+    def _compute_best_price(self):
+        for record in self:
+            prices = record.offer_ids.mapped("price")
+            record.best_price = max(prices) if prices else 0.0
+
+    # On change
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        self.garden_area = 10 if self.garden else 0
+        self.garden_orientation = "north" if self.garden else "n/a"
```
qucol-odoo: You're checking the same condition twice. Not a big issue, but might as well use a classic if/else to avoid it.

estate/models/estate_property.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def action_property_sold(self):
+        for record in self:
```
qucol-odoo: Most actions should only be called on a single record, and thus should start with [`self.ensure_one()`](https://github.com/odoo/odoo/blob/189df3538aa07d22f683008522b9143421f411a2/odoo/orm/models.py#L5931) instead of a for loop.

estate/models/estate_property.py
```diff
@@ -0,0 +1,155 @@
...
+            if record.state == "canceled":
+                raise UserError(
+                    _("This property has already been canceled. It can not be sold!")
```
qucol-odoo: Good job finding the translation function! It is a better practice to use it as `self.env._(...)` than simply `_`, as explained in [its docstring](https://github.com/odoo/odoo/blob/189df3538aa07d22f683008522b9143421f411a2/odoo/orm/environments.py#L318-L324), for  the reasons explained in [this commit](https://github.com/odoo/odoo/commit/b794f0f332f473deb2c04eba60baf4761db3b508).

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,92 @@
...
+    def _inverse_date_deadline(self):
+        for record in self:
+            base_date = (
+                record.create_date.date() if record.create_date else fields.Date.today()
+            )
+            if record.date_deadline and base_date:
+                record.validity = (record.date_deadline - base_date).days
+
+    def action_confirm_offer(self):
+        for offer in self:
```
qucol-odoo: Same comment as for the other action method, would it make sense to confirm multiple offers at once?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,92 @@
...
+
+            # Any accepted offer?
+            if any(o.status == "accepted" for o in offer.property_id.offer_ids):
+                raise UserError(_("An offer have already been accepted!"))
+
+            # Accept offer
+            offer.status = "accepted"
+            offer.property_id.state = "offer_accepted"
+            offer.property_id.selling_price = offer.price
+            offer.property_id.buyer_id = offer.partner_id
```
qucol-odoo: You can batch those writes, it slightly more efficient
```suggestion
            offer.property_id.write({
                'state': 'offer_accepted',
                'selling_price': offer.price,
                'buyer_id': offer.partner_id.id,
            })
```

.gitignore
qucol-odoo, line 1: This and the `dprint.json` file shouldn't be here 😇

estate_account/security/ir.model.access.csv
qucol-odoo, line 1: There's no point in creating the access file if you don't create any rule alongside it. It would make more sense to create the file when needing to add access groups related to the module :)

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,181 @@
...
+    <field name="arch" type="xml">
+      <search string="Estate Property Search">
+        <field name="name" />
+        <field name="postcode" />
+        <field name="expected_price" />
+        <field name="bedrooms" />
+        <field
+          name="living_area"
+          string="Min Living Area"
+          filter_domain="[('living_area', '>=', self)]"
```
qucol-odoo: There are a few symbols that can mess with the XML parser, including `>` and `<`, which, as a safety measure, should be replaced by `&gt;` and `&lt;`. 
```suggestion
          filter_domain="[('living_area', '&gt;=', self)]"
```

qucol-odoo (commented):
Good job, you really flew through the tutorial with ease 💪 I highlighted a few things worth mentioning, but overall everyone is really good 👌 

In a real case scenario, it wouldn't make much sense to have commits fixing issues from other commits belonging to the same PR, because it means that you are introducing an issue and fixing it simultaneously: it would be better to simply not introduce it at all. It would be a great exercise for you to try to get rid of all the [FIX] commits in your PR (hint: you should probably use [`git rebase -i`](https://git-scm.com/docs/git-rebase#_interactive_mode)). Ultimately you could also squash commits to have one per chapter, or even squash everything as one big [ADD] commit, it's kinda up to you, as long as you experiment a bit with the interactive rebasing :)

estate_account/security/ir.model.access.csv
Emivvvvv: Haahah you're correct. I left it since I copy/pasted the new module from the old one. But yeah will remove it with it's references 👍🏻

estate/models/estate_property.py
Emivvvvv: Thanks for the references! This seems like an interesting topic, will definitely check the resources that you've sent! (and also update my usace as `self.env._`)

estate/models/estate_property.py
Emivvvvv: Good catch 😭

estate/models/estate_property.py
Emivvvvv: I kinda prefer specifying the argument names since python don't have strong type system. But yeah thanks for pointing it out. I'll meet at the middle with you and ensure all `comodel_name` arguments will be the first argument. If the more idiomatic way is not having them ofc I can remove them too.

estate/models/estate_property_offer.py
Emivvvvv: Yes you're right. I might have missed this one. Never seen `self.ensure_one()` before. Will update accordingly. Thanks.

PR #1402, opened 2026-09-02
[Add] Real Estate property Model

losbo-odoo:


awesome_dashboard/__manifest__.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 {
-    'name': "Awesome Dashboard",
+    'name': "Awesome Dashboard1",
```
bit-odoo: unneccary changes.

estate/models/__init__.py
bit-odoo: Can you please import it as per Odoo standard?

PR #1412, opened 2026-09-15
[ADD] : Creation of the new module

macai-odoo:
Module creation as requested in the tutorial


estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_property
```
msho-odoo: Always add a new line for end of files :)

estate/models/estate_property.py
```diff
@@ -0,0 +1,24 @@
...
+    bedrooms = fields.Integer('Bedrooms')
+    living_area = fields.Integer('Living Area')
+    facades = fields.Integer('Facades')
+    garage = fields.Boolean('Garage')
+    garden = fields.Boolean('Garden')
+    garden_area = fields.Integer('Garden Area')
+    garden_orientation = fields.Selection(
+        string='Garden Orientation',
+        selection=[('North','North'),('South','South'),('East','East'),('West','West')]
+    )
```
msho-odoo, line 32: The one line on the selection field is okay but when you have many values, it's better to have it like this.
Also, it's preferable to have the key as all small letters, so no confusion happens when they are used inside the code in `if` statements for example.

so for example the selection can be  ('east', "East") instead of ('east', 'East') but some teams don't do this, they just stick to their own convention which is all single quotes (if possible) or all double quotes
But I see you have all single quotes so good for me but thought to mention it so you know.

```suggestion
    garden_orientation = fields.Selection(
        string='Garden Orientation',
        selection=[
            ('north', 'North'),
            ('south', 'South'),
            ('east', 'East'),
            ('west', 'West')
        ]
    )
```

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate,estate,model_estate_property,base.group_user,1,1,1,1
```
msho-odoo: EOF new line please for all files :) 
Also alwyas refer to your runbot `ci/style`, if it's red, it will be telling you in the logs what's to fix :)

msho-odoo: better to have it like this

```suggestion
access_estate_property_user,access.estate.property,model_estate_property,base.group_user,1,1,1,1
```

The `access_` thing is just a convention for this and I added `_user` at the end because this access rule is for the base user, you might have other rules on the same model but for `_manager` for example and son on.

estate/__manifest__.py
```diff
@@ -0,0 +1,6 @@
+{
+    'name':'Real Estate',
+    'version':'1.0',
+    'depends':['base'],
+    'data':['security/ir.model.access.csv']
```
msho-odoo:
```suggestion
    'data': [
        'security/ir.model.access.csv',
    ]
```
there is also a space after each column `: ` :)

msho-odoo (changes requested):
Nice work 🔥 

For the commit message we write in a way that aligns with `when this commit is merged it will .....` and we add the name of the modules before the column `: `
so you might change it to something like [IMP] estate: add estate property model or however you like to describe your changes :)
Also, left you some comments, thanks!

msho-odoo:
@macai-odoo  please mention me here when you think the PR is ready for another review so I know, typically, after each chapter :)

macai-odoo:
Hello !
@msho-odoo I just finished the chapter 6 of the training !

Thank you !

estate/models/estate_property.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class EstateProperty(models.Model):
+    _name = 'estate.property'
+    _description = 'Real Estate property model'
+
+    name = fields.Char('Name', required=True)
+    description = fields.Text('Description')
+    postcode = fields.Char('Postcode', required=True)
+    date_availability = fields.Date('Availability', copy=False, default=fields.Date.add(fields.Date.today(), months=3))
```
msho-odoo: For the default in this case the date will be valuated when first imported only, you need to call a function for the default here, check `lambda` on the code base and see how it's used in the default of the fields and use it here :)

estate/models/estate_property.py
```diff
@@ -0,0 +1,38 @@
...
+    facades = fields.Integer('Facades')
+    garage = fields.Boolean('Garage')
+    garden = fields.Boolean('Garden')
+    garden_area = fields.Integer('Garden Area')
+    garden_orientation = fields.Selection(
+        string='Garden Orientation',
+        selection=[('north', 'North'),
+                   ('south', 'South'),
+                   ('east', 'East'),
+                   ('west', 'West')]
```
```suggestion
        selection=[
            ('north', 'North'),
            ('south', 'South'),
            ('east', 'East'),
            ('west', 'West'),
        ]
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+    active = fields.Boolean('Active', default=True)
+    state = fields.Selection(
+        string='State',
+        selection=[('new', 'New'),
+                   ('offer_received', 'Offer Received'),
+                   ('offer_accepted', 'Offer accepted'),
+                   ('sold', 'Sold'),
+                   ('cancelled', 'Cancelled')],
```
msho-odoo: same styling as previous selection field :)

estate/views/estate_menu.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<odoo>
+    <menuitem id="test_menu_root" name="Real Estate">
+        <menuitem id="test_first_level_menu" name="Advertisements">
```
msho-odoo: ids should be a bit more descriptive than that so other developers know mostly what it does from the name :)

macai-odoo:
This is the chapter 7 of the training with the implementation of : 

- Property Type
- Property Tags
- Property Offers

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,106 @@
...
+                        <page string="Description">
+                            <group>
+                                <field name="description" string="Description"/>
+                            </group>
+                            <group>
+                                <field name="bedrooms" string="Bedrooms"/>
+                            </group>
+                            <group>
+                                <field name="living_area" string="Living Area (sqm)"/>
+                            </group>
```
msho-odoo: hmmm I think you don't need to add each field to a `<group>` tag, you do this to style different groups of fields that you want to show together, feel free to have a look at one of odoo's existing form view in any app and have a look at it's form xml record to see it better.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,106 @@
...
+                            <group>
+                                <field name="facades" string="Facades"/>
```
msho-odoo, line 71: it's totally correct to have `string` attribute on the field in the xml view but you can have it on the field definition on the python model and don't use it here so you don't have to add the `string` attribute every time you use it in a view (unless you want to show it differently in certain views)

estate/models/estate_property.py
```diff
@@ -0,0 +1,38 @@
...
+    name = fields.Char('Name', required=True)
+    description = fields.Text('Description')
+    postcode = fields.Char('Postcode', required=True)
+    date_availability = fields.Date('Availability', copy=False, default=fields.Date.add(fields.Date.today(), months=3))
+    expected_price = fields.Float('Expected Price', required=True)
+    selling_price = fields.Float('Selling Price', readonly=True, copy=False)
+    bedrooms = fields.Integer('Bedrooms', default=2)
+    living_area = fields.Integer('Living Area')
+    facades = fields.Integer('Facades')
+    garage = fields.Boolean('Garage')
```
msho-odoo: It's better practice to have boolean fields as `is_something` or `has_something`, so you may rename the fields to `has_garden` and `has_garage`

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,106 @@
...
+            <search string="Properties">
+                <field name="name" string="Title"/>
+                <field name="postcode" string="Postcode"/>
+                <field name="expected_price" string="Expected Price"/>
+                <field name="bedrooms" string="Bedrooms"/>
+                <field name="living_area" string="Living Area (sqm)"/>
+                <field name="facades" string="Facades"/>
+                <separator/>
+                <filter string="Available" name="active" domain="['|', ('state', '=', 'new'),
+                                                                        ('state', '=', 'offer_received')]"/>
```
msho-odoo: Stick to the correct convention you were following `name` attribute before `string` attribute :)
Also, the domain like this is more readable (yours is still correct)
```suggestion
                <filter name="active" string="Available" domain="[('state', 'in',  ('new', 'offer_received'))]"/>
```

estate/__manifest__.py
```diff
@@ -0,0 +1,12 @@
...
+    'name': 'Real Estate',
+    'version': '1.0',
+    'depends': ['base'],
+    'application': True,
+    'installable': True,
+    'data': [
+        'security/ir.model.access.csv',
+        'views/estate_property_views.xml',
+        'views/estate_menu.xml',
+            ]
```
msho-odoo:
```suggestion
    'data': [
        'security/ir.model.access.csv',
        'views/estate_property_views.xml',
        'views/estate_menu.xml',
    ]
```

msho-odoo (changes requested):
Thanks for the nice work and for the green `ci/style` on runbot :)
Have a look a the warning in `ci/tutorials` and see how to get rid of them, it's stated in the logs there.

Left you some comments :)
Remember to ping again when ready for review, Thanks!

macai-odoo:
@msho-odoo Hello !
I've completed the chapter 8, thank you in advance for your feedback !

I also have a small question regarding the compute field.
In most of the exemple, they used a for each loop on self, is it mandatory to do so ? 
(Got the answer in chapter 9 :) )
Thanks !

aboo-odoo:
Hey I'll be managing your review now to spread the load among the 4 of us :smile:

> I also have a small question regarding the compute field. In most of the exemple, they used a for each loop on self, is it mandatory to do so ? 

Glad you got the answer :+1: Indeed, `self` is a recordset, i.e. a group of one or more records (several estate properties for example). Due to ORM optimizations, we try to manage recordset as much as possible instead of managing individual records independently.

macai-odoo:
@aboo-odoo
Hey !
I think I'm ready for another review,
Thank you in advance :)

estate/models/__init__.py
```diff
@@ -0,0 +1,5 @@
...
+from . import estate_property_type
+from . import estate_property_tag
+from . import estate_property_offer
+from . import res_user
```
aboo-odoo: Always leave an empty line at the end of a file, 

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
+from odoo import models, fields, api
```
aboo-odoo: [Imports](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#imports)
The imports are ordered as: 
- Imports of odoo submodules

```suggestion
from odoo import api, fields, models
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+from odoo.tools import float_is_zero
+from odoo.tools.float_utils import float_compare
+
+
+class EstateProperty(models.Model):
+    _name = 'estate.property'
+    _description = 'Real Estate property model'
+    _order = 'id desc'
+
+#   Property information
```
aboo-odoo: Not needed 😄

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+    name = fields.Char('Name', required=True)
```
aboo-odoo, line 12: For your personal knowledge, you could also

```suggestion
    name = fields.Char(required=True)
```
By default, the string of a field is that field name 😄 But leaving it is not an error either, it's personal preferences 👍

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+    description = fields.Text('Description')
+    postcode = fields.Char('Postcode', required=True)
+    date_availability = fields.Date('Availability', copy=False, default=lambda self: fields.Date.add(fields.Date.today(), months=3))
```
aboo-odoo, line 15: Nice o use a lambda method here for `default`, good job 👍

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+    has_garage = fields.Boolean('Garage')
+    has_garden = fields.Boolean('Garden')
+    garden_area = fields.Integer('Garden Area')
+    garden_orientation = fields.Selection(
+        string='Garden Orientation',
+        selection=[('north', 'North'),
+                   ('south', 'South'),
+                   ('east', 'East'),
+                   ('west', 'West'),
+        ]
```
aboo-odoo, line 30: Related to the comment of Moamen: some teams like UI strings to be double quoted; this also applies to all your other strings. Don't change them all because it's not meaningful but just keep it in mind.
```suggestion
        selection=[
            ('north', "North"),
            ('south', "South"),
            ('east', "East"),
            ('west', "West"),
        ]
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+    )
+    total_area = fields.Integer(compute='_compute_total_area', string='Total Area')
```
aboo-odoo, line 32: Nitpick; string first
```suggestion
    total_area = fields.Integer(string='Total Area', compute='_compute_total_area')
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+        string='State',
+        selection=[('new', 'New'),
+                   ('offer_received', 'Offer Received'),
+                   ('offer_accepted', 'Offer Accepted'),
+                   ('sold', 'Sold'),
+                   ('cancelled', 'Cancelled'),
+        ],
+        required=True,
+        copy=False,
+        default='new'
```
aboo-odoo, line 44: Like for the EOL, always add a comma at the end of multi-line dict, list, function parameter, etc.
```suggestion
        default='new',
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+    )
+    property_type_id = fields.Many2one("estate.property.type", string='Property Type')
+
+#   Other Information
+    salesman_id = fields.Many2one('res.users', string='Salesman', default=lambda self: self.env.user)
+    buyer_id = fields.Many2one('res.partner', string='Buyer', copy=False)
+
+#   Tags
+    tag_ids = fields.Many2many('estate.property.tag', string='Tags')
```
aboo-odoo, line 53: Good use of the suffixes `_id.s` ! Well done 😄

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+                   ('sold', 'Sold'),
+                   ('cancelled', 'Cancelled'),
+        ],
+        required=True,
+        copy=False,
+        default='new'
+    )
+    property_type_id = fields.Many2one("estate.property.type", string='Property Type')
+
+#   Other Information
```
aboo-odoo, line 48:
```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+    salesman_id = fields.Many2one('res.users', string='Salesman', default=lambda self: self.env.user)
+    buyer_id = fields.Many2one('res.partner', string='Buyer', copy=False)
+
+#   Tags
```
aboo-odoo, line 52:
```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+    )
+
+#   Python:
+    @api.constrains('selling_price')
+    def _check_selling_price(self):
+        for record in self:
+            if not float_is_zero(record.selling_price, 2):
+                if float_compare(record.selling_price, record.expected_price, 2) == -1:
+                    raise UserError('The Selling Price must be at least 90% of the Expected Price,'
+                                    ' you must change the expected price to accepted this offer')
```
aboo-odoo, line 77: Most strings are translated by default in Odoo. Unfortunately, this is not case of Error Messages. To make the string translatable, you need to call the `_()` method, like so
```suggestion
                    raise UserError(self.env._(
                        'The Selling Price must be at least 90% of the Expected Price,'
                        ' you must change the expected price to accepted this offer'
                    ))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+
+#   Computation Fields:
+
+    @api.depends('living_area', 'garden_area')
+    def _compute_total_area(self):
+        for record in self:  # QUESTION is the for each loop here necessary ? or can I just use self ?
```
aboo-odoo, line 83: >> ANSWER: No, you need to use the loop since _compute methods are called with self being a recordset; a set of record. Thus, since you cannot perform your addition on all the records at once, you need to loop through them.

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends('offer_ids.price')
+    def _compute_best_price(self):
+        for record in self:
+            if len(record.offer_ids) > 0:
+                record.best_price = max(record.offer_ids.mapped('price'))
+            else:
+                record.best_price = 0.
```
aboo-odoo, line 92: Could be simplified into 
```suggestion
            record.best_price = max(record.offer_ids.mapped('price'), default=0)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+        if self.has_garden:
+            self.garden_area = 10
+            self.garden_orientation = 'north'
+        else:
+            self.garden_area = None
+            self.garden_orientation = None
+
+#   Actions buttons
+    def action_cancel_property_state(self):
+        for record in self:
```
aboo-odoo, line 105: Good, like for _compute methods, this could be called with a recordset which would fail without the loop

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+            if record.state == 'cancelled':
+                raise UserError('Cancelled properties cannot be cancelled again !')
+            elif record.state == 'sold':
+                raise UserError('Sold Properties cannot be cancelled !')
+            record.state = 'cancelled'
+            return True
```
aboo-odoo, line 111: Careful here that you want to check all the records before returning
```suggestion
```

Another, slightly better way, would be to do like so:
```suggestion
        properties_to_cancel = self.filtered(lambda p: p.state not in ['sold', 'cancelled'])
        if self - properties_to_cancel:
            raise UserError(...)
        else:
            properties_to_cancel.state = 'cancelled'
        return True
```

Advantages are: 

estate/models/estate_property.py
aboo-odoo, line 83: Here you have quite small compute methods. In the future, you might have huge once, at that point, it's important to have meaningful parameter naming.
 ```suggestion
         for estate in self:  # QUESTION is the for each loop here necessary ? or can I just use self ?
```

**Comment by macai-odoo** [comment]
Hello ! 
I think I'm ready for another code review :) 
I might fix the style a bit later if the runbot raise a warning because he takes some time :)

File: estate/models/estate_property.py [hunk 30]
```diff
@@ -0,0 +1,149 @@
...
+        return True
+
+    def action_sell_property_state(self):
+        for record in self:
+            if record.state == 'sold':
+                raise UserError('Sold Properties cannot be sold again !')
+            elif record.state == 'cancelled':
+                raise UserError('Cancelled properties cannot be sold !')
+            record.state = 'sold'
+            return True
```
aboo-odoo, line 121:
```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+
+#   Actions :
+    def action_accept_selling_offer(self, buyer, price):
+        for record in self:
+            if record.buyer_id:
+                raise UserError('An offer is already accepted for this house')
+            record.buyer_id = buyer
+            record.selling_price = price
+            record.state = 'offer_accepted'
+            return True
```
aboo-odoo, line 132:
```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+        return True
+
+    def action_set_offer_received(self):
+        for record in self:
+            if record.state == 'new':
+                record.state = 'offer_received'
+                return True
```
aboo-odoo, line 139:
```suggestion
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,149 @@
...
+        return True
+
+#   CRUD methods:
+    @api.ondelete(at_uninstall=False)
+    def _delete_property(self):
+        for record in self:
+            if not record.state in ('new', 'cancelled'):
+                raise UserError('Property that are not new or cancelled cannot be deleted !')
+        #return super().unlink()
```
aboo-odoo: why was it commented ?
```suggestion
        return super().unlink()
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,61 @@
...
+        for record in self:
+            if record.status == 'accepted':
+                raise UserError('Accepted offer cannot be refused !')
+            record.status = 'refused'
+            return True
+        return True
+
+#   CRUD
+    @api.model
+    def create(self, vals):
```
aboo-odoo, line 59: Check out
```suggestion
    @api.model_create_multi
    def create(self, vals_list):
```

instead, to make usage of recordset creation and optimize your code 😄

estate/models/res_user.py
```diff
@@ -0,0 +1,10 @@
+from odoo import models, fields
+
+
+class ResUser(models.Model):
+    _inherit = 'res.users'
+
+    property_ids = fields.One2many('estate.property',
+                                   'salesman_id',
+                                   'Salesperson',
+                                   domain = lambda self: [('state', 'in', ('new', 'offer_received'))])
```
aboo-odoo:
```suggestion
    property_ids = fields.One2many(
        'estate.property',
        'salesman_id',
        'Salesperson',
        domain = lambda self: [
            ('state', 'in', ('new', 'offer_received')),
        ],
    )
```

estate/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    'data': [
+        'security/ir.model.access.csv',
+        'views/estate_property_views.xml',
+        'views/estate_property_offer_views.xml',
+        'views/estate_property_type_views.xml',
+        'views/estate_property_tag_views.xml',
+        'views/res_users_views.xml',
+        'views/estate_menu.xml',
+    ],
+    'author': 'macai',
```
aboo-odoo: You don't own what you code here ahah; it's all for Odoo S.A. (which is the default value)
```suggestion
```

estate/__manifest__.py
aboo-odoo, line 1: It's missing a description 😄 Don't you have an error message when you install your module ?

aboo-odoo (commented):
Overall, excellent job! 🥳 

Good naming of your PR! 

Could you squash all your commits together into a single one ? This is the default behavior here at Odoo: one task -> one PR -> one commit 👍 

estate/models/estate_property.py
macai-odoo: I commented it (And then deleted it) because calling super().unlink() start an infinite recursion as 
calling unlink would calling the _ondelete, ect...
I thought that "ondelete" executes before the deleting and should mainly raise Errors and no returning anything. 
Am I wrong ?

Thanks in advance :)

macai-odoo:
Hey @aboo-odoo !
A big thank you for the reviews !

I've squashed all the commits above, and (tried to) implemented the fixes/rework as you suggested.
I hope I did it the right way :).

I only have one question regarding this part : 
https://github.com/odoo/tutorials/pull/1412#discussion_r4060284046

About the return with a function with the api.ondelete decorator
Should the function return something ? Or call super().unlink() ?
If I call the super().unlink(), I think that occur an infinite recursion loop, but maybe I did something wrong ? Or understand something the wrong way. That's why I didn't call it :)

Thank you in advance :)

aboo-odoo:
> About the return with a function with the api.ondelete decorator Should the function return something ? Or call super().unlink() ? If I call the super().unlink(), I think that occur an infinite recursion loop, but maybe I did something wrong ? Or understand something the wrong way. That's why I didn't call it :)

It's normal, the `api.ondelete` is a check to see if you can delete the record. If you call unlink in it, you call the check again, and so on. :smile: 

Don't hesitate to fix the other comments I put down :smile:

macai-odoo:
Hello @aboo-odoo, , @lost-odoo 
I've written the small test case with the 2 offers.
Can you please check it ? See if everything is ok ? 
I'm also not sure about the commit name, as is only add a test.

Thanks you in advance !

macai-odoo:
Sorry I had to "clean-up" all the merge message, that I created trying to clean up the styling and commit name.
If it is something that shouldn't be done, sorry :'(

estate/models/estate_property_offer.py
macai-odoo: Good to know thank you :)

estate/__manifest__.py
macai-odoo: I added the description, I didn't notice any warnings tho ?

PR #1413, opened 2026-09-15
Technical Training - alnav

Kermut572:
Creation of the required manifest and init files for the real estate module.


estate/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    'category': 'Productivity',
+    'description': """
+    Our brand new real estate app!
+    """,
+    'application': True,
+    # data files always loaded at installation
+    'data': [],
+    # data files containing optionally loaded demonstration data
+    'demo': [],
+}
```
msho-odoo: Please add a new line at the end of files :)
Also always refer to your [runbot](https://runbot.odoo.com/runbot/batch/2765462) to address the styling errors ans the warnings

estate/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    'author': "alnav",
+    'category': 'Productivity',
+    'description': """
+    Our brand new real estate app!
+    """,
+    'application': True,
+    # data files always loaded at installation
+    'data': [],
+    # data files containing optionally loaded demonstration data
+    'demo': [],
```
msho-odoo: I guess no need to add them if they are not used (unless it gave you warnings on the log?)

estate/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+{
+    'name': "Real Estate",
+    'version': '1.0',
+    'depends': ['base'],
+    'author': "alnav",
```
msho-odoo: usually author is `Odoo S.A.` :)

msho-odoo (changes requested):
Great work!

Left you a couple of comments.
And for the commit message, it usually should align with `if this commit is merged it will ....` so you can write something like `[ADD] estate: add estate property module` or whatever message you think is better :)
refer to the [Git Guidelines](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html)

estate/models/estate_property.py
```diff
@@ -0,0 +1,19 @@
...
+    name = fields.Char("Property Name", required=True)
+    descritpion = fields.Text("Property Description")
+    postcode = fields.Char("Postcode")
+    date_availability = fields.Date("Available Date")
+    expected_price = fields.Float("Expected Price")
+    selling_price = fields.Float("Selling Price", required=True)
+    bedrooms = fields.Integer("Number of Bedrooms")
+    living_area = fields.Integer("Living Area")
+    facades = fields.Integer("Number of Facades")
+    garage = fields.Boolean("Has a Garage")
```
msho-odoo: good practice for boolean fields to be `is_something` or `has_something` so when someone reads the name in the middle of the code know it's a boolean, so you might make it `has_garden` and `has_garage`

estate/models/estate_property.py
```diff
@@ -0,0 +1,19 @@
+from odoo import fields, models
+
+class Property(models.Model):
```
msho-odoo, line 9: before `class ...` declaration we expect 2 empty lines

estate/models/estate_property.py
```diff
@@ -0,0 +1,19 @@
...
+    date_availability = fields.Date("Available Date")
+    expected_price = fields.Float("Expected Price")
+    selling_price = fields.Float("Selling Price", required=True)
+    bedrooms = fields.Integer("Number of Bedrooms")
+    living_area = fields.Integer("Living Area")
+    facades = fields.Integer("Number of Facades")
+    garage = fields.Boolean("Has a Garage")
+    garden = fields.Boolean("Has a Garden")
+    garden_area = fields.Integer("Garden Area")
+    garden_orientation = fields.Selection(string="Garden Orientation", selection=[('north', "North"), ("south", "South"), ("east", "East"), ("west", "West")])
```
```suggestion
    garden_orientation = fields.Selection(
     string="Garden Orientation",
        selection=[
            ("north", "North"),
            ("south", "South"),
            ("east", "East"),
            ("west", "West"),
        ]
    )
```

msho-odoo (changes requested):
Nice work, left a couple of tiny comments

When you try to fix some code after reviewing, there is no need to create a new commit just for the fix, it makes the history unnecessarily crowded, try using `git commit --amend` instead of `git commit`, this command lets you edit the last commit and also pushes your current changes to it without making a new one :)

estate/models/estate_property.py
msho-odoo: this is a better and conventional styling at odoo for bracketed objects if you would say :) 
nitpick regarding the quotes: some teams might stick to single quotes being only on technical strings (the ones that the user don't see) and double quotes are for the strings that the user can see)
so for example the selection can be  ('east', "East") instead of ('east', 'East') but some teams don't do this, they just stick to their own convention which is all single quotes (if possible) or all double quotes so you can pick a convention and stick to it for now :)

msho-odoo:
@Kermut572  just forgot to tell you, you can mention me here when you think the PR is ready for another review, typically, after each chapter :)

Kermut572:
> @Kermut572 just forgot to tell you, you can mention me here when you think the PR is ready for another review, typically, after each chapter :)

Alright I'll keep that in mind, thanks!

Kermut572:
@msho-odoo just finished chapter 5, PR is ready for review :)

estate/models/estate_property.py
```diff
@@ -0,0 +1,46 @@
...
+    garden_area = fields.Integer("Garden Area")
+    garden_orientation = fields.Selection(
+        string="Garden Orientation",
+            selection=[
+                ("north", "North"),
+                ("south", "South"),
+                ("east", "East"),
+                ("west", "West"),
+            ],
+    )
```
msho-odoo, line 39:
```suggestion
    garden_orientation = fields.Selection(
        string="Garden Orientation",
        selection=[
            ("north", "North"),
            ("south", "South"),
            ("east", "East"),
            ("west", "West"),
        ],
    )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,46 @@
...
+    state = fields.Selection(
+        string="Property State",
+            selection=[
+                ("new", "New"),
+                ("offer_received", "Offer Received"),
+                ("offer_accepted", "Offer Accepted"),
+                ("sold", "Sold"),
+                ("cancelled", "Cancelled"),
+            ],
+        default="new",
```
msho-odoo, line 50:
```suggestion
    state = fields.Selection(
        string="Property State",
        selection=[
            ("new", "New"),
            ("offer_received", "Offer Received"),
            ("offer_accepted", "Offer Accepted"),
            ("sold", "Sold"),
            ("cancelled", "Cancelled"),
        ],
        default="new",
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,46 @@
...
+
+class Property(models.Model):
+    _name = "estate.property"
+    _description = "Real Estate Property"
+
+    name = fields.Char("Property Name", required=True)
+    descritpion = fields.Text("Property Description")
+    postcode = fields.Char("Postcode")
+
+    date_availability = fields.Date("Available Date", copy=False, default=fields.Date.add(fields.Date.today(), months=3))
```
msho-odoo: Using `default=fields.Date.add(fields.Date.today(), months=3)` makes this only get valued first time it's imported and it's stuck there, you need to use `lambda` function instead, you will find multiple examples in the codebase, it's good practice to search for previous code there as well :)

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<odoo>
+    <menuitem id="estate_menu_root" name="Real Estate">
+        <menuitem id="estate_menu_first_level" name="Advertisements">
```
msho-odoo: It's better to give your ids a more descriptive name so that it's are understood when it's read somewhere in the code without needing to go back to its definition

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "depends": ["base"],
+    "author": "Odoo S.A.",
+    "category": "Productivity",
+    "description": """
+    Our brand new real estate app!
+    """,
+    "application": True,
+    "data": [
+        "security/ir.model.access.csv",
+
```
msho-odoo: why the extra line :)

msho-odoo (changes requested):
Nice work 💯 
left you some comments

nitpick about the commit message, I would make it `add` instead of `added` just to align with guidelines because you would say `This commit will add property model actions...` not `This commit will added property model actions` but it's not a big thing, well done.
Also have a look at the runbot, see why `ci/tutorials` is giving you a warning ant try to solve it.
Keep up the good work 🔥

estate/__manifest__.py
Kermut572: In the account_accountant enterprise module they have something like this
`'data/account_accountant_tour.xml',
  'data/ir_config_parameters.xml',

  'security/account_accountant_security.xml',

  'views/account_account_views.xml',
  'views/account_fiscal_year_view.xml',
  ...`
  
  I thought this was the standard way to format it, but looking at the other modules it seems to be more of an exception than a rule so I'll change it :D

msho-odoo: okaay, maybe some teams stick to that if you saw it somewhere :)
but I think mostly teams don't :)

Kermut572:
@msho-odoo chapter 6 ready for review :)

estate/models/estate_property.py
```diff
@@ -0,0 +1,50 @@
+from datetime import datetime
+
```
msho-odoo: No need to leave empty lines after each import line.
If you must, you can leave a line between the external import lines and the odoo import line, but that's just a nitpick from me

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<odoo>
+    <menuitem id="estate_main_menu_button" name="Real_Estate">
+        <menuitem id="estate_app_top_bar_property_model_button" name="Advertisements">
```
msho-odoo: I wouldn't add `_button` at the end of the id of the menuitem, it could get misunderstood with a real action button, I would stick to the naming guidelines [here](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html?highlight=guidelines#xml-ids-and-naming)

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,96 @@
...
+            <search string="Properties">
+                <field name="name"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="facades"/>
+                <separator/>
+                <filter string="Available" name="active" domain="[('active', '=', True)]"/>
+                <filter string="New" name="active" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"/>
```
msho-odoo: Usually, we write the `name` attribute before the `string` :)
Your domain is correct, but I think this is more readable:

```suggestion
                <filter name="active" string="New" domain="[('state', 'in', ('new', 'offer_received'))]"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,96 @@
...
+                <group>
+                    <filter string="Postcode" name="postcode" context="{'group_by':'postcode', 'residual_visible':True}"/>
+                </group>
+            </search>
+        </field>
+    </record>
+
+    <record id="estate_property_model_list_view" model="ir.ui.view">
```
msho-odoo:
```suggestion
    <record id="estate_property_view_list" model="ir.ui.view">
```
just to align with the [naming guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html?highlight=guidelines#xml-ids-and-naming) 
You can have a quick check for the id names in the files in general

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,96 @@
...
+        </field>
+    </record>
+
+    <record id="estate_property_model_form_view" model="ir.ui.view">
+        <field name="name">Properties</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form string="Property">
+                <sheet>
+                    <field name="name" style="font-size:20pt;"/>
```
msho-odoo: in-line styling is not encouraged, you can use the tag `<h1> </h1>` instead for this case or the ready to use classes (you can check some views in the codebase if you like)

msho-odoo (changes requested):
Nice work 👍 
Just for the commit message we use `[ADD]` tag when adding a brand new module in the commit but adding some fields, views or logic/feature we use `[IMP]` which you will be using most of the time as we don't add a new module everyday at odoo :) but we do a lot of improvements
I also left you some comments but they are not big, just some nitpicks.
Good job 🔥

Kermut572:
@msho-odoo chap. 7 ready ! I also implemented the changes you suggested, I tried my best to follow naming conventions for the xml ids but I am not too sure about the ones in estate_menus, could you tell me whether they are better now or if I should change them? Thanks!

estate/security/ir.model.access.csv
```diff
@@ -1,2 +1,5 @@
 id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
 access_estate_property_model,access_estate_property_model,model_estate_property,base.group_user,1,1,1,1
+access_estate_property_type_model,access_estate_property_type_model,model_estate_property_type,base.group_user,1,1,1,1
```
msho-odoo: nitpick: I would rename the id to be `access_<mode_name>_<user_type>` so it will be `access_estate_property_type_user` because you might add a rule of the same model but for manager foe example, then it will be `access_estate_property_type_manager`.

estate/views/estate_menus.xml
```diff
@@ -1,8 +1,12 @@
...
-    <menuitem id="estate_main_menu_button" name="Real_Estate">
-        <menuitem id="estate_app_top_bar_property_model_button" name="Advertisements">
-            <menuitem id="estate_app_property_model_action" action="estate_property_model_action"/>
+    <menuitem id="estate_main_menu" name="Real_Estate">
+        <menuitem id="estate_property_action_list" name="Advertisements">
+            <menuitem id="estate_property_trigger_action" action="estate_property_action"/>
+        </menuitem>
+        <menuitem id="estate_settings_action_list" name="Settings">
+            <menuitem id="estate_app_property_type_trigger_action" action="estate_property_type_action"/>
+            <menuitem id="estate_app_property_tag_trigger_action" action="estate_property_tag_action"/>
```
msho-odoo, line 9: oh don't make two menuitems have the same id

estate/views/estate_menus.xml
```diff
@@ -1,8 +1,12 @@
 <?xml version="1.0"?>
 <odoo>
-    <menuitem id="estate_main_menu_button" name="Real_Estate">
-        <menuitem id="estate_app_top_bar_property_model_button" name="Advertisements">
-            <menuitem id="estate_app_property_model_action" action="estate_property_model_action"/>
+    <menuitem id="estate_main_menu" name="Real_Estate">
+        <menuitem id="estate_property_action_list" name="Advertisements">
+            <menuitem id="estate_property_trigger_action" action="estate_property_action"/>
+        </menuitem>
+        <menuitem id="estate_settings_action_list" name="Settings">
```
msho-odoo: if you refer to this section in coding guidelines: [XML IDs and naming](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html?highlight=coding%20guidelines#xml-ids-and-naming)
it says: "For a menu: `<model_name>_menu`, or `<model_name>_menu_do_stuff` for submenus"
So you might call this one for example `estate_property_menu_open_settings`

estate/views/estate_property_views.xml
```diff
@@ -38,22 +38,29 @@
...
+    <record id="estate_property_view_form" model="ir.ui.view">
         <field name="name">Properties</field>
         <field name="model">estate.property</field>
         <field name="arch" type="xml">
             <form string="Property">
                 <sheet>
-                    <field name="name" style="font-size:20pt;"/>
+                    <h1><field name="name"/></h1>
+                    <group>
+                        <field name="tags_ids" nolabel="1" widget="many2many_tags" style="font-siye:14pt;"/>
```
msho-odoo: in-line styling is not encouraged, you can use ready-to-use classes or some tags for different font sizes for example, have  a look at the codebase for some many2many fields, you will probably find an example

msho-odoo (commented):
Nice work 🚀 
left you small comments.
Just don't forget to add the modules you edit in the commit message, so it's `[IMP] estate: add property tags, types and offers`

Kermut572:
hey @msho-odoo, chapters 8 and 9 ready for review :)

estate/models/estate_property.py
```diff
@@ -56,3 +61,53 @@ class Property(models.Model):
...
+    _check_expected_price = models.Constraint("CHECK(expected_price > 0)", "Expected price must be greater than zero")
+    _check_selling_price = models.Constraint("CHECK(selling_price >= 0)", "Selling price must be greater or equal to zero")
+
+    @api.constrains("expected_price", "selling_price")
+    def _check_selling_price_proportion(self):
+        for record in self:
+            if record.selling_price == 0:
+                continue  # 0 means no offer accepted
+
+            if float_compare(record.selling_price, 0.9 * record.expected_price, 2) < 0:
```
msho-odoo, line 75: Totally correct, you can also make one if statement saying if selling price is not zero **and** the selling price is less than 90% of the expected price, then raise the error

estate/models/estate_property.py
```diff
@@ -56,3 +61,53 @@ class Property(models.Model):
...
+                raise ValidationError(message="Selling price must be at least 90 percent of expected price")
```
msho-odoo: The `_` I added is imported `from odoo` and is used to support translation for strings shown to the user like this error message, apply it in all error messages
```suggestion
               raise ValidationError(_("Selling price must be at least 90 percent of expected price"))
```

estate/models/estate_property.py
```diff
@@ -56,3 +61,53 @@ class Property(models.Model):
...
+
+    def action_mark_as_sold(self):
+        for record in self:
+            if record.state == "cancelled":
+                raise UserError(message="Can't sell a cancelled auction.")
+
+            record.state = "sold"
```
msho-odoo: Correct, but you can do the check with all the records (self) one time, to check if there's any record has that state, you raise the error, instead of looping over them and checking and assigning one by one

estate/models/estate_property.py
```diff
@@ -56,3 +61,53 @@ class Property(models.Model):
...
+    def action_cancel_selling(self):
+        for record in self:
+            if record.state == "sold":
+                raise UserError(message="Can't cancel a sold auction.")
+
+            record.state = "cancelled"
+
+        return True
+
+    @api.depends("offer_ids")
```
```suggestion
    @api.depends("offer_ids.price")
```

estate/models/estate_property.py
```diff
@@ -56,3 +61,53 @@ class Property(models.Model):
...
+
+    @api.onchange("has_garden")
+    def _onchange_has_garden(self):
+        if self.has_garden:
+            self.garden_area = 10
+            self.garden_orientation = "north"
+            return
+
+        self.garden_area = 0
+        self.garden_orientation = ""
```
```suggestion
        self.garden_orientation = False
```

estate/models/estate_property_offer.py
```diff
@@ -15,5 +18,51 @@ class PropertyOffer(models.Model):
...
+
+    def action_accept_offer(self):
+        for record in self:
+            linked_property = record.property_id
+            states = linked_property.offer_ids.mapped("status")
+
+            if "accepted" in states:
+                raise UserError(message="An offer was already accepted")
+
+            if linked_property.state == "cancelled":
```
msho-odoo: why not? 🤔 
```suggestion
            if "cancelled" in states:
```

estate/models/estate_property_offer.py
```diff
@@ -15,5 +18,51 @@ class PropertyOffer(models.Model):
...
+
+        return True
+
+    def action_reject_offer(self):
+        for record in self:
+            record.status = "refused"
+
+        return True
+
+    @api.depends("validity")
```
msho-odoo: it also depends on the `create_date`

estate/views/estate_menus.xml
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0"?>
 <odoo>
     <menuitem id="estate_main_menu" name="Real_Estate">
-        <menuitem id="estate_property_action_list" name="Advertisements">
+        <menuitem id="estate_menu_open_advertisements" name="Advertisements">
```
msho-odoo: I would refer to the [coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#xml-ids-and-naming) for the naming of menus

msho-odoo (changes requested):
Nice work!
Left you a couple of comments, thanks!

Kermut572:
@msho-odoo hey! hope you had a nice weekend :) I should be done with the tutorial, I'll still keep an eye on this PR though so I can correct any potential  mistake you might point out!

estate/models/estate_property.py
```diff
@@ -0,0 +1,122 @@
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models, _
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare
+
+
+class Property(models.Model):
+    _name = "estate.property"
```
msho-odoo: Always match the name of the model to the name of the class
```suggestion
class EstateProperty(models.Model):
    _name = "estate.property"
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,122 @@
...
+    _description = "Real Estate Property"
+    _order = "id desc"
+
+    name = fields.Char("Property Name", required=True)
+    description = fields.Text("Property Description")
+    postcode = fields.Char("Postcode")
+
+    date_availability = fields.Date("Available Date", copy=False, default=lambda _x: datetime.now() + relativedelta(months=+3))
```
msho-odoo: nitpick
```suggestion
    date_availability = fields.Date("Available Date", copy=False, default=lambda _x: datetime.now() + relativedelta(months=3))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,122 @@
...
+            if float_compare(record.selling_price, 0.9 * record.expected_price, 2) < 0:
+                raise ValidationError(_("Selling price must be at least 90 percent of expected price"))
+
+    def action_mark_as_sold(self):
+
+        if "cancelled" in self.mapped("state"):
+            raise UserError(_("Can't sell a cancelled auction."))
+
+        for record in self:
+            record.state = "sold"
```
msho-odoo: Better to do this right away
```suggestion
        self.state = "sold"
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,122 @@
...
+
+        return True
+
+    def action_cancel_selling(self):
+
+        if "sold" in self.mapped("state"):
+            raise UserError(_("Can't cancel a sold auction."))
+
+        for record in self:
+            record.state = "cancelled"
```
```suggestion
        self.state = "cancelled"
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,83 @@
...
+        selection=[
+            ("accepted", "Accepted"),
+            ("refused", "Refused"),
+        ],
+    )
+
+    validity = fields.Integer("Validity (in days)", default=7)
+
+    date_deadline = fields.Date("Deadline", compute="_compute_date_deadline", inverse="_inverse_date_deadline")
+    create_date = fields.Datetime()
```
msho-odoo: `create_date` is one of the columns which is auto created by the orm during creation of the object, no need to declare it here, you can use right away

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,83 @@
+from datetime import datetime, timedelta
+
+from odoo import api, fields, models, _
+from odoo.exceptions import UserError
+
+
+class PropertyOffer(models.Model):
+    _name = "estate.property.offer"
```
msho-odoo: same here with the names

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,83 @@
...
+            record.status = "accepted"
+            linked_property.selling_price = record.price
+            linked_property.buyer_id = record.partner_id
+            linked_property.state = "offer_accepted"
+
+        return True
+
+    def action_reject_offer(self):
+        for record in self:
+            record.status = "refused"
```
msho-odoo: same, you can use `self.` right away

estate/models/estate_property.py
```diff
@@ -0,0 +1,122 @@
...
+
+        return True
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
```
msho-odoo, line 70: Have a look at the[ coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions) (see the order part), all the compute, onchange and private functions comes before the public functions

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,83 @@
...
+        for record in self:
+            offer_date = record.create_date
+            if not offer_date:
+                offer_date = datetime.now()
+            record.validity = (record.date_deadline - offer_date.date()).days
+
+    @api.model
+    def create(self, vals):
+        for offer in vals:
+            linked_property = self.env["estate.property"].browse(offer["property_id"])
```
msho-odoo, line 70: correct, but if you can create offers without property_id specified, you might need a check first that "property_id" is in vals

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,83 @@
...
+            if offer["price"] < linked_property.best_price:
```
msho-odoo: Better use `float_compare`, specially that it allows you to specify precision

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,31 @@
...
+
+    offer_count = fields.Integer(string="Number of Offers", compute="_compute_offer_count")
+
+    property_ids = fields.One2many("estate.property", "property_type_id", string="Properties")
+    offer_ids = fields.One2many("estate.property.offer", "property_type_id", string="Offers")
+
+    @api.depends("offer_ids")
+    def _compute_offer_count(self):
+        for record in self:
+            record.offer_count += len(record.offer_ids)
```
msho-odoo: `+=` means every time you change the offers, it **adds** its number
```suggestion
            record.offer_count = len(record.offer_ids)
```

estate/models/res_users.py
```diff
@@ -0,0 +1,7 @@
+from odoo import fields, models
+
+
+class Users(models.Model):
+    _inherit = "res.users"
+
+    property_ids = fields.One2many("estate.property", "seller_id", string="Listed Properties", domain=[("active", "=", True)])
```
msho-odoo: hmmm, I think you don't need to explicitly add that domain

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="estate_property_type_view_form" model="ir.ui.view">
+        <field name="name">Property Types</field>
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <form string="Properties">
+                <header>
+                    <!-- This looks a bit cursed, maybe there's a way to make it cleaner? -->
+                    <button name="action_open_related_offers" type="object" icon="fa-money" class="oe_stat_button oe_right" invisible="offer_count == 0"> Offers: <field name="offer_count"/></button>
```
msho-odoo, line 10: haha, I think it's good, there also another way to do this without adding an action to the python file since it uses the list view anyway.
You can use 
```xml
<button name="%(estate.estate_property_offer_action)d" ..... >
</button>
```
See [example](https://github.com/odoo/odoo/blob/bf6fea1fb37dc07ee7adcdf22a7e31f020e8f63e/addons/crm/views/crm_lead_views.xml#L16) in crm

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,160 @@
...
+            <search string="Properties">
+                <field name="name"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area" filter_domain="[('living_area', '>=', self)]"/>
+                <field name="facades"/>
+                <separator/>
+                <filter name="active" string="Available" domain="[('active', '=', True)]"/>
+                <filter name="new" string="New" domain="[('state', 'in', ('new','offer_received'))]"/>
```
msho-odoo, line 17: you mean `name="available" string="Available"` ? there should be a filter with name ~~"name"~~ "new"?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,160 @@
...
+                <group>
+                    <filter name="postcode" string="Postcode" context="{'group_by':'postcode', 'residual_visible':True}"/>
```
msho-odoo: what's the `residual_visible` used for?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,160 @@
...
+                    <t t-name="card">
+                        <div>
+                            <h2>
+                                <field name="name"/>
+                            </h2>
+                            <div>Expected Price: <field name="expected_price"/></div>
+                            <div t-if="record.state.raw_value == 'offer_received'">Best price: <field name="best_price"/></div>
+                            <div t-elif="record.state.raw_value == 'offer_accepted'">Selling price: <field name="selling_price"/></div>
+                            <div t-elif="record.state.raw_value == 'sold'">Sold at: <field name="selling_price"/></div>  
+                            <div t-else=""></div>
```
msho-odoo: nitpick if you like: you can remove the `t-else` and the `t-elif` and only use `t-if` in all of them since all of them are on the same record and if one is true, you know the rest is false

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,160 @@
...
+                </header>
+                <sheet> 
+                    <h1><field name="name"/></h1>
+                    <group>
+                        <field name="tags_ids" nolabel="1" widget="many2many_tags" options="{'color_field': 'color'}"/>
+                    </group>
+                    
+                    <group>
+                        <group>
+                            <field name="property_type_id" options="{'no_create': true}"/>
```
msho-odoo: We also shouldn't be able to edit it

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,160 @@
...
+
+    <record id="estate_property_view_form" model="ir.ui.view">
+        <field name="name">Properties</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form string="Property">
+                <header>
+                    <button name="action_mark_as_sold" type="object" string="Sold" invisible="state in ('sold', 'cancelled')"/>
+                    <button name="action_cancel_selling" type="object" string="Cancel" invisible="state in ('sold', 'cancelled')"/>
+                    <field name="state" widget="statusbar"/>
```
msho-odoo: I think the statusbar should be visible in some states only, lookup `statusbar_visible`

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,160 @@
...
+                            </group>
+                            <group>
+                                <field name="garden_area" invisible="not has_garden"/>
+                            </group>
+                            <group>
+                                <field name="garden_orientation" invisible="not has_garden"/>
+                            </group>
+                            <group>
+                                <field name="total_area"/>
+                            </group>
```
msho-odoo, line 108: No need to add the `<group>` tag over each field, the tag is used to group some fields in the view together when it's more convenient, you can look it up in some form views in the codebase of odoo and see the UI, to notice its effect.

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,160 @@
...
+        <field name="arch" type="xml">
+            <search string="Properties">
+                <field name="name"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area" filter_domain="[('living_area', '>=', self)]"/>
+                <field name="facades"/>
+                <separator/>
+                <filter name="active" string="Available" domain="[('active', '=', True)]"/>
```
msho-odoo, line 16: you have the `search_default_active` on the main action, do you need this here?

estate/views/res_users_views.xml
```diff
@@ -0,0 +1,28 @@
...
+                        <field name="property_ids">
+                            <list>
+                                <field name="name"/>
+                                <field name="property_type_id"/>
+                                <field name="postcode"/>
+                                <field name="tags_ids"/>
+                                <field name="bedrooms"/>
+                                <field name="living_area"/>
+                                <field name="expected_price"/>
+                                <field name="selling_price"/>
```
msho-odoo: Do you explicitly need to add the list of fields here?

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class EstateProperty(models.Model):
+    _inherit = "estate.property"
+
+    def action_mark_as_sold(self):
+        for record in self:
+            self.env['account.move'].create(
+                {
+                    "partner_id": record.buyer_id.id,
```
msho-odoo, line 12: you are missing a `journal_id` field on the account_move.
you should get the account_journal with type 'sale' and assign it the journal_id on the account move

estate_account/__manifest__.py
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "Real Estate Accounting",
+    "version": "1.0",
+    "depends": ["base", "estate", "account"],
```
msho-odoo, line 4: you can remove "base", it's implicitly implied from "estate" and  "account" since they depend on it

estate_account/__manifest__.py
```diff
@@ -0,0 +1,11 @@
...
+    "author": "Odoo S.A.",
+    "category": "Productivity",
+    "description": """
+    Accounting module for Real Estate!
+    """,
+    "license": "OPL-1",
+}
```
msho-odoo, line 11: hmm you are missing a field that allows you to have this module installed automatically once the modules it depends on are installed

msho-odoo (changes requested):
Thank you for your work!

Regarding some commit messages, they shouldn't exceed 50 characters, refer to [the git guidelines](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html#commit-message-structure), and feel free to add a commit description after that in case you need more illustration on your changes.

I also left you some comments, good work.
Request a review from me when you're ready again, thanks!

estate/models/estate_property.py
Kermut572: I guess this is also a matter of team? There are a few occurrences in hr/accounting where they also use '+', but again it seems like a minority so I'll just remove it here :)

estate/models/estate_property.py
Kermut572: I'll admit I am a bit confused. In the tutorial, for the actions, they say

"Also note that we loop on self. Always assume that a method can be called on multiple records; it’s better for reusability."

Should we assume that for these actions they'll never be used on multiple records so it's fine not to loop on self?

Kermut572: The coding guidelines also say this:

> Action method : an object action method is prefix with action_. Since it uses only one record, add self.ensure_one() at the beginning of the method.

It answers my question then, although there seems to be a contradictory thing going on here :D

estate/models/estate_property_offer.py
Kermut572: Alright! In this case since property_id is a required field for offer it should be good though right?

estate/views/estate_property_type_views.xml
Kermut572: Great I'll keep that in mind thanks! It certainly looks cleaner than having the action returning the view like I did :)

estate/views/estate_property_views.xml
Kermut572: woops probably something I forgot to remove when I used some of the odoo source code

estate/views/estate_property_views.xml
Kermut572: I think I do, because `search_default_active` is linked to the active filter isn't it? Or did I misunderstand the question/tutorial :)?

Kermut572:
@msho-odoo thanks for your detailed feedback! Fixes are available for review, I went through all of your comments and I also left some comments on them :)

estate/models/estate_property.py
msho-odoo: Yes, it's just a nitpick from me, I just wanted to comment on it in case you thought it's mandatory for the `relativedelta` method, but it's fine to leave it if you want

estate/models/estate_property.py
msho-odoo: @Kermut572 
Oh the `self.state = "sold"` works also for multiple records.
The tutorial is telling you "always assume multiple records" is correct. The problem is sometimes you can't work on the whole `self` at once.
For example, assume here your `state` field in this action will be decided by some other field let's assume it's `garden_oreintation`, in that case you can't simple use a condition on `self.garden_oreintation` since that field might be different per each record inside `self`, but in our case here in the file, you want to set `state` field for all records to the same value `sold`, so it's safe to directly set `self.state`.

Kermut572: @msho-odoo ohh okay I get it now thanks! I didn't understand it that way but it makes total sense, I'll keep that in mind next time I have to use this :saluting_face:

msho-odoo: Another case where you need the loop or (ensure_one() if it's correct) is when some relational fields on some models are expected to be of size 1 always, for example you'r expecting `offer.property_id` to always return one property_d since you can't have the same offer for multiple properties. In that case it's wrong to use `self.property_id` (where self refer to offer model) because that will return multiple records if self itself is multiple records, so you have to loop in that case or use `mapped` for example.
Hope it's clearer now. If not let me know, please :)

estate/models/estate_property_offer.py
msho-odoo: Yes :+1:

estate/models/estate_property.py
Kermut572: @msho-odoo okay so basically the "only" time where I would use self directly is if I wanted to unconditionally alter some common field of all the given records right? As soon as I need to check a specific value of the record then I'd have to loop or use mapped, correct :)?

msho-odoo: Yes, you can just simply think:
Is self here a recordset  or one record? (in case `self.ensure_one()` for example you're sure it's one record otherwise assume it's multiple records)
Does it make sense to access/assign `self.<field_name>` for all the records right away or it will break my logic and I have to do it record by record?

Kermut572: great I get it now, thanks a lot for the explanation!

estate/views/estate_property_views.xml
msho-odoo: Yes yes you're correct, sorry, I am not sure why I wrote that

msho-odoo:
Thank you very much for your work and questions/replies on the comments, well done.
It's time for some JS for the other tutorial now, at least the start of it if you haven't already started it :sweat_smile:

Kermut572:
> Thank you very much for your work and questions/replies on the comments, well done. It's time for some JS for the other tutorial now, at least the start of it if you haven't already started it 😅

Thanks for all of your feedback! I'm halfway through the first chapter of the js tutorial, takes me a bit more time since I'm more familiar with pyhton than js :face_with_peeking_eye:

PR #1414, opened 2026-09-15
ollet - Technical Training

ollet-odoo:
task-6573598


estate/data/estate_menus.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<odoo>
+    <menuitem id="estate_menu_root" name="Real Estate">
+        <menuitem id="estate_first_level_menu" name="Advertisements">
+            <menuitem id="estate_property_menu_action" action="estate_property_action"/>
+        </menuitem>
+    </menuitem>
+</odoo>
```
msho-odoo: Always add a new line at end of files :)

estate/data/estate_menus.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<odoo>
+    <menuitem id="estate_menu_root" name="Real Estate">
+        <menuitem id="estate_first_level_menu" name="Advertisements">
```
msho-odoo: it's better to give ids a descriptive name rather than first_level_menu :)

estate/data/estate_property_views.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="estate_property_action" model="ir.actions.act_window">
+        <field name="name">Properties</field>
+        <field name="res_model">estate.property</field>
+        <field name="view_mode">list,form</field>
+    </record>
+        <record id="estate_property_view_tree" model="ir.ui.view">
```
msho-odoo:
```suggestion
    <record id="estate_property_view_tree" model="ir.ui.view">
```
it's not gonna break the xml but it's better aligned to be debugged easier in the future

estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from .property import Property
```
msho-odoo: EOF new line in all files :)

estate/models/property.py
```diff
@@ -0,0 +1,27 @@
+from odoo import models, fields
+
+
+class Property(models.Model):
+    _name = "estate.property"
+    _description = "Real estate property"
+
+    name = fields.Char(required=True, string='Title')
```
msho-odoo: it's just a convention to have the string as the first attribute

estate/models/property.py
```diff
@@ -0,0 +1,27 @@
...
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(copy=False, default=lambda x: fields.Date.add(fields.Date.today(), months=3),
+                                    string="Available From")
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float(readonly=True, copy=False)
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer(string="Living Area (sqm)")
```
msho-odoo, line 22: Stick to using single quotes (like you did in `name` field) or double quotes as you did here.
some teams might stick to single quotes being only on technical strings (the ones that the user don't see) and double quotes are for the strings that the user can see)
so here they will use double quotes as that is the string shown to the user

estate/models/property.py
```diff
@@ -0,0 +1,27 @@
...
+    faces = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
```
msho-odoo: Good practice to have boolean fields on the form `is_something` or `has_something` so you might change this to `has_garden` and `has_garage`

estate/models/property.py
```diff
@@ -0,0 +1,27 @@
...
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        selection=(('north', 'North'), ('east', 'East'), ('south', 'South'), ('west', 'West')))
```
msho-odoo:
```suggestion
    garden_orientation = fields.Selection(
        selection=[
            ('north', 'North'),
            ('east', 'East'),
            ('south', 'South'),
            ('west', 'West'),
        ]
    )
```
We usually use the `[]` for the selection attribute and style it like this with multiple values.

if you want to apply the nitpick mentioned in the previous commit which is used by some teams, you will have ` ('north', "North")` instead of  `('north', 'North')`

estate/models/property.py
```diff
@@ -0,0 +1,27 @@
...
+    active = fields.Boolean(default=True)
+    state = fields.Selection(default='new', required=True, copy=False,
+                             selection=(('new', 'New'), ('offer_received', 'Offer Received'),
+                                        ('offer_accepted', 'Offer Accepted'), ('sold', 'Sold'),
+                                        ('cancelled', 'Cancelled')))
```
msho-odoo: Apply same styling here :)

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,0
```
msho-odoo: no need to add `estate.` here in the id, the file is already defined in that module, but you may want to add `_user` at the end of the id so that it's `access_estate_property_user`, it's just gives you the chance to make other rules for the same module with different users like `_manager` for example

estate/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+# noinspection statement-effect
+{
+    'name': 'estate',
+    'depends': ['base'],
+    'data': [
+        'security/ir.model.access.csv',
+        'data/estate_property_views.xml',
+        'data/estate_menus.xml'
+    ],
+    'author': "Olivier Le Thanh Duong",
```
msho-odoo: author is usually `Odoo S.A.`

msho-odoo (changes requested):
Good job 🔥 

Just for some commit messages, we write it in a way that aligns with `when this commit is merged it will .....` so for example `estate: Basic security` can be `add access rules to estate_property` or however you would like to describe your change :)

I also left you some comments :)
feel free to ping me when you need another review, also always make sure your runbot is green ahead please :)

ollet-odoo:
@msho-odoo Thanks for the review, i have corrected according to your remarks. I don't know if I should mark the conversation as resolved when I fix them or if you will do it when you do the review again

msho-odoo:
> @msho-odoo Thanks for the review, i have corrected according to your remarks. I don't know if I should mark the conversation as resolved when I fix them or if you will do it when you do the review again

Usually resolving the comment would depend on the team or the reviewer and what they prefer :)
but here if you're sure you've done the change, feel free to resolve it :+1:

estate/data/estate_menus.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<odoo>
+    <menuitem id="estate_menu_root" name="Real Estate">
+        <menuitem id="estate_Advertisements_menu" name="Advertisements">
```
msho-odoo:
```suggestion
        <menuitem id="estate_advertisements_menu" name="Advertisements">
```
I would stick to the snake case here :), let's make it all small letters

estate/data/estate_property_views.xml
```diff
@@ -0,0 +1,92 @@
...
+        <field name="arch" type="xml">
+            <form string="Property">
+                <sheet>
+                    <div class="oe_title">
+                        <h1>
+                            <field name="name"/>
+                        </h1>
+                    </div>
+                    <group>
+
```
msho-odoo: nitpick: when xml becomes bigger, it makes it easier to read if we get rid of these inside empty lines 😅
You can leave an empty line between records maybe but not preferably here :)

estate/models/property.py
```diff
@@ -0,0 +1,38 @@
...
+    active = fields.Boolean(default=True)
+    state = fields.Selection(default='new', required=True, copy=False,
+                             selection=[
+                                 ('new', "New"),
+                                 ('offer_received', "Offer Received"),
+                                 ('offer_accepted', "Offer Accepted"),
+                                 ('sold', "Sold"),
+                                 ('cancelled', "Cancelled")
+                             ]
+                            )
```
msho-odoo:
```suggestion
    state = fields.Selection(
         default='new',
         required=True,
         copy=False,
         selection=[
             ('new', "New"),
             ('offer_received', "Offer Received"),
             ('offer_accepted', "Offer Accepted"),
             ('sold', "Sold"),
             ('cancelled', "Cancelled")
         ]
     )
```

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property_user,access_estate_property,model_estate_property,base.group_user,1,1,1,0
```
msho-odoo: EOF line :)

msho-odoo: No need to add the `estate.` before the id, it's implicit since it's in the same module.
I believe you should give `unlink` access to the user, here you're setting the `perm_unlink` to 0

estate/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+# noinspection statement-effect
+{
+    'name': 'estate',
+    'depends': ['base'],
+    'data': [
+        'security/ir.model.access.csv',
+        'data/estate_property_views.xml',
```
msho-odoo: The views file and the menus file should be in `views` folder not `data` folder

msho-odoo (changes requested):
Good work 👍 
Just try to use `git commit --amend` instead of just `git commit` when fixing few stuff like spaces, names, quotes, etc. because the same commit still applies, no need to add the fixes to a new commit.
Also, left you a few changes.
Keep it up 🔥

estate/security/ir.model.access.csv
ollet-odoo: I was using the line suggested in the log as a base.
Is it a consistency rule to note have it in it's own module or just a hint you are giving me?

ollet-odoo:
robodoo rebase-merge

estate/security/ir.model.access.csv
msho-odoo: I am just giving you the reasoning why you don't need the `<module_name>.` before the id in this case but it's not a big thing

estate/models/property.py
```diff
@@ -0,0 +1,106 @@
...
+
+class Property(models.Model):
+    _name = 'estate.property'
+    _description = "Real estate property"
+
+    name = fields.Char(string="Title", required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(copy=False, default=lambda x: fields.Date.add(fields.Date.today(), months=3),
+                                    string="Available From")
```
msho-odoo: nitpick: we usually stick to making the `string` attribute the first attribute unless it's a relational field like the `Many2one` we put the comodel first and we usually indent new lines for the same block with only a tap, BUT that's not a big thing

estate/models/property.py
```diff
@@ -0,0 +1,106 @@
...
+        copy=False,
+        selection=[
+            ('new', "New"),
+            ('offer_received', "Offer Received"),
+            ('offer_accepted', "Offer Accepted"),
+            ('sold', "Sold"),
+            ('cancelled', "Cancelled")
+        ]
+    )
+    property_type_id = fields.Many2one("estate.property.type", string="Type")
```
msho-odoo, line 48: here you are using double quotes and in `property_offer.py` you are using single quotes, stick to a convention and follow it

estate/models/property.py
```diff
@@ -0,0 +1,106 @@
...
+    offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers")
+
+    total_area = fields.Float(compute='_compute_total_area')
+
+    @api.depends('garden_area', 'living_area')
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.garden_area + record.living_area
+
+    best_price = fields.Float(string="Best Offer", compute="_compute_best_price")
```
msho-odoo: we put all the fields definitions before the methods, see the [coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions) and check the order part

estate/models/property.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for record in self:
+            if record.offer_ids:
+                record.best_price = max(record.offer_ids.mapped("price"))
+            else:
+                record.best_price = None
```
msho-odoo: More pythonic way :)

```suggestion
            if record.offer_ids:
                record.best_price = max(record.offer_ids.mapped("price")) if record.offer_ids else 0.0
```

estate/models/property.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @api.onchange('has_garden')
+    def _onchange_has_garden(self):
+        if self.has_garden:
+            self.garden_area = 10
+            self.garden_orientation = 'north'
+        else:
+            self.garden_area = 0
+            self.garden_orientation = ''
```
msho-odoo: I think this would be `self.garden_orientation = False` since keys are already strings

estate/models/property.py
```diff
@@ -0,0 +1,106 @@
...
+            else:
+                record.state = 'sold'
+        return True
+
+    def action_do_cancel(self):
+        for record in self:
+            if record.state == 'sold':
+                raise UserError('Sold property cannot be canceled')
+            else:
+                record.state = 'cancelled'
```
msho-odoo: hmmm you may return True at the end of this action as well

estate/models/property.py
```diff
@@ -0,0 +1,106 @@
...
+        'CHECK(selling_price >= 0)',
+        'Expected price must be positive',
+    )
+
+    @api.constrains('expected_price', 'selling_price')
+    def _check_expected_price_selling_price(self):
+        for record in self:
+            if not float_is_zero(record.selling_price, precision_digits=2):
+                if float_compare(record.expected_price * 0.90, record.selling_price, precision_digits=2) == 1:
+                    raise ValidationError("Selling price must be above 90% of expected price")
```
msho-odoo: You can `and` the two if conditions instead of the two nested if statements.
Also when using `float_compare` we check against 0, so your check would be `float_compare(....) > 0`
Also all these private methods should not be after the public methods, check the coding guidelines as mentioned above and fix it :)

estate/models/property_offer.py
```diff
@@ -0,0 +1,51 @@
...
+        copy=False,
+        selection=[
+            ('accepted', "Accepted"),
+            ('refused', "Refused"),
+        ]
+    )
+    partner_id = fields.Many2one('res.partner', string="Partner", required=True)
+    property_id = fields.Many2one('estate.property', required=True)
+    validity = fields.Integer(default=7)
+    date_deadline = fields.Date(compute='_compute_deadline', inverse="_inverse_deadline")
```
msho-odoo: nitpick: when naming the compute function we just add `_compute_` as a prefix to the name of the field so it would be `_compute_date_deadline` and so is the inverse

estate/models/property_offer.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @api.depends("validity", "create_date")
```
msho-odoo, line 34: @ decorators like onchange and depends stock also to single quotes or double quotes :)

estate/models/property_offer.py
```diff
@@ -0,0 +1,51 @@
...
+    def _compute_deadline(self):
+        for record in self:
+            base_date = record.create_date or fields.Date.today()
+            record.date_deadline = fields.Date.add(base_date, days=record.validity)
+
+    def _inverse_deadline(self):
+        for record in self:
+            base_date = fields.Date.to_date(record.create_date) if record.create_date else fields.Date.today()
```
msho-odoo: hmm do you need the `Date.to_date()` here?

estate/models/property_offer.py
```diff
@@ -0,0 +1,51 @@
...
+            record.validity = (record.date_deadline - base_date).days
+
+    def action_accept(self):
+        for record in self:
+            if record.property_id.state in ('offer_accepted', 'sold', 'cancelled'):
+                err_msg = f"Cannot accept offer on property that is {record.property_id.state}"
+                raise exceptions.UserError(err_msg)
```
msho-odoo: It's better to import `UserError` from `odoo.exceptions` at the beginning of the file and use it, no need to import the whole `odoo.exceptions`

estate/models/property_offer.py
```diff
@@ -0,0 +1,51 @@
...
+            record.status = record.status = 'accepted'
+            record.property_id.buyer_id = self.partner_id
+            record.property_id.selling_price = record.price
+            record.property_id.state = 'offer_accepted'
```
msho-odoo: Here you are looping over the records in self which is correct but a lot of actions in odoo are made so that they can be applied on multiple records at the same time (for example, selecting multiple offers and rejecting them all at once), in that case your code will work fine but it's best to deal with recordset when you can better than a single record at a time (sometimes it's not possible in case you need to evaluate some value for each record before doing the action).
Maybe here you can check on all the states at once if they hold some value? or maybe after the checks in the loop you go outside the loop and write your values to `self` using `self.write` instead or `record`

estate/models/property_type.py
```diff
@@ -0,0 +1,12 @@
...
+
+class PropertyType(models.Model):
+    _name = 'estate.property.type'
+    _description = "Property Type"
+
+    name = fields.Char(string="Name", required=True)
+    _name_uniq = models.Constraint(
+        'UNIQUE (name)',
+        "The name of the type must be unique!",
+    )
```
msho-odoo: oh please ALWAYS make sure your runbot is green, EOF line here for example will make your runbot red and it will tell your the reason with the place where you should add the line :)

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,35 @@
...
+        <field name="model">estate.property.offer</field>
+        <field name="arch" type="xml">
+            <list>
+                <field name="price"/>
+                <field name="partner_id"/>
+                <field name="validity"/>
+                <field name="date_deadline"/>
+                <button name="action_accept" title="Accept" type="object" icon="fa-check"/>
+                <button name="action_refuse" title="Refuse" type="object"
+                        icon="oi-close"/>
```
msho-odoo: nitpick
```suggestion
                <button name="action_refuse" title="Refuse" type="object" icon="oi-close"/>
```

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,35 @@
...
+        <field name="name">estate.property.offer.list</field>
+        <field name="model">estate.property.offer</field>
+        <field name="arch" type="xml">
+            <list>
+                <field name="price"/>
+                <field name="partner_id"/>
+                <field name="validity"/>
+                <field name="date_deadline"/>
+                <button name="action_accept" title="Accept" type="object" icon="fa-check"/>
+                <button name="action_refuse" title="Refuse" type="object"
```
msho-odoo: Wouldn't you want these two buttons to disappear when you either accept or refuse an offer? how to do so? 🤔

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,43 @@
...
+            <list string="Properties">
+                <field name="name"/>
+            </list>
+        </field>
+    </record>
+    <record id="estate_property_tag_view_form" model="ir.ui.view">
+        <field name="name">estate.property.tag.form</field>
+        <field name="model">estate.property.tag</field>
+        <field name="arch" type="xml">
+            <form string="Property">
```
msho-odoo: you but a string='property' on every list and form views? intended? or copy-paste noise :) ?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,111 @@
...
+                <field name="expected_price"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area"/>
+                <field name="faces"/>
+                <separator/>
+                <filter name="Available properties" domain="[('state', 'in', ['offer_received', 'new'])]"/>
+                <group>
+                    <filter string="Postcode" name="postcode" context="{'group_by':'postcode'}"/>
+                    <filter string="Type" name="property_type" context="{'group_by':'property_type_id'}"/>
```
msho-odoo: nitpick, usually the name before the string but not a big thing

estate/models/__init__.py
```diff
@@ -0,0 +1,4 @@
...
+from .property_type import PropertyType
+from .property_tag import PropertyTag
+from .property_offer import PropertyOffer
```
msho-odoo:
```suggestion
from . import property
from . import property_type
from . import property_tag
from . import property_offer
```
import the file here not the model

msho-odoo (changes requested):
Thank you for the good work 🔥

Left you a couple of comments (some of them are nitpicks so, you can attend to the important one first :))

Also please ping me when you finish a chapter for a review so that changes are not too much to review and I know whether you missed something in the code or you didn't get to that chapter yet:)
but thanks, keep it up 👍

estate/models/property.py
ollet-odoo: Yeah i will switch to double quote everywhere, otherwhise it will drive me crazy :sweat_smile:

estate/models/property.py
ollet-odoo: Do we need the if block in that way?

estate/models/property_offer.py
ollet-odoo: If i remember correctly I was hitting an error without it as it is a date time and date_deadline is a date.

estate/views/estate_property_offer_views.xml
ollet-odoo: it is done later in the chapter ;)

estate/models/property_type.py
ollet-odoo: I need to figure a way to properly set my IDE for this. otherwise I seems to always remove them without realising

estate/views/estate_property_tag_views.xml
ollet-odoo: Copy-paste noise, I will remove them. I don't think they do much anyway

estate/models/property_offer.py
ollet-odoo: the self.write thing isn't covered in the tutorial it seems ;)
Will look into it

estate/models/property.py
msho-odoo: I guess a traceback will occur if you try to access `mapped('price`) on a non existing field

estate/models/property_offer.py
msho-odoo: okay :+1: 
you can also use `.date()` in both methods also if you like

estate/models/property_offer.py
msho-odoo: yeah have a look in the codebase, it's fine, you can write the field write away as you did for record but on self using `self.` but the write looks more readable when editing multiple fields at the same time.
Also, please don't resolve the comment if you reply to it, it's hard to see you replies to my comments when they are resolved :cry:

ollet-odoo: What things that's not clear to me, if outside the loop I do
```python
self.property_id.buyer_id = self.partner_id
```
will that works as expected and write different value properly or will it break or be undefined?

estate/models/property_type.py
ollet-odoo: I think it's configured properly now, at least for python files.

estate/models/property_offer.py
msho-odoo: Yes, I see what you mean, the self.partner_id recordset will be all assigned to the buyer_id of the property if it doesn't check for singleton recordset :+1: 
so your code is correct to loop on the records in this case, just make sure to assign `record.partner_id` not `self.partner_id`

ollet-odoo:
I have finished chapter 11 and the tests are passing if you could please review again.

estate/models/property_offer.py
ollet-odoo: Yes good catch I corrected that also

estate/models/property_offer.py
```diff
@@ -0,0 +1,74 @@
...
+            record.validity = (record.date_deadline - base_date).days
+
+    @api.model
+    def create(self, vals_list):
+        for vals in vals_list:
+            property_id = self.env['estate.property'].browse(vals['property_id'])
+            property_id.state = 'offer_received'
+            for record in property_id.offer_ids:
+                if record.price > vals['price']:
+                    raise UserError("Cannot create an offer with a lower value than an existing one")
```
msho-odoo: It's correct of course, you can also use `max()` and `mapped()`, you will do the same thing in one line and without for loops, if you like

estate/models/property_type.py
```diff
@@ -0,0 +1,22 @@
+from odoo import models, fields
+
+
+class PropertyType(models.Model):
+    _name = "estate.property.type"
+    _description = "Property Type"
+    _order = "sequence asc"
```
msho-odoo: I think the type also wants order by name

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="estate_property_offer_view_tree" model="ir.ui.view">
+        <field name="name">estate.property.offer.list</field>
+        <field name="model">estate.property.offer</field>
+        <field name="arch" type="xml">
+            <list editable="bottom"
+                  decoration-success="status == 'accepted'"
+                  decoration-danger="status == 'refused'"
+            >
```
msho-odoo: We don't like a very long line but this one is fine, it makes it easier to read the file :)
```suggestion
            <list editable="bottom" decoration-success="status == 'accepted'" decoration-danger="status == 'refused'">
```

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,40 @@
...
+                <field name="price"/>
+                <field name="partner_id"/>
+                <field name="validity"/>
+                <field name="date_deadline"/>
+                <button name="action_accept" title="Accept" type="object" icon="fa-check"
+                        invisible="property_id_state in ('offer_accepted', 'sold', 'cancelled')"
```
msho-odoo, line 14: surely it's correct, but usually teams avoid adding new fields to the database unless absolutely necessary. It's no big issue here, it's a related field already but thought to mention it
For example here, if you already don't show an offer for cancelled or sold properties, you can just make the invisible condition here be something like `state != False`

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,129 @@
...
+    <record id="estate_property_view_form" model="ir.ui.view">
+        <field name="name">estate.property.form</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <form>
+                <header>
+                    <button name="action_do_sold" type="object" class="oe_highlight" string="Sold"
+                            invisible="state in ('sold', 'cancelled', 'new')"/>
+                    <button name="action_do_cancel" type="object" string="Cancel"
+                            invisible="state in ('sold', 'cancelled',)"/>
```
msho-odoo: nitpick
```suggestion
                            invisible="state in ('sold', 'cancelled')"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,129 @@
...
+                        <field name="tag_ids" widget="many2many_tags" options="{'color_field': 'color'}"/>
+                    </div>
+                    <group>
+                        <!--
+                        FIXME The attributes are documented but don't see to do anything
+                        https://www.odoo.com/documentation/19.0/developer/reference/frontend/javascript_reference.html#relational-fields:~:text=can_create,%3A
+                        -->
+                        <field name="property_type_id" widget="many2one" can_create="false"
+                               can_write="false"
+                               options="{'no_create': true}"
```
msho-odoo: I believe yours is correct, here is also a more used way if you like
```suggestion
                        <field name="property_type_id" widget="many2one" options="{'no_create': True, 'no_edit': True}"/>
```

msho-odoo (changes requested):
Thank you, Good work!
Left you small suggestions and nitpicks for chapter 11, no big thing

estate/views/estate_property_offer_views.xml
ollet-odoo: Initlay I was trying to use the property_id.state but it didn't seem to work. we cannot put condition on field of other object correct?

msho-odoo: Yes, correct I believe, if you must access that field like you did you have to have the related field, this is why my suggestion didn't include it at all :) 
you can leave your solution as is if you like, just thought to bring your attention to it.
I believe you also need to add that field to the view to use it like you did :thinking: (like add it and make it always invisible), is your current file working as expected?

ollet-odoo:
@msho-odoo I have made the fix asked,implemented the chapter 12 and the CI is passing, could you please review again?

estate/views/estate_property_offer_views.xml
ollet-odoo: I think I did have it as some point and removed it and I just tested it and it seems to be working. But the docs seems to agree with you that it shouldn't work.

estate/models/property.py
```diff
@@ -0,0 +1,113 @@
+from odoo import models, fields, api
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+
+class Property(models.Model):
+    _name = "estate.property"
```
msho-odoo, line 7: these two lines of the model and class name should align, so it's `class EstateProperty` and it's the case in all models.

P.S. sorry if I miss old points and I raise them now, I might have not noticed some stuff with the quick pace of the onboarding 😅

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,165 @@
...
+                        <field class="fw-bold fs-5" name="name"/>
+                        <group>
+                            <group>
+                                Expected price:
+                                <field name="expected_price"/>
+                                <div t-if="['offer_received', 'offer_accepted', 'sold'].indexOf(record.state.raw_value) >=  0">
+                                    Best price:
+                                    <field name="best_price"/>
+                                </div>
+                                <div t-if="['offer_accepted', 'sold'].indexOf(record.state.raw_value) >=  0">
```
msho-odoo, line 52: I think you can just check on the selling price and show it if not 0

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,165 @@
...
+                                    Selling price:
+                                    <field name="selling_price"/>
+                                </div>
+                                <div t-if="['sold'].indexOf(record.state.raw_value) >=  0">
+                                    <widget name="web_ribbon" title="Sold" bg_color="text-bg-danger"/>
+                                </div>
```
msho-odoo, line 58: There is an invisible attribute for the widget, you don't have to add a t- directive to check

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,165 @@
...
+    </record>
+    <record id="estate_property_view_kanban" model="ir.ui.view">
+        <field name="name">estate.property.kanban</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <kanban
+                    default_group_by="property_type_id"
+                    edit="False"
+                    records_draggable="False"
+            >
```
msho-odoo, line 39: just a nitpick. We don't like very long lines but don't go extreme :)
```suggestion
            <kanban default_group_by="property_type_id" edit="False" records_draggable="False">
```

estate/models/property_type.py
```diff
@@ -0,0 +1,22 @@
+from odoo import models, fields
+
+
+class PropertyType(models.Model):
+    _name = "estate.property.type"
+    _description = "Property Type"
+    _order = "sequence asc, name asc"
```
msho-odoo, line 7: order is ascending by default
```suggestion
    _order = "sequence, name"
```

estate/models/property_offer.py
```diff
@@ -0,0 +1,73 @@
...
+            )
+            record.validity = (record.date_deadline - base_date).days
+
+    @api.model
+    def create(self, vals_list):
+        for vals in vals_list:
+            property_id = self.env['estate.property'].browse(vals['property_id'])
+            property_id.state = 'offer_received'
+            if property_id.offer_ids and max(property_id.offer_ids.mapped('price')) > vals['price']:
+                raise UserError("Cannot create an offer with a lower value than an existing one")
```
msho-odoo, line 55: oh I forgot to mention this but when we show strings like that error to the user we want to support translation as well so we add this `_` imported `from odoo` to do that.

```suggestion
                raise UserError(_("Cannot create an offer with a lower value than an existing one"))
```

estate_account/models/__init__.py
```diff
@@ -0,0 +1,33 @@
+from odoo import models, Command
```
msho-odoo, line 1: this file shouldn't be called `__init__.py`

estate_account/models/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+                        }),
+                        Command.create({
+                            "price_unit": 100.00,
+                            "quantity": "1",
+                            "name": "Administrative fee"
+                        }),
+                    ]
+                }
+            )
+        super().action_do_sold()
```
msho-odoo, line 33: This super call should be at the beginning of the method, you don't want to create an account_move and then the super raises an error

estate_account/models/__init__.py
```diff
@@ -0,0 +1,33 @@
+from odoo import models, Command
+
+
+class Property(models.Model):
+    _inherit = "estate.property"
+
+    def action_do_sold(self):
+        move_type = "out_invoice"  # Customer Invoice
+        journals = self.env["account.journal"].search([("type", "=", "sale")])
+        journal = next(iter(journals))
```
msho-odoo, line 10: Why add this line? you typically have a journal for sale that you add moves to it with every sale but it's the same one journal

estate_account/models/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+        for record in self:
+            partner_id_id = record.buyer_id.id
+            self.env["account.move"].create(
+                {
+                    "journal_id": journal.id,
+                    "move_type": move_type,
+                    "partner_id": partner_id_id,
```
msho-odoo, line 17: we don't have `_id_id` for names, it's a bit confusing
```suggestion
                    "partner_id": record.buyer_id.id,
```

msho-odoo (changes requested):
Thank you for the good work 👍 
I left you some comments and nitpicks to look it.
If you find a comment worth applying in other places, please do.

Thanks, really good work!

estate/views/estate_property_views.xml
ollet-odoo: I was thinking we might still want to display it if we sold if for 0 euro :rofl:

estate_account/models/__init__.py
ollet-odoo: I need a journal, at first I was doing journals[0] but the linter was complaining. one of the example had a _get_default_journal but it doesn't seem to exist anymore (or yet in 19.0) ?

msho-odoo: I see, you can use `limit=1` on the search method, but it's weird I think the runbot complains about that :thinking:

msho-odoo:
@ollet-odoo 
Hi, I see you requested a review from me yesterday but I think some of my comments are not yet addressed, can you double check them and ping me when you're done 
Thanks!

ollet-odoo:
I double checked just now and found some things to fix. Sorry for not answering earlier, I was concentrated on the web tutorial.
I hadn't understood that when using limit=1 I didn't need to item on the list anymore. Everything should be correct now I hope.

msho-odoo:
No worries, well done on the changes and thank you for the hard work :rocket:
I will have a quick look at your web tutorial PR, didn't notice, sorry.

ollet-odoo:
> No worries, well done on the changes and thank you for the hard work 🚀 I will have a quick look at your web tutorial PR, didn't notice, sorry.

No problem. I just did an in person review of the web PR with another coach in the office. I will make corrections. So you don't need to review it again for now

PR #1415, opened 2026-09-15
19.0 tutorial delje

Delvaux-Jean-Baptiste:
[REF] Updated branch name
[FIX] Added extra end of file line


estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
+from odoo import fields, models
+
```
msho-odoo, line 6: Before `class ...` we conventionally leave 2 lines :)

estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
...
+    name= fields.Char('Estate Name', required= True)
+    description= fields.Text()
+    postcode= fields.Char()
+    date_availability= fields.Date()
+    expected_price= fields.Float()
+    selling_price= fields.Float()
+    bedrooms= fields.Integer()
+    living_area= fields.Integer()
+    facades= fields.Integer()
+    garage= fields.Boolean()
```
msho-odoo: Good practice to have boolean fields on the form `is_something` or `has_something` so you might change this to `has_garden` and `has_garage`

estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
...
+    garden= fields.Boolean()
+    garden_orientation= fields.Selection(string='Type',
+                                        selection=[('east', 'East'),
+                                                    ('west', 'West'),
+                                                    ('north', 'North'),
+                                                    ('south', 'South')
+                                                    ]
+                                        )
+
```
leclerc-leo: Small nitpick, We usually try to have a formatting like this:
```suggestion
    garden_orientation= fields.Selection(
        string='Type',
        selection=[
            ('east', 'East'),
            ('west', 'West'),
            ('north', 'North'),
            ('south', 'South'),
        ],
    )

```
You can install the `ruff` extension to help with formatting.
You can check for `ruff.toml` files for the config example in the codebase.

estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
...
+    postcode= fields.Char()
+    date_availability= fields.Date()
+    expected_price= fields.Float()
+    selling_price= fields.Float()
+    bedrooms= fields.Integer()
+    living_area= fields.Integer()
+    facades= fields.Integer()
+    garage= fields.Boolean()
+    garden= fields.Boolean()
+    garden_orientation= fields.Selection(string='Type',
```
msho-odoo: It's good, but it's better to commit to a convention and follow it all the way, so for example in this field you wrote the attribute `string` before `='Type'` but you didn't do that in name field `Char('Estate Name')`

estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
...
+                                        selection=[('east', 'East'),
+                                                    ('west', 'West'),
+                                                    ('north', 'North'),
+                                                    ('south', 'South')
+                                                    ]
+                                        )
```
msho-odoo: It's correct but we style it like this (see the suggestion).

so for example the selection can be ` ('east', "East")` instead of  `('east', 'East')` but some teams don't do this, they just stick to their own convention which is all single quotes (if possible) or all double quotes

```suggestion
    garden_orientation= fields.Selection(
        string='Type',
        selection=[
           ('east', 'East'),
           ('west', 'West'),
           ('north', 'North'),
           ('south', 'South'),
         ]
      )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
+from odoo import fields, models
+
+class EstatePropertyModel(models.Model):
+    _name = "estate_property"
+    _description = "The details of a property"
+
+    name= fields.Char('Estate Name', required= True)
```
msho-odoo: We do spaces before and after the operator so it's `name = fields` not `name= fields`.
Better to check all style errors in your [runbot](https://runbot.odoo.com/runbot/batch/2765725/build/125250892) and fix it :)
The log there provides a description for the reason of the error

msho-odoo (changes requested):
@Delvaux-Jean-Baptiste 
Nice job on the branch situation 🔥 

For the commit message we write in a way that aligns with `when this commit is merged it will .....`
so you might change `[IMP] estate: Added models` to something like `[IMP] estate: add estate property model` or however you like to describe your changes :)

but really nice job, well done 👍 
also left some comments for you to check

msho-odoo:
@Delvaux-Jean-Baptiste please mention me here when you think the PR is ready for another review, typically, after each chapter :)
Also remember to check the runbot for styling issues :wink:

Delvaux-Jean-Baptiste:
@msho-odoo I believe I am ready for the last review of the day ;-)

estate/models/estate_property.py
```diff
@@ -1,25 +1,41 @@
...
 
-    name = fields.Char('Estate Name', required = True)
+    name = fields.Char('Estate Name', required=True)
     description = fields.Text()
     postcode = fields.Char()
-    date_availability = fields.Date()
-    expected_price = fields.Float(required = True)
-    selling_price = fields.Float()
-    bedrooms = fields.Integer()
+    date_availability = fields.Date(copy=False, default=fields.Date.today()+relativedelta(months=3))
```
msho-odoo: directly valuating the fields.Date at default will work when imported but it will be stuck at this value, you should use `lambda` function. Search the codebase for it and let me know if you have any question :)

estate/views/estate_menus_views.xml
```diff
@@ -0,0 +1,11 @@
...
+<odoo>
+    <menuitem id="estate_menu_root" name="Real Estate">
+        <menuitem id="estate_fist_level_menu" name="Advertisements">
+            <menuitem 
+            id="estate_property_menu_action"
+            action="estate_property_action"
+            />
+        </menuitem>
+    </menuitem>
+</odoo>
```
msho-odoo: Please Always check your runbot and make sure it's not red :)
it should be telling you you are missing a new line on the EOF here :)

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,98 @@
...
+    <!-- List View -->
+    <record id="estate_property_list" model="ir.ui.view">
+        <field name="name">estate.property.list</field>
+        <field name="model">estate_property</field>
+        <field name="arch" type="xml">
+            <list string="Estates">
+                <field string="Estate" name="name"/>
+                <field name="postcode"/>
+                <field name="bedrooms"/>
+                <field string="Living Area (sqm)" name="living_area"/>
```
msho-odoo: nitpick: usually we have `name` attribute before `string` so other developers see right away what field you are referring to

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,98 @@
...
+        <field name="arch" type="xml">
+            <search string="Search properties">
+                <field string="Title" name="name"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field string="Living Area (sqm)" name="living_area"/>
+                <field name="facades"/>
+
+                <filter string="Available" name="available" domain="['|',('state', '=', 'New'), ('state', '=', 'offer_received')]"></filter>
```
msho-odoo: The key you check on is `'new'` not `'New'`
This could be easier to read
```suggestion
                <filter string="Available" name="available" domain="[('state', 'in', ('new', 'offer_received'))]"></filter>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,98 @@
...
+    <record id="[ID given to this record]" model="[Model being modified]">
+        <field name="name">[name for this field]</field>
+        <field name="model">[Model from which to gather data]</field>
+        <field name="arch" type="[Type of data architecture]">
+            <list string="Estates">
+                <field string="[Column String Visible]" name="[Data name from Model]"/>
+            </list>
+        </field>
+    </record>
+-->
```
msho-odoo, line 121: Delete unneeded comments please :)

.gitignore
```diff
@@ -127,3 +127,4 @@ dmypy.json
```
msho-odoo, line 127: ops, I shouldn't see this file here :)
environmental files like this one (or .vscode) for example shouldn't be pushed with your commit

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,98 @@
...
+<odoo>
+    <!-- Action Handler -->
+    <record id="estate_property_action" model="ir.actions.act_window">
+        <field name="name">Properties</field>
+        <field name="res_model">estate_property</field>
+        <field name="view_mode">list,form</field>
+    </record>
+
+    <!-- List View -->
+    <record id="estate_property_list" model="ir.ui.view">
```
msho-odoo: Should be `id="estate_property_view_list"` according to [coding guidelines](url)
I am writing this to grab your attention to the existing of such guidelines, you can have a look there although you might come across some existing files in the codebase that are not following this comment for example (maybe they are old enough or the team is not strict on conventions) but you should always follow guidelines for new diffs in the code :)

msho-odoo (changes requested):
Thanks for the good work!

Regarding the commit message we use `[ADD]` for adding whole new modules, but in this case (adding some field, view or feature/logic) we use `[IMP]` tag (for improvement) which you will be using most of the time in your team :)

I also left you some comments to check and please Always remember to green your runbot before pinging of the review 😉

Delvaux-Jean-Baptiste:
@msho-odoo I may or may not have forgotten to ask for review ysterday....
Anyway, chapters 7 to 10 are in here ;-)

estate/models/estate_property.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class EstatePropertyModel(models.Model):
+    _name = "estate_property"
+    _description = "The details of a property"
+
+    name = fields.Char('Estate Name', required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(copy=False, default=fields.Date.today() + relativedelta(months=3))
```
msho-odoo, line 15: using the `fields.Date` directly on the default like this, will make it only evaluated when imported and the value will be stuck, you should use a function on the default instead of the value directly hear, see `lambda` function in the codebase and use it here

estate/models/estate_property.py
```diff
@@ -0,0 +1,103 @@
...
+    has_garden = fields.Boolean()
+    garden_area = fields.Integer(default=0)
+    garden_orientation = fields.Selection(
+        string='Garden Orientation',
+        selection=[
+           ('east', 'East'),
+           ('west', 'West'),
+           ('north', 'North'),
+           ('south', 'South'),
+           ('', ''),
```
msho-odoo, line 31: hmm, no need to add noise values like this one, thinking about resetting the field at some point in the code? you can set to `False`

estate/models/estate_property.py
```diff
@@ -0,0 +1,103 @@
...
+        for record in self:
+            record.total_area = record.garden_area + record.living_area
+
+    @api.depends("offers_ids")
+    def _compute_best_price(self):
+        for record in self:
+            if record.offers_ids:
+                record.best_price = max(record.offers_ids.mapped('price'))
+            else:
+                record.best_price = 0
```
msho-odoo, line 65: totally correct, here is a more pyhtonic way if you like
```suggestion
        for record in self:
            record.best_price = max(record.offers_ids.mapped('price')) if record.offer_ids else 0.0
```
or you can use the `default` parameter in the max function if you like to look it up

estate/models/estate_property.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @api.onchange("has_garden")
+    def _onchange_has_garden(self):
+        self.garden_area = 10 if self.has_garden else 0
+        self.garden_orientation = 'south' if self.has_garden else ''
```
msho-odoo, line 70:
```suggestion
        self.garden_orientation = 'south' if self.has_garden else False
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def action_sell_property(self):
+        for record in self:
+            if record.state == 'cancelled':
+                UserError("Cannot sell a cancelled property")
```
msho-odoo, line 75: This `_` I added is imported `from odoo` and is used to support translation when showing text like that to the user
```suggestion
                UserError(_("Cannot sell a cancelled property"))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,103 @@
...
+                return False
+            record.state = 'sold'
```
msho-odoo, line 77: It's correct but **when we can** work on the recordset without looping on the records, we prefer it, it's more efficient to read and write to the database.
I think in this case we can check if any records has a state `cancelled`and in that case we raise the error (this is even more convenient on the business side, imagine the user is selling multiple properties at the same time after he selected some of them, it makes sense to raise the error if any of them is cancelled, instead of selling some and then when we hit a cancelled record, we stop. The user might be confused which ones got sold and which ones didn't)

estate/models/estate_property.py
```diff
@@ -0,0 +1,103 @@
...
+        'CHECK(expected_price > 0)',
+        'The price of a property must be strictly positive',
+    )
+
+    _check_selling_price_constraint = models.Constraint(
+        'CHECK(selling_price >= 0)',
+        'The selling price must be positive',
+    )
+
+    @api.constrains('selling_price', 'expected_price')
```
msho-odoo, line 98: Just the order of the constraints, public methods, private methods, etcs. has a convention, refer to [coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions) and check it

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,54 @@
...
+    _description = "An offer for an estate"
+
+    price = fields.Float()
+    status = fields.Selection(string='State',
+            selection=[
+                ('refused', 'Refused'),
+                ('accepted', 'Offer Accepted'),
+            ],
+            copy=False,
+        )
```
msho-odoo, line 19:
```suggestion
    status = fields.Selection(string='State',
        selection=[
            ('refused', 'Refused'),
            ('accepted', 'Offer Accepted'),
        ],
        copy=False,
    )
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,54 @@
...
+    partner_id = fields.Many2one("res.partner", string="Buyer")
+    property_id = fields.Many2one("estate_property", string="Property")
+    validity = fields.Integer(default=7)
+    date_deadline = fields.Date(compute="_compute_deadline", inverse="_inverse_deadline")
+
+    @api.depends("validity")
```
msho-odoo, line 25: it also depends on the `create_date`

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,54 @@
...
+    def _compute_deadline(self):
+        for record in self:
+            if record.create:
+                record.date_deadline = datetime.now() + relativedelta(days=record.validity)
```
msho-odoo, line 29: hmm won't this make the date_deadline here becomes `datetime` and the "else" part will make it `date`?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,54 @@
...
+            else:
+                record.date_deadline = record.create_date.date() + relativedelta(days=record.validity)
+
+    def _inverse_deadline(self):
+        for record in self:
+            record.validity = (record.date_deadline - record.create_date.date()).days
+
+    @api.depends("property_id")
```
msho-odoo, line 37: public action like this one don't take decorations `depends`, only the _compute functions does because the automatically change of you change that "depends" field but actions no, we use them to apply some logic

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,54 @@
...
+    def action_accept_offer(self):
+        for record in self:
+            if record.status == "accepted" or record.property_id.state in ['cancelled', 'sold', 'offer_accepted']:
+                return False
```
msho-odoo, line 41: you can raise an error telling the user they can't accept an offer for such a property and include the state of that property for example

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,54 @@
...
+            record.property_id.state = 'offer_accepted'
+            record.property_id.buyer_id = record.partner_id
+            record.status = 'accepted'
+        return True
+
+    @api.depends("property_id")
+    def action_refuse_offer(self):
+        for record in self:
+            if record.property_id.state in ['cancelled', 'sold']:
+                return False
```
msho-odoo, line 52: same thing with the depends and error message

estate/views/estate_menus_views.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <menuitem id="estate_menu_root" name="Real Estate">
```
msho-odoo, line 3: this one slipped out of the naming convention :)

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,33 @@
...
+        <field name="name">estate.property.tag.form</field>
+        <field name="model">estate_property_tag</field>
+        <field name="arch" type="xml">
+            <form>
+                <sheet>
+                    <h1><field name="name"></field></h1>
+                </sheet>
+            </form>
+        </field>
+    </record>
```
msho-odoo, line 32: Is it needed to add such views for list and form view to add changes to them? I am not sure if a chapter asks you to edit something in them but if not then you have them by default

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,122 @@
...
+    <!-- List View -->
+    <record id="estate_property_view_list" model="ir.ui.view">
+        <field name="name">estate.property.list</field>
+        <field name="model">estate_property</field>
+        <field name="arch" type="xml">
+            <list string="Estates">
+                <field name="name" string="Estate"/>
+                <field name="postcode"/>
+                <field name="bedrooms"/>
+                <field name="living_area" string="Living Area (sqm)"/>
```
msho-odoo, line 19: if you have the string attribute on the field definition in the python file, you can omit it here and in every other view that uses the same field (unless you want it different in a certain view)

estate/views/estate_property_views.xml
msho-odoo: @Delvaux-Jean-Baptiste  Please delete the comments in your file if they are not useful :)

.gitignore
msho-odoo: @Delvaux-Jean-Baptiste still can see the file 👀

msho-odoo (changes requested):
Thank you for your work!

Don't forget to add the module name in your [FIX] commits as well and follow the git guidelines with the message.

Left you some comments, but please apply the comments whenever they apply, I might have not pointed every point in every file, but if one of my comments apply in some other file, apply it there as well (like the translation `_` on every error message)

Thanks, keep it up 🔥

estate/views/estate_menus_views.xml
Delvaux-Jean-Baptiste: @msho-odoo Could you point me at which naming convention? I looked into a few other add-ons and I couldn't find some sort of pattern

estate/views/estate_property_tag_views.xml
Delvaux-Jean-Baptiste: I believe it was asked at some point to put a list. I know it is by default, but I do believe it asked for a custom one for some reason.

Delvaux-Jean-Baptiste:
@msho-odoo New release. Fixed as much as I could. What I haven't fixed I don't know how to, I'd be gad to have some guidance ;-)
How can I remove the gitignore file for example?

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+
+class EstatePropertyModel(models.Model):
+    _name = "estate_property"
```
msho-odoo, line 9: The name of the model and the class should match, also the name of the model uses `.` instead of `_`. Apply it in all models, please :)
```suggestion
class EstateProperty(models.Model):
    _name = "estate.property"
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+    _description = "The details of a property"
+    _order = "id desc"
+
+    name = fields.Char('Estate Name', required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(copy=False, default=lambda self: fields.Date.today() - relativedelta(months=3))
```
```suggestion
    date_availability = fields.Date(copy=False, default=lambda self: fields.Date.today() + relativedelta(months=3))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+           ('sold', 'Sold'),
+           ('cancelled', 'Cancelled'),
+        ],
+        default='new',
+    )
+    active = fields.Boolean(default=True)
+    property_type_id = fields.Many2one("estate_property_type", string="Type")
+    buyer_id = fields.Many2one("res.partner", string="Buyer")
+    salesperson_id = fields.Many2one('res.users', string='Salesperson', index=True, default=lambda self: self.env.user)
+    property_tags_ids = fields.Many2many("estate_property_tag", string="Tags")
```
msho-odoo, line 49: We pass the name of the model with dots, please apply it everywhere
```suggestion
    property_tags_ids = fields.Many2many("estate.property.tag", string="Tags")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+    offers_ids = fields.One2many("estate_property_offer", "property_id", string="Offers")
+    total_area = fields.Integer(compute="_compute_total_area", readonly=True)
+    best_price = fields.Float(compute="_compute_best_price", string="Best Offer", readonly=True)
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.garden_area + record.living_area
+
+    @api.depends("offers_ids")
```
```suggestion
    @api.depends("offers_ids.price")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+    def _onchange_has_garden(self):
+        self.garden_area = 10 if self.has_garden else 0
+        self.garden_orientation = 'south' if self.has_garden else False
+
+    def action_sell_property(self):
+        for record in self:
+            if record.state == 'cancelled':
+                UserError(self.env._("Cannot sell a cancelled property"))
+                return False
+            record.state = 'sold'
```
msho-odoo: It's better for the database to read and write in recordset instead of record by record **when you can**, you can here check if any of the states in self is cancelled and raise the error and later also assign the state 'sold' to `self` one time

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+        return True
+
+    def action_cancel_property(self):
+        for record in self:
+            if record.state == 'sold':
+                UserError(self.env._("Cannot cancel a sold property"))
+                return False
+            record.state = 'cancelled'
```
msho-odoo: same thing here as above

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+    )
+
+    @api.constrains('selling_price', 'expected_price')
+    def _check_selling_price(self):
+        for record in self:
+            if (not float_is_zero(record.selling_price, 2)) and float_compare(record.selling_price, record.expected_price * 0.9, 2) < 0:
+                raise ValidationError(self.env._("The selling price is less than 90 percent of the asked price"))
+
+    @api.ondelete(at_uninstall=False)
+    def _check_deletion(self):
```
msho-odoo, line 102: don't start it with `_check`, that's for constrains, you can start it with `_unlink` maybe `_unlink_if_not_new_or_cancelled`

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,67 @@
...
+            ],
+            copy=False,
+        )
+    partner_id = fields.Many2one("res.partner", string="Buyer")
+    property_id = fields.Many2one("estate_property", string="Property")
+    validity = fields.Integer(default=7)
+    date_deadline = fields.Date(compute="_compute_deadline", inverse="_inverse_deadline")
+    property_type_id = fields.Many2one("estate_property_type", related="property_id.property_type_id", store=True)
+
+    @api.depends("validity")
```
msho-odoo, line 28: it also depends on `create_date`

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,67 @@
...
+    def _compute_deadline(self):
+        for record in self:
+            if record.create:
```
msho-odoo, line 31: I think you swapped the if-else blocks? 🤔 
Also:
```suggestion
            if record.create_date:
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,67 @@
...
+                record.date_deadline = datetime.now() + relativedelta(days=record.validity)
+            else:
+                record.date_deadline = record.create_date.date() + relativedelta(days=record.validity)
+
+    def _inverse_deadline(self):
+        for record in self:
+            record.validity = (record.date_deadline - record.create_date.date()).days
```
msho-odoo, line 38: You might also add an if-else here for the `create_date` as you did in the _compute method

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,67 @@
...
+
+    @api.depends("property_id")
+    def action_accept_offer(self):
+        for record in self:
+            if record.status == "accepted" or record.property_id.state in ['cancelled', 'sold', 'offer_accepted']:
+                return False
```
msho-odoo, line 44: You can raise an error to the user telling him why their action is not executed instead of silently returning False

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,67 @@
...
+            if record.create:
+                record.date_deadline = datetime.now() + relativedelta(days=record.validity)
+            else:
+                record.date_deadline = record.create_date.date() + relativedelta(days=record.validity)
+
+    def _inverse_deadline(self):
+        for record in self:
+            record.validity = (record.date_deadline - record.create_date.date()).days
+
+    @api.depends("property_id")
```
msho-odoo, line 40: depends decorator is for the _compute methods, public action methods like this one don't get a depends decorator, they are just triggered when you click on the button.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,67 @@
...
+        for record in self:
+            if record.status == "accepted" or record.property_id.state in ['cancelled', 'sold', 'offer_accepted']:
+                return False
+            record.property_id.selling_price = record.price
+            record.property_id.state = 'offer_accepted'
+            record.property_id.buyer_id = record.partner_id
+            record.status = 'accepted'
+        return True
+
+    @api.depends("property_id")
```
msho-odoo, line 51: same with the decorator

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,67 @@
...
+        for record in self:
+            if record.property_id.state in ['cancelled', 'sold']:
+                return False
+            record.status = 'refused'
+        return True
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            property = self.env['estate_property'].browse(vals['property_id'])
```
msho-odoo, line 62: you might add a check if 'property_id` exist in the vals, in case you can create an offer with no property_id.

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,67 @@
...
+            if property.state == "new":
+                property.state = "offer_received"
```
msho-odoo, line 64: Be careful to change the state here before the check and raising the error.
Here in this case it's okay, since at least the first offer should always pass the check and one offer is enough to change the state, but just to bring your attention to it.

estate/models/estate_property.py
msho-odoo, line 102: Refer to t[he coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions) ( the order section). I think these check constrains and private methods should come before public methods (like the action here)

estate/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import test_unit_test
```
msho-odoo: The EOF line :)

estate/views/estate_property_offer_views.xml
```diff
@@ -0,0 +1,20 @@
...
+<odoo>
+    <!-- List View -->
+    <record id="estate_property_offer_view_list" model="ir.ui.view">
+        <field name="name">estate.property.offer.list</field>
+        <field name="model">estate_property_offer</field>
+        <field name="arch" type="xml">
+            <list string="Tags" editable="top" decoration-success="status == 'accepted'" decoration-danger="status == 'refused'">
+                <field string="Buyer" name="partner_id"/>
+                <field string="Price" name="price"/>
+                <field string="Validity" name="validity"/>
```
msho-odoo, line 11: nitpick if you like: You can add the `string` in the fields definition in the python model, and don't need to add them here again or any other view that shows the field.

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,53 @@
...
+            <form>
+                <sheet>
+                    <h1><field name="name"></field></h1>
+                    <div class="oe_button_box" name="button_box">
+                        <button class="oe_stat_button" 
+                                type="object"
+                                name="action_view_offer"
+                                icon="fa-money">
+                            <field name="offer_count" widget="statinfo" srring="Offers"/>
+                        </button>
```
msho-odoo, line 36: I think it's good, there also another way to do this without adding an action to the python file since it uses the normal views anyway.
You can use 
```xml
<button name="%(estate.estate_property_offer_action)d" ..... >
</button>
```
See [example](https://github.com/odoo/odoo/blob/bf6fea1fb37dc07ee7adcdf22a7e31f020e8f63e/addons/crm/views/crm_lead_views.xml#L16) in crm

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,113 @@
+
+<odoo>
+    <!-- Action Handler -->
+    <record id="estate_property_action" model="ir.actions.act_window">
+        <field name="name">Property</field>
+        <field name="res_model">estate_property</field>
+        <field name="view_mode">list,form</field>
+        <field name="context">{'search_default_estate_state': 1}</field>
```
msho-odoo, line 8: What does this line do? 🤔

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,113 @@
...
+     <record id="estate_property_search" model="ir.ui.view">
+        <field name="name">estate.property.search</field>
+        <field name="model">estate_property</field>
+        <field name="arch" type="xml">
+            <search string="Search properties">
+                <field string="Title" name="name"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field string="Living Area (sqm)" name="living_area" filter_domain="[('living_area', '&gt;', self)]"/>
```
msho-odoo, line 103: This is greater than, make it greater than or equal

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,113 @@
...
+                <field name="facades"/>
+
+                <filter name="estate_state" string="Available" domain="[('state', 'in', ('new', 'offer_received'))]"></filter>
```
msho-odoo: There is no `estate_state` field defined, did you mean `state`?

estate/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+{
+    'name': "Estate",
+    'author': "delje",
```
msho-odoo: author should be `Odoo S.A.` :)

estate_account/__manifest__.py
```diff
@@ -0,0 +1,10 @@
+{
+    'name': "Estate_account",
+    'author': "delje",
+    'license': "LGPL-3",
+    'depends': ["base", "estate", "account"],
```
msho-odoo: no need to add the `base`, it's implicit since you are using estate and account which depend on base

estate_account/__manifest__.py
```diff
@@ -0,0 +1,10 @@
...
+    'application': True,
+    'installable': True,
+    'data': [
+    ],
+}
```
msho-odoo, line 11: There is another key-value missing, which allows the module to be automatically installed when the modules it depends on are installed.

estate_account/models/res_users.py
```diff
@@ -0,0 +1,8 @@
+from odoo import fields, models
+
```
msho-odoo: I see the `__init__.py` file empty, it should include the model import

msho-odoo (changes requested):
Thank you for your work!
It's good I already can' see the `.gitignore` file anymore :) nice work but I can see the `.vscode` folder 👀, that also shouldn't be here.

If you need help with that or any other comment, don't hesitate to come directly to me, we will work it together 😉

Keep up the good work! and remember always make your runbot green :)

estate/models/estate_property.py
Delvaux-Jean-Baptiste: But that was on the tutorial :'(
That is going to be a nightmare to find them all :p

msho-odoo: It's okay, time is tight, you can skip this comment, and continue with the tutorials :+1:

estate/models/estate_property.py
Delvaux-Jean-Baptiste: @msho-odoo 
Something like that?
```
        for record in self:
            if record.state == 'cancelled':
                UserError(self.env._("Cannot sell a cancelled property"))
                return False
        self.state = 'sold'
        return True
```

msho-odoo: I see you did `self.state = 'sold'` in bulk outside the loop, yes that's correct, it's better.
Here also another way if you like:
```python
if "cancelled" in self.mapped('state'):
    raise UserError("You can't sell a cancelled property")
self.state = "sold"
return True
```

msho-odoo:
Thank you for the hard work and changes you made based on the comments, well done!
It's time to do some JS for the framework tutorial if you haven't already started :rocket:

PR #1418, opened 2026-09-15
surap - Technical Training

srapyx:
Chapter 5


estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_property
```
leclerc-leo: We try to always have an empty line at the end of every file.
You can use the Ruff extension with the config from https://runbot211.odoo.com/runbot/static/build/125366383-19-0/logs/ruff_config.txt to have warning about style inside your editor.
You can also check the ci/style from the runbot (https://runbot.odoo.com/runbot/bundle/190-tuto-surap-512658) to check what it says.

estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
...
+    bedroom: fields.Integer = fields.Integer()
+    living_area: fields.Integer = fields.Integer()
+    facades: fields.Integer = fields.Integer()
+    garden: fields.Boolean = fields.Boolean()
+    garden_area: fields.Integer = fields.Integer()
+    garage: fields.Boolean = fields.Boolean()
+    garden_orientation: fields.Selection = fields.Selection(
+        selection=[("north", "North"), ("south", "South"), ("east", "East"), ("west", "West")],
+        string='Type',
+        help="Type is used to separate Leads and Opportunities")
```
leclerc-leo, line 28: It's best to close those in a new line to make it easier to find where it ends
```suggestion
    garden_orientation: fields.Selection = fields.Selection(
        selection=[("north", "North"), ("south", "South"), ("east", "East"), ("west", "West")],
        string='Type',
        help="Type is used to separate Leads and Opportunities",
    )
```

estate/data/.gitkeep
leclerc-leo, line 1: it's best not to push .gitkee, __pycache__, ...

estate/models/estate_property.py
```diff
@@ -0,0 +1,25 @@
+
+
+from odoo import fields, models
```
leclerc-leo: a bit of nitpick
```suggestion
from odoo import fields, models
```

leclerc-leo (commented):
👋  Good Work
It's better to have a PR content that's a bit more explicit than `check`.

estate/models/estate_property.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @api.depends('garden_area', 'living_area')
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends('offer_ids.price')
+    def _compute_best_price(self):
+        for record in self:
+            record.best_price = max([offer.price for offer in record.offer_ids] + [0])
```
leclerc-leo, line 54: you can use mapped to return the field in a list.
```suggestion
            record.best_price = max([0, *record.offer_ids.mapped('price')])
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,84 @@
...
+
+class EstateProperty(models.Model):
+    _name: str = "estate.property"
+    _active = True
+    _description: str | None = None
+
+    name: fields.Char = fields.Char()
+    description: fields.Text = fields.Text()
+    postcode: fields.Char = fields.Char()
+    date_availability: fields.Date = fields.Date('Availability Date', copy=False, default=date_utils.add(fields.Date.today(), months=3))
```
leclerc-leo, line 16: You need to put it inside a function/lambda otherwise today is resolved only when the server start so it will diverge if the server run for more than a day.
```suggestion
    date_availability= fields.Date('Availability Date', copy=False, default=lambda self: date_utils.add(fields.Date.today(), months=3))
```

Also you dont need to type it, if you use odoo-ls it will be typed correctly.

estate/models/estate_property.py
```diff
@@ -0,0 +1,84 @@
...
+            self.garden_orientation = 'north'
+        else:
+            self.garden_area = 0
+            self.garden_orientation = None
+
+    @api.constrains('selling_price')
+    def _check_selling_price(self):
+        for record in self:
+            if float_compare(record.selling_price, record.expected_price * 0.9, 2) < 0:
+                raise ValidationError("Can not accept offer if offered price is lower than 90% of expected")
```
leclerc-leo, line 69: You can use `_` to translate the string for the user.
```suggestion
                raise ValidationError(_("Can not accept offer if offered price is lower than 90% of expected"))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def _check_state_compatibility_because(self, record, state, message):
+        if record.state == state:
+            raise UserError(message)
+
+    def sell(self):
+        for record in self:
+            self._check_state_compatibility_because(record, "cancelled", "Canceled property can not be sold")
```
leclerc-leo, line 77: I find it easier to understand to use directly the condition rather than a function that just check the state.
```suggestion
            if record.state == 'cancelled':
                raise UserError(_("Canceled property can not be sold"))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,84 @@
+from odoo import api, fields, models
+from odoo.exceptions import UserError
+from odoo.orm.utils import ValidationError
+from odoo.tools import date_utils, float_compare
+
+
+class EstateProperty(models.Model):
+    _name: str = "estate.property"
```
leclerc-leo, line 8: We dont type anything, it's usually only done in the framework. This is because typing everything make things slower and usually it's not needed.
```suggestion
    _name = "estate.property"
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,84 @@
...
+    facades: fields.Integer = fields.Integer()
+    garden: fields.Boolean = fields.Boolean()
+    garden_area: fields.Integer = fields.Integer()
+    garage: fields.Boolean = fields.Boolean()
+    garden_orientation: fields.Selection = fields.Selection(
+        selection=[("north", "North"), ("south", "South"), ("east", "East"), ("west", "West")],
+        string='Type',
+        help="Type is used to separate Leads and Opportunities")
+    state = fields.Selection(
+        [(word.lower(), word) for word in ["New", "Offer Received", "Offer Accepted", "Sold", "Cancelled"]],
```
leclerc-leo, line 30: it's better to explicitly write them rather than use (word.lower(), word).
For example, here "Offer Received" could be just "received". (and "offer received" contains a space which we dont usually do)

leclerc-leo (commented):
👋  Good Work here are some more comments.

PR #1419, opened 2026-09-15
aalon - Technical Onboarding 

Elio1104:
[ADD] '.idea' to gitignore


.gitignore
```diff
@@ -127,3 +127,4 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+.idea
```
leclerc-leo: Try not to push changes to .gitignore, you can add a .gitignore in the folder that contains tutorials, odoo (community) and enterprise folder instead and it should still register.

If you only have tutorial opened, you should probably open a folder that contains those 3 folders as you can use it to search in the odoo/enterprise codebase for example which will help greatly later.

estate/models/estate_property.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate Property"
+
+    name = fields.Char('Name', required=True)
+    description = fields.Text('Description')
+    postcode = fields.Char('Postcode')
+    date_availability = fields.Date('Availability Date', copy=False, default=lambda self: fields.Date.add(fields.Date.today(), days=90))
```
leclerc-leo: We try to keep the lines around ~120 characters. If you use vscode you can configure the settings for `rulers` to add one there.
Nitpick: some people like to have `string=` before the string to make it clearer. 
```suggestion
    date_availability = fields.Date(
        string='Availability Date',
        copy=False,
        default=lambda self: fields.Date.add(fields.Date.today(), days=90)<
    )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,41 @@
...
+    garage = fields.Boolean('is Garage')
+    garden = fields.Boolean('is Garden')
+    garden_area = fields.Integer('Garden Area')
+    garden_orientation = fields.Selection(
+        string='Garden Orientation',
+        selection=[
+            ("north", "North"),
+            ("south", "South"),
+            ("east", "East"),
+            ("west", "West")],
```
leclerc-leo: nitpick
```suggestion
        selection=[
            ("north", "North"),
            ("south", "South"),
            ("east", "East"),
            ("west", "West"),
        ],
```

leclerc-leo (commented):
👋  Great work

aboo-odoo:

estate/models/estate_property.py
```diff
@@ -0,0 +1,140 @@
+from odoo import fields, models, api
```
aboo-odoo, line 1: [Imports](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#imports)
The imports are ordered as: 
- Imports of odoo submodules

```suggestion
from odoo import api, fields, models
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,140 @@
...
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools import float_compare
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate Property"
+
+    name = fields.Char('Title', required=True)
```
aboo-odoo, line 11: For your personal knowledge, you could also

```suggestion
    name = fields.Char(required=True)
```
By default, the string of a field is that field name 😄 But leaving it is not an error either, it's personal preferences 👍

estate/models/estate_property.py
```diff
@@ -0,0 +1,140 @@
...
+    description = fields.Text('Description')
```
aboo-odoo, line 12: Some teams like UI strings to be double quoted; this also applies to all your other strings. Don't change them all because it's not meaningful but just keep it in mind.


```suggestion
    description = fields.Text("Description")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,140 @@
...
+    postcode = fields.Char('Postcode')
+    property_type_id = fields.Many2one(
```
aboo-odoo, line 14: Nice use of `_id.s` suffixes 👍

estate/models/estate_property.py
```diff
@@ -0,0 +1,140 @@
...
+        string='Property Type',
+    )
+    tag_ids = fields.Many2many(
+        comodel_name='estate.property.tag',
+        string='Tags',
+    )
+    date_availability = fields.Date(
+        string='Available From',
+        copy=False,
+        default=lambda self: fields.Date.add(
```
aboo-odoo, line 25: Good use of a lambda function here! Well done 😄

estate/models/estate_property.py
```diff
@@ -0,0 +1,140 @@
...
+    bedrooms = fields.Integer('Bedrooms', default=2)
+    living_area = fields.Integer('Living Area (sqm)')
+    facades = fields.Integer('Number of Facades')
+    garage = fields.Boolean('is Garage')
+    garden = fields.Boolean('is Garden')
+    garden_area = fields.Integer('Garden Area')
+    garden_orientation = fields.Selection(
+        string='Garden Orientation',
+        selection=[
+            ("north", "North"),
```
aboo-odoo, line 48: Either you put single quote everywhere, double quote everywhere or double quote UI strings and single quote the rest but **be consistant**

estate/models/estate_property.py
```diff
@@ -0,0 +1,140 @@
...
+    )
+
+    _check_selling_price = models.Constraint(
+        definition='CHECK(selling_price >= 0)',
+        message='The selling price must be positive',
+    )
+
+    @api.constrains("selling_price", "expected_price")
+    def _check_selling_price(self):
+        for record in self:
```
aboo-odoo, line 106: Here you have quite small compute methods. In the future, you might have huge once, at that point, it's important to have meaningful parameter naming.
```suggestion
        for estate in self:
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,140 @@
...
+            if float_compare(record.selling_price, record.expected_price * 0.9, precision_rounding=0.01) < 0:
+                raise ValidationError('The selling price must be at least 90% of the Expected price')
```
aboo-odoo: Most strings are translated by default in Odoo. Unfortunately, this is not case of Error Messages. To make the string translatable, you need to call the `_()` method, like so

```suggestion
                raise ValidationError(self.env._('The selling price must be at least 90% of the Expected price'))
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,92 @@
...
+                ('id', '!=', record.id),
+                ('status', '=', 'accepted'),
+            ])
+
+            accepted_offer.status = 'refused'
+
+            record.property_id.state = 'offer_accepted'
+            record.property_id.selling_price = record.price
+            record.property_id.buyer = record.partner_id
+        return True
```
aboo-odoo, line 85:
```suggestion
        offers_to_accept = self.filtered(lambda p: p.state not in ['sold', 'cancelled'])
        if self - offers_to_accept:
            raise UserError(...)
        else:
            offers_to_accept.property_id.state = 'offer_accepted'
            offers_to_accept.property_id.selling_price = record.price
            offers_to_accept.property_id.buyer = record.partner_id
        return True
```

Advantages are: 

Now, let's talk about your search. A search is an SQL operation in the database. This means that you are doing an SQL transaction in a loop which is a performance killer. 

What you could do to help is to search before the loop to "pre-process" data and use that data instead. Like so:

 ```suggestion
    def action_accept(self):
        offers_to_accept = self.filtered(lambda p: p.state not in ['sold', 'cancelled'])
        if not(self - offers_to_accept):  # not all the offers are to be accepted
            raise UserError(...)
        
        already_accepted_offers = self.search([
            ('property_id', 'in', offers_to_accept.property_id.ids),
            ('id', 'not in', offers_to_accept.ids),
            ('status', '=', 'accepted'),
        ])
        already_accepted_offers.status = 'refused'

        offers_to_accept.property_id.state = 'offer_accepted'
        offers_to_accept.property_id.selling_price = record.price
        offers_to_accept.property_id.buyer = record.partner_id
        return True
```

Now, mind that there might be an issue if 2 offers for the same property are accepted at the same time (unlikely but possible); this needs safeguard. For example, by grouping the records by `property_id`, and if several offers relate to the same property, accept only one and refuse the others.

**Review by aboo-odoo — COMMENTED** [review]
Overall, excellent job! 🥳

Global comments:

File: estate/tests/test_estate_property.py [hunk 13]
```diff
@@ -0,0 +1,47 @@
...
+            },{
+                'price': 400000,
+                'partner_id': cls.buyers.id,
+                'validity': 10,
+                'property_id': cls.property.id
+            }
+        ])
+
+    def test_accept_offer(self):
+        self.property.offer_ids[0].action_accept()
```
aboo-odoo: Your test don't assert anything 😄 I guess it does check that there is no error when run ahah

aboo-odoo (commented):
Thank you for the test, it could be improved a little but it's a good start 👍

PR #1420, opened 2026-09-15
[ADD] estate : add Real estate module to mangage estate

Rs-appez:


estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_property, estate_property_tag, estate_property_type
```
leclerc-leo: nitpick, we usually do something like this.
```suggestion
from . import estate_property
from . import estate_property_tag
from . import estate_property_type
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,50 @@
+from odoo import fields, models
+
+
+class Property(models.Model):
+    _name = "estate_property"
```
leclerc-leo: We usually use `name.of.module.name.of.model`
```suggestion
    _name = "estate.property"
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,50 @@
...
+    _description = "Estate Property"
+
+    name = fields.Char(required=True)
+    postcode = fields.Char()
+    description = fields.Text()
+    date_availability = fields.Date(
+        copy=False,
+        default=lambda _: fields.Date.add(fields.Date.today(), months=3),
```
leclerc-leo: `_` is reserved for translations (check `_(` in the odoo codebase for it's usage)
```suggestion
        default=lambda self: fields.Date.add(fields.Date.today(), months=3),
```

estate/views/estate_menus.xml
```diff
@@ -0,0 +1,14 @@
...
+            <menuitem id="estate_model_menu_action" action="estate_property_model_action" />
+        </menuitem>
+        <menuitem id="estate_setings_menu_first_level" name="Settings">
+            <menuitem id="estate_property_type_model_menu_action"
+                action="estate_property_type_model_action" />
+            <menuitem id="estate_property_tag_model_menu_action"
+                action="estate_property_tag_model_action" />
+        </menuitem>
+    </menuitem>
+</odoo>
```
leclerc-leo: we add an extra empty line at the end of every file to not have this symbol appear on github.

estate/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "estate",
+    "license": "LGPL-3",
+    "author": "alpir",
```
```suggestion
    "author": "Odoo S.A.",
```

leclerc-leo (commented):
👋  Great Work, just some nitpick

aboo-odoo:

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+    _order = "id desc"
+
+    _check_positive_expected_price = models.Constraint(
+        "CHECK(expected_price > 0)",
+        "The expected price must be stricly positive",
+    )
+    _check_positive_selling_price = models.Constraint(
+        "CHECK(selling_price >= 0)",
+        "The selling price must be positive",
+    )
```
aboo-odoo: Order of a python file is :
3. Field declarations
9. Action methods

So all your SQL constraints need to be after the fields

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+
+    name = fields.Char(required=True)
+    postcode = fields.Char()
+    description = fields.Text()
+    date_availability = fields.Date(
+        copy=False,
+        default=lambda self: fields.Date.add(fields.Date.today(), months=3),
```
aboo-odoo, line 18: Good use of lambda here, well done!

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+        required=True,
+        default="new",
+    )
+    property_type_id = fields.Many2one(
+        "estate.property.type",
+        string="Property Type",
+    )
+    tags_ids = fields.Many2many("estate.property.tag")
+    buyer = fields.Many2one("res.partner", copy=False)
+    salesman = fields.Many2one("res.users", default=lambda self: self.env.user)
```
```suggestion
    salesman_id = fields.Many2one("res.users", default=lambda self: self.env.user)
```

Convention is the suffix relational fields by `_id.s`

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+        copy=False,
+        required=True,
+        default="new",
+    )
+    property_type_id = fields.Many2one(
+        "estate.property.type",
+        string="Property Type",
+    )
+    tags_ids = fields.Many2many("estate.property.tag")
+    buyer = fields.Many2one("res.partner", copy=False)
```
```suggestion
    buyer_id = fields.Many2one("res.partner", copy=False)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+    salesman = fields.Many2one("res.users", default=lambda self: self.env.user)
+    offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers")
+
+    @api.depends("living_area", "living_area")
+    def _compute_total_area(self):
+        for record in self:
```
```suggestion
        for property in self:
```
Here you have quite small compute methods. In the future, you might have huge once, at that point, it's important to have meaningful parameter naming.

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids.price")
+    def _find_best_price(self):
+        for record in self:
+            if not record.offer_ids:
+                record.best_price = 0
+            else:
+                record.best_price = max(record.offer_ids.mapped("price"))
```
aboo-odoo: Could be simplified into 

```suggestion
            record.best_price = max(record.offer_ids.mapped("price"), default=0)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = None
```
```suggestion
            self.garden_orientation = False
```

`False` is the default empty value for fields

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+                if (
+                    float_compare(
+                        record.selling_price,
+                        record.expected_price * (THRESOLD_MIN_SELLING_POURCENT / 100),
+                        2,
+                    )
+                    < 0
+                ):
+                    raise ValidationError(
+                        _(
```
aboo-odoo: We now use the env._ for translations 
```suggestion
                        self.env._(
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+    # ------------------------------------------------------------
+    # ACTIONS
+    # ------------------------------------------------------------
+
+    def action_sold_property(self):
+        for record in self:
+            if record.state == "cancelled":
+                raise UserError(_("Cancelled property cannot be sold."))
+            record.state = "sold"
+        return True
```
aboo-odoo: A slightly better way but more advanced way, would be to do like so:
```suggestion
        properties_to_cancel = self.filtered(lambda p: p.state not in ['sold', 'cancelled'])
        if self - properties_to_cancel:
            raise UserError(...)
        else:
            properties_to_cancel.state = 'cancelled'
        return True
```

Advantages are: 

estate/models/estate_property.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def action_cancel_property(self):
+        for record in self:
+            record.state = "cancelled"
```
aboo-odoo: writes can be done in recordset, only reads needs to be sequential
```suggestion
        self.state = "cancelled"
```

aboo-odoo (commented):
Phenomenal work so far! Keep going

Generic comment:

Rs-appez:
@aboo-odoo 
I added the test, can you check it ?

aboo-odoo (commented):
Excellent! Thanks you for the test 😄

PR #1422, opened 2026-09-15
19.0 technical training coleo

coleo-odoo:


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
msho-odoo: ops, I should't see that file :)
see how to avoid adding such environment files in git pushes

estate/models/estate_property.py
```diff
@@ -0,0 +1,23 @@
...
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date()
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float()
+    bedrooms = fields.Integer()
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
```
msho-odoo: Good practice to have boolean fields on the form `is_something` or `has_something` so you might change this to `has_garden` and `has_garage`

estate/models/estate_property.py
```diff
@@ -0,0 +1,23 @@
...
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection(
+        string='Type',
+        selection=[('north', 'North'), ('south', 'South'), ('east', 'East'), ('west', 'West')]
+    )
```
msho-odoo, line 38: The styling is fine by me but when we have multiple values it's better to have them on separate files.

```suggestion
    garden_orientation = fields.Selection(
        string='Garden Orientation',
        selection=[
            ('north', 'North'),
            ('south', 'South'),
            ('east', 'East'),
            ('west', 'West'),
        ]
    )
```

so for example the selection can be  ('east', "East") instead of ('east', 'East') but some teams don't do this, they just stick to their own convention which is all single quotes (if possible) or all double quotes.
I see you are stick to single quotes, fine be me (though it's not always possible :) )

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+"id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
+"estate.access_estate_property","access_estate_property","estate.model_estate_property","base.group_user",1,1,1,1
```
msho-odoo: Always add new line at the end of files.
no need to add the `estate.`, the file is already in that module.
no need to add the double quotes as well, you can remove them.
Also as an improvement you can make the id `access_estate_property_user` (I added `_user` at the end) in case you want to add another rule on the same module but for the `_manager` not the base user

estate/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+{
+    'name': 'Real Estate',
+    'author': 'coleo',
```
msho-odoo: author is usually `Odoo S.A.` :)

msho-odoo (changes requested):
Nice work 🔥 
Well done on the commit messages format, just a nitpick you can ignore for now, that `CLN` tag is usually for heavy cleaning of the code base but usual small cleaning can have the `IMP` tag.

But it's worth to note that usually when you fix something after a review, no need to add a new commit, you can use `git commit --amend` which allows you to edit your commit message and also include your current staged changes to the same commit without making a new one.

Also left you some comments, and always remember to have your runbot checked when you push and make sure it's green :)

msho-odoo:
@coleo-odoo just forgot to tell you, you can mention me here when you think the PR is ready for another review, typically, after each chapter :)

coleo-odoo:
@msho-odoo I think the PR is ready for review.
I finished Chapter 5. Sorry for the number of commits, I'll try to keep it to one per chapter.

estate/models/estate_property.py
```diff
@@ -8,21 +8,41 @@ class Property(models.Model):
     name = fields.Char(required=True)
     description = fields.Text()
     postcode = fields.Char()
-    date_availability = fields.Date()
+    date_availability = fields.Date(
+        copy=False, 
+        default=fields.Date.add(fields.Date.today(), months=3)
```
msho-odoo: In this default the date will get valued when this is first imported and will not change later with each creation, you should use `lambda` function to fix this. Search the codebase for lambda functions used in the default values of fields and apply it here :)

estate/views/estate_menu.xml
```diff
@@ -0,0 +1,7 @@
+<odoo>
+    <menuitem id="estate_menu_root" name="Real Estate">
+        <menuitem id="estate_first_level_menu" name="Advertisements">
+            <menuitem id="estate_menu_action" action="estate_action" name="Properties"/>
+        </menuitem>
+    </menuitem>
+</odoo>
```
msho-odoo: Always remember the EOF new line and Always remember to fix red runbot to be green before reviews :)

estate/views/estate_menu.xml
```diff
@@ -0,0 +1,7 @@
+<odoo>
+    <menuitem id="estate_menu_root" name="Real Estate">
+        <menuitem id="estate_first_level_menu" name="Advertisements">
```
msho-odoo: Better to make ids more descriptive so that they indicate what they are without having to read the definition

estate/views/estate_property_views.xml
```diff
@@ -1,7 +1,7 @@
 <odoo>
     <record id="estate_action" model="ir.actions.act_window">
-        <field name="name">Porperty</field>
+        <field name="name">Porperties</field>
         <field name="res_model">estate.property</field>
-        <field name="view_mode">tree,form</field>
+        <field name="view_mode">list,form</field>
     </record>
 </odoo>
```
msho-odoo: EOF line in **ALL** files please: and remember the red runbot is already guiding you to fix this stuff so please check it before the review :)

estate/models/estate_property.py
```diff
@@ -8,21 +8,41 @@ class Property(models.Model):
...
+    state = fields.Selection(
+        string='State',
+        selection=[
+            ('new', 'New'),
+            ('offer_received', 'Offer Received'),
+            ('offer_accepted', 'Offer Accepted'),
+            ('sold', 'Sold'),
+            ('cancelled', 'Cancelled')
+        ],
+        default="new",
```
msho-odoo, line 49: Stick to a convention about quotes :)

estate/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+{
+    'name': 'Real Estate',
+    'author': 'Odoo S.A.',
+    'license': 'LGPL-3',
+    'depends': [
```
msho-odoo: nitpick: usually the `depends` list is written in one line since it's usually just small module names but no biggie

msho-odoo (changes requested):
Thanks for the nice work 👍
Left you a couple of comments.

nitpick: I would stick the commit message to be `estate: create...` instead of `estate: created...` since it should make a whole sentence when it's concatenated with `this commit will ....` according to [git guidlines](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html?highlight=git%20)

coleo-odoo:
@msho-odoo The PR should be ready for review

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
...
+        default=lambda self: fields.Date.add(
+            fields.Date.today(),
+            months=3,
+        ),
+    )
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float(
+        readonly=True,
+        copy=False,
+    )
```
msho-odoo: nitpick: we don't like very long lines but I think that field on one line is okay :)

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
...
+        ],
+        default="new",
+        required=True,
+        copy=False,
+    )
+    type_id = fields.Many2one("estate.property.type")
+    partner_id = fields.Many2one(
+        "res.partner",
+        string="Buyer",
+        index=True,
```
msho-odoo: hmm why the index?

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
...
+        string="Salesperson",
+        index=True,
+        default=lambda self: self.env.user,
+    )
+    tag_ids = fields.Many2many("estate.property.tag")
+    offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers")
+    best_price = fields.Float(compute="_compute_best_price")
+
+    _check_expected_price_positive = models.Constraint(
+        "CHECK(expected_price >= 0)",
```
msho-odoo, line 70: Expected price should be **strictly** positive (it can't be 0)

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @api.constrains("selling_price")
+    def _check_selling_price(self):
+        for record in self:
+            if (not float_is_zero(record.selling_price, 2)) and float_compare(
+                record.selling_price, (record.expected_price * 0.9), 2,
+            ) < 0:
+                msg = (
+                    r"The selling price cannot be lower than 90% of the expected price"
+                )
```
msho-odoo: Suggestion: No need for the extra brackets in the first condition and the extra newlines as well, this ways is shorter and easy to read
But most importantly note the `_` at the beginning of the string, this is to support translation into other languages but you need to import it `from odoo` (like fields and api imported)

```suggestion
            if not float_is_zero(record.selling_price, 2) and\
                float_compare(record.selling_price, (record.expected_price * 0.9), 2) < 0:
                raise ValidationError(_("The selling price cannot be lower than 90% of the expected price"))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
...
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids")
+    def _compute_best_price(self):
+        for record in self:
+            if record.offer_ids:
+                record.best_price = max(record.offer_ids.mapped("price"))
+            else:
+                record.best_price = 0
```
msho-odoo, line 98: Totally correct but here is a more pythonic way if you like
```suggestion
            record.best_price = max(record.offer_ids.mapped("price")) if record.offer_ids else 0.0
```
or you can use the `default` parameter in the max function if you like, you can look it up

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @api.onchange("has_garden")
+    def _onchange_hjas_garden(self):
```
msho-odoo: :)
```suggestion
    def _onchange_has_garden(self):
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
...
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = False
+
+    def cancel_property(self):
+        if self.state == "sold":
+            msg = "A sold property cannot be cancelled"
+            raise UserError(msg)
```
msho-odoo: Same translation underscore thing for all the strings raised to the user like this one

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
...
+        self.state = "cancelled"
```
msho-odoo: using `self.state` directly is dangerous, you can enter this method with a recordset instead of a single record (we always assume a user can apply the action on multiple records they selected at the same time which odoo can do), so you have to always assume you're working on a multiple records and do `for property in self:` and then work on the property.

That being said, we prefer to work on the recordset as a whole instead of looping on the records **when we can** which is not always possible. But here I think it's possible, as it makes sense to raise an error to the user if any of the records had `state` as `sold`, try to do that instead of `self.state == ...` because it won't work in this case.

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
...
+    @api.onchange("has_garden")
+    def _onchange_hjas_garden(self):
+        if self.has_garden:
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = False
+
+    def cancel_property(self):
```
msho-odoo: it's more of a convention to have these buttons with public methods to start with `action_` so you may name it `action_cancel_property` if you like

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
...
+        if self.state == "sold":
```
msho-odoo: you might want to prevent cancelling if the property is also already cancelled for example but that's a business decision so fine by me 👍

estate/models/estate_property.py
```diff
@@ -0,0 +1,129 @@
...
+            msg = "A sold property cannot be cancelled"
+            raise UserError(msg)
+        self.state = "cancelled"
+        return True
+
+    def sell_property(self):
+        if self.state == "cancelled":
+            msg = "A cancelled property cannot be sold"
+            raise UserError(msg)
+        if not ("accepted" in self.offer_ids.mapped("status")):
```
msho-odoo: Nice, you are doing the recordset check I was talking about (instead of looping) in the previous method :), you can make it also like this instead of the `self ==` if possible

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,65 @@
...
+        string="Partner",
+        index=True,
+        required=True,
+    )
+    property_id = fields.Many2one("estate.property", required=True)
+    validity = fields.Integer(default=7)
+    date_deadline = fields.Date(
+        compute="_compute_deadline",
+        inverse="_inverse_deadline",
+    )
```
msho-odoo: preferable to make the fields declarations in one line when it's not too long, it makes it a bit easier to scroll through them, looking at the name and type quickly but that's a nitpick from me :)

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,65 @@
...
+
+    _check_price_positive = models.Constraint(
+        "CHECK(price >= 0)",
```
msho-odoo: should be strictly positive as well

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,65 @@
...
+            record.validity = (
+                record.date_deadline - fields.Date.to_date(record.create_date)
+            ).days
+
+    def confirm_offer(self):
+        if "accepted" in self.property_id.offer_ids.mapped("status"):
+            msg = "Property already sold"
+            raise UserError(msg)
+        self.status = "accepted"
+        self.property_id.selling_price = self.price
```
msho-odoo: I think you should also set the buyer_id on the property to be the partner_id on the offer.

Also, I think it might cause an issue here to use `self` instead of looping through the records, since `self.price` or `self.partner_id` might be multiple prices or multiple partners and you will be assigning that to each price and partner_id in the property which is not correct (also the field on the property_id is `selling_price` not `selling_prices` :)

estate/views/estate_menu.xml
```diff
@@ -0,0 +1,11 @@
+<odoo>
+    <menuitem id="estate_menu_root" name="Real Estate">
+        <menuitem id="estate_first_level_menu_advertisements" name="Advertisements">
+            <menuitem id="estate_menu_action_property_list" action="estate_property_show_list" name="Properties"/>
+        </menuitem>
+        <menuitem id="estate_first_level_menu_settings" name="Settings">
+            <menuitem id="estate_menu_action_property_type_list" action="estate_property_type_show_list" name="Property Types"/>
+            <menuitem id="estate_menu_action_property_tag_list" action="estate_property_tag_show_list" name="Property Tags"/>
```
msho-odoo: you can refer to the [naming guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html?highlight=coding%20guidelines#xml-ids-and-naming) for better naming for menus

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,7 @@
+<odoo>
+    <record id="estate_property_tag_show_list" model="ir.actions.act_window">
+        <field name="name">Tags</field>
+        <field name="res_model">estate.property.tag</field>
+        <field name="view_mode">list</field>
```
msho-odoo: Is `form` not needed?

estate/views/estate_property_tag_views.xml
```diff
@@ -0,0 +1,7 @@
+<odoo>
+    <record id="estate_property_tag_show_list" model="ir.actions.act_window">
```
msho-odoo: you can call this `estate_property_tag_action` but no need to add the `list` part because you can add multiple `view_mode` other than list later

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,7 @@
+<odoo>
+    <record id="estate_property_type_show_list" model="ir.actions.act_window">
+        <field name="name">Types</field>
+        <field name="res_model">estate.property.type</field>
+        <field name="view_mode">list</field>
```
msho-odoo: same list and form thing, unless chapter asked you to only do list now 🤔 but the id shouldn't reflect that in this case, since changing ids in stable versions is not recommended

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,100 @@
...
+        <field name="arch" type="xml">
+            <search string="Search Property">
+                <field name="name" string="Title"/>
+                <field name="postcode"/>
+                <field name="expected_price"/>
+                <field name="bedrooms"/>
+                <field name="living_area" string="Living Area (sqm)"/>
+                <field name="facades"/>
+                <separator/>
+                <filter string="Available" name="date_availability" domain="['|', ('state', '=', 'new'), ('state', '=', 'offer_received')]"/>
```
msho-odoo: correct, but this is more readable
```suggestion
                <filter string="Available" name="date_availability" domain="[('state', 'in', ('new', 'offer_received'))]"/>
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,100 @@
...
+                <filter string="Post Code" name="postcode" context="{'group_by':'postcode'}"/>
```
msho-odoo, line 22: usually `name` comes before `string`

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,100 @@
...
+
+    <record id="estate_property_view_tree" model="ir.ui.view">
+        <field name="name">estate.property.list</field>
+        <field name="model">estate.property</field>
+        <field name="arch" type="xml">
+            <list string="Properties" editable="bottom" open_form_view="True">
+                <field name="name"/>
+                <field name="postcode"/>
+                <field name="bedrooms"/>
+                <field name="living_area" string="Living Area (sqm)"/>
```
msho-odoo: you can add the `string` to the python field definition and avoid adding it in the xml view each time you use the field (unless you wan tit different in a certain view)

msho-odoo (changes requested):
Thank you for the good work !

For the commit message header, it shouldn't be that long, ideally it should be < 50 characters, refer to [git guidelines](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html#commit-message-structure)

I left you a couple of comments and nitpicks, if you find one of my comments is valid for other parts of the code, please apply it there as well, I didn't point out every single point in case I mentioned it before in some other file :)

Keep up the good work, thanks 🔥

coleo-odoo:
@lost-odoo  I've added the tests on the PR

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
+from odoo import api, fields, models
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
+
+
+class Property(models.Model):
+    _name = "estate.property"
```
msho-odoo: class name and model name should match, make sure it's the case in all models
```suggestion
class EstateProperty(models.Model):
    _name = "estate.property"
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+    )
+    _check_selling_price_positive = models.Constraint(
+        "CHECK(selling_price > 0)",
+        "The Selling Price should be positive.",
+    )
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_if_property_sold_or_cancelled(self):
+        for record in self:
+            if not record.state in "cancelled sold":
```
msho-odoo: You can also use sets and check on the difference or intersections instead of looping record by record.
Also use `in ("cancelled", "sold")` if you will leave it like this

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+        for record in self:
+            if (
+                not float_is_zero(record.selling_price, 2)
+                and float_compare(
+                    record.selling_price,
+                    (record.expected_price * 0.9),
+                    2,
+                )
+                < 0
+            ):
```
msho-odoo, line 97: Way easier to read :)
I don't like very long lines either but don't go to far haha
```suggestion
                not float_is_zero(record.selling_price, 2)
                and float_compare(record.selling_price, (record.expected_price * 0.9), 2) < 0):
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+                raise ValidationError(
+                    self.env._(
+                        r"The selling price cannot be lower than 90% of the expected price",
+                    ),
+                )
```
msho-odoo: since we're talking about oneliners, 5 lines for an error is a bit long :) but it's not a big thing anyway
```suggestion
                raise ValidationError(self.env._("The selling price cannot be lower than 90% of the expected price"))
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids")
```
```suggestion
    @api.depends("offer_ids.price")
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+                record.garden_orientation = "north"
+            else:
+                record.garden_area = 0
+                record.garden_orientation = False
+
+    def action_cancel_property(self):
+        for record in self:
+            if record.state == "sold":
+                raise UserError(self.env._("A sold property cannot be cancelled"))
+            record.state = "cancelled"
```
msho-odoo, line 126: Better to work on a recordset instead of looping when you can, you might check if "sold" is among any of the state in `self` and raise an error. Also after the error you can assign `.state = "cancelled` in bulk (aka on `self`)

estate/models/estate_property.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def action_sell_property(self):
+        for record in self:
+            if record.state == "cancelled":
+                raise UserError(self.env._("A cancelled property cannot be sold"))
+            if not ("accepted" in record.offer_ids.mapped("status")):
+                raise UserError(
+                    self.env._("A offer need to be accepted before selling"),
+                )
+            record.state = "sold"
```
msho-odoo, line 137: I think same computation on recordset here is also applicable

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,79 @@
...
+    @api.model
+    def create(self, vals_list):
+        for vals in vals_list:
+            if self.env["estate.property"].browse(vals["property_id"]).state == "new":
+                self.env["estate.property"].browse(
+                    vals["property_id"],
+                ).state = "offer_received"
+        return super().create(vals_list)
+
+    def _inverse_deadline(self):
```
msho-odoo, line 49: Refer to the [coding guidelines](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#symbols-and-conventions), I believe the order of the methods here is not correct :)

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,79 @@
...
+                record_date,
+                days=record.validity,
+            )
+
+    @api.model
+    def create(self, vals_list):
+        for vals in vals_list:
+            if self.env["estate.property"].browse(vals["property_id"]).state == "new":
+                self.env["estate.property"].browse(
+                    vals["property_id"],
```
msho-odoo: You can apply `browse` only once, fetch the records and work on them. It's not ideal to hit the database multiple times when you can do it only once

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,79 @@
...
+                ).state = "offer_received"
```
msho-odoo: You should check if the offer you're creating is the biggest so fat, otherwise, you tell the user the property already has a higher offer

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,79 @@
...
+        inverse="_inverse_deadline",
+    )
+    property_type_id = fields.Many2one(related="property_id.type_id", store=True)
+
+    _check_price_positive = models.Constraint(
+        "CHECK(price > 0)",
+        "The Price of an Offer should be positive.",
+    )
+
+    @api.depends("validity")
```
msho-odoo: It also depends on `create_date`

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,79 @@
...
+    def _compute_deadline(self):
+        for record in self:
+            if record.create:
```
msho-odoo:
```suggestion
            if record.create_date:
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,79 @@
...
+            record.property_id.selling_price = record.price
+            record.property_id.state = "offer_accepted"
+            record.property_id.partner_id = record.partner_id
+        return True
+
+    def action_refuse_offer(self):
+        for record in self:
+            if record.status == "accepted":
+                record.property_id.selling_price = 0
+            record.status = "refused"
```
msho-odoo: Why not set the selling_price to 0 without the check and do it directly on `self`?

estate/models/res_users.py
```diff
@@ -0,0 +1,11 @@
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    property_ids = fields.One2many(
+        "estate.property",
+        "user_id",
+        string="Real Estate Properties",
```
msho-odoo, line 10: You can add `domain` here to achieve the chapter's goal and show only available properties

estate/tests/test_estate_property.py
```diff
@@ -0,0 +1,110 @@
...
+        self.assertEqual(
+            record.date_availability,
+            fields.Date.add(fields.Date.today(), months=3),
+        )
+        self.assertEqual(record.selling_price, 0)
+        self.assertEqual(record.best_price, 0)
+        self.assertEqual(record.active, True)
+
+        partner_id = self.env["res.partner"].create({"name": "Test Partner"}).id
+        property_id = record.id
```
msho-odoo: in odoo codebase, when someone sees the valriable `partner_id`, they expect it's the object itself not the `id` identifier in the database. Later if you want to use it, you can write `partner_id.id`

estate/tests/test_estate_property.py
```diff
@@ -0,0 +1,110 @@
...
+        has_garage = True
+        has_garden = True
+        garden_area = 5
+        garden_orientation = "north"
+        price_offer1 = 9000.00
+        status_offer1 = "refused"
+        price_offer2 = 10000.00
+        status_offer2 = "accepted"
+
+        record = self.env["estate.property"].create(
```
msho-odoo: It's not a big thing, but you can create this object in the `setUpClass(cls)` of test class which allows you to use it in case you created multiple tests, so you don't have to create an object like this every time you add a new test

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,146 @@
+<odoo>
+    <record id="estate_property_action" model="ir.actions.act_window">
+        <field name="name">Porperties</field>
+        <field name="res_model">estate.property</field>
+        <field name="context">{'search_default_date_availability': True}</field>
```
msho-odoo, line 5: I think the chapter wanted the default search view on `active`? or no?

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,146 @@
...
+                                <field name="has_garden"/>
+                                <field name="garden_area" invisible="not has_garden"/>
+                                <field name="garden_orientation" invisible="not has_garden"/>
+                                <field name="total_area"/>
+                                <field name="active"/>
+                            </group>
+                        </page>
+                        <page string="Offers">
+                            <field name="offer_ids"
+                                readonly="state == 'sold' or state == 'cancelled'"/>
```
msho-odoo: you can also don `state in ('sold', 'cancelled')`

estate_account/models/estate_property.py
```diff
@@ -0,0 +1,30 @@
...
+class Property(models.Model):
+    _inherit = "estate.property"
+
+    def action_sell_property(self):
+        for record in self:
+            self.env["account.move"].create(
+                {
+                    "partner_id": record.partner_id.id,
+                    "move_type": "out_invoice",
+                    "invoice_line_ids": [
```
msho-odoo, line 20: there is also `journal_id` field where the account_move belong.
You can fetch it from the database using [Search()](https://www.odoo.com/documentation/18.0/developer/reference/backend/orm.html#odoo.models.Model.search)
to have something like that:
`journal_id = self.env["account.journal"].search([], limit=1)`

estate_account/__manifest__.py
```diff
@@ -0,0 +1,10 @@
+{
+    'name': 'Real Estate Account',
+    'author': 'Odoo S.A.',
+    'license': 'LGPL-3',
+    'depends': [
+        'base',
+        'estate',
+        'account',
+    ],
+}
```
msho-odoo, line 11: don't forget `auto_install`

msho-odoo (changes requested):
Nice work 🔥 
Left you some comments,
keep up the good work and feel free to come to any of the coaches if you need help.

estate/views/estate_property_views.xml
coleo-odoo: The chapter says "we want a filter for the ‘Available’ properties".
Idk could be understood both way

msho-odoo: Either way you got how default filters work :) nice work!

PR #1423, opened 2026-09-15
ardec - Technical Training

Rutradc:


estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_property, estate_property_type, estate_property_tag, estate_property_offer
```
lost-odoo: We usually tend to separate them on different lines, so when we add an import, we do not change git history on something we did not really change

estate/models/estate_property.py
```diff
@@ -0,0 +1,101 @@
+from odoo.exceptions import UserError, ValidationError
+from odoo.tools import float_compare, float_is_zero
+
+from odoo import api, fields, models
```
lost-odoo: For imports, we usually try to separate and order them in different categories.
1- Simple imports (import datetime)
2- From imports from external modules (from datetime import ....)
3- Odoo imports

And inside each of those categories we try to order them in alphabetical order

estate/models/estate_property.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate property informations"
```
lost-odoo: A good convention is to use single quotes for technical strings, and double for user interfaced strings. Basically _name would be single and _description double. 

This applies everywhere :)

estate/models/estate_property.py
```diff
@@ -0,0 +1,101 @@
...
+
+    name = fields.Char(required=True, string="Title")
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(copy=False, default=fields.Date.add(fields.Date.today(), months=3), string="Available From")
```
lost-odoo: You should use a lambda function for your default otherwise the default would be static and set up when the server starts not everytime we create a new record.

estate/models/estate_property.py
```diff
@@ -0,0 +1,101 @@
...
+            ('new', 'New'),
+            ('offer_received', 'Offer Received'),
+            ('offer_accepted', 'Offer Accepted'),
+            ('sold', 'Sold'),
+            ('cancelled', 'Cancelled'),
+        ],
+        required=True,
+        default='new',
+    )
+    property_type_id = fields.Many2one("estate.property.type", string="Property Type")
```
lost-odoo: Try to always name your parameter so it is easier to read.

estate/models/estate_property.py
```diff
@@ -0,0 +1,101 @@
...
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("offer_ids.price")
+    def _compute_best_price(self):
+        for record in self:
+            record.best_price = max(record.offer_ids.mapped('price')) if len(record.offer_ids) > 0 else 0
```
```suggestion
            record.best_price = max(record.offer_ids.mapped('price')) if record.offer_ids else 0
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,101 @@
...
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = None
+
+    active = fields.Boolean(default=True)
```
lost-odoo: Do not separate your fields like this put them together at the top of the class. Then you can put your computes, on changes, ...

estate/models/estate_property.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def sell_property(self):
+        for record in self:
+            if record.state == "cancelled":
+                raise UserError("Cancelled properties cannot be sold.")
```
lost-odoo: You need to translate this using self.env._("...").

This applies at other places too.

estate/models/estate_property.py
```diff
@@ -0,0 +1,101 @@
...
+
+    _check_positive_expected_price = models.Constraint(
+        'CHECK(expected_price >= 0)',
+        'The expected price should be positive.'
+    )
+
+    _check_positive_selling_price = models.Constraint(
+        'CHECK(selling_price >= 0)',
+        'The selling price should be positive.'
+    )
```
lost-odoo: Same put them after the fields but before any functions.

estate/models/estate_property.py
```diff
@@ -0,0 +1,101 @@
...
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "Estate property informations"
+
+    name = fields.Char(required=True, string="Title")
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(copy=False, default=fields.Date.add(fields.Date.today(), months=3), string="Available From")
+    expected_price = fields.Float(required=True)
+    selling_price = fields.Float(readonly=True, copy=False)
```
lost-odoo: For monetary values, you should use a Monetary field ans add a currency field.

estate/models/estate_property.py
```diff
@@ -0,0 +1,101 @@
...
+
+    _check_positive_selling_price = models.Constraint(
+        'CHECK(selling_price >= 0)',
+        'The selling price should be positive.'
+    )
+
+    @api.constrains('selling_price', 'expected_price')
+    def _check_selling_price(self):
+        for record in self:
+            if float_compare(record.selling_price, record.expected_price * 0.9, 2) <= 0 and not float_is_zero(record.selling_price, 2):
```
lost-odoo: if you use a currency, you should use the currency comparison which are more accurate and have different roundings per currencies

estate/models/estate_property.py
lost-odoo: This applies also at multiple places.

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_estate_property,access.estate.property,model_estate_property,base.group_user,1,1,1,1
```
lost-odoo: You should suffix the id and name by the group name for instance here it should be `access_estate_property_user` because you could have `access_estate_property_manager` later on.

estate/security/security.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- <record id="groups_privilege_real_estate" model="base.module_category_real_estate_brokerage">
+        <field name="name">Real Estate</field>
+    </record> -->
```
lost-odoo: 👀

estate/views/estate_property_type_views.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="estate_property_type_view_form" model="ir.ui.view">
+        <field name="name">estate.property.type.form</field>
+        <field name="model">estate.property.type</field>
+        <field name="arch" type="xml">
+            <form>
+                <sheet>
+                    <div class="oe_title">
+                        <h1 class="mb32">
```
lost-odoo, line 10: Does this class mb32 exists ?

estate/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
+    'name': "Estate",
+    'author': "Arthur De Clerck",
+    'license': "LGPL-3",
+    'depends': ['base'],
+    'category': "Real Estate/Brokerage",
+    'application': True,
+    'data': [
+        'security/ir.model.access.csv',
+        # 'security/security.xml',
```
lost-odoo: 👀

lost-odoo (changes requested):
Hi, good job.

I made a first pass, and overall it is pretty good.
I had some comments, some are valid at multiple places.

Do not hesitate to ask me questions if needed.

estate/models/estate_property.py
Rutradc: Does it also applies in the xml files ?

estate/models/estate_property.py
Rutradc: I'm not sure how to implement it, should I add a currency_id field in the model like this
```python
currency_id = fields.Many2one(
        "res.currency", string="Currency",
        default=lambda self: self.env.company.currency_id,
    )
```
and then make the field like this
```python
expected_price = fields.Monetary(currency_field='currency_id', required=True)
```
or is there a better way to do it ?

estate/models/estate_property.py
Rutradc: Do you mean putting a string="name" in each field ?

PR #1425, opened 2026-09-16
19.0 technical training demat

DenisMth:


estate/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import estate_property
```
leclerc-leo: We try to always have an empty line at the end of every file.
You can use the Ruff extension with the config from https://runbot326.odoo.com/runbot/static/build/125365592-19-0/logs/ruff_config.txt to have warning about style inside your editor.
You can also check the ci/style from the runbot (https://runbot.odoo.com/runbot/bundle/190-technical-training-demat-512697) to check what it says.

estate/models/estate_property.py
```diff
@@ -0,0 +1,22 @@
+from odoo import fields, models
+import datetime
+
+class TestModel(models.Model):
```
leclerc-leo: nitpick (check ci/ctyle) but before classes, 2 empty lines are asked
```suggestion


class TestModel(models.Model):
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,22 @@
...
+    _name = "estate_property_model"
```
leclerc-leo: We name models `name.of.module.name.of.model`.
```suggestion
    _name = "estate.property"
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,22 @@
...
+    selling_price = fields.Float(copy=False, readonly=True)
+    bedrooms = fields.Integer(default=2)
+    living_area = fields.Integer()
+    facades = fields.Integer()
+    garage = fields.Boolean()
+    garden = fields.Boolean()
+    garden_area = fields.Integer()
+    garden_orientation = fields.Selection([("north", "North"), ("south", "South"), ("west", "West"), ("east", "East")])
+    active = fields.Boolean(default=True)
+    state = fields.Selection([("new", "New"), ("offer_received", "Offer Received"), ("offer_accepted", "Offer Accepted"), ("sold", "Sold"), ("cancelled", "Cancelled")], default="new")
```
leclerc-leo: We try to keep line bellow ~120 characters. If you have using vs code you can add a `rulers` in the setting to have a line and visualize it.
```suggestion
    state = fields.Selection(
        selection=[
            ("new", "New"),
            ("offer_received", "Offer Received"),
            ("offer_accepted", "Offer Accepted"),
            ("sold", "Sold"),
            ("cancelled", "Cancelled"),
        ],
        default="new",
    )
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,22 @@
...
+import datetime
+
+class TestModel(models.Model):
+    _name = "estate_property_model"
+    _description = "Estate property model test description"
+
+    name = fields.Char(default="Unknown", required=True)
+    description = fields.Text("Description of the Estate Propert Model")
+    postcode = fields.Char("Post code")
+    date_availability = fields.Date("Availability date", default=datetime.date.today() + datetime.timedelta(days=90), copy=False)
```
leclerc-leo: If not using lambda, `datetime.date.today()` will resolve when the server start, so if your server keept running for a few days, the date would be wrong.
Also for adding date it's best to use `fields.Date.add` instead of `+`
```suggestion
    date_availability = fields.Date("Availability date", default=lambda self: datetime.date.today() + datetime.timedelta(days=90), copy=False)
```

estate/views/menu_estate_property.xml
```diff
@@ -0,0 +1,30 @@
+<odoo>
+    <menuitem id="test_menu_root" name="Real Estate">
```
leclerc-leo: To prevent any issue it's best to prefix any view id with the name of the module
```suggestion
    <menuitem id="estate_test_menu_root" name="Real Estate">
```

leclerc-leo (commented):
👋 Good Work

aboo-odoo:

estate/models/estate_property.py
```diff
@@ -0,0 +1,90 @@
+from odoo import fields, models, api
```
aboo-odoo: [Imports](https://www.odoo.com/documentation/19.0/contributing/development/coding_guidelines.html#imports)
The imports are ordered as: 
- Imports of odoo submodules

```suggestion
from odoo import api, fields, models
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,90 @@
...
+    _name = "estate.property"
+    _description = "Estate property model test description"
+
+    total_area = fields.Integer(compute="_compute_total_area")
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        self.total_area = self.garden_area + self.living_area
+
+    best_price = fields.Float(compute="_compute_best_price")
```
aboo-odoo, line 13: Order of a python file is :
3. Field declarations
9. Action methods

So all your fields need to be together

estate/models/estate_property.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @api.depends("offer_ids")
+    def _compute_best_price(self):
+        for record in self:
+            if record.offer_ids:
+                record.best_price = max(record.offer_ids.mapped('price'))
+            else:
+                record.best_price = 0
```
aboo-odoo: Correct be could be simplified into 

```suggestion
            record.best_price = max(record.offer_ids.mapped('price'), default=0)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        self.total_area = self.garden_area + self.living_area
+
+    best_price = fields.Float(compute="_compute_best_price")
+
+    @api.depends("offer_ids")
+    def _compute_best_price(self):
+        for record in self:
```
aboo-odoo: Here you have quite small compute methods. In the future, you might have huge once, at that point, it's important to have meaningful parameter naming.
```suggestion
        for estate in self:
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,90 @@
...
+                record.best_price = 0
+
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = ""
```
aboo-odoo: Empty is not a valid value from your selection. Use

```suggestion
            self.garden_orientation = False
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def action_cancel(self):
+        for record in self:
+            if record.state != "sold":
+                record.state = "cancelled"
+            else:
+                raise UserError("A sold property cannot be cancelled.")
```
aboo-odoo: Most strings are translated by default in Odoo. Unfortunately, this is not case of Error Messages. To make the string translatable, you need to call the `_()` method, like so

estate/models/estate_property.py
```diff
@@ -0,0 +1,90 @@
...
+        return True
```
aboo-odoo: Correct but you could also do this:

```suggestion
        properties_to_cancel = self.filtered(lambda p: p.state != 'sold')
        if self - properties_to_cancel:  # at least one property is 'sold'
            raise UserError(...)
        else:
            properties_to_cancel.state = 'cancelled'
        return True
```
Advantages are:

estate/models/estate_property.py
```diff
@@ -0,0 +1,90 @@
...
+            else:
+                raise UserError("A cancelled property cannot be sold.")
+        return True
+
+    name = fields.Char(required=True)
+    description = fields.Text("Description of the Estate Propert Model")
+    postcode = fields.Char("Post code")
+    date_availability = fields.Date(
+        "Availability date",
+        default=lambda self: datetime.date.today() + datetime.timedelta(days=90),
```
aboo-odoo, line 19: Good use of a lambda function here 👍

estate/models/estate_property.py
```diff
@@ -0,0 +1,90 @@
...
+        selection=[
+            ("new", "New"),
+            ("offer_received", "Offer Received"),
+            ("offer_accepted", "Offer Accepted"),
+            ("sold", "Sold"),
+            ("cancelled", "Cancelled")
+        ],
+        default="new",
+    )
+    property_type_id = fields.Many2one("estate.property.type", string="Property Type")
```
aboo-odoo, line 49: Good use of `_id.s` suffixes! Well done 👍

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,45 @@
...
+    _name = "estate.property.offer"
+    _description = "Estate Property Offer description test"
+
+    validity = fields.Integer()
+    date_deadline = fields.Date(compute="_compute_validity_date", inverse="_inverse_validity_date")
+
+    @api.depends("create_date", "validity")
+    def _compute_validity_date(self):
+        for record in self:
+            record.date_deadline = fields.Date.add(fields.Date.today(), days=record.validity)
```
aboo-odoo: In the specs, doesn't it say that the validity depends on the create_date value ?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def _inverse_validity_date(self):
+        for record in self:
+            record.validity = (record.date_deadline - fields.Date.to_date(record.create_date)).days
+
+    def action_accept_offer(self):
+        for record in self:
+
+            if record.property_id.selling_price == 0 and not record.property_id.buyer_id:
```
aboo-odoo: Always Use `float_compare` to compare float

aboo-odoo: Why using the `property_id.selling_price` and not the `property_id.state` ? Isn't the goal of the state to be used on condition like this ?

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,45 @@
...
+    @api.depends("create_date", "validity")
+    def _compute_validity_date(self):
+        for record in self:
+            record.date_deadline = fields.Date.add(fields.Date.today(), days=record.validity)
+
+    def _inverse_validity_date(self):
+        for record in self:
+            record.validity = (record.date_deadline - fields.Date.to_date(record.create_date)).days
+
+    def action_accept_offer(self):
```
aboo-odoo, line 31: Also, in the specs, if an offer is accepted, shouldn't the property by sold?

estate/views/form_estate_property.xml
```diff
@@ -0,0 +1,101 @@
...
+                            </group>
+                            <group>
+                                <field string="Garden Area (sqm)" name="garden_area"/>
+                            </group>
+                            <group>
+                                <field string="Garden Orientation" name="garden_orientation"/>
+                            </group>
+                            <group>
+                                <field string="Total Area" name="total_area"/>
+                            </group>
```
aboo-odoo, line 56: That's a lot of groups !

```suggestion
                            <group>
                                <field string="Description" name="description"/>
                                <field string="Bedrooms" name="bedrooms"/>
                                <field string="Living Area (sqm)" name="living_area"/>
                                <field string="Facades" name="facades"/>
                                <field string="Garage" name="garage"/>
                                <field string="Garden" name="garden"/>
                                <field string="Garden Area (sqm)" name="garden_area"/>
                                <field string="Garden Orientation" name="garden_orientation"/>
                                <field string="Total Area" name="total_area"/>
                            </group>
```
Should do the trick

estate/views/list_estate_property_offer.xml
```diff
@@ -0,0 +1,15 @@
...
+            <list>
+                <field name="price" string="Price"/>
+                <field name="partner_id" string="Partner"/>
+                <field name="status" string="Status"/>
+                <field name="validity"/>
+                <field name="date_deadline"/>
+            </list>
+        </field>
+    </record>
+</odoo>
```
aboo-odoo: Always leave an empty line at the end of a file,

estate/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+{
+    'name': "Real Estate",
+    'depends': ['base'],
+    'application': True,
+    'installable': True,
+    'author': "Demat",
```
aboo-odoo: You don't own what you code here ahah; it's all for Odoo S.A. 
```suggestion
    'author': "Odoo S.A.",
```

aboo-odoo (commented):
Good job so far! Keep going

Generic comment:

estate/models/estate_property_offer.py
DenisMth: I changed to make the state change to "offer_accepted" because otherwise this state is useless, if it goes directly from "offer_received" to "sold" but maybe I forgot something in the flow

PR #1426, opened 2026-09-16
19.0 tutorial mialm

MiguelitoM:


estate/models/estate_property.py
```diff
@@ -0,0 +1,24 @@
+from odoo import fields, models
+
+
+class EstateProperty(models.Model):
+    _name = "estate_property"
+    _description = "This is the model for the real estate property :)"
+    name = fields.Char(required=True)
```
leclerc-leo, line 8: We use `name.of.module.name.of.model`
```suggestion
class EstateProperty(models.Model):
    _name = "estate.property"
    _description = "This is the model for the real estate 

    name = fields.Char(required=True)
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,24 @@
...
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(copy=False, default=fields.Datetime.add(fields.Datetime.today(), months=3))
```
leclerc-leo: Otherwise fields.Datetime.today is resolved when the server start and you will have issue if it runs for more than a day.
```suggestion
    date_availability = fields.Date(copy=False, default=lambda self: fields.Datetime.add(fields.Datetime.today(), months=3))
```

estate/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+{
+    'name': "Real Estate",
+    'depends': ["base"],
+    'application': True,
+    'installable': True,
+    'author': 'mialm',
```
```suggestion
    'author': 'Odoo S.A.',
```

leclerc-leo (commented):
👋 Good Work, you can check ci/style too.
If you need to setup a linter you can use ruff (there's an extension in vscode/pycharm) and use the odoo.toml present in odoo/community.

PR #1427, opened 2026-09-16
19.0 technical onboarding macau

mcauz:


estate/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+from . import (
+    estate_property,
+    estate_property_offer,
+    estate_property_tag,
+    estate_property_type,
+)
```
leclerc-leo: Nitpick: We usually do something like this instead
```suggestion
from . import estate_property
from . import estate_property_offer
from . import estate_property_tag
from . import estate_property_type
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,114 @@
+from odoo import _, api, fields, models
+
+
+class EstateProperty(models.Model):
+    _name = 'estate_property'
```
leclerc-leo: We usually use `name.of.module.name.of.model`
```suggestion
    _name = 'estate.property'
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,114 @@
...
+        string='Name',
+        required=True,
+        default="Unknown",
+    )
+    description = fields.Text(
+        string='Description',
+    )
+    postcode = fields.Char(
+        string='Postcode',
+    )
```
leclerc-leo: nitpick, if it's possible we try to keep lines as one (if they dont go over ~120 characters)
```suggestion
    postcode = fields.Char(string='Postcode')
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,114 @@
...
+    date_availability = fields.Date(
+        string='Date availability',
+        copy=False,
+        default=fields.Date.add(fields.Date.today(), months=3),
```
leclerc-leo: If not used inside a function `fields.Date.today()` is resolved when the server starts so if it runs for a few days you will end up with errors.
```suggestion
        default=lambda self: fields.Date.add(fields.Date.today(), months=3),
```

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,estate.model_estate_property,base.group_user,1,1,1,1
+estate.access_estate_property_offer,access_estate_property_offer,estate.model_estate_property_offer,base.group_user,1,1,1,1
+estate.access_estate_property_tag,access_estate_property_tag,estate.model_estate_property_tag,base.group_user,1,1,1,1
+estate.access_estate_property_type,access_estate_property_type,estate.model_estate_property_type,base.group_user,1,1,1,1
```
leclerc-leo: We try to always have an empty line at the end of every file.
You can use the Ruff extension with the config from https://runbot326.odoo.com/runbot/static/build/125365592-19-0/logs/ruff_config.txt to have warning about style inside your editor.
You can also check the ci/style from the runbot (https://runbot.odoo.com/runbot/bundle/190-technical-training-demat-512697) to check what it says.

estate/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+{
+    'name': 'Real Estate',
+    'author': 'macau',
```
```suggestion
    'author': 'Odoo S.A.',
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,114 @@
+from odoo import _, api, fields, models
```
leclerc-leo: nitpick: useless import
```suggestion
from odoo import api, fields, models
```

leclerc-leo (commented):
👋 Good Work

estate/models/estate_property.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @api.depends('living_area', 'garden_area')
+    def _compute_total_area(self):
+        for record in self:
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends('offers.price')
+    def _compute_best_price(self):
+        for record in self:
+            record.best_price = max(record.offers.mapped('price')) if record.offers else 0
```
leclerc-leo, line 64: nitpick, both are valid
```suggestion
            record.best_price = max([0, *record.offers.mapped('price')])
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @api.constrains('expected_price', 'selling_price')
+    def _check_selling_price_relation_to_expected_price(self):
+        for record in self:
+            if float_compare(record.expected_price * 0.9, record.selling_price, 2) == 1 and not float_is_zero(record.selling_price, 2):
+                raise ValidationError(self.env._(r'The selling price cannot be lower than 90% of the expected price.'))
+
+    @api.ondelete(at_uninstall=False)
+    def _unlink_if_status_is_new_or_cancelled(self):
+        if any(record.state not in ['new', 'cancelled'] for record in self):
```
leclerc-leo, line 107: When using `in`, It's usually best to use a set rather than a list
```suggestion
        if any(record.state not in {'new', 'cancelled'} for record in self):
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,108 @@
...
+            raise UserError(self.env._('Can\'t delete a property that is not new or cancelled.'))
```
leclerc-leo, line 108: nitpick
```suggestion
            raise UserError(self.env._("Can't delete a property that is not new or cancelled."))
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,90 @@
...
+            'name': self.env._('Properties'),
+            'views': [[False, 'list'], [False, 'form']],
+            'domain': [('id', '=', self.property_id.id)],
+        }
+
+    @api.constrains('price')
+    def _check_price(self):
+        for record in self:
+            if float_compare(record.property_id.expected_price * 0.9, record.price, 2) == 1:
+                raise ValidationError(self.env._(r'The price of an offer cannot be lower than 90% of the expected price of the property.'))
```
leclerc-leo, line 80: your IDE probably doesn't like it but python accept it fine
```suggestion
                raise ValidationError(self.env._('The price of an offer cannot be lower than 90% of the expected price of the property.'))
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @api.model
+    def create(self, vals_list):
+        for vals in vals_list:
+            property = self.env['estate.property'].browse(vals['property_id']).with_prefetch(self.ids)
+            if any(offer.price > vals['price'] for offer in property.offers):
```
leclerc-leo, line 86: price is not a required field so it may be missing (probably blocked by the constraint tho)
```suggestion
            if any(offer.price > vals.get('price', 0) for offer in property.offers):
```

estate/models/estate_property_tag.py
```diff
@@ -0,0 +1,15 @@
+from odoo import fields, models
+
+
+class EstatePropertyTag(models.Model):
+    _name = 'estate.property.tag'
+    _description = 'Estate Property Tag'
+    _order = 'name'
+
+    name = fields.Char(string='Name', required=True, default="Unknown")
```
leclerc-leo, line 9: nitpick, by default Odoo using the technical name of the field with `_` replaced with spaces and the first letter of each word capitalized.
```suggestion
    name = fields.Char(required=True, default="Unknown")
```

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,23 @@
...
+
+    _unique_type_name = models.Constraint(
+        'unique (name)',
+        'The name of a property type should be unique.',
+    )
+
+    @api.depends('offers')
+    def _compute_offers_count(self):
+        for record in self:
+            record.offers_count = len(record.offers)
```
leclerc-leo, line 23: nitpick, this is probably too much for now, but here's what this would look like instead.
Using _read_group allow to fetch the count directly (not the offer so it's a bit faster) and make only one request to the database.
The write cannot be batched tho.
```suggestion
        counts = self.env['estate.property.offer']._read_group(
            domain=[('property_type', 'in', self.ids)],
            groupby=['property_type'],
            aggregates=['__count'],
        )
        mapped = {property_type.id: count for property_type, count in counts}
        for record in self:
            record.offers_count = mapped.get(record.id, 0)
```

leclerc-leo (commented):
👋 Great Work. not much too say really.

estate/models/estate_property_offer.py
mcauz: The problem is with the percent character that is a special character. If I escape it with double percent characters, it appear two time in the string. So the only solution that I find is to use the raw string.

estate/models/estate_property_type.py
mcauz: I love this type of optimization. Thanks !

estate/models/estate_inherited_users.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class EstateInheritedUsers(models.Model):
+    _inherit = 'res.users'
+
+    property_ids = fields.One2many(
+        comodel_name='estate.property',
+        inverse_name='salesperson',
+        string='Estate Properties',
+        domain=['|', ('state', '=', 'new'), ('state', '=', 'offer_received')],
```
leclerc-leo:
```suggestion
        domain=[('state', 'in', ('new', 'offer_received'))],
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+    )
+    living_area = fields.Integer()
+    name = fields.Char(required=True, default="Unknown")
+    offers = fields.One2many(comodel_name='estate.property.offer', inverse_name='property_id')
+    postcode = fields.Char()
+    property_type = fields.Many2one(comodel_name='estate.property.type')
+    salesperson = fields.Many2one(comodel_name='res.users', default=lambda self: self.env.user)
+    selling_price = fields.Float(readonly=True, copy=False)
+    state = fields.Selection(
+        selection=[('new', 'New'), ('offer_received', 'Offer received'), ('offer_accepted', 'Offer accepted'), ('sold', 'Sold'), ('cancelled', 'Cancelled')],
```
leclerc-leo: We try to keep the length of a line bellow 120 characters.
If you are using vscode you can add a "ruler" in the setting for that many character to easily catch those.

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+    @api.onchange('garden')
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = 'north'
+        else:
+            self.garden_area = 0
+            self.garden_orientation = None
+
+    def sold_property(self):
```
leclerc-leo: We prefix every function that will be called from a button an action and prefix it in the function to make it clearer it will probably be called by the user.
```suggestion
    def action_sold_property(self):
```

estate/models/estate_property.py
```diff
@@ -0,0 +1,105 @@
...
+        for record in self:
+            if record.state == 'cancelled':
+                raise UserError(self.env._('A cancelled property cannot be sold.'))
+            record.state = 'sold'
```
leclerc-leo: You can batch the write here as they all get written the same value.
```suggestion
        self.state = 'sold'
```

estate_account/models/EstateInheritedProperty.py
```diff
@@ -0,0 +1,27 @@
...
+                        'price_unit': record.selling_price * 0.06,
+                    }),
+                    Command.create({
+                        'name': self.env._('Administrative fees'),
+                        'quantity': 1,
+                        'price_unit': 100,
+                    }),
+                ],
+            })
+        return super().sold_property()
```
leclerc-leo, line 29: You can batch create the moves.
```suggestion
        journal_id = self.env['account.journal'].search([('code', '=', 'INV')], limit=1).id
        moves = []
        for record in self:
            moves.append({
                'partner_id': record.buyer.id,
                'move_type': 'out_invoice',
                'journal_id': journal_id,
                'invoice_line_ids': [
                    Command.create({
                        'name': self.env._('Purchase of %s', record.id),
                        'quantity': 1,
                        'price_unit': record.selling_price * 0.06,
                    }),
                    Command.create({
                        'name': self.env._('Administrative fees'),
                        'quantity': 1,
                        'price_unit': 100,
                    }),
                ],
            })
        self.env['account.move'].create(moves)
        return super().sold_property()
```

leclerc-leo (commented):
👋 Great work.

PR #1429, opened 2026-09-16
19.0 technical training jever

JeromeVerkyndt:


estate/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
```
lost-odoo: Hi do not forget to follow those [Git guidelines](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html) when doing you commit messages

estate/models/__init__.py
```diff
@@ -0,0 +1,3 @@
...
+
+from . import estate_property
```
lost-odoo: Missing final new line at end of file

estate/models/__init__.py
lost-odoo: `# -*- coding: utf-8 -*-` not needed anymore

estate/models/estate_property.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+
+from odoo import fields, models
+from datetime import timedelta, date
+
+class Property(models.Model):
+    _name = "estate.property"
+    _description  = "Estate property model"
+
```
lost-odoo, line 7: Do not forget to make you runbot green, here ci/style would complain

estate/models/estate_property.py
```diff
@@ -0,0 +1,34 @@
...
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(copy=False, default=date.today()+ timedelta(days=90))
```
lost-odoo: You should use a lambda here for your default parameter otherwise, it would be fixed from the server start. Using a lambda here would be dynamic at creation.

lost-odoo (changes requested):
Hi I already did a small review, overall it's good just minor things. I always made one comment for a specific issue but it applies for the whole pull request.

So for instance the final newline occurs at multiple places.

Good Job :D

JeromeVerkyndt:
@lost-odoo Here are my unit test you asked for the extra challenge:)

PR #1430, opened 2026-09-16
[ADD] Estate: Adding new estate module

GerolD122:
Creating new estate module, which consists from property estate model with types, tags, and also offers to buy thos estate modules. More info [in here](https://www.odoo.com/documentation/19.0/developer/tutorials/server_framework_101.html)
Here is code for Finished Chapters 1-11

[[Tutorial] Server framework 101](https://www.odoo.com/odoo/action-1592/new/26755/tasks/6574031)


estate/models/estate_property.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "This is estate property model"
+
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(copy=False, default=fields.Datetime.today().replace(month=fields.Datetime.today().month +3))
```
leclerc-leo: without lambda self, `fields.Datetime.today` is resolved when the server start so you will have some issue at some point. Also fields.Date.add is shorter and easier to understand that replace  
```suggestion
    date_availability = fields.Date(copy=False, default=lambda self: fields.Date.add(fields.Date.today(), months=3))
```

estate/views/__init__.py
```diff
@@ -0,0 +1 @@
+
```
leclerc-leo: this file is not needed

.gitignore
```diff
@@ -127,3 +127,4 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+estate/.vscode/settings.json
```
leclerc-leo: try not to commit changes of generic files

leclerc-leo (commented):
👋 Good Work, you can check ci/style too.
If you need to setup a linter you can use ruff (there's an extension in vscode/pycharm) and use the odoo.toml present in odoo/community.

aboo-odoo:

estate/models/estate_property.py
```diff
@@ -0,0 +1,92 @@
...
+                raise UserError(_("Sold property cannot be cancelled"))
+            else:
+                record.state = "cancelled"
+
+    def action_sell_to_offer(self, offer_buyer_id, offer_selling_price):
+        self.buyer_id = offer_buyer_id
+        self.selling_price = offer_selling_price
+        self.state = "sold"
+        
+        return True
```
GerolD122, line 91: @aboo-odoo I have a question about implementation of this logic. I decided to leave any changing of  data inside of estate.property model for inner methods (because of encapsulation), and didnt looped through self inside of this method.
Can i have any thoughts about this implementation? I kinda not sure if this approach is correct, because i not sure what practices are used in odoo framework, and how some things work "under the hood" :/
Thank you

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,53 @@
...
+        for record in self:
+            if (
+                record.property_id.state == "sold"
+                or record.property_id.state == "cancelled"
+            ):
+                raise UserError(_("Property already %s") % record.property_id.state)
+            elif record.property_id.action_sell_to_offer(
+                offer_buyer_id=record.partner_id, offer_selling_price=record.price
+            ):
+                record.status = "accepted"
```
GerolD122, line 49: place where action_sell_to_offer() is called

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+
+class EstateProperty(models.Model):
+    _name = "estate.property"
+    _description = "This is estate property model"
+    _expected_price_constraint = models.Constraint(
+        "CHECK (expected_price >= 0)", "Expected Price must be positive or zero."
+    )
+    _selling_price_constraint = models.Constraint(
+        "CHECK (selling_price >= 0)", "Selling Price must be positive."
+    )
```
aboo-odoo: Order of a python file is :
3. Field declarations
9. Action methods

So all your SQL constraints need to be after the fields

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+
+    property_type_id = fields.Many2one("estate.property.type", string="Type")
+    name = fields.Char(required=True)
+    description = fields.Text()
+    postcode = fields.Char()
+    date_availability = fields.Date(
+        copy=False, default=lambda self: fields.Date.add(fields.Date.today(), months=3)
```
aboo-odoo, line 18: Good use of lambda here!

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+            ("new", "New"),
+            ("offer_recieved", "Offer Recieved"),
+            ("offer_accepted", "Offer Accepted"),
+            ("sold", "Sold"),
+            ("cancelled", "Cancelled"),
+        ],
+    )
+    active = fields.Boolean(default=True)
+    total_area = fields.Integer(compute="_compute_total_area")
+    buyer_id = fields.Many2one("res.partner", string="Buyer", copy=False)
```
aboo-odoo, line 51: Good use of `_id.s` suffixes fir relational fields, well done

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+    salesperson_id = fields.Many2one(
+        "res.users", string="Salesperson", default=lambda self: self.env.user
+    )
+    tag_ids = fields.Many2many("estate.property.tag")
+    offer_ids = fields.One2many("estate.property.offer", "property_id", string="Offers")
+
+    @api.depends("living_area", "garden_area")
+    def _compute_total_area(self):
+        for record in self:
```
```suggestion
        for property in self:
```
Here you have quite small compute methods. In the future, you might have huge once, at that point, it's important to have meaningful parameter naming.

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+            record.total_area = record.living_area + record.garden_area
+
+    @api.depends("salesperson_id")
+    def _find_best_price(self):
+        for record in self:
+            record.best_price = max(record.offer_ids.mapped("price"), default=0)
```
aboo-odoo: 👍

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+
+    @api.onchange("garden")
+    def _onchange_garden(self):
+        if self.garden:
+            self.garden_area = 10
+            self.garden_orientation = "north"
+        else:
+            self.garden_area = 0
+            self.garden_orientation = ""
```
aboo-odoo, line 86:
```suggestion
            self.garden_orientation = False
```

False is the default value for selection

estate/models/estate_property.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def action_sold_property(self):
+        for record in self:
+            if record.state == "cancelled":
+                raise UserError(_("Cancelled property cannot be sold"))
+            else:
+                record.state = "sold"
```
aboo-odoo, line 114: A slightly better but more advanced way, would be to do like so:
```suggestion
        properties_to_sell = self.filtered(lambda p: p.state != 'cancelled')
        if self - properties_to_sell:
            raise UserError(...)
        else:
            properties_to_sell.state = 'sold'
        return True
```

Advantages are: 

estate/models/estate_property.py
aboo-odoo: You could like that and placing all the method updating property values in the property models (it's actually better in my opinion too). But at the end, it comes down to logic, personal preferences, and team preferences. 

Here, your implementation works without loop because you only preform write operation on the records. If you were reading, you would have had issue when self is recordset -> writes can be performed on a recordset, reads need to be sequential.

estate/security/__init__.py
aboo-odoo, line 1: No needed, you can delete this file

estate/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+estate.access_estate_property,access_estate_property,model_estate_property,base.group_user,1,1,1,1
+estate.access_estate_property_type,access_estate_property_type,model_estate_property_type,base.group_user,1,1,1,1
+estate.access_estate_property_tag,access_estate_property_tag,model_estate_property_tag,base.group_user,1,1,1,1
+estate.access_estate_property_offer,access_estate_property_offer,model_estate_property_offer,base.group_user,1,1,1,1
```
aboo-odoo: Always leave an empty line at the end of a file, 

estate/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "Estate module",
+    "depends": ["base"],
+    "application": True,
+    "author": "vadem",
```
aboo-odoo:
```suggestion
    "author": "Odoo S.A.",
```
Odoo owns you work, not you ahha

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,55 @@
...
+                record.create_date.date() if record.create_date else fields.Date.today()
+            )
+            record.validity = (record.date_deadline - base_date).days
+
+    def action_confirm(self):
+        for record in self:
+            if (
+                record.property_id.state == "sold"
+                or record.property_id.state == "cancelled"
+            ):
```
aboo-odoo:
```suggestion
            if (
                record.property_id.state in ["sold", "cancelled"]
            ):
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,55 @@
...
+                raise UserError(_("Property already sold, or cancelled"))
+            elif record.property_id.action_sell_to_offer(
+                offer_buyer_id=record.partner_id, offer_selling_price=record.price
+            ):
```
aboo-odoo: Since `action_sell_to_offer` is doing stuff, I wouldn't put the return value as a condition. Anyway, if it were to raise an Error, the rest of your method wouldn't get executed


```suggestion
            record.property_id.action_sell_to_offer(
                offer_buyer_id=record.partner_id,
                offer_selling_price=record.price,
            )
```

estate/models/estate_property_offer.py
```diff
@@ -0,0 +1,55 @@
...
+                record.status = "accepted"
+
+    def action_refuse(self):
+        for record in self:
+            record.status = "refused"
```
aboo-odoo:
```suggestion
        self.status = "refused"
```

As mentioned, writes can be done on recordset

aboo-odoo (commented):
Overall, excellent job! 🥳

Global comments:

aboo-odoo (commented):
Nice ! Thank you for the test 😄

GerolD122:
@aboo-odoo Thanks for review! I checked your comments and made changes according to them, but i'm gonna squash all my commits after i'll finish all chapters and will have "finished job"

PR #1431, opened 2026-09-16
19.0 technical training manab

manab-odoo:
First PR with Chapter 1 to 5 done and Chapter 6 half done


estate/models/estate_property.py
```diff
@@ -0,0 +1,43 @@
+from odoo import models, fields
+
```
lost-odoo, line 2: Do not forget the make your runbot green, for instance here it should complain because it does not have two lines before the class definition.

estate/models/estate_property.py
```diff
@@ -0,0 +1,43 @@
...
+class Property(models.Model):
+    _name = "estate_property"
+    _description = "estate property model"
+
+    name = fields.Char('Nom', required=True)
+    description = fields.Text('Description')
+    postcode = fields.Char('Post Code')
+    date_availability = fields.Date('Date Availability', copy=False, default=fields.Date.add(fields.Date.today(), months=3))
```
lost-odoo, line 11: here default should use a lambda function otherwise the default would be static depending on when the server started not at record creation.

estate/models/estate_property.py
```diff
@@ -0,0 +1,43 @@
...
+    garage = fields.Boolean('Has Garage')
+    garden = fields.Boolean('Has Garden')
+    garden_area = fields.Integer('# Garden Areas')
+    garden_orientation = fields.Selection(
+        string = 'Garden Orientation',
+        selection=[
+        ('north', 'North'), 
+        ('south','South'),
+        ('east','East'),
+        ('west','West')
```
lost-odoo: You should use single quotes for technical string and double quotes for user interface strings.

estate/models/estate_property_type.py
```diff
@@ -0,0 +1,7 @@
+from odoo import models, fields
+
+class EstatePropertyType(models.Model):
+    _name = "estate.property.type"
+    _description = "Types of Estate Property"
+
+    name = fields.Char('Nom', required=True)
```
lost-odoo: Do not forget the final newline.

estate/views/estate_property_type_views.xml
lost-odoo, line 1: You should avoid pushing empty files but I guess it is because you are working on it. But just in case 😄

estate/models/estate_property.py
```diff
@@ -0,0 +1,43 @@
...
+            ('new', 'New'), 
+            ('offer_received', 'Offer Received'), 
+            ('offer_accepted', 'Offer accepted'), 
+            ('sold', 'Sold'), 
+            ('cancelled', 'Cancelled')
+            ],
+        required=True,
+        copy=False,
+        default='new'
+    )
```
lost-odoo, line 42:
```suggestion
    state = fields.Selection(
        string='State',
        selection=[
            ('new', 'New'), 
            ('offer_received', 'Offer Received'), 
            ('offer_accepted', 'Offer accepted'), 
            ('sold', 'Sold'), 
            ('cancelled', 'Cancelled')
        ],
        required=True,
        copy=False,
        default='new',
    )
```

Be careful at your indentation as well as the final comma, it is better to always add it so that later if someone wants to add a new line in here, it won't trigger a diff on the final line you added.

estate/models/estate_property.py
```suggestion
        ('west','West'),
```

estate/views/estate_property_views.xml
```diff
@@ -0,0 +1,106 @@
...
+                            <group>
+                                <field name="garden" string="Garden"/>
+                            </group>
+                            <group>
+                                <field name="garden_area" string="Garden Area (sqm)"/>
+                            </group>
+                            <group>
+                                <field name="garden_orientation" string="Garden Orientation"/>
+                                <field name="state" string="State"/>
+                            </group>
```
lost-odoo, line 69: Why doing so many groups ?

lost-odoo (changes requested):
Hi thanks for you pull request,

I made a small review, sometimes I put a comment for one thing but it applies to other files as well.

Good Job :D

